PR #1, opened 2017-03-15
Remove prefer-stable

chillu:
It checks out outdated tags on framework in builds,
which work differently from a `composer create-project` (module is composer root).
At least prior to beta1, we want to test against latest versions of dependencies.

Related build failure: https://travis-ci.org/silverstripe/silverstripe-admin/jobs/211094670#L445


dhensby:
see https://github.com/silverstripe/silverstripe-framework/pull/6712 for how we think this is best resolved.

PR #3, opened 2017-03-19
Drop PHP 5.5 support

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/6705


tractorcow:
Build isn't green but not going to block merge.

PR #6, opened 2017-03-21
Re-instate the insert oembed function

flamerohr:
Removes unnecessary entwine code as they were rebuilt with react in the asset-admin

See https://github.com/silverstripe/silverstripe-framework/issues/6480 for more details.

Depends on this PR:
https://github.com/silverstripe/silverstripe-framework/pull/6727

Dependent PRs:
https://github.com/silverstripe/silverstripe-asset-admin/pull/406


client/src/lib/Injector.js
```diff
@@ -28,6 +28,8 @@ class Injector {
    */
   getComponentByDataType(dataType) {
     switch (dataType) {
+      case 'String':
+        return this.components.LabelField;
```
tractorcow: String is single-line editable text, not single-line uneditable text. This should default to `TextField`, and Labels should be marked as Structural (SCHEMA_DATA_TYPE_STRUCTURAL).

client/src/components/FieldHolder/FieldHolder.js
```diff
@@ -123,26 +143,34 @@ function fieldHolder(Field) {
...
+    id: PropTypes.string,
+    description: PropTypes.any,
+    hideLabels: PropTypes.bool,
+    message: PropTypes.shape({
+      extraClass: PropTypes.string,
+      value: PropTypes.any,
+      type: PropTypes.string,
     }),
+    data: PropTypes.oneOfType([
+      PropTypes.array,
```
tractorcow: Should data be object not array It could be an object without prefix / suffix right?

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -11,6 +11,20 @@ class FormBuilderModal extends SilverStripeComponent {
     this.handleSubmit = this.handleSubmit.bind(this);
     this.handleHide = this.handleHide.bind(this);
     this.clearResponse = this.clearResponse.bind(this);
+    this.handleLoadingError = this.handleLoadingError.bind(this);
+  }
+
+  handleLoadingError(schema) {
+    if (this.props.showErrorMessage) {
```
tractorcow: Not defined in props. Where is this turned on?

client/src/lib/Injector.js
flamerohr: oh right, thanks for that, I completely forgot it was the way it's edited

client/src/components/FieldHolder/FieldHolder.js
flamerohr: The trouble stems from php converting empty associative arrays into *gasp* arrays :(
So where originally it could be PropTypes.shape alone, if none of the fields inside are required, it could be returned to us as an array...

client/src/components/FormBuilderModal/FormBuilderModal.js
flamerohr: Added missing definition, it's used in `InsertEmbedModal`.

PR #8, opened 2017-03-27
Fix missing permissions codes for ModelAdmins

wilr:
When creating a new ModelAdmin subclass in 4.0 there is not a matching permission checkbox in the CMS. This is because the root LeftAndMain::providePermissions() handles the creating of the codes and $this->getRequiredPermissions is only called on the root LeftAndMain rather than the target class.


sminnee:
No chance of a test? :)

PR #13, opened 2017-04-02
ENHANCEMENT Support text formatting in modal response

phalkunz:
https://github.com/silverstripe/silverstripe-framework/issues/6115

### DIFF
--- a/client/src/legacy/AddToCampaignForm.js
+++ b/client/src/legacy/AddToCampaignForm.js
@@ -26,6 +26,16 @@ jQuery.entwine('ss', ($) => {
     },
   });
 
+  // This is required because the React version of e.preventDefault() doesn't work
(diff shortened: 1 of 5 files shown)


client/src/legacy/AddToCampaignForm.js
```diff
@@ -26,6 +26,16 @@ jQuery.entwine('ss', ($) => {
     },
   });
 
+  // This is required because the React version of e.preventDefault() doesn't work
+  // this is to prevent PJAX request to occur when clicking a link the modal
+  $('.add-to-campaign-modal .add-to-campaign-modal__nav-link').entwine({
```
tractorcow: Another way I found that this can be circumvented is to add a `target` property (non-empty).

See https://github.com/visionmedia/page.js/blob/aa605501e240727a0afaef7ec1641f60c8d7e8a5/page.js#L575

PR #16, opened 2017-04-04
BUG Correctly validate date in datefield shiv

tractorcow:
lack of `isValid` check ment that "invalid date" was being passed back as the date value.


tractorcow:
After merge restart https://github.com/silverstripe/silverstripe-framework/pull/6774

PR #19, opened 2017-04-06
API Move tree handling code to CMS module

tractorcow:
See https://github.com/silverstripe/silverstripe-cms/issues/1510

API Introduce new custom CMS form handler to reduce reliance on session. This is used in cms and asset-admin modules.

Clean up LeftAndMain::getEditForm


code/LeftAndMain.php
```diff
@@ -1910,18 +1507,25 @@ public function getSilverStripeNavigator()
...
             // see GridFieldDetailForm::ItemEditForm
             return $this->getRequest()->requestVar('CMSMainCurrentPageID');
         } elseif (isset($this->urlParams['ID']) && is_numeric($this->urlParams['ID'])) {
             return $this->urlParams['ID'];
-        } elseif (Session::get($this->sessionNamespace() . ".currentPage")) {
+        }
+
+
+        /** @deprecated */
+        if (Session::get($this->sessionNamespace() . ".currentPage")) {
```
chillu: Should we deprecate `CMSMain->sessionNamespace()` as well? AFAICT it's not used elsewhere.

chillu:
`TranslatableCMSMainExtension` uses `currentPage()`, which is already using the new mechanism. `LeftAndMainSubsites` still uses sessions: https://github.com/silverstripe/silverstripe-subsites/search?p=2&q=session&type=&utf8=%E2%9C%93. How would you see this working with additional context data? `LeftAndMainFormRequestHandler` could be replaced by Injector with additional arguments, but it's construction is still "hardcoded" to just pass in `[$id]`. Maybe we need a `LeftAndMain->getRequestHandlerContext()` with an `updateRequestHandlerContext`? 

```
// Use custom reqest handler for LeftAndMain requests;
// CMS Forms cannot be identified solely by name, but also need ID (and sometimes OtherID)
$form->setRequestHandler(
  LeftAndMainFormRequestHandler::create($form, [$id])
);
```

The `[$id]` assumption is also hardcoded in `CMSMain::$url_handlers` though:

```
private static $url_handlers = [
  'EditForm/$ID' => 'EditForm',
];
```

tractorcow:
> How would you see this working with additional context data? LeftAndMainFormRequestHandler could be replaced by Injector with additional arguments, but it's construction is still "hardcoded" to just pass in [$id]. Maybe we need a LeftAndMain->getRequestHandlerContext() with an updateRequestHandlerContext?

We could achieve this, but probably will need to limit decorated "context" to querystring arguments. E.g.

```
class LeftAndMainSubsites extends Extension {
	public function updateEditForm(Form $form) {
		$handler = $form->getRequestHandler();a
		$handler->addContext('SubsiteID', $this->subsiteID()); // would add ?SubsiteID=X querystring arg
	}
}
```

tractorcow:
Added this concern as a (sparse) issue at https://github.com/silverstripe/silverstripe-admin/issues/29

PR #21, opened 2017-04-06
React date/datetime field

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/6140


client/src/components/DateField/DateField.js
```diff
@@ -0,0 +1,119 @@
...
+        localDate = dateObject.format(localFormat);
+      }
+    }
+    return localDate;
+  }
+
+}
+
+DateField.propTypes = {
+  lang: React.PropTypes.string,
```
tractorcow: Can you expand on this, and declare the `data` key (see my review on framework for context) with the necessary subkeys? We should also add top level standard keys here.

I'll create an issue for this.

tractorcow: See https://github.com/silverstripe/silverstripe-admin/issues/43

client/src/components/DateTimeField/DateTimeField.js
```diff
@@ -0,0 +1,73 @@
...
+
+class DateTimeField extends DateField {
+
+  getInputProps() {
+    const placeholder = i18n.inject(
+      i18n._t('DateTimeField.DateTimeFormatExample', 'Example: {datetime}'),
+      { datetime: moment().endOf('month').format(localFormat) }
+    );
+    const props = {};
+    Object.assign(props, super.getInputProps());
```
tractorcow: Can we chain Object.assign()?

```
return Object.assign(
  {},
  super.getInputProps(),
  {
    type: this.props.html5 ? 'datetime-local' : 'text',
    placeholder,
  }
);
```

client/src/components/DateField/DateField.js
```diff
@@ -0,0 +1,119 @@
...
+  getInputProps() {
+    const placeholder = i18n.inject(
+      i18n._t('DateField.DateFormatExample', 'Example: {datetime}'),
+      { datetime: moment().endOf('month').format(localFormat) }
+    );
+    const props = {};
+
+    let val = this.props.value;
+
+    if (!this.props.html5 || (this.hasNativeSupport() && this.props.html5)) {
```
tractorcow: `this.props.data.html5` ?

chillu: Done

client/src/components/DateTimeField/DateTimeField.js
chillu: Yep, I've done that now

PR #31, opened 2017-04-13
Fix asset thumbnail size

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/6430

### DIFF
--- a/client/src/components/ListGroup/ListGroup.scss
+++ b/client/src/components/ListGroup/ListGroup.scss
@@ -57,15 +57,25 @@
 .list-group-item__thumbnail {
   width: 64px;
   height: auto;
-  display: block;
-  background: $gray-lighter;
+
+  // Horizonally and vertically center
+  display: flex;
+  justify-content: center;
+  align-items: center;
+
   float: left;
   margin: -12px 12px -12px #{0 - $grid-gutter-width / 2};
 
   > img {
-    max-height: 65px;
-    height: 100%;
+    max-height: 100%;
+    max-width: 100%;
   }
+}
+
+.list-group-item__details {
+  // To make sure the thumbnail width (not actual images widths) is consistent
+  // across the list when the filename is really long
+  max-width: calc(100% - 64px);
 }
 
 .list-group-item__info {
--- a/client/src/components/ListGroup/ListGroup.scss
+++ b/client/src/components/ListGroup/ListGroup.scss
@@ -25,6 +25,7 @@
   cursor: pointer;
   text-decoration: none;
   transition: opacity .2s ease-in-out;
+  overflow: hidden;
 
   &:hover {
     text-decoration: none;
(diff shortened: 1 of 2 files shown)


client/src/components/ListGroup/ListGroup.scss
```diff
@@ -55,11 +56,16 @@
...
+  height: auto;
   display: block;
   background: $gray-lighter;
   float: left;
-  margin: -12px 12px 0 #{0 - $grid-gutter-width / 2};
+  margin: -12px 12px -12px #{0 - $grid-gutter-width / 2};
+
+  > img {
+    max-height: 65px;
+    height: 100%;
```
flamerohr: I suggest making this `max-height`, as the image could be smaller height than `100%`

phalkunz: The thumbnails can end up looking like:

silverstripe_-_campaigns

And I think this is quite different from the design -
 https://projects.invisionapp.com/d/main#/projects/boards/1547737/71700347

flamerohr: The mocks do not cater for when an image is too width or too short, or perhaps it's cropped on the sides (the car picture)
I do not agree that the picture should be stretched if it wasn't tall enough, are we able to have it vertically-aligned middle instead?

phalkunz: It looks to me the car picture in the design is the case where it's resized and cropped to fill the designated space. I think stretching (aspect ration is respected) the image to fit is still better than having image different sizes. We can get Jared's input on this since Paul is away.

flamerohr: yeah, my counter argument is that stretching an image is misrepresentation of what the image really is and we do not stretch the preview in the AssetAdmin Gallery either.

phalkunz: silverstripe_-_files

True, I don't mind not stretching the image so they're consistent everywhere but this is what I see in the AssetAdmin at the moment. Pretty sure the last image is vertically stretched or is it a bug?

flamerohr: hmm, doesn't look stretched to me, it's possible that the thumbnail is 2x size because of retina support and could be a bug with that instead?
The second file you have is what I'm describing :)

phalkunz: You can see in the detail view that the last image is really wide. Anyways, I'm happy not crop and stretch the image to keep them consistent.

flamerohr: We can investigate the really wide last image later on and see what's happening there :)

Sorry, didn't mean to dismiss that the last image is enlarged, I've got a few instances of that in my local as well

phalkunz: @flamerohr I've created an issue to cover this inconsistency in the asset admin https://github.com/silverstripe/silverstripe-asset-admin/issues/434

PR #32, opened 2017-04-18
Fix throw error rather than return empty object

flamerohr:
Related https://github.com/silverstripe/silverstripe-asset-admin/pull/431


tractorcow:
There's also an `onLoadingError` in FormBulderModal.js (asset-admin), but that can be fixed separately.

PR #33, opened 2017-04-20
Fix config sections is no longer an associative array

flamerohr:
This is so that order priority is enforced when json.stringify is used.

Pull requests which rely on this:
https://github.com/silverstripe/silverstripe-asset-admin/pull/435
https://github.com/silverstripe/silverstripe-campaign-admin/pull/10
https://github.com/silverstripe/silverstripe-frameworktest/pull/33

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -275,7 +275,7 @@ class LeftAndMain extends Controller implements PermissionProvider
         $cmsClassNames = CMSMenu::get_cms_classes(self::class, true, CMSMenu::URL_PRIORITY);
 
         foreach ($cmsClassNames as $className) {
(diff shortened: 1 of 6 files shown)


code/LeftAndMain.php
```diff
@@ -306,6 +306,7 @@ public function getClientConfig()
         return [
             // Trim leading/trailing slash to make it easier to concatenate URL
             // and use in routing definitions.
+            'name' => $this->class,
```
tractorcow: `$this->class` is going soon as we remove Object class.

flamerohr: good point, I've changed to `static:class` :D

tractorcow:
Looks good, will need a rebase soon. :)

tractorcow:
Tests are broken and I think the comments need updating in BootRouter.js. Can you please fix @flamerohr ?

PR #35, opened 2017-04-21
MINOR: Fix order of operations for filesize calculations

lukereative:
Fixes a bug where filesizes were being incorrectly calculated as the maths order of operations was not taken into consideration, this also remove decimals for KB

**Related PR:** https://github.com/silverstripe/silverstripe-assets/pull/10
**Related Issue:** https://github.com/silverstripe/silverstripe-asset-admin/issues/355


dhensby:
awesome, thanks!

PR #58, opened 2017-05-09
React implementation for insert link

flamerohr:
First part of the work to get Insert link modals working with React

See https://github.com/silverstripe/silverstripe-framework/issues/5947 for more details

Pull requests related
https://github.com/silverstripe/silverstripe-framework/pull/6914
https://github.com/silverstripe/silverstripe-cms/pull/1815


client/src/containers/InsertLinkModal/InsertLinkModal.js
```diff
@@ -0,0 +1,134 @@
...
+  setOverrides(props = null) {
+    if (!props) {
+      // clear any overrides that may be in place
+      const schemaUrl = props && props.schemaUrl || this.props.schemaUrl;
+      if (schemaUrl) {
+        this.props.actions.schema.setSchemaStateOverrides(schemaUrl, null);
+      }
+    }
+    if (props && props.schemaUrl) {
+      const attrs = Object.assign({}, props.fileAttributes);
```
flamerohr: this property was already named like this from InsertMediaModal
I'm hesitant to rename it here, as I feel we could move all this `setOverrides()` logic to a HOC to reduce code repetition :)

clarkepaul:
Some UI points:
Use CSS class .form--no-dividers on the forms to remove the dividing lines.
Can we have custom titles for each modal? 
* Link to an external URL
* Link to a page 
* Link to a file
* Link to an email address
Change checkbox text to "Open in new window/tab"

client/src/containers/InsertLinkModal/InsertLinkModal.js
flamerohr: note: I'd look at moving to HOC not part of this story, as this has gotten quite big already

flamerohr:
@clarkepaul done :)

code/Forms/EditorExternalLinkFormFactory.php
```diff
@@ -0,0 +1,82 @@
...
+            ),
+        ]);
+        
+        return $fields;
+    }
+    
+    protected function getFormActions($controller, $name, $context)
+    {
+        $actions = FieldList::create([
+            FormAction::create('insert', _t('CMSMain.INSERT_LINK', 'Insert link'))
```
tractorcow: Please use namespaced i18n keys.

_config.php
```diff
@@ -60,7 +60,10 @@
     ->enablePlugins(array(
         'contextmenu' => null,
         'image' => null,
-        'sslink' => FRAMEWORK_ADMIN_DIR . '/client/dist/js/TinyMCE_sslink.js'
-    ));
+        'sslink' => FRAMEWORK_ADMIN_DIR . '/client/dist/js/TinyMCE_sslink.js',
```
tractorcow: Can we make FRAMEWORK_ADMIN_DIR more risilient to being root, either by:

  - `ltrim($link, '/')` (probably just do this)
  - Not using FRAMEWORK_ADMIN_DIR at all (see https://github.com/silverstripe/silverstripe-asset-admin/commit/7bc13947054c08a892c410c3ba835011eeda8483#diff-8a7315557cd9f672e36f7e8f0ce2c29e for proof of concept).

code/Forms/EditorExternalLinkFormFactory.php
```diff
@@ -0,0 +1,85 @@
...
+        return $form;
+    }
+    
+    protected function getFormFields($controller, $name, $context)
+    {
+        $fields = FieldList::create([
+            TextField::create('Link', _t('SilverStripe\\Forms\\HTMLEditor\\HTMLEditorField.URL', 'URL'), 'http://'),
+            TextField::create(
+                'Description',
+                _t('SilverStripe\\Forms\\HTMLEditor\\HTMLEditorField.LINKDESCR', 'Link description')
```
tractorcow: How do you feel about removing cross-module keys and just using `__CLASS__` as standard? I'm sensing that the module-refactoring we did a while back may be too easily un-done otherwise.

flamerohr: I think this could lead to many duplication instances, but maybe I'm just paranoid.
I wouldn't be completely opposed to it.
e.g. I prefer `_t(LeftAndMain::class.'.something')` over `_t(self::class.'.something')` or `_t(__CLASS__.'.something')` for clarity.

_config.php
flamerohr: I've implemented the Self executing Closure method you linked

code/Forms/EditorExternalLinkFormFactory.php
tractorcow: The problem is that these strings only exist in this module, so they may not be collected.

tractorcow:
Will merge on green.

client/src/lib/TinyMCEActionRegistrar.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * Acts as top level action registrar component for tinymce
+ */
+class TinyMCEActionRegistrar {
```
chillu: @tractorcow @flamerohr We should decide if we call these things `registrar`, `register` or `registry` - we already have `ReactRouteRegister` and `ReducerRegister`. There's design pattern is called [registry](https://martinfowler.com/eaaCatalog/registry.html), described as:

> A well-known object that other objects can use to find common objects and services.

That seems to apply here?

chillu:
Great work! Just a niggle: There's no loading indicator when the form schema is loading. But rather than fixing that, I think we should work on caching those forms better to avoid the leader in the first place: https://github.com/silverstripe/silverstripe-admin/issues/71

Also, we're loading *five* new JS dist files for this stuff - can we please combine them into the the `asset-admin` and `silverstripe-admin` dist files? We should only load *one* dist file per SS module.

client/src/legacy/TinyMCE_sslink-external.js
```diff
@@ -0,0 +1,101 @@
...
+        this[0]
+      );
+    },
+
+    buildAttributes(data) {
+      const attributes = this._super(data);
+
+      let href = attributes.href;
+      // Prefix the URL with "http://" if no prefix is found
+      if (!href.match(/^(https?)?:\/\//)) {
```
chillu: What if you insert a link to another protocol? This is too narrow.

tractorcow:
> Also, we're loading five new JS dist files for this stuff - can we please combine them into the the asset-admin and silverstripe-admin dist files? We should only load one dist file per SS module.

Probably should, just need to make sure it can still be easily enabled / disabled even if it's combined by default.

client/src/legacy/TinyMCE_sslink-external.js
tractorcow: just change it to match '://` (detect if a protocol exists)

client/src/lib/TinyMCEActionRegistrar.js
flamerohr: Yeah, I guess Register would be the appropriate name for it

Sticking to a standard for this would be nice 👍

client/src/legacy/TinyMCE_sslink-external.js
flamerohr: yeah, that can be changed

flamerohr:
> Also, we're loading five new JS dist files for this stuff - can we please combine them into the the asset-admin and silverstripe-admin dist files? We should only load one dist file per SS module.

Yeah, I had thought about combining them...
My concern is if we're combining the plugin files, we're working against (which we already are) what tinymce does, we could end up losing the ability to conditionally add the item to the dropdown when needed due to race conditions between Registering the item into the menu and tinymce booting up.

It makes clear the separation of concern for this case, setting a clear example for other modules to add to the "Insert links" dropdown list.

The reason it's like this is because that's how tinymce's plugin system works, it would only load the link plugin if the plugin is listed in the config... Although, I think if one tinymce instance loads the plugin, all will include the plugin link due to the Register being global. So it's a kind of all-or-nothing for the plugin.

(sorry, this reply turned out longer than I thought it would)

client/src/legacy/TinyMCE_sslink-external.js
chillu: @flamerohr Do you have this on your own todo list, or do you want to create a new ticket?

chillu:
@flamerohr Please investigate this, it's a medium to high priority performance issue in my opinion: https://github.com/silverstripe/silverstripe-admin/issues/74

client/src/legacy/TinyMCE_sslink-external.js
flamerohr: @chillu it was on my "to create pull request for" list :)
https://github.com/silverstripe/silverstripe-admin/pull/75

PR #61, opened 2017-05-12
BUGFIX: Get rid of special dev webpack plugins

unclecheese:
They're not being used and they're breaking things, e.g. `ExtractTextPlugin`


tractorcow:
Unlike CMS, the admin module actually has a real vendor.js.

@flamerohr can you confirm this PR is ok?

flamerohr:
yeah, I think the intention of the `CommonsChunkPlugin` was good, but it made things illogical for logistics and trying to debug the issue, also I could see this causing trouble for dependency injection.

Travis didn't find any problems with a `diff` between dist files, so I think this PR is ok to merge.

PR #62, opened 2017-05-12
Fix update ss template FormActions

flamerohr:
To be closer to React's implementation

Fixes https://github.com/silverstripe/silverstripe-framework/issues/6439

Dependencies:
https://github.com/silverstripe/silverstripe-framework/pull/6907
https://github.com/silverstripe/silverstripe-cms/pull/1819

### DIFF
--- a/themes/cms-forms/templates/SilverStripe/Forms/FormAction.ss
+++ b/themes/cms-forms/templates/SilverStripe/Forms/FormAction.ss
@@ -1,6 +1,6 @@
 <% if $UseButtonTag %>
 	<button $getAttributesHTML('class') class="btn<% if $extraClass %> $extraClass<% end_if %>">
-		<% if $ButtonContent %>$ButtonContent<% else %>$Title.XML<% end_if %>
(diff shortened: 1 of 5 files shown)


themes/cms-forms/templates/SilverStripe/Forms/FormAction.ss
```diff
@@ -1,6 +1,6 @@
 <% if $UseButtonTag %>
 	<button $getAttributesHTML('class') class="btn<% if $extraClass %> $extraClass<% end_if %>">
-		<% if $ButtonContent %>$ButtonContent<% else %>$Title.XML<% end_if %>
+		<span><% if $ButtonContent %>$ButtonContent<% else %>$Title.XML<% end_if %></span>
```
tractorcow: See comment on https://github.com/silverstripe/silverstripe-framework/pull/6907#discussion_r116392449

If we set html directly via $ButtonContent, then we shouldn't wrap that in a span.

PR #64, opened 2017-05-12
BUGFIX: Remove duplicate ID

mikenz:
The ID gets applied to the <input> having it on the <div> results in a duplicate ID on the page


dhensby:
thanks

PR #67, opened 2017-05-15
Feature to adding root as an available option and cache invalidation

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/217

Adds the feature to select root as an option in the `TreeDropdownField`
Also invalidates cache after changing location (through form schema)

Related pull requests
https://github.com/silverstripe/silverstripe-asset-admin/pull/476
https://github.com/silverstripe/silverstripe-framework/pull/6915

### DIFF
--- a/client/src/components/TreeDropdownField/TreeDropdownField.js
+++ b/client/src/components/TreeDropdownField/TreeDropdownField.js
@@ -122,8 +122,8 @@ class TreeDropdownField extends Component {
       let selectedOption = options.find((option) => (option.id === value));
       if (!selectedOption) {
         selectedOption = this.getSelectedOption();
-        options.unshift(selectedOption);
       }
+      options.unshift(selectedOption);
     }
 
     if (this.props.data.hasEmptyDefault && !this.props.visible.length) {
@@ -520,7 +520,7 @@ TreeDropdownField.propTypes = {
   loading: PropTypes.array, // List of nodes marked as loading
   failed: PropTypes.array, // List of nodes that failed to load
   data: PropTypes.shape({
-    cacheKey: PropTypes.string.isRequired,
+    cacheKey: PropTypes.string,
     urlTree: PropTypes.string.isRequired,
     emptyString: PropTypes.string,
     valueObject: PropTypes.shape({
--- a/client/src/components/TreeDropdownField/TreeDropdownField.js
+++ b/client/src/components/TreeDropdownField/TreeDropdownField.js
@@ -104,20 +104,27 @@ class TreeDropdownField extends Component {
    *
    * @return {Array}
    */
-  getDropdownOptions() {
+  getDropdownOptions(value) {
     // force renderMenu() to handle rendering even if options are empty
     const node = this.getVisibleTree();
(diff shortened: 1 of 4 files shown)


tractorcow:
What does the root option look like when you're viewing the dropdown? Screenshot?

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -405,10 +423,15 @@ class TreeDropdownField extends Component {
    */
   renderMenu(renderMenuOptions) {
     // Build root node
-    const tree = this.getVisibleTree() || {};
+    const visibleTree = this.getVisibleTree();
+    const tree = Object.assign({}, visibleTree, {
+      // we only want to show options with a title
```
tractorcow: Where is this logic coming from? Should a file with no title be hidden? A record can theoretically have an empty title.

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -405,10 +423,15 @@ class TreeDropdownField extends Component {
    */
   renderMenu(renderMenuOptions) {
     // Build root node
-    const tree = this.getVisibleTree() || {};
+    const visibleTree = this.getVisibleTree();
```
tractorcow: If you have `showRootOption = true`, does the root appear in this tree?

client/src/components/TreeDropdownField/TreeDropdownField.js
flamerohr: An empty title is fine and that wouldn't get filtered out, is there a case when a File would have a title of `null` instead of an empty string?

P.S. Please comment on the correct line you're referring to, I couldn't see the context of your question

client/src/components/TreeDropdownField/TreeDropdownField.js
flamerohr: are you talking about this line? https://github.com/silverstripe/silverstripe-admin/pull/67/files#diff-3eeee0f0f28cfe1453d1922e6333dc5fR129

I'm not sure I understand why the comment was placed here
root appears when `showRootOption` is true and you're on the root directory.

tractorcow: Oh, I was wondering if you wanted the visibleTree to be one level deeper (i.e. another dummy level for the root object) but I realised I was mis-understanding the point of the api change. :)

client/src/components/TreeDropdownField/TreeDropdownField.js
tractorcow: Sorry, it was the line down. :) I'll do that in future.

flamerohr:
A screenshot of the implementation
screen shot 2017-05-15 at 7 39 32 pm

tractorcow:
Is (root) visible when viewing the second level, or only on the first?

flamerohr:
only on the first

tractorcow:
ok, sounds good. ;)

PR #70, opened 2017-05-16
WIP: Pulls/4/react dependency injection

unclecheese:
~~Don't merge. API in active flux and discussion.~~


unclecheese:
## Issues, limitations, and open questions

Based on my observations so far...

### Only one component gets to influence `render()`

While the props shape can represent the aggregate influence of all customisations of the component, `render()` doesn't quite work that way. See [experiment 2](https://github.com/open-sausages/experimental-react-di/blob/master/js/src/TextFieldCharacterCounterCreator.js) and [experiment 3](https://github.com/open-sausages/experimental-react-di/blob/master/js/src/TextFieldPasswordStrengthCheckerCreator.js) for how this becomes a problem. If the password strength checker component wanted to render the alert message as UI, it won't be able to rely on the core `TextField` UI. That UI may or may not have already been updated by `TextFieldCharacterCountCreator`.

### Order of execution / priority
Everything in the DI layer will suffer from this problem, but it's something to keep in mind. Should we keep this to a simple `priority` parameter to `customise()`? That might invite an arms race between module developers, e.g. `{priority: 999999}` etc..

### The need for context
Right now, all customisations are global. Every single `FormBuilder` rendered text field in the CMS now has a character count. This is not good enough. We need an API that allows the both the consumer of the injected component to specify context, as well as the provider of the injected component. Rough API:
```js
this.context.injector.get('TextField.CampaignAdmin.EditForm.Title');

Injector.customise('TextField.CampaignAdmin.EditForm.Title', NewComponentCreator);
```

Ideally, we could make this more flexible with wildcarding:
```js
Injector.customise('TextField.CampaignAdmin.*') // All text fields in CampaignAdmin

Injector.customise('TextField.AssetAdmin.SearchForm.*') // All text fields in AssetAdmin search

Injector.customise('DropdownField.*') // I don't like Chosen.
```

### propTypes collisions

If I want the title of my `GalleryItem` to be an inline image, I'm hosed right now, as the `title` propType is declared upstream as `string`. We need an API for influencing that and resolving conflicts. This becomes a bit more problematic with things like `oneOfType()` and `shape()`.

### We should throw when `register()` is called more than once for the same service
`register()` is used to declare the service for the first time, and defines the service that will be fed to the front of the middleware chain (`customise()`). It is conceivable that you would want something so specific that you just want to override the entire service. This would conceivably nuke all the existing middleware definitions, or at least put them at risk of failure. Possible solution:

```
Injector.register('TextField', TextField); // OK

Injector.register('TextField', AnotherTextField);

Injector.register('TextField', AnotherTextField, { force: true }); // OK
```

### Components are far too monolithic to make DI worthwhile

Without the opportunity to inject on small components, e.g. `<GalleryItemTitle />`, we're inviting injection by subclass, which we really want to discourage.

Future story: Break up large components into smaller, more agnostic pieces.

client/src/boot/registerComponents.js
```diff
@@ -0,0 +1,50 @@
...
+  Injector.register('HtmlReadonlyField', HtmlReadonlyField);
+  Injector.register('LookupField', LookupField);
+  Injector.register('CompositeField', CompositeField);
+  Injector.register('Tabs', Tabs);
+  Injector.register('TabItem', TabItem);
+  Injector.register('FormAction', FormAction);
+  Injector.register('LabelField', LabelField);
+  Injector.register('TreeDropdownField', TreeDropdownField);
+
+  Injector.freeze();
```
flamerohr: Do we want to freeze registering more?

`asset-admin` injects a few more fields as well as potentially other modules.
Or perhaps only freeze protect `customize`

client/src/lib/dependency-injection/withInjector.js
```diff
@@ -0,0 +1,55 @@
...
+            props[dependencies[i]] = resolved[i];
+          }
+        }
+      }
+
+      return React.createElement(
+        Component,
+        Object.assign({}, this.props, props),
+        this.props.children
+      );
```
flamerohr: Is `React.createElement()` preferred over `<Component {...this.props} />`?

flamerohr:
@unclecheese I'm curious what went wrong with both experiments?
experiment 3 seems like all sorts of no with the `setTimeout()` and `alert()` lol

Context would be a tricky one, even more so if you throw in wildcards. Most forms are named specifically and don't know their own context except that it could be inferred by the field's `id` property.

PropTypes could be opened up to be slightly more forgiving, e.g. display types like `title` could be handled by the `castStringToElement()` API that we have, which casts strings to `<Element>{string}</Element>` or if it's contained in an object it will be casted as such:
`{ react: ReactComponent }` as a react component
`{ html: HtmlString }` as unescaped html
I forget the other allowed types, but this could be used a bit more to our advantage.
I wouldn't recommend some data PropTypes to be changed though, such as count properties.

I agree with the throwing of an already registered service, I don't agree with a force flag to allow ourselves or modules to override it as such and would be more keen to opt for a `customise()` call to handle overriding it (this is assuming that priority has been solved somehow first)

Totally agree with breaking up these monolithic components to smaller ones, I'm also keen to start moving presentational components to stateless functional components and move away from es6 classes for them.
If possible, when breaking up these components, work on opening up the PropTypes as mentioned above to be more forgiving when display certain types of data.

client/src/containers/App/App.js
```diff
@@ -8,9 +9,10 @@ import SilverStripeComponent from 'lib/SilverStripeComponent';
 class App extends SilverStripeComponent {
   render() {
     // TODO re-add <div className="app"> wrapper when applying to document.body
-    const Child = React.Children.only(this.props.children);
-    return (Child);
+    return <div>{this.props.children}</div>;
```
unclecheese: Yeah, I haven't seen any issues yet, but I have noted it as a temporary hack that needs to be reviewed before merge.

client/src/boot/registerComponents.js
unclecheese: So the idea here is that we would discourage every module but `silverstripe-admin` from making Injector mutations on document load. Admin is the only module that gets that privilege, ensuring it comes last. The other way we could do it, if we wanted to be a bit more permissive is to do it on document load, wrapped in a `setTimeout(.., 0);` to force it to a new thread.

Neither of those are ideal. We need some kind of contract between modules to ensure `silverstripe-admin` comes last, though.

client/src/lib/dependency-injection/withInjector.js
unclecheese: Either is fine. Sometimes I find `React.createElement` more expressive, but we should be consistent and use JSX.

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -202,6 +202,42 @@ class FormBuilder extends SilverStripeComponent {
...
 
   /**
+   * Default data type to component mappings.
+   * Used as a fallback when no component type is provided in the form schema.
+   *
+   * @param string dataType - The data type provided by the form schema.
+   * @return object|null
+   */
+  getComponentForDataType(dataType) {
+    const { injector: { get } } = this.context;
```
chillu: Yay, great way to solve this (`withInjector` context HoC)

client/src/lib/dependency-injection/container.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Register a dependency. This is the initial version of a depenency that will be
+ * passed to the first link in the middleware chain (if any customisations exist)
+ * @param key The name of the dependency to register
+ * @param Component The component to register
+ */
+const register = (key, value) => {
+  di.factory(
+    key,
+    () => compose(...(middlewares[key] || []))(value)
```
chillu: Such ES6, so compose! ❤️

unclecheese: :) I got a bit cheeky here.

unclecheese:
> I'm curious what went wrong with both experiments?
experiment 3 seems like all sorts of no with the setTimeout() and alert() lol

Yeah, it's not meant to be anything you'd use in the real world. Just a proof of concept. `setTimeout()` is used to throttle the `onChange` event (so it doesn't trigger every keystroke), and the use of `alert()` was alluded to in the issue I described with `render()`. Only one customisation really gets to influence the UI. My password checker _could_ render that message as UI (e.g. an error message below the field), but I have no guarantee that `TextField` at that point is the standard core TextField. It could have been (and is, in that experiment) already customised to render new UI.

> I agree with the throwing of an already registered service, I don't agree with a force flag to allow ourselves or modules to override it as such and would be more keen to opt for a customise() call to handle overriding it (this is assuming that priority has been solved somehow first)

I tend to agree. The only case I see for having a `force` parameter is that it's consistent with the Injector pattern we have in the backend, which is "Whomever gets there last gets to have their implementation." If five modules inject on `Mailer`, four of those are completely irrelevant to the execution chain.

>Another note about priority, what do you thinking about having a "name" param for the customize() API, which can help identify the customisations a bit more succinctly, the idea is from the way SS handle config.yml files.

Love this idea. We could use it on the `displayName` property to make the React console a lot more informative, as well.

unclecheese:
I've added the note about ensuring `silverstripe-admin` authority to the OP.

client/src/lib/dependency-injection/withInjector.js
```diff
@@ -0,0 +1,55 @@
...
+      let props = {};
+      if (dependencies) {
+        if (!Array.isArray(dependencies)) {
+          throw new Error(`
+            withInjector() passed an argument for dependencies that is ${typeof dependencies}. 
+            Must be an array of named dependencies.
+          `);
+          return;
+        }
+        const resolved = dependencies.map(this.context.injector.get);
```
chillu: So the constructed container could declare dependencies as propTypes, and this HoC resolves them automatically. Does that mean most container components and some presentational components would be wrapped this way in their default export? I assume that doesn't interfere with the ability to further wrap these exported components with `customise()`?

chillu:
> If I want the title of my GalleryItem to be an inline image, I'm hosed right now, as the title propType is declared upstream as string. We need an API for influencing that and resolving conflicts. This becomes a bit more problematic with things like oneOfType() and shape().

I think part of the answer here is making components more granular (such as a `GalleryItemTitle`), and allowing to replace them completely (or HoC wrap them). PropTypes are our API, and the typing should be specific enough to ensure the component can operate with the input variations. If it expects `title` as a `string`, that's because it's `render()` function will expect a string. If you want it to be a component, that's changing some fairly fundamental assumptions of the `render()` function (and potentially other logic in the component). I'm not a fan of Chris' suggestion on putting casting info in the propType shape (if I understand this correctly). 

> Order of execution / priority

Could we just sketch this out for now, and leave implementation to a follow-up card? I'd be keen to get clarity if we go down `priority` vs `before/after` as part of this cart, and making sure that we don't paint ourselves in a corner. If the consensus is `before/after`, then we need a JS constraint resolver library, which sounds like scope creep for this card.

> The need for context

Service names in DI container are usually just strings. In SS4, we are using suffixes as well (`<interface-fqcn>.<arbitrary-suffix>`). So this would be a consistent approach - just that we need multiple layers of suffixes here. Wildcarding makes the system a bit more complex, but as long as we don't use regexes or something it feels like this is an appropriate level of flexibility. In particular for forms, we need context specific behaviour based on various criteria. My only worry is that it'll become quite opaque which implementation is chosen - can you think about diagnostics and debugging for this please?

`FormBuilder` and `FormBuilderLoader` don't know anything about the current section name, they only care about the schema URL. I guess the schema could contain this metadata? Ideally we keep section names consistent with their FQCN reference as well, which would look ugly in service name references. Outside of form builder, would you see e.g. `<GalleryTile>` knowing about section names? I want to avoid components knowing too much. Outside of `FormBuilder`, is it important to know in which section context a component is applied?

client/src/lib/dependency-injection/withInjector.js
unclecheese: Yes. The idea is that almost everything should be wrapped in `withInjector()`, particularly containers, as they by definition have dependencies. Components that shouldn't would be simple presentational components that render only HTML. e.g. `const Title = (props) => <h2>{props.children}</h2>`;

There's nothing to my knowledge that interferes with an injector-consuming component being customised.

client/src/lib/dependency-injection/container.js
```diff
@@ -0,0 +1,209 @@
...
+    (key, wrapper) => customise(meta, key, wrapper)
+  );
+};
+
+/**
+ * Resolve all of the middleware constraints and freeze the DI layer
+ */
+const load = function load() {
+  for(const key in middlewares) {
+    if(middlewares.hasOwnProperty(key)) {
```
flamerohr: how about using `Object.keys(middlewares).forEach(key => {`?
I'd normally reserve `for...in` and `for...of` for cases where breaking is beneficial.

client/src/lib/dependency-injection/withInjector.js
```diff
@@ -0,0 +1,53 @@
...
+  class Injectable extends React.Component {
+    render() {
+      let props = {};
+      if (dependencies) {
+        if (!Array.isArray(dependencies)) {
+          throw new Error(`
+            withInjector() passed an argument for dependencies that is ${typeof dependencies}. 
+            Must be an array of named dependencies.
+          `);
+          return;
```
flamerohr: this `return` is unreachable.
it's also an inconsistent return value it should be `return null;` if it were still around ;)

client/src/lib/dependency-injection/withInjector.js
```diff
@@ -0,0 +1,53 @@
...
+            props[dependencies[i]] = resolved[i];
+          }
+        }
+      }
+      const newProps = Object.assign({}, this.props, props);
+      return <Component {...newProps} />;
+    }
+  }
+
+  Component.contextTypes = injectorContext;
```
flamerohr: I have a feeling you may be overwriting other existing contextTypes for the injected Component, would be safer to have `Component.contextTypes = Object.assign({}, Component.contextTypes, injectorContext);`

but also, this will fail ESLint. As above add the `// eslint-disable-next-line rule-here` comment

flamerohr:
> I'm not a fan of Chris' suggestion on putting casting info in the propType shape (if I understand this correctly).

Not quite how it works, for reference it's done with fieldHolder's title fields (leftTitle, rightTitle, Description) and allows the PHP to define html (or HOC to define react) for these fields if required rather than string. It does mean the propType would need to open up and allow for it though, yes...
I think it might be worth creating our own "displayText" prop type, so it can be reused, but that's outside the scope of this story :D

client/src/lib/dependency-injection/container.js
```diff
@@ -0,0 +1,235 @@
...
+  return container[key];
+};
+
+/**
+ * Updates the injector by callback. First parameter should contain
+ * an object with keys for name, and (optional) "before" and "after" declarations
+ * e.x.
+ * Injector.update(
+ *   {
+ *     name: 'my-module',
```
chillu: I think this should be called `module` rather than `name`, with `beforeModule` and `afterModule` counterparts. It's a bit confusing to use `name` here, I first thought it was the component name.

client/src/lib/dependency-injection/container.js
```diff
@@ -0,0 +1,235 @@
...
+const validateMeta = (meta) => {
+  if (typeof meta.name === undefined) {
+    throw new Error('Injector.update() called with no "name" property specified.');
+  }
+  PRIORITIES.forEach(k => {
+    if (
+      meta[k] !== undefined &&
+      (typeof meta[k] !== 'string' && !Array.isArray(meta[k]))
+    ) {
+      throw new Error(`Injector.update() key ${k} must be a string or array`);
```
chillu: Thanks for doing these descriptive error messages, that really helps!

client/src/lib/dependency-injection/container.js
```diff
@@ -0,0 +1,235 @@
...
+ * Applies a topological sort to the middlewares and resolves
+ * priority declarations
+ * @param middlewareList
+ * @returns {Array}
+ */
+const sortMiddlewares = (middlewareList) => {
+  const graph = [];
+  let sortedMiddlewares = [];
+
+  middlewareList.forEach(entry => {
```
chillu: So if we apply a HoC in `silverstripe-admin/` on a "base component" (can't think of a use case for this at the moment...), does the `*` allow modules to apply their customisations *before*?

client/src/lib/dependency-injection/container.js
```diff
@@ -0,0 +1,235 @@
...
+  get,
+  load,
+  update: protect(update),
+  register: protect(register),
+};
+if (process.env.NODE_ENV !== 'production') {
+  ContainerAPI.__reset__ = reset;
+}
+
+export default ContainerAPI;
```
chillu: In AirBnB standards, ES6 module exports should match the file name.

client/src/lib/dependency-injection/provideInjector.js
```diff
@@ -0,0 +1,29 @@
...
+  InjectorProvider.childContextTypes = {
+    injector: React.PropTypes.shape({
+      get: React.PropTypes.func,
+    }),
+  };
+
+  return InjectorProvider;
+}
+
+export {
```
chillu: Why don't you make this the default export?

client/src/lib/dependency-injection/withInjector.js
```diff
@@ -0,0 +1,55 @@
...
+  };
+  Injectable.contextTypes = injectorContext;
+  Injectable.displayName = `withInjector(
+    ${(Component.displayName || Component.name || 'Component')}
+  )`;
+
+  return Injectable;
+}
+
+export {
```
chillu: Same here, should this be the default export?

client/src/lib/dependency-injection/container.js
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * Updates the injector by callback. First parameter should contain
+ * an object with keys for name, and (optional) "before" and "after" declarations
+ * e.x.
+ * Injector.update(
+ *   {
+ *     name: 'my-module',
+ *     before: ['another-module']
+ *   },
+ *   wrap => {
```
chillu: That's a pretty weird syntax - particularly if you present it to PHP developers who might misread this as an array (like I just did). It's also mixing terminologies between "update", "callback" and "wrap".

I've looked at this for a few minutes, and have to honestly say that I can't figure out what gets executed when and how without actually stepping through a debugger. Can we make the API surface of this simpler?

chillu: We ended up with this syntax:

```
// module-a.js
Injector.transform('module-a.fields', update => {
    update('TextField', ComponentA);
    update('DateField', ComponentB);
});
Injector.transform('module-a.core', update => {
    update('Menu', ComponentC);
}, {beforeModule: '*'});


// module-b.js
Injector.transform('module-b', update => {
    update('TextField', ComponentD);
}, {beforeModule: 'module-a'});
```

client/src/lib/dependency-injection/container.js
unclecheese: You can't do anything in front of the base component. It's stored outside of the middleware chain completely. `before: '*'` will be executed as the first customisation _of the base component_ but not before.

client/src/lib/dependency-injection/container.js
unclecheese: Per our discussion on Slack, `name` is now the first parameter of `Injector.transform()`.

client/src/lib/dependency-injection/provideInjector.js
unclecheese: Yeah these should be default exports. They're named exports in the main `Injector` file, and i think this is just legacy.

client/src/lib/dependency-injection/container.js
chillu:
```
module-a/Component.js
module-a/HoC-a.js (applies to Component)
module-b/HoC-b.js (applies to Component)
```

Can I get `HoC-b` to apply before `HoC-a`?

flamerohr: Assuming you're using injector for both cases, you'd put `{before: 'HOC-a-group-name'}` :)

unclecheese: Yeah, that would be 
```
Injector.transform('module-b', update => {
  update('Component', HOC);
}, { before: 'module-a' });
```

chillu:
@unclecheese build is failing on NPM_BUILD:

> modulesModuleNotFoundError: Module not found: Error: Cannot resolve 'file' or 'directory' ./dependency-injection/Container in /home/travis/build/silverstripe/silverstripe-admin/client/src/lib

client/src/lib/dependency-injection/inject.js
```diff
@@ -0,0 +1,51 @@
...
+        ...this.props,
+        ...props,
+      };
+      return <Component {...newProps} />;
+    }
+  }
+  Injected.contextTypes = injectorContext;
+  Injected.displayName = `inject(
+    ${(Component.displayName || Component.name || 'Component')}
+  )`;
```
flamerohr: This is really nice to have as a default

chillu:
This caused a regression which was fixed with https://github.com/silverstripe/silverstripe-admin/pull/116

PR #76, opened 2017-05-18
Add unsaved changes message

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/6683


client/src/boot/BootRoutes.js
```diff
@@ -92,6 +94,22 @@ class BootRoutes {
       </ApolloProvider>,
       document.getElementsByClassName('cms-content')[0]
     );
+
+    history.listenBeforeUnload(() => {
+      if (this.shouldConfirmBeforeUnload()) {
+        return i18n._t('Admin.CONFIRMUNSAVEDSHORT');
```
tractorcow: Can you please use the default string as well?

`return i18n._t('Admin.CONFIRMUNSAVEDSHORT', 'WARNING: Your changes have not been saved.');`

client/src/boot/BootRoutes.js
```diff
@@ -92,6 +94,22 @@ class BootRoutes {
...
+      }
+
+      return null;
+    });
+
+    history.listenBefore(() => {
+      if (this.shouldConfirmBeforeRoute()) {
```
tractorcow: This code is duplicated from above; Can you put this into a common method?

client/src/boot/BootRoutes.js
```diff
@@ -102,9 +120,17 @@ class BootRoutes {
     const store = this.store;
 
     pageRouter('*', (ctx, next) => {
-      // eslint-disable-next-line no-param-reassign
-      ctx.store = store;
-      next();
+      if (this.shouldConfirmBeforeRoute()) {
+        if (window.confirm(i18n._t('Admin.CONFIRMUNSAVEDSHORT'))) {
```
tractorcow: If you re-shuffle your logic, you can combine these blocks.

```
if (!this.shouldConfirmBeforeRoute() || window.confirm(i18n._t('Admin.CONFIRMUNSAVEDSHORT')) {
	ctx.store = store;
	next();
} else {
	// whatever?
}
```

client/src/boot/BootRoutes.js
```diff
@@ -140,8 +166,53 @@ class BootRoutes {
       });
     });
 
+    window.onbeforeunload = () => {
+      if (this.shouldConfirmBeforeUnload()) {
+        return i18n._t('Admin.CONFIRMUNSAVEDSHORT');
```
tractorcow: Third time this code is duplicated in this file. :)

tractorcow:
Ok, this is weird... I can't even click links in the CMS anymore. I think you need another next() in your page.js callback?

tractorcow:
image

tractorcow:
Core change detection has broken in non-react sections.

- Go to pages section
- Edit any page
- Modify the title
- click "assets" section
- No change detection alert, you navigate immediately

I was expecting to see this alert:

image

tractorcow:
This is the code which handles this (in LeftAndMain.EditForm.js)

```js
// Can't bind this through jQuery
window.onbeforeunload = function(e) {
  var form = $('.cms-edit-form');
  form.trigger('beforesubmitform');
  if(form.is('.changed') && ! form.is('.discardchanges')) {
    return i18n._t('Admin.CONFIRMUNSAVEDSHORT');
  }
};
```

client/src/boot/BootRoutes.js
```diff
@@ -140,8 +156,64 @@ class BootRoutes {
       });
     });
 
+    window.onbeforeunload = this.handleBeforeUnload;
```
tractorcow: I think this is overwriting the `window.onbeforeunload` assigned via LeftAndMain.EditForm.js. See my comment.

tractorcow:
Some more feedback:

When navigating in a react section I get this dialog:

image

When I navigate (hook based on window.onbeforeunload) I get this error.

image


Can we consolidate the messages to be the same?

When we use `window.confirm` we need to add an extra line to disclose the effects of "cancel" and "ok". note that this doesn't need to be exactly the same as the chrome default message (since that's added by the browser, and we have no control over it) but it needs to meet the same need.

tractorcow:
Oh, it's already there.

use `Admin.CONFIRMUNSAVED` instead of `Admin.CONFIRMUNSAVEDSHORT`. Done :)

I'll push it up and see how it looks.

client/src/boot/BootRoutes.js
```diff
@@ -140,8 +158,79 @@ class BootRoutes {
...
+   * with unsaved changes. Return `false` otherwise.
+   *
+   * @return {Boolean}
+   */
+  shouldConfirmBeforeRoute() {
+    const state = this.store.getState();
+    const unsaveds = state.unsavedForms || {};
+    const locationBeforeTransitions = state.routing.locationBeforeTransitions;
+    const pathname = locationBeforeTransitions && locationBeforeTransitions.pathname;
+    let ret = false;
```
flamerohr: @phalkunz for future reference, a meaningful variable name would be helpful, e.g. `shouldConfirm`

client/src/boot/BootRoutes.js
```diff
@@ -140,8 +158,79 @@ class BootRoutes {
       });
     });
 
+    // Legacy support – make sure we don't overwrite the `onbeforeunload`
+    // handler in `LeftAndMain.EditForm.js` but run it after the new handler.
+    const currBeforeUnload = window.onbeforeunload;
+    window.onbeforeunload = () => {
+      if (this.shouldConfirmBeforeUnload()) {
```
flamerohr: a bit of code duplication here: e.g. `let message = this.handleBeforeUnload(); if (message) { return message; }`

client/src/boot/BootRoutes.js
phalkunz: 👍

client/src/boot/BootRoutes.js
```diff
@@ -140,8 +158,79 @@ class BootRoutes {
...
+
+  /**
+   * Return `true` if there are forms with unsaved changes. Return `false`
+   * otherwise.
+   *
+   * @return {Boolean}
+   */
+  shouldConfirmBeforeUnload() {
+    const state = this.store.getState();
+    const unsaveds = state.unsavedForms || {};
```
tractorcow: An oversight here is that not every unsaved form should block navigation if un-submitted; E.g. search forms.

PR #82, opened 2017-05-21
Upgrade to webpack2 and cleanup config file

flamerohr:
The webpack config file needed a cleanup, so I tried out upgrading to webpack2 while I was at it...
running `yarn watch` or `npm run watch` gets ~2sec build times after intial build, I've been exploring ways to improve this. Separating js and css streams helped improve it.

Fixes https://github.com/silverstripe/silverstripe-framework/issues/6478

Notable changes:
- Utilise babel's native imports support, which allows enables some bytes saved through the tree shaking feature.
- Now uses `-p` flag for building to production, I had found occasional false-negatives during rebases which is not good, and it was only caught by Travis' build diff comparison test.
- Removed the use of UglifyJS, the `-p` flag does most of everything we want except for removing all comments.


webpack/js/modules.js
```diff
@@ -0,0 +1,47 @@
...
+    rules: [
+      eslintLoader,
+      {
+        // .js and .jsx files are caught
+        test: /\.jsx?$/,
+        exclude: /(node_modules|thirdparty)/,
+        loader: 'babel-loader',
+        options: {
+          presets: [
+            ['es2015', { modules: false }],
```
unclecheese: Should this be `es2015-webpack` now? https://gist.github.com/sokra/27b24881210b56bbaff7#babel-and-webpack

webpack/js/plugins.js
```diff
@@ -0,0 +1,34 @@
...
+    }),
+    // Most vendor libs are loaded directly into the 'vendor' bundle (through require()
+    // calls in vendor.js). This ensures that any further require() calls in other
+    // bundles aren't duplicating libs.
+    new webpack.optimize.CommonsChunkPlugin({
+      name: 'vendor',
+      minChunks: module => module.context && module.context.indexOf('/node_modules/') > -1,
+    }),
+    uglifyPlugin,
+  ].filter(plugin => plugin);
```
unclecheese: What does this do? In what scenario is `plugin` falsey?

webpack.config.js
```diff
@@ -1,207 +1,57 @@
...
-      new webpack.optimize.CommonsChunkPlugin({
-        name: 'vendor',
-        minChunks: Infinity,
-      }),
-    ],
+    devtool: (ENV !== 'production') ? 'source-map' : '',
+    resolve: require('./webpack/js/resolve')(ENV, PATHS),
+    externals: require('./webpack/js/externals')(ENV, PATHS),
+    module: require('./webpack/js/modules')(ENV, PATHS),
+    plugins: require('./webpack/js/plugins')(ENV, PATHS),
```
flamerohr: yup, fair call, I'll look at doing that
this was something I had implemented at the start and hadn't revisited since

webpack/js/modules.js
flamerohr: interesting, I hadn't run into that during my research for upgrading...
I had only found the `{ modules: false }` config, which stops babel from transforming all imports/exports to commonjs
I'll look into this a bit more, thanks!

webpack/js/plugins.js
flamerohr: `uglifyPlugin` is null in non-production builds, uglifyPlugin is only passed in to remove all comments from the `dist` build file.

webpack/js/modules.js
flamerohr: Ah hah, I had used `es2015-webpack` actually, but then found that the later versions of babel supports it with what I have currently :D
Check out the readme for https://github.com/gajus/babel-preset-es2015-webpack

flamerohr:
@unclecheese I've rebased and moved the require calls to the top of the file

tractorcow:
Tests are broken. ;)

flamerohr:
Rebased and fixed tests breaking

PR #85, opened 2017-05-25
Add function for formatting numbers in written form

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/6685


client/lang/en.js
```diff
@@ -29,6 +29,16 @@ if (typeof(ss) === 'undefined' || typeof(ss.i18n) === 'undefined') {
     "Admin.LINKLABEL_EXTERNALURL": "Link to external URL",
     "Admin.LINKLABEL_EMAIL": "Link to email address",
     "Admin.LINK_EXTERNAL": "Insert external link",
-    "Admin.LINK_EMAIL": "Insert email link"
+    "Admin.LINK_EMAIL": "Insert email link",
+    "Admin.WRITTEN_NUMBER_ZERO": "zero",
```
phalkunz: Not quite sure what you mean. Is there an example I can see.

client/src/styles/_layout.scss
```diff
@@ -113,3 +113,15 @@ body {
...
 }
+
+
+/**
+ * Workaround for tooltip placement. React bootstrap add class `left`, `right`,
+ * `top` and `bottom` instead of `tooltip-left`, `tooltip-right`, `tooltip-top`
+ * and `tooltip-bottom`.
+ */
+@each $dir in [left, right, top, bottom] {
+  .tooltip.#{$dir} {
```
phalkunz: But there isn't a tooltip component? Will it be just scss file in the folder?

client/lang/en.js
flamerohr: So you'll find a list of translations (with code surrounding this list) named `client/lang/en.js`
You'll also find another file named `client/lang/src/en.json` which contains the same list of translations (but without the code surrounding the list)

client/src/styles/_layout.scss
flamerohr: Yeah, that's fine :)
We had something similar with `Label/Label.scss` (actually, that could be cleaned up to be `LabelField/LabelField.scss` later on)
And I think it's better practice to have component specific styles handled like a component specific style, even if we don't have a javascript component for it.

phalkunz: 👍

client/lang/en.js
phalkunz: I see. Thanks :)

PR #86, opened 2017-05-26
TinyMCE external link, fix hash only links

mikenz:


dhensby:
Looks like the distance files need building

PR #94, opened 2017-05-31
Code cleanup and separate schemaStateOverride logic to HOC

flamerohr:
See https://github.com/silverstripe/silverstripe-framework/issues/6743 for more details

Setting field defaults for insert modals belonged in a HOC, so there's less copy+paste code and higher re-usability


client/src/bundles/bundle.js
```diff
@@ -83,4 +82,4 @@ require('../legacy/HtmlEditorField.js');
 require('../legacy/TabSet.js');
 require('../legacy/GridField.js');
 
-require('../boot/index.js');
+require('boot');
```
flamerohr: it uses the module path in webpack, rather than a relative path
and `index.js` is unnecessary, because require checks for a index.js when a directory is specified.

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Compares the current properties with received properties and determines if overrides need to be
+   * cleared or added.
+   *
+   * @param {object} props
+   */
+  setOverrides(props = null) {
+    if (!props) {
+      // clear any overrides that may be in place
+      const schemaUrl = props && props.schemaUrl || this.props.schemaUrl;
```
flamerohr: yup, you're right :)

client/src/bundles/bundle.js
phalkunz: Thanks :)

PR #96, opened 2017-06-02
Fix print button in reports

flamerohr:
Fixes the print button to open in a new window

Relies on: https://github.com/silverstripe/silverstripe-framework/pull/6988 to stop the crashing


code/Forms/GridFieldPrintButtonExtension.php
```diff
@@ -0,0 +1,13 @@
...
+namespace SilverStripe\Admin\Forms;
+
+use SilverStripe\Core\Extension;
+use SilverStripe\View\Requirements;
+
+class GridFieldPrintButtonExtension extends Extension
+{
+    public function updatePrintData($data)
+    {
+        Requirements::css(ltrim(FRAMEWORK_DIR . '/admin/client/dist/styles/GridField_print.css', '/'));
```
tractorcow: This was moved to admin, it's not in framework anymore.

tractorcow:
Pushed up a fix.

PR #97, opened 2017-06-02
ENHANCEMENT Update style of CMSLogin form

tractorcow:
Admin component for https://github.com/silverstripe/silverstripe-framework/issues/6387


client/src/legacy/LeftAndMain.js
```diff
@@ -959,78 +962,66 @@ $.entwine('ss', function($) {
...
+      // Build properties
+      const tempid = $('body').data('member-tempid');
+      const url = $.path.addSearchParams('CMSSecurity/login', {
+        tempid,
+        BackURL: window.location.href,
+      });
+
+      // todo add URL to iframe
+      ReactDOM.render(
+        <LoginDialog className="login-dialog" bodyClassName="login-dialog__body"
```
flamerohr: could we have these classes added by the component? and concatenate any other respective classNames.

Also, please apply each attribute to a new line for readability purposes

client/src/bundles/bundle.js
```diff
@@ -33,6 +33,7 @@ require('expose-loader?ListGroupItem!components/ListGroup/ListGroupItem');
 require('expose-loader?TextField!components/TextField/TextField');
 require('expose-loader?LiteralField!components/LiteralField/LiteralField');
 require('expose-loader?Toolbar!components/Toolbar/Toolbar');
+require('expose-loader?LoginDialog!components/LoginDialog/LoginDialog');
```
flamerohr: does this need exposing currently?

flamerohr:
Additional feedback @tractorcow :)
I've found a `onHide={this.handleHide}` but `handleHide` is not defined, this has been shifted to `onHide={this.props.onHide}` so it continues to be extensible with react-bootstrap's modal.

I've shifted LoginDialog to a generic IframeDialog component, we can reuse this for other iframe modals (at least have the option for using it)

Issue #100, opened 2017-06-05
Remove constrained viewport size


clarkepaul:
Requires https://github.com/silverstripe/silverstripe-admin/issues/100 to be done first 
Part of Reponsive CMS issue https://github.com/silverstripe/silverstripe-framework/issues/6267

Remove constrained viewport of 720px by swapping `<meta name="viewport" content="width=720, maximum-scale=1.0" />` for `<meta name="viewport" content="width=device-width, maximum-scale=1.0" />` [~1pt]

---

Pull requests

- [x] https://github.com/silverstripe/silverstripe-admin/pull/103

tractorcow:
Fixed with https://github.com/silverstripe/silverstripe-admin/pull/103

PR #103, opened 2017-06-06
Do not zoom out to content when the screen is smaller than 720px

phalkunz:


tractorcow:
Please ensure that you link your PR back to the issue that you are resolving, @phalkunz 

I.e. https://github.com/silverstripe/silverstripe-admin/issues/100

PR #104, opened 2017-06-07
Fix check for readonly and disabled flags before loading the tree data

flamerohr:


dhensby:
Looks like the dist file wasn't generated correctly?

flamerohr:
yeah, apparently I had some different vendor packages from npm... needed to update those.

tractorcow:
This looks fine. Please merge on green.

PR #112, opened 2017-06-20
Feature/react from presentation

unclecheese:
See https://github.com/silverstripe/silverstripe-framework/issues/5630

Dependencies:
- https://github.com/silverstripe/silverstripe-asset-admin/pull/506
- https://github.com/silverstripe/silverstripe-cms/pull/1860
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/30
- https://github.com/silverstripe/webpack-config/pull/5


client/src/lib/dependency-injection/inject.js
```diff
@@ -1,11 +1,14 @@
 import React from 'react';
 import injectorContext from './injectorContext';
 
-const inject = (Component, dependencies, mapDependenciesToProps) => {
+const inject = (dependencies, mapDependenciesToProps, contextArg) => (Component) => {
```
flamerohr: I don't think the asset-admin has been updated with this new signature?

please add JSDoc block for the HOC at least

client/src/lib/dependency-injection/inject.js
```diff
@@ -1,11 +1,14 @@
...
   // eslint-disable-next-line react/prefer-stateless-function
   class Injected extends React.Component {
     render() {
       let props = {};
+      const context = typeof contextArg === 'function' ?
```
flamerohr: I don't like mixed params, it convolutes the API for the sake of reducing a couple of characters typed out for the end-user...

Please enforce a type rather than open for mixed types, at least for the new param that is added, dependencies could be left as is because it's an existing API... although would be nice to get that standardised to a single type as well.

client/src/lib/dependency-injection/tests/SchemaStateManager-test.js
```diff
@@ -0,0 +1,92 @@
+/* global jest, jasmine, describe, beforeEach, it, pit, expect, process */
+import SchemaStateManager from '../SchemaStateManager';
+
+describe('SchemaStateManager', () => {
+  let manager = null;
+  const state = {
```
flamerohr: build state in the `beforeEach()` call, so any mutations that could possibly happen in unit tests are shielded from other unit tests

client/src/components/FormAction/FormAction.js
```diff
@@ -1,6 +1,7 @@
 import React from 'react';
 import SilverStripeComponent from 'lib/SilverStripeComponent';
 import castStringToElement from 'lib/castStringToElement';
+import cx from 'classnames';
```
unclecheese: This is, or at least _was_ a convention back when it was part of the main React repo. I think it's just force of habit.

client/src/boot/index.js
```diff
@@ -1,134 +1,82 @@
...
   }
+
+  // Bootstrap routing
+  const routes = new BootRoutes(store, apolloClient);
+
+  // Force this to the end of the execution queue to ensure it's last.
+  window.setTimeout(() => {
+    registerComponents();
+    Injector.load();
+    routes.start(window.location.pathname);
```
chillu: Yeah we could have a more granular boot process, e.g. where components can hook in after injector setup, but before routing. I'm fine with this `setTimeout()` solution for the moment though. Or rather, I don't see how registering `jQuery.ready()` style callbacks would change this. You'd still need the `setTimeout()` to say "do this after executig all other procedural code in this event loop cycle - which is exactly what we want.

client/src/lib/dependency-injection/inject.js
unclecheese: Yeah, good catch. Enforced types for both args. `dependencies` was an array everywhere it was being used anyway.

client/src/lib/createClassMap.js
```diff
@@ -0,0 +1,11 @@
+const createClassMap = (classesStr) => {
+  const classConfig = {};
+  if (classesStr) {
+    classesStr.split(' ').forEach(c => {
+      classConfig[c.trim()] = true;
```
flamerohr: `c` to `classStr`
don't need `trim()` because the `split(' ')` would do the trimming for you, if wanted could add a `classStr !== ''` check

client/src/lib/dependency-injection/inject.js
```diff
@@ -1,24 +1,29 @@
...
+const inject = (dependencies, mapDependenciesToProps, getContext) => (Component) => {
   // eslint-disable-next-line react/prefer-stateless-function
   class Injected extends React.Component {
     render() {
       let props = {};
-      const context = typeof contextArg === 'function' ?
-        contextArg(this.props) :
-        contextArg;
-      let deps = dependencies;
+      const context = getContext && typeof contextArg === 'function' ?
```
flamerohr: could have a `getContext && typeof getContext !== 'function'` check above this to throw an exception, but not necessary.
this could then be simplified to `const context = (getContext) ? getContext(this.props) : null;`

client/src/lib/createClassMap.js
unclecheese: Yes, great catch.

client/src/boot/index.js
flamerohr: Yeah, that's cool
I've always felt uneasy with using `window.onload` for `appBoot` in general, but it seems unavoidable currently.

flamerohr:
Squashed some of the commits, as a few were just rebase commits

PR #114, opened 2017-06-20
Custom redux state in injector

flamerohr:
This allows plugging in extensions to registered reducers so that actions could end up saving a different set of data.

Please see https://github.com/silverstripe/silverstripe-framework/issues/6809 for more details

Relies on Aaron's other injector work to be merged first https://github.com/silverstripe/silverstripe-admin/pull/112

Please merge these pull requests afterwards:
- https://github.com/silverstripe/silverstripe-asset-admin/pull/508
- https://github.com/silverstripe/silverstripe-framework/pull/7042
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/28

Proof of concept module made here:
https://github.com/open-sausages/injected-redux-pocs


client/src/boot/registerReducers.js
```diff
@@ -0,0 +1,33 @@
...
+export default (extra = {}) => {
+  const FormReducer = applyFormMiddleware(
+    combineReducers({
+      formState: ReduxFormReducer,
+      formSchemas: SchemaReducer,
+    })
+  );
+
+  Object.keys(extra).forEach((key) => {
+    const reducer = extra[key];
```
flamerohr: Yup, good catch!

flamerohr: I think the one `registerReducers` file is fine, the `extra` in the params is mainly for the `apollo.reducer()` which I had no way to access that otherwise.
I don't see any value in splitting out the formReducers from "coreReducers", to me it's just as core

client/src/boot/registerReducers.js
```diff
@@ -0,0 +1,36 @@
...
+    treeDropdownField: TreeDropdownFieldReducer,
+    mobileMenu: MobileMenuReducer,
+    // @todo - Restore this once we address https://github.com/silverstripe/silverstripe-admin/issues/90
+    // unsavedForms: UnsavedFormsReducer,
+    ...extra,
+  });
+
+  Injector.transform('admin', (updater) => {
+    updater.reducer('form', applyFormMiddleware);
+  });
```
unclecheese: Nice. Glad this is in there. Makes a lot more sense.

client/src/lib/dependency-injection/buildReducerContainer.js
```diff
@@ -1,7 +1,69 @@
...
+  /**
+   * The redux store which will be providing the global context for the reducer middleware
+   * @type {object}
+   */
+  store: null,
+
+  setStore(store) {
+    if (this.initialised) {
+      throw new Error('Cannot mutate DI container after it has been initialised');
+    }
```
unclecheese: What happened to the `protect()` wrapper? This kind of thing comes up a lot, and it would be nice to have that all contained in one place.

client/src/lib/dependency-injection/inject.js
```diff
@@ -24,34 +31,21 @@ const inject = (dependencies, mapDependenciesToProps, getContext) => (Component)
...
-              mapDepedenciesToProps parameter passed to inject() 
-              should return an object that maps prop names to dependencies
-             `);
-          }
-        } else {
+      if (dependencies) {
+        const resolved = dependencies.map(dep => this.context.injector.get(dep, context));
+
+        props = (mapDependenciesToProps)
+          ? mapDependenciesToProps(...resolved)
```
unclecheese: The check that `mapDependenciesToProps` returns an object has been removed?

client/src/lib/dependency-injection/buildReducerContainer.js
flamerohr: It got impractical to have the `protect()` wrapper for the methods, which then needed rebinding to that object.
I guess it could have a `base.protect.call(this);` helper method if desired.

client/src/lib/dependency-injection/inject.js
flamerohr: oh true, I may have left it out and confused it with the `dependencies` check which I shifted above, will add it back in

flamerohr:
Have added changes based on feedback

PR #116, opened 2017-06-21
Fix app container was not filling the whole content

flamerohr:
This fixes the react container div not stretching to the full width and height of the parent container (which it needs to)

Screenshot of before (campaign-admin):
screen shot 2017-06-02 at 11 29 35 am


chillu:
Rebasing now

chillu:
Ran `yarn run lint` and `yarn run test` locally

PR #118, opened 2017-06-26
FIX namespace in PHP docs

phalkunz:
Issue: silverstripe/silverstripe-framework#5757


phalkunz:
This is not priority at the moment and should use short name as much as possible by using `use` at the top even for `{@link}`.

PR #131, opened 2017-06-30
Fix sanitize change detection values, TinyMCE produces false-negatives

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/1877


flamerohr:
test html, might need to insert to mysql directly:
```
<div class="captionImage leftAlone" style="width: 795px;"><img class="leftAlone" title="SilverStripe CMS interface" src="assets/Uploads/_resampled/ResizedImage795406-screen-about.png" alt="SilverStripe CMS interface" width="795" height="406"><p class="caption leftAlone">SilverStripe CMS interface</p></div><h3>Friendly and flexible open source CMS</h3><p>The SilverStripe CMS is designed to promote accessible site design, semantic markup, and HTML5 use. With its template flexibility, we're confident that it will be able to accommodate whatever is coming next. It has been downloaded well over 500,000 times and powers websites for government agencies, corporations, non-profits and a large number of much smaller websites.</p><p>We believe the open source model simply produces better web software and in turn, better meets our own needs and those of other developers who use the software. As with all open source software, anyone has access to the source code, and a global community of developers can share best practices, code, documentation, roadmap ideas and so on.</p><p>You never have to pay a licensing fee for SilverStripe software, and you never have to worry about vendor lock-in. Under the BSD license, which is one of the most flexible open source licenses and Open Source Initiative approved, you benefit from the contributions of others in the community, and the transparent way in which the software is developed in consultation with its users.</p>
```

PR #136, opened 2017-07-03
ENHANCEMENT: Simplify form transforms, make factories more customisable

unclecheese:
## Simplify form transforms
The form schema transforms were needlessly complex and verbose:

__current__
```js
Injector.transform(
  'my-custom-form',
  (updater) => {
    updater.form.alterSchema(
      'AssetAdmin.*',
      (updateSchema) => (form, values) => (
        updateSchema(
          form.updateField('Title', {
            myCustomProp: true
          })
          .getState()
        )
     )
  )
});
```

This is confusing and bloated. There's unnecessary currying and a bizarre call to `updateSchema` when it feels like the function should just return a new graph. The reason these calls are so complicated is because these transformations are loaded as middleware (`(next) => (params) => next(params)`) when they're essentially mini-reducers. They apply a sequence of state transforms and return new state. By making that change, the transforms become more streamlined and legible. 

__new__
```js
Injector.transform(
  'my-custom-form',
  (updater) => {
    updater.form.alterSchema(
      'AssetAdmin.*',
      (values, form) =>
        form.updateField('Title', {
            myCustomProp: true
        })
        .getState()
      )
    }
);
```

## Make factoires more customisable

In order to accomplish the above, containers need to have more control over the factories they create. Currently, `MiddlewareRegistry` creates factories, which should not be a concern for that service. The `getFactory` method has therefore been removed from that class and added to each container. This allows the form container to create a "reducer" style factory rather than using middleware, which allows the above. Further, this allows us to strip domain-specific concerns out of the factories, such as `displayName` and hand off that responsibility to the container. The only job of the middleware registry is now to store, sort, and fetch middleware by context.

## Documentation updates
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7112


client/src/lib/dependency-injection/applyFormMiddleware.js
```diff
@@ -1,7 +1,6 @@
 import Injector from '../Injector';
 import getIn from 'redux-form/lib/structure/plain/getIn';
 import setIn from 'redux-form/lib/structure/plain/setIn';
-import SchemaStateManager from './SchemaStateManager';
 
 const applyFormMiddleware = (reducer) => () => (state, action) => {
```
flamerohr: I do wonder if we should exclude executing this when `action.type == '@@redux-form/REGISTER_FIELD'`, as it's a fair amount of overhead for little gain for this to be firing every redux-form action or at least for registering a field.

client/src/lib/dependency-injection/tests/MiddlewareRegistry-test.js
```diff
@@ -129,12 +77,23 @@ describe('MiddlwareRegistry', () => {
 
       registry.sort();
 
-      const factory = registry.getFactory();
+      const matches = registry.getMatchesForContext().map(m => m.factory);
+      const factory = compose(...matches)(service);
```
unclecheese: Yes, much cleaner, and a more accurate test to boot. Have updated all the tests and eliminated the `compose` dependency.

client/src/lib/dependency-injection/applyFormMiddleware.js
unclecheese: Yeah, welcome to redux. Every reducer, every action. I've added this filter as a black list, in case there are more actions we want to filter out. 

Relevant: https://github.com/omnidan/redux-ignore

client/src/lib/dependency-injection/applyFormMiddleware.js
```diff
@@ -28,27 +31,11 @@ const applyFormMiddleware = (reducer) => () => (state, action) => {
...
-      }
-
-      return {
-        ...field,
-        value: update,
-      };
-    }),
-  });
+  const updates = formSchemaMiddleware(formState.values, schemaState);
+  newState = setIn(newState, `formSchemas.${schemaKey}.state`, updates);
```
flamerohr: do we need a `if (updates)` check for this, in case a rogue middleware registered and forgot to return the state?

PR #144, opened 2017-07-06
NEW Create a component that handles click event outside of a component

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-asset-admin/issues/519


client/src/components/ClickOutComponent/ClickOutComponent.js
```diff
@@ -0,0 +1,76 @@
...
+    if (!this.wasClicked) {
+      this.props.onClickOut.call(this.props.callerComponent);
+    } else {
+      this.wasClicked = false;
+    }
+  }
+
+  render() {
+    return Array.isArray(this.props.children) ?
+      <div>{this.props.children}</div> :
```
flamerohr: perhaps add a `className` property, so the `<div>` doesn't kill all stylings that may require direct child styles (e.g. flexbox)

client/src/components/ClickOutComponent/ClickOutComponent.js
```diff
@@ -0,0 +1,76 @@
...
+    el.removeEventHandler('click', this.handleDocumentClick);
+  }
+
+  handleElementClick() {
+    this.wasClicked = true;
+  }
+
+  handleDocumentClick() {
+    if (!this.wasClicked) {
+      this.props.onClickOut.call(this.props.callerComponent);
```
flamerohr: pass in the event object to `onClickOut()` so it's extensible if necessary, and you shouldn't need the `.call()` part if done like my comment in the other pull request.

client/src/components/ClickOutComponent/ClickOutComponent.js
```diff
@@ -0,0 +1,76 @@
...
+  componentDidMount() {
+    const el = ReactDOM.findDOMNode(this);
+    el.addEventListener('click', this.handleElementClick);
+    document.addEventListener('click', this.handleDocumentClick);
+  }
+
+  componentWillUnmount() {
+    const el = ReactDOM.findDOMNode(this);
+    el.removeEventHandler('click', this.handleElementClick);
+    el.removeEventHandler('click', this.handleDocumentClick);
```
flamerohr: I think this should be `document.removeEventHandler`

client/src/components/ClickOutComponent/ClickOutComponent.js
```diff
@@ -0,0 +1,76 @@
...
+ *       }
+ *    < /ClickOutComponent>
+ *  ```
+ */
+
+class ClickOutComponent extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.wasClicked = false;
```
flamerohr: have you tried move this to `this.state` and use `this.setState`, I can see why it potentially could be undesired, but might be worth using.

client/src/components/ClickOutComponent/ClickOutComponent.js
```diff
@@ -0,0 +1,76 @@
...
+  handleDocumentClick() {
+    if (!this.wasClicked) {
+      this.props.onClickOut.call(this.props.callerComponent);
+    } else {
+      this.wasClicked = false;
+    }
+  }
+
+  render() {
+    return Array.isArray(this.props.children) ?
```
flamerohr: maybe to keep this consistent and less confusing, just have `<div>{this.props.children}</div>`
with the `className` property, there's less reason to need to hack around `Children.only`

client/src/components/ClickOutComponent/ClickOutComponent.js
flamerohr: scratch that, can't with the bubbling event handling below :D

client/src/components/ClickOutComponent/ClickOutComponent.js
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  handleElementClick() {
+    this.wasClicked = true;
+  }
+
+  handleDocumentClick() {
+    if (!this.wasClicked) {
+      this.props.onClickOut.call(this.props.callerComponent);
+    } else {
```
flamerohr: maybe remove the `else` so that `wasClicked` can be repeatedly used?
At the moment, it looks like once `wasClicked` goes true, it stays true.

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  componentDidMount() {
+    const el = ReactDOM.findDOMNode(this);
+
+    // Click or touch event handlers on `el` and `document` will be both
+    // triggered when user clicked or touched inside the compoent's DOM. The
+    // handler on `el` is triggered before the one on `document` based on event
+    // bubbling.
+    ['click', 'touchstart'].forEach((eventName) => {
```
flamerohr: I'm a bit skeptical using the touch events, they almost always lead to undesired UX problems...
e.g. `touchstart` triggers long before a `click` event and they can trigger together. `touchstart` can also cause closing the modal when you accidentally try scrolling the screen by using the backdrop edge of the modal, etc...

Might be worth breaking this out to another story and leave this as `click` only for this work. Investigate a polyfill to handle touch events maybe?
https://github.com/jquery/PEP

PS: forgot to remove `touchstart` events below on unmount ;)

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  componentWillUnmount() {
+    const el = ReactDOM.findDOMNode(this);
+    el.removeEventHandler('click', this.handleElementClick);
+    el.removeEventHandler('click', this.handleDocumentClick);
+  }
+
+  handleElementClick() {
+    this.wasClicked = true;
```
flamerohr: Perhaps leave a comment about relying on bubbling here :)

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -0,0 +1,84 @@
...
+      this.props.onClickOut();
+    } else {
+      this.wasClicked = false;
+    }
+  }
+
+  render() {
+    return Array.isArray(this.props.children) ?
+      <div>{this.props.children}</div> :
+      React.Children.only(this.props.children);
```
flamerohr: I think this should be kept simple and do a simple `return <div className={this.props.className}>{this.props.children}</div>;`

where className is default `''` and it will enable users that rely on direct children styles (flexbox)

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -0,0 +1,84 @@
...
+  }
+}
+
+Focusedzone.propTypes = {
+  children: PropTypes.object,
```
flamerohr: I would suggest `PropTypes.any` for children, because children could be 0, 1 or many (class or function) components or even just a display string... and it's a headache to validate them all properly. `PropTypes.object` kind of masks the variable nature of `children`

client/src/components/Focusedzone/Focusedzone.js
phalkunz: I put a comment in the component's constructor now.

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  handleElementClick() {
+    this.wasClicked = true;
+  }
+
+  handleDocumentClick() {
+    if (!this.wasClicked) {
+      this.props.onClickOut();
+    } else {
```
flamerohr: remove the `else` :) otherwise if `wasClicked` has been set true once, it won't reset afterwards

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -0,0 +1,91 @@
...
+    // handler on `el` is triggered before the one on `document` based on event
+    // bubbling.
+    el.addEventListener('click', this.handleElementClick);
+    document.addEventListener('click', this.handleDocumentClick);
+  }
+
+  componentWillUnmount() {
+    const el = ReactDOM.findDOMNode(this);
+    el.removeEventHandler('click', this.handleElementClick);
+    el.removeEventHandler('click', this.handleDocumentClick);
```
flamerohr: this should be `document.removeEventHandler()` I believe :)

PR #146, opened 2017-07-11
check for already set config in tinymce

andrewandante:
This means that a user can set config in `mysite` or a module that comes alphabetically before `silverstripe-admin` and have it apply, without having to resort to the `zzzmysite` workaround.

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -9,14 +9,11 @@ use SilverStripe\Admin\CMSProfileController;
 use SilverStripe\Forms\HTMLEditor\TinyMCEConfig;
 
 // Default CMS HTMLEditorConfig
-TinyMCEConfig::get('cms')->setOptions(array(
+$tinyMCEDefaultOptionsMap = [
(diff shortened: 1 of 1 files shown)


flamerohr:
oh I see, so you've just changed it to not aggressively override things so module `_config.php` loading order is less of a problem...
Yeah, this will be a temporary workaround to the actual problem.

flamerohr:
Oh btw, @andrewandante you should recreate this pull request to the `1` branch, otherwise we won't see it until `silverstripe-admin@2.0` comes out ;)

andrewandante:
@flamerohr here you go https://github.com/silverstripe/silverstripe-admin/pull/147

tractorcow:
So I just found out about this bug; This means that any value set as the default inside `TinyMCEConfig::$settings` is ignored; E.g. `'friendly_name' => '(Please set a friendly name for this config)',` stays set here.

Since we have module priority now, we should rely on that instead.

PR #150, opened 2017-07-13
Fix keyboard accessibility broke with current version of react-select

flamerohr:
Enhancement added the ability to search for TreeDropdownField
Enhancement updated styles for TreeDropdownField, to browse inside a folder it will now be a button
API added mapHighlight to castStringToElement library

Please see https://github.com/silverstripe/silverstripe-framework/issues/6842 for more details


chillu:
Looking awesome overall. A bit of UX and design feedback (cc @clarkepaul). Might not block this PR, some of these are pre-existing issues. But please ensure they're actioned soon.

Loading is misaligned

image

Can we add "type to search" when you focus the cursor in the field? It's a bit weird that it shows the current selection, but then doesn't allow to edit the title itself:

image

To unselect the current item via the keyboard, you have to press ESC. It's not possible to keyboard focus on the "delete" icon. This seems like an odd way to solve this. Have you researched if there's an aria construct for this?

flamerohr:
> Can we add "type to search" when you focus the cursor in the field?
I've tried, but that's how `react-select` works... It can also be considered weird to appear like you've lost your current selection by focusing on the field as well. So it's a bit of two-sides of the same coin situation here.
The "type to search" is less relevant during a multi-select situation as well.

> To unselect the current item via the keyboard, you have to press ESC. It's not possible to keyboard focus on the "delete" icon.
This is also how `react-select` works, it relies on always having focus on the one select field, so that it can capture keyboard accessibility events.
It's not an obvious feature and could use some prompt somewhere probably... but you can browse options by up/down keys, and move in/out of folders with right/left keys respectively. Would be difficult to enable this feature if focus can jump to the close button or arrow button.

flamerohr:
I've updated the alignment of the loading text :)

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -20,38 +23,51 @@ class TreeDropdownField extends Component {
...
+      return;
+    }
+
+    if (nextProps.data.cacheKey !== this.props.data.cacheKey) {
       // invalidate the tree cache, as paths have changed
-      this.loadTree([]);
+      this.loadTree([], false, nextProps.search);
+    }
+
+    if (this.props.search !== nextProps.search) {
```
flamerohr: no, I don't think it's possible in an application setting.
the `cacheKey` comes from schema, so it'll be two separate redux update calls and it'd only be only or the other updated and unlikely both at the same time.

just thinking about the order, perhaps it could be safer with `(this.props.search !== nextProps.search)` being checked first, but not necessary :)

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -75,6 +91,9 @@ class TreeDropdownField extends Component {
     // No more path means this is the complete tree
     let node = this.props.tree;
     for (const next of this.props.visible) {
+      if (!node.children) {
+        break;
+      }
       node = node.children.find((child) => (child.id === next));
```
flamerohr: I think that's the point of `node` though, as it is traversing down the hierarchy `this.props.tree` recursively to get to the last entry of visible
we might be able to refactor this to use the `.reduce()` function so that it is self-contained but I think it's not really worth the effort for this case

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -293,14 +316,14 @@ class TreeDropdownField extends Component {
       }
     }
 
-    return this.loadTree(path);
+    return this.loadTree(path, false);
   }
 
-  loadTree(path) {
+  loadTree(path, loadNewPath, search) {
```
flamerohr: fair call, I'll add some docs in

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -293,14 +324,22 @@ class TreeDropdownField extends Component {
       }
     }
 
-    return this.loadTree(path);
+    return this.loadTree(path, false);
```
tractorcow: Can you either add a default for the third arg, or pass in `''`

tractorcow:
client/src/components/TreeDropdownField/TreeDropdownField.scss
  12:14  error  Whitespace required before !important  space-before-bang

PR #152, opened 2017-07-13
FIX Allow buttons to retain their font size on small screens if they don't have an icon

robbieaverill:
~~We set `font-size: 0` for small devices on buttons inside a `.btn-toolbar` element. This change allows you to add `.btn-retain-text-sm` to a button in this case to let it retain its original font size.~~

*Update:* only apply `font-size: 0` to buttons that don't have a font icon.

Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/533


dhensby:
needs rebasing and rebuilding

robbieaverill:
Done, thanks

robbieaverill:
Thanks for the feedback @phalkunz, updated

PR #154, opened 2017-07-17
Fix check that an entry is found before destructuring it

flamerohr:
Happens primarily when search is cleared in asset-admin


tractorcow:
I'll re-test and merge once I'm happy with the testing.

PR #156, opened 2017-07-17
NEW Allow help link to be disabled in user code

robbieaverill:
_Initial approach removed_

Users can now disable the CMS help link in user code:

```php
# File: mysite/_config.php
use SilverStripe\Admin\LeftAndMain;

Config::modify()->set(LeftAndMain::class, 'help_link', false);
```

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -532,17 +532,19 @@ class LeftAndMain extends Controller implements PermissionProvider
             i18n::set_locale($member->Locale);
         }
 
-        // can't be done in cms/_config.php as locale is not set yet
-        CMSMenu::add_link(
-            'Help',
-            _t('SilverStripe\\Admin\\LeftAndMain.HELP', 'Help', 'Menu title'),
-            LeftAndMain::config()->uninherited('help_link'),
-            -2,
-            array(
-                'target' => '_blank'
-            ),
-            'font-icon-help-circled'
-        );
+        if ($helpLink = LeftAndMain::config()->uninherited('help_link')) {
+            // can't be done in cms/_config.php as locale is not set yet
+            CMSMenu::add_link(
+                'Help',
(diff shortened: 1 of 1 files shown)


robbieaverill:
Ok, it's is a small break in that the public title property won't return a string any more 😄 if the team likes this approach I can make it protected and overload to return from the public getter?

code/LeftAndMain.php
```diff
@@ -525,18 +525,6 @@ protected function init()
             i18n::set_locale($member->Locale);
         }
 
-        // can't be done in cms/_config.php as locale is not set yet
```
tractorcow: Why can't we just put this whole block around an

```php
if (LeftAndMain::config()->uninherited('help_link')) {}
```

And call it a day?

_config.php
```diff
@@ -69,3 +70,14 @@
 });
 
 CMSMenu::remove_menu_class(CMSProfileController::class);
+
+CMSMenu::add_link(
+    'Help',
+    function () {
+        return _t(LeftAndMain::class . '.HELP', 'Help', 'Menu title');
```
tractorcow: Remember class constants can't be text collected. Best to avoid if you can for now! Sorry it's not obvious.

code/LeftAndMain.php
robbieaverill: Ok, let's do that

_config.php
robbieaverill: Good to know!

tractorcow: I will fix this, one day.

code/LeftAndMain.php
robbieaverill: @tractorcow updated, thanks 😄

_config.php
robbieaverill: Do we have an issue logged for this?

tractorcow:
Merge on green, if green.

PR #157, opened 2017-07-17
FIX Add getter and setter for the version provider dependency on LeftAndMain

robbieaverill:
Related PR that made this change: #153


tractorcow:
I would prefer to fix the broken injection than to make it silently not fail.

PR #167, opened 2017-07-23
Custom alert style

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-campaign-admin/issues/16


client/src/components/FormAction/FormAction.js
```diff
@@ -113,6 +136,26 @@ class FormAction extends SilverStripeComponent {
...
+   * @return {String}|null
+   */
+  getSavedTitle() {
+    // In case this is specified directly
+    return this.props.savedTitle || this.props.data.savedTitle || null;
+  }
+
+  getSavedIcon() {
+    // In case this is specified directly
+    return this.props.savedIcon || this.props.data.savedIcon || null;
```
tractorcow: Why are all these fields in two places? They should either be on this.props or this.props.data.

tractorcow: It looks like there's a bug that you're working around rather than fixing.

client/src/components/FormAction/FormAction.js
```diff
@@ -72,6 +86,17 @@ class FormAction extends SilverStripeComponent {
   }
 
   /**
+   * @return {boolean}
+   */
+  isConstructive() {
```
flamerohr: I think `isPrimary()` or `isPrimaryType()` is a better name, it took me a while to work out what "constructive" meant and I doubt I'll remember a month later :)

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -363,7 +365,11 @@ function mapStateToProps(state, ownProps) {
 
   const stateOverrides = schema && schema.stateOverride;
   const loading = schema && schema.metadata && schema.metadata.loading;
-  return { schema, submitting, values, stateOverrides, loading };
+  const changed = !!(
+    state.unsavedForms &&
+    state.unsavedForms.find(form => ownProps.identifier === form.name)
+  );
```
flamerohr: Could we use the `isDirty()` API provided by redux-forms?
`isDirty(form.name, getFormState)(state);`

client/src/components/FormAction/FormAction.js
```diff
@@ -149,10 +192,15 @@ FormAction.propTypes = {
   name: React.PropTypes.string,
   handleClick: React.PropTypes.func,
   title: React.PropTypes.string,
+  // Action text when there's no changes detected in the form
+  savedTitle: React.PropTypes.string,
+  savedIcon: React.PropTypes.string,
+  savedClasses: React.PropTypes.string,
```
flamerohr: missing definition in the `data` prop

client/src/components/FormAction/FormAction.js
```diff
@@ -149,10 +192,15 @@ FormAction.propTypes = {
...
   type: React.PropTypes.string,
   loading: React.PropTypes.bool,
   icon: React.PropTypes.string,
   disabled: React.PropTypes.bool,
+  changed: React.PropTypes.bool,
```
flamerohr: this should be named `formChanged` I think, not to be confused with this actual component changing somehow

client/src/boot/transforms.js
```diff
@@ -0,0 +1,27 @@
...
+      updater.form.alterSchema('*', form => {
+        form.mutateField('action_save', (field) => {
+          const isPristine = form.isPristine();
+          const classes = classnames({
+            [field.data.pristineClass]: isPristine,
+            [field.data.dirtyClass]: !isPristine,
+          });
+          return {
+            ...field,
+            title: isPristine ? field.data.pristineTitle : field.data.dirtyTitle,
```
tractorcow: What if pristineTitle and dirtyTitle aren't set? FormField.php doesn't set these by default either.

tractorcow: You could get rid of `dirtyTitle` and just have `title` (for dirty / normal) and optional `pristineTitle`

unclecheese: Have updated this to have fallback value

client/src/boot/index.js
```diff
@@ -8,6 +8,7 @@ import { setConfig } from 'state/config/ConfigActions';
 import registerComponents from 'boot/registerComponents';
 import registerReducers from 'boot/registerReducers';
 import applyDevtools from 'boot/applyDevtools';
+import transforms from './transforms';
```
flamerohr: import from `boot/transforms` would be following existing conventions, could this be named `applyTransforms`

client/src/boot/transforms.js
```diff
@@ -0,0 +1,27 @@
...
+        form.mutateField('action_save', (field) => {
+          const isPristine = form.isPristine();
+          const classes = classnames({
+            [field.data.pristineClass]: isPristine,
+            [field.data.dirtyClass]: !isPristine,
+          });
+          return {
+            ...field,
+            title: (isPristine ? field.data.pristineTitle : field.data.dirtyTitle) || field.title,
+            icon: (isPristine ? field.data.pristineIcon : field.data.dirtyIcon) || field.icon,
```
flamerohr: could we move the `isPristine ? x : y` parts to `customTitle/Icon` constants?

PR #171, opened 2017-07-26
Fix remove clearSchema, as it wasn't doing anything

flamerohr:
code clean up while working on https://github.com/silverstripe/silverstripe-asset-admin/pull/565


client/src/state/schema/SchemaReducer.js
```diff
@@ -12,6 +12,18 @@ export default function schemaReducer(state = initialState, action = null) {
       }));
     }
 
+    case ACTION_TYPES.CLEAR_SCHEMA: {
+      return deepFreeze(Object.entries(state).reduce((prev, [key, value]) => {
```
phalkunz: Requires quite a bit of thinking. Can you help break this down a bit and/or a unit test would help as well.

flamerohr: fair enough, I can move this to a helper function... this is basically removing an object property the same as `const newState = { ...state }; delete state[action.payload.id];` but without creating the `newState` object.

PR #175, opened 2017-07-31
Expand injector form API to be more state aware with FormStateManager

unclecheese:
Required for silverstripe/silverstripe-campaign-admin#16
Resolves: silverstripe/silverstripe-framework#7226


client/src/lib/dependency-injection/FormStateManager.js
```diff
@@ -19,25 +35,25 @@ class SchemaStateManager {
...
+   * @returns {FormStateManager}
    */
   mutateField(fieldName, updater) {
-    const fieldList = this.schemaState.fields || [];
+    const fieldList = this.schema.state.fields || [];
     const fieldIndex = fieldList.findIndex(field => field.name === fieldName);
     if (fieldIndex < 0) {
       return this;
     }
     const field = fieldList[fieldIndex];
```
flamerohr: just noticed this should be `{ ...fieldList[fieldIndex] }` or below have `updater({ ...field })` in case of mutations in the `updater()`

client/src/lib/dependency-injection/FormStateManager.js
```diff
@@ -1,16 +1,32 @@
...
+class FormStateManager {
 
   /**
    * Constructor
-   * @param {object} schemaState
+   * @param {object} schema
    */
-  constructor(schemaState) {
-    this.schemaState = { ...schemaState };
+  constructor(schema, reduxFormState) {
```
flamerohr: missing JSDoc for `reduxFormState`

PR #180, opened 2017-08-04
Fix FormBuilderLoader calls initialize on redux-form after setting a new schema

flamerohr:
Enhancement appleFormMiddleware now execute middleware for all forms with a matching name
Enhancement code clean up for schemaReducer

This is needed to be able to action this issue https://github.com/silverstripe/silverstripe-asset-admin/issues/309 properly
Otherwise the "Update file" label will only show on load and disappear after loading


client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -276,6 +256,13 @@ class FormBuilderLoader extends Component {
             this.props.identifier
           );
 
+          const schemaData = formSchema.schema || this.props.schema.schema;
+          const formData = schemaFieldValues(schemaData, overriddenSchema.state);
+
+          // need to initialize the form again in case it was loaded before
+          // this will re-trigger Injector.form APIs, reset values and reset pristine state as well
+          this.props.actions.reduxForm.initialize(this.props.identifier, formData);
+
```
unclecheese: By re-initialising, you're overwriting the `submitSucceeded` prop, which may be useful for displaying messaging. Consider:

`initialize(this.props.iderntifier, formData, false, { keepSubmitSucceeded: true })`

client/src/components/Modal/Modal.scss
```diff
@@ -73,7 +73,6 @@
   width: 100%;
   animation: modal--slideup .6s 1;
   background: $background-main;
-  display: block;
```
flamerohr: skimmed this while looking through the Modal css, it's a benign change and small clean up.

Also, there's a `display: flex` in this same selector block, a few lines down

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
flamerohr: nice one, will make the change

flamerohr: also, as an aside note, what is with `redux-form` always having these magical hidden parameters?!?

Issue #184, opened 2017-08-09
Add schema's form name to identifier


flamerohr:
We've found that we needed to check the schema's name for a "variation" of the `fileEditForm` in asset-admin during this story https://github.com/silverstripe/silverstripe-asset-admin/issues/309
See code:
```
  const schemaName = schema.schema.name;
  if (schemaName !== 'fileInsertForm') {
    return schema;
  }
  // do transformations here
```

Ideally we should be able to utilise the Injector's selector API to target this variation, as it would not be ideal to have the above code for other variations of other forms.
This came up because the `fileEditorForm` is defined inside another re-useable component, so outside components have no control of the identifier.

Suggestion would be to append the schema name to the end of the identifier by `FormBuilder`.

## Pull Requests
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/188

PR #188, opened 2017-08-10
Feature/form builder identifier schema

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-admin/issues/184


client/src/containers/FormBuilderLoader/createIdentifier.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Creates a dot-separated identifier for forms generated
+ * with schemas (e.g. FormBuilderLoader)
+ *
+ * @param props
+ * @returns {string}
+ */
+const createIdentifier = (props) => {
+  const { identifier, schema } = props;
```
flamerohr: maybe deconstruct in the params? could do `{ identifier, schema = {} }` so can remove at least one exists check below

client/src/containers/FormBuilderLoader/tests/createIdentifier-test.js
```diff
@@ -0,0 +1,35 @@
...
+    expect(createIdentifier(props)).toBe('one.two');
+    props = {
+      identifier: 'one.two',
+      schema: {
+        schema: {
+          name: 'three',
+        },
+      },
+    };
+    expect(createIdentifier(props)).toBe('one.two.three');
```
flamerohr: one last test for if identifier is empty and schema.schema.name is populated :D

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -339,7 +352,7 @@ function mapStateToProps(state, ownProps) {
   const reduxFormState =
     state.form &&
     state.form.formState &&
-    getIn(state.form.formState, ownProps.identifier);
+    getIn(state.form.formState, createIdentifier(ownProps));
```
flamerohr: could even reassign the identifier value here so `const identifier = createIdentifier(ownProps);`

it does have the cross-wiring, so I'm a little adverse to that idea, but it's possible and it wouldn't stack/recurse

client/src/containers/FormBuilderLoader/createIdentifier.js
```diff
@@ -0,0 +1,16 @@
+/**
+ * Creates a dot-separated identifier for forms generated
+ * with schemas (e.g. FormBuilderLoader)
+ *
+ * @param props
+ * @returns {string}
+ */
+const createIdentifier = (props) => {
```
flamerohr: I guess I'd prefer more defined params for helper functions like this... object is fine, but the nested name is weird and specific to `FormBuilder`

unclecheese: Yeah, I agree. I had originally named it `getIdentifierFromProps` but that seemed a bit too specific. I really can't imagine this being used anywhere else. The whole purpose of abstracting it was because sometimes you don't have the `this` context, e.g. `mapStateToProps`. It really should be a static member of `FormBuilderLoader`.

How about I move it into the `FormBuilderLoader` file, and rename it to `createFormIdentifierFromProps`?

client/src/containers/FormBuilderLoader/tests/createIdentifier-test.js
unclecheese: In theory that should never happen because `identifier` is required. If we move the method to a `FormBuilderLoader` member, it's probably safe to rely on that being true.

client/src/containers/FormBuilderLoader/createIdentifier.js
flamerohr: sure, that sounds good :)

PR #190, opened 2017-08-15
Update font icons

tractorcow:
Parent story: https://github.com/silverstripe/silverstripe-asset-admin/issues/578


tractorcow:
Updated with new icons

image

PR #195, opened 2017-08-17
TreeDropdownField API improvements

unclecheese:
Partially resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/578

* Extract tree parsing utility methods into separate file, so they can be called statically
* Export utility methods as part of the `TreeDropdownField` module
* Add new `findTreeByPath` as prop

To do (someday):

* More internal computations like `findTreeByPath` should be props, e.g. `getDropdownOptions`


client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -343,7 +246,7 @@ class TreeDropdownField extends Component {
     }
 
     // If ancestor node is already loaded (and non-empty) then don't re-trigger
-    const foundTree = this.findTreeByPath(this.props.tree, path);
+    const foundTree = findTreeByPath(this.props.tree, path);
```
flamerohr: shouldn't this be `this.props.findTreeByPath()`? Seems weird to be from props in some parts and not in other parts

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -774,6 +677,7 @@ TreeDropdownField.propTypes = {
   readOnly: PropTypes.bool,
   disabled: PropTypes.bool,
   tree: PropTypes.shape(TreeDropdownFieldNode.propTypes), // Root node of tree
+  findTreeByPath: PropTypes.func.isRequired, // Finds the node given the tree and a path
```
flamerohr: remove the `.isRequired` part, since the `defaultProps` entry covers that

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -843,6 +748,14 @@ function mapDispatchToProps(dispatch) {
...
-export { TreeDropdownField, ConnectedTreeDropdownField, MULTI_EMPTY_VALUE, SINGLE_EMPTY_VALUE };
+export {
+  TreeDropdownField,
+  ConnectedTreeDropdownField,
+  MULTI_EMPTY_VALUE,
+  SINGLE_EMPTY_VALUE,
+  findTreePath,
+  findTreeByID,
+  findTreeByPath,
+};
```
flamerohr: I think it would be worth exposing `findTreePath` and `findTreeByID` as customisable props as well, what do you reckon?

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -116,7 +114,7 @@ class TreeDropdownField extends Component {
    * @return {Object}
    */
   getVisibleTree() {
-    return this.findTreeByPath(this.props.tree, this.props.visible);
+    return this.props.findTreeByPath(this.props.tree, this.props.visible);
```
tractorcow: I'm not sure why this is both a prop and an import. Surely it should just be `findTreeByPath` not `this. findTreeByPath` nor `this.props. findTreeByPath`?

client/src/components/TreeDropdownField/TreeDropdownField.js
unclecheese: Yeah, definitely. Was just trying to do the minimum to get the other story over the line.

client/src/components/TreeDropdownField/TreeDropdownField.js
unclecheese: It's a bit of a halfway measure, mostly owing to the limitations of Injector. The idea is to give the user some say in how the visible tree is computed and rendered. (https://github.com/silverstripe/silverstripe-asset-admin/pull/592/files#diff-5ec27365abb27c9225a02be39af02f85R23)

It abstracts away how the tree is parsed, but assigns the utilities as default props so that it doesn't impose undue composition on the developer.

client/src/components/TreeDropdownField/TreeDropdownField.js
unclecheese: yup, missed that one.

PR #198, opened 2017-08-18
Improve error handling of ajax errors when providing form schema

tractorcow:
Parent story: https://github.com/silverstripe/silverstripe-asset-admin/issues/528


code/LeftAndMain.php
```diff
@@ -1250,7 +1296,8 @@ public function delete($data, $form)
 
         $record->delete();
 
-        $this->getResponse()->addHeader('X-Status', rawurlencode(_t('SilverStripe\\Admin\\LeftAndMain.DELETED', 'Deleted.')));
+        $this->getResponse()->addHeader('X-Status',
+            rawurlencode(_t('SilverStripe\\Admin\\LeftAndMain.DELETED', 'Deleted.')));
```
flamerohr: where's the `__CLASS__` replacement here? :P

code/LeftAndMain.php
```diff
@@ -386,29 +388,76 @@ public function schema($request)
...
+            'code' => $errorCode,
+        ];
+        if ($errorMessage) {
+            $error['value'] = $errorMessage;
+        }
+
+        // Support explicit error handling with status = error, or generic message handling
+        // with a message of type = error
+        $result = [
+            'status' => 'error',
```
flamerohr: I think this `'status'` doesn't do anything and isn't used anywhere currently, but it's a good start for an explicit error handle

PR #205, opened 2017-08-24
FIX Add ability to ignore GET params when view grid field item in cms

phalkunz:


client/src/legacy/LeftAndMain.js
```diff
@@ -1212,8 +1212,11 @@ $.entwine('ss', function($) {
   /**
    * Overload the default GridField behaviour (open a new URL in the browser)
    * with the CMS-specific ajax loading.
+   *
+   * To opt out of this overload by adding `show-item-ignore-get-params` html
+   * attribute to the grid field.
    */
-  $('.cms .grid-field').entwine({
+  $('.cms .grid-field:not([show-item-ignore-get-params])').entwine({
```
flamerohr: I think the attribute name is a little confusing, something like `cms-loading-ignore-params` would explain what it's trying to do straight away

PR #214, opened 2017-08-30
Add TreeDropdownField from React to Entwine

flamerohr:
Also fixes Page back button being hidden in mobile view


tractorcow:
I've noticed this weird behaviour:

If I select multiple items, it only saves the first one.

Before saving:

image

After saving:

image

Can you please have a look into this?

flamerohr:
@tractorcow thanks, issue should be fixed now

chillu:
image

Hooray!

PR #216, opened 2017-08-31
ENHANCEMENT Allow campaigns to be created inline

tractorcow:
Build on top of https://github.com/silverstripe/silverstripe-admin/pull/212, which is also required

Fixes localisation issues so that admin doesn't rely on non-dependencies.


tractorcow:
I've updated FormBuilderLoader.js with a fix for re-submission breaking form state, and have updated the comments to better explain the logic.

tractorcow:
Pushed up a fix for CI working in trusty.

PR #219, opened 2017-09-01
Update font icons

sachajudd:
Updated icon screen shot 2009-12-26 at 5 25 51 pm


dhensby:
The NPM build sanity checks have failed :(

sachajudd:
I'll take another look at this on Monday 🙂

tractorcow:
It looks as though this changes removes a recently added font icon, too.

tractorcow:
Ok, I've investigated the issue with the missing icon. It looks like it was renamed incorrectly, but the damage is done now, so I think we can just leave it as is for now.

Issue #160, opened 2017-07-20
Validation on fields are visually attached to the fields (requires gap)


clarkepaul:
There should be a gap between validation messages and the field. We are using Bootstrap defaults as far as I'm aware so referring to that might help.
![image](https://user-images.githubusercontent.com/555033/28401185-665b6ed0-6d6c-11e7-8e31-a995e0b09087.png)

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-admin/pull/234
~https://github.com/silverstripe/silverstripe-framework/pull/7354~

clarkepaul:
It seems we are no longer using Bootstrap messages for fields, we should re-align ourselves with Bootstrap 4 methods and classes (ie. use .is-invalid and .is-valid) as demonstrated here https://getbootstrap.com/docs/4.0/components/forms/#validation

chillu:
See https://github.com/silverstripe/silverstripe-framework/issues/6223 for a wider task for agreeing on an error message style

PR #234, opened 2017-09-08
Style field validation message

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-admin/issues/160


client/src/components/Form/Form.js
```diff
@@ -45,6 +45,7 @@ class Form extends Component {
     const className = ['form'];
     if (valid === false) {
       className.push('form--invalid');
+      className.push('was-validated');
```
flamerohr: what's the purpose of this class? how come it's not BEM?

client/src/components/TextField/TextField.js
```diff
@@ -31,6 +31,7 @@ class TextField extends SilverStripeComponent {
       value: this.props.value,
       placeholder: this.props.placeholder,
       autoFocus: this.props.autoFocus,
+      required: Boolean(this.props.meta && this.props.meta.error),
```
flamerohr: nice, do we need to repeat this for other fields as well?

flamerohr:
I was also curious if we could use something from bootstrap that doesn't require the `:valid`/`:invalid` selector, as they won't work for customised fields like `react-select`

I've found this in the docs for bootstrap4:
> As a fallback, `.is-invalid` and `.is-valid` classes may be used instead of the pseudo-classes for server side validation. They do not require a .was-validated parent class.

Have you tried looking into using that instead @phalkunz ?

phalkunz:
@flamerohr Did you mean using `.is-invalid` class instead of setting `required` prop?

client/src/components/TextField/TextField.js
phalkunz: Yes for the fields that are not subclass of `TextField`. Depends on the other comment, I might need to replicate this in other field components.

flamerohr:
that's right :) I wonder if that would mean we don't need to the `noValidate` on the form element anymore

phalkunz:
@flamerohr Good idea. I'll look into fixing on Monday :)

client/src/components/Form/Form.js
```diff
@@ -59,6 +59,7 @@ class Form extends Component {
     );
 
     return (
+      // Needs `noValidate` here in order to prevent browser's message
```
flamerohr: sweet, don't need `noValidate` anymore :)

client/src/components/FieldHolder/FieldHolder.js
```diff
@@ -115,7 +118,12 @@ function fieldHolder(Field) {
     }
 
     renderField() {
-      const field = <Field {...this.props} />;
+      const props = { ...this.props };
+      const hasErrors = Boolean(this.props.meta && this.props.meta.error);
+      if (hasErrors) {
+        props.extraClass = `${props.extraClass} is-invalid`;
```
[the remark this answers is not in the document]
flamerohr: I'd also recommend replicating the `renderMessage()` logic to the `hasErrors` check:
- checking there is a message
- adding the `meta.touched` and `meta.dirty` checks

so you don't get a red highlighted field with no message

PR #235, opened 2017-09-08
Verify exist element before add new item

expouic:
Verify exists indication to remove item before add new

ex: 
_config.php
CMSMenu::remove_menu_item('Help');

### DIFF
--- a/code/CMSMenu.php
+++ b/code/CMSMenu.php
@@ -342,11 +342,15 @@ class CMSMenu implements IteratorAggregate, i18nEntityProvider
             $item->setAttributes($attributes);
         }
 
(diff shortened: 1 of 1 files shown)


dhensby:
Please fix the code style issues.

@tractorcow how do you feel about this?

tractorcow:
If I were using this API, I would expect all changes to be processed in order.

changes:
- add
- remove
- add

Expected: added
Result after these changes: removed

Maybe if you explained the problem you were trying to solve I could give you some advice? :D

robbieaverill:
Thanks for contributing @expouic, feel free to reopen this PR if you'd like to re-address it =)

Issue #126, opened 2017-06-29
Pagination on gridfields breaks at smaller screen sizes


clarkepaul:
A quick fix would be to align pagination to the left on smaller screens, if there was still too much going on I suggest removing the actions which take the user the end or beginning as they can also do this via the page selector.
<img width="225" alt="pasted_image_29_06_17__4_36_pm" src="https://user-images.githubusercontent.com/555033/27671943-c8143c7e-5ceb-11e7-8c3c-670c33b682d1.png">

## Notes

This should be reviewed with:
- https://github.com/silverstripe/silverstripe-asset-admin/issues/625
- https://github.com/silverstripe/silverstripe-admin/issues/242

## Pull request

 - [x] https://github.com/silverstripe/silverstripe-admin/pull/241

PR #241, opened 2017-09-11
FIX Fine-tune button styles

phalkunz:
Fixes:
 - https://github.com/silverstripe/silverstripe-asset-admin/issues/625
 - https://github.com/silverstripe/silverstripe-admin/issues/126
 - https://github.com/silverstripe/silverstripe-admin/issues/242


tractorcow:
Can you please rebase and ensure these are tested on IE? I've noticed some issues with icon / button height in IE that should be verified.

PR #247, opened 2017-09-12
Fix incorrect data type obtained sometimes through html dataset

flamerohr:


tractorcow:
Tested and it seems to do no harm; Seems logical to do a bit of type-coercing in the entwine area, where things are a bit "wild west-ey".

PR #250, opened 2017-09-12
Update airbnb rules for linting

flamerohr:
Requires https://github.com/silverstripe/webpack-config/pull/9

Please see https://github.com/silverstripe/silverstripe-framework/issues/6427 for more details


package.json
```diff
@@ -73,25 +75,21 @@
     "validator": "^6.1.0"
   },
   "devDependencies": {
-    "@silverstripe/webpack-config": "^0.2.9",
-    "babel-core": "^6.24.1",
+    "@silverstripe/webpack-config": "file:../../webpack-config",
```
flamerohr: For development purposes, to replace with the proper `0.3.0` version when that is published on the `webpack-config` package

client/src/components/FieldHolder/FieldHolder.js
```diff
@@ -53,7 +77,7 @@ function fieldHolder(Field) {
     /**
      * Build title label
      *
-     * @returns {Component}
+     * @returns {React}
```
phalkunz: Shouldn't it be `Component` instead? Is `React` a type?

client/src/boot/BootRoutes.js
```diff
@@ -1,6 +1,6 @@
-/* global window */
+/* global window, document */
 
-import $ from 'jQuery';
+import $ from 'jquery';
```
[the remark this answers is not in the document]
flamerohr: it's to standardise calling jquery in module code
there were always issues of confusion with the case being `jQuery` or `jquery`, so this makes it all `jquery`.

client/src/components/CheckboxSetField/CheckboxSetField.js
```diff
@@ -132,6 +131,6 @@ CheckboxSetField.defaultProps = {
   value: [],
 };
 
-export { CheckboxSetField };
+export { CheckboxSetField as Component };
```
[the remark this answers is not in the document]
flamerohr: yup, there's an eslint rule where `import { CheckboxField }`and `import CheckboxField` cannot both exist, which makes sense because it's easy to confuse between having the `{ }` wrappers and without it
So it made more sense to name the item export as `Component` here

client/src/components/FieldHolder/FieldHolder.js
flamerohr: React is the javascript class, which is what happens here... I'm still exploring this, might be more valid to be `Symbol` I think, but `Component` isn't correct.
JSDocs take `Component` as HTML Web Component iirc

PR #252, opened 2017-09-13
BUG Fix multi-selection treefield

tractorcow:
Fixes:
 - Crash when multi-selecting
 - Ability to select a value more than once
 
 Fixes CMS behat build


client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -496,9 +496,14 @@ class TreeDropdownField extends Component {
...
-      ? this.props.value
-      : [this.props.value];
+
+    // Always coerce value to array for menu (even if not multi-select)
+    let value = [];
+    if (this.props.value) {
+      value = Array.isArray(this.props.value)
+        ? this.props.value
+        : [this.props.value];
+    }
```
flamerohr: Even after the discussion we've had, I don't agree with coercing the value to array is correct for `TreeDropdownMenu`.

It makes more sense to me that the value proptype should be accepting a string or array, rather than coercing to an array value.

I'll push up an update soon to address this.

Issue #217, opened 2017-08-31
Toolbar title is not aligned for Files and Campaigns sections


clarkepaul:
This one seems to have been 'fixed" a few times now but is again be out of alignment. All the sections use different HTML structure so maybe it makes sense to fix this properly. It seems to get broken when the breadcrumbs are updated so it would make sense to separate the title out of the breadcrumbs (this is a legacy pattern). ... After inspecting the code it seems we have empty divs where the breadcrumbs would be if they existed but on main levels, but we don't need them so best to remove the empty elements.

Files area and Campaigns area are out of alignment:
<img width="187" alt="pasted_image_31_08_17__3_58_pm" src="https://user-images.githubusercontent.com/555033/29905879-3ea4c85e-8e65-11e7-8d74-d7598ff46c87.png">

<img width="179" alt="pasted_image_31_08_17__3_57_pm" src="https://user-images.githubusercontent.com/555033/29905852-25569f12-8e65-11e7-8ef0-a1a83ff2966a.png">

CMS

![silverstripe_-_edit_page_](https://user-images.githubusercontent.com/318116/30258752-455314a4-9710-11e7-8bda-dc7bf6e04fa1.png)

## Excluded

- South bar height https://github.com/silverstripe/silverstripe-asset-admin/issues/634

## Pull requests

- [ ] https://github.com/silverstripe/silverstripe-admin/pull/254
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/630
- [ ] https://github.com/silverstripe/silverstripe-reports/pull/82
- [ ] https://github.com/silverstripe/silverstripe-siteconfig/pull/71
- [ ] https://github.com/silverstripe/silverstripe-cms/pull/1968
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7391

clarkepaul:
Also slightly out of alignment with breadcrumbs
<img width="171" alt="pasted_image_31_08_17__4_14_pm" src="https://user-images.githubusercontent.com/555033/29906200-809070f4-8e67-11e7-880b-89bb785329b4.png">

clarkepaul:
These south toolbars are also misaligned (top borders should align), caused from the buttons.
![image](https://user-images.githubusercontent.com/555033/30352682-6e4ff0aa-9875-11e7-959d-2965489b1e7e.png)

clarkepaul:
[...]

PR #254, opened 2017-09-14
Pulls/1.0/toolbar style layout

phalkunz:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/217


dhensby:
NPM tests failing :(

client/src/legacy/LeftAndMain.Content.js
```diff
@@ -89,7 +89,9 @@ $.entwine('ss', function($){
 
       // Fix dimensions to actual extents, in preparation for a relayout via jslayout.
       this.height('auto');
-      this.height(this.innerHeight()-this.css('padding-top')-this.css('padding-bottom'));
+      const paddingTop = parseInt(this.css('padding-top'));
+      const paddingBottom = parseInt(this.css('padding-bottom'));
```
flamerohr: I'd recommend using `parseInt(css, 10)` or `Number(css)` which will use base 10

flamerohr:
pages with a parent seem to have a bit too much padding/margin on the left:
screen shot 2017-09-18 at 2 04 06 pm

Not sure if the this was an effect of the toolbar fixes, but see if you can fix this as well:
screen shot 2017-09-18 at 2 02 27 pm

Security has too much gap between back button and breadcrumbs
screen shot 2017-09-18 at 2 02 04 pm

flamerohr:
Cross browser testing didn't reveal anything new with these pull requests, but it did reveal a regression from another story, which I don't think should be a blocker for this.

screen shot 2017-09-18 at 3 35 48 pm

client/src/styles/legacy/_actionTabs.scss
```diff
@@ -215,7 +215,8 @@ $border: 1px solid darken(#D9D9D9, 15%);
 			of ss-ui-action-tabset
 		****************************************************************/
 		&.action-menus.ss-tabset {
-			margin-top: 5px;
+			margin-top: 4px;
+			float: right;
```
flamerohr: is the float necessary?
I don't see a difference with or without it

Issue #65, opened 2017-05-14
New UX insert link dialog - Insert anchors


chillu:
Follow up task from https://github.com/silverstripe/silverstripe-framework/issues/6740

## Acceptance Criteria

 * I can insert an anchor into the CMS rich text editor
 * I can choose from existing anchors both on the current page, and another page
 * By default, the current page is selected
 * I can add a custom anchor
 * I can set a link description (prepopulated to selected text)
 * I can edit an existing anchor link by placing the cursor in the link and using the "insert" dialog
 * I can choose to open the anchor in a new window/tab

## Notes

 * Designs: https://projects.invisionapp.com/share/BGAYOTRFY#/screens/224873109

## Related

 * [Inline edit/remove of links in HTML editor](https://github.com/silverstripe/silverstripe-admin/issues/66)
 * [New UX insert link dialog - internal URLs](https://github.com/silverstripe/silverstripe-framework/issues/6740)
 
## PRs

 - [x] https://github.com/silverstripe/silverstripe-admin/pull/255
 - [ ] https://github.com/silverstripe/silverstripe-cms/pull/1966

tractorcow:
Won't be able to get this done in time for beta3. Moved to next milestone.

WIP at https://github.com/open-sausages/silverstripe-cms/tree/pulls/4.0/anchor-modal

tractorcow:
@clarkepaul question about the designs; I have spoken with @flamerohr and he suggested to look into a new approach.

Instead of a dropdown + textfield approach, could we use a single comobox (typeable dropdown) as a means of selecting an existing / creating a new value.

Change:

![image](https://user-images.githubusercontent.com/936064/30404369-873d77de-993a-11e7-903b-932b0e920fb6.png)

To:

![image](https://user-images.githubusercontent.com/936064/30404407-a9783898-993a-11e7-8a30-62564ecda807.png)

A demo of how the combobox could work can be tested at https://jedwatson.github.io/react-select/ (bottom of page, select single select).

clarkepaul:
Yeah that would be good if it was possible. I thought we didn't have that UI possibility though as I already asked for this pattern to be introduced for adding items to campaigns (select or add new) but I was told it was too hard to do, I think @flamerohr said it was too hard to do.

clarkepaul:
[...]

PR #255, opened 2017-09-14
Update link modals

tractorcow:
Required for https://github.com/silverstripe/silverstripe-admin/issues/65


client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -16,14 +18,17 @@ class FormBuilderModal extends SilverStripeComponent {
   }
 
   handleLoadingError(schema) {
-    if (this.props.showErrorMessage) {
+    const providesOnLoadingError =
+      typeof this.props.onLoadingError === 'function'
+      && this.props.onLoadingError !== noop;
```
flamerohr: Can simplify to `this.props.onLoadingError !== noop`, we have the propTypes validation which checks for a function type.
Users shouldn't be passing in `null` to the prop anyway.

tractorcow: Updated.

PR #260, opened 2017-09-15
Editing sslink-email, subject is still URI encoded

mikenz:
Allow email subjects with spaces in them to be edited without all the spaced becoming %20


dhensby:
Not sure how the npm test is failing from this. Is it pre existing?

mikenz:
npm test is probably failing as the dist files need updating - which I can't do from the github online editor.

PR #261, opened 2017-09-16
API: Pass the member to alternateAccessCheck

mikenz:


dhensby:
Thanks for all these tweaks @mikenz

PR #263, opened 2017-09-17
Catch invalid JSON in $(window).on("message")...

cjsewell:
I am using [webpack-dev-server](https://github.com/webpack/webpack-dev-server) to help with development. 
It uses the message event, sending non-json strings. Causing LeftAndMain.js to throw an exception.

This just add a simple try catch around the `JSON.parse` and rejects the message if it fails.


dhensby:
thanks :)

Issue #264, opened 2017-09-17
Performance recommendation: Avoid Config.update in LeftAndMain::init()


tractorcow:
Parent story: https://github.com/silverstripe/silverstripe-admin/issues/199

LeftAndMain::init() sets the following config changes:

- SSViewer.rewrite_hash_links
- ContentNegotiator.enabled
- SSViewer.themes

Since the first time the config is modified it switches from the cached config to a new uncached copy, it becomes an immediate performance hit.

I suggest to move these values to protected get/setters which have their defaults provided via config, but do not require config modification to set from their default.

For example, i18n::set_locale uses default_locale config to set the default, but modifying the current locale doesn't require a config update.

Expected performance saving: **>50ms** on /admin/pages load (expected saving should be more, since any subsequent read of these configs should rely on the base cached collection).

Reference: https://blackfire.io/profiles/dea4e266-e469-4976-abf9-c7844503b04b/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=SilverStripe%5CCore%5CConfig%5CConfig_ForClass%3A%3Aupdate&callname=main()

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7392
- [x] https://github.com/silverstripe/silverstripe-admin/pull/269

PR #269, opened 2017-09-21
Enahncement use the new setter and getter for LeftAndMain init

flamerohr:
This is instead of updating config... every request.

Please check https://github.com/silverstripe/silverstripe-admin/issues/264 for more details

Depends on https://github.com/silverstripe/silverstripe-framework/pull/7392

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -580,7 +580,7 @@ class LeftAndMain extends Controller implements PermissionProvider
     {
         parent::init();
 
-        SSViewer::setRewriteHashLinks(false);
+        SSViewer::setRewriteHashLinksDefault(false);
         ContentNegotiator::setEnabled(false);
 
         // set language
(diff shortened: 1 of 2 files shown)


code/LeftAndMain.php
```diff
@@ -733,9 +733,9 @@ protected function init()
...
         // Load the editor with original user themes before overwriting
         // them with admin themes
-        $themes = HTMLEditorConfig::config()->get('user_themes');
+        $themes = HTMLEditorConfig::getThemes();
         if (empty($themes)) {
-            HTMLEditorConfig::config()->set('user_themes', SSViewer::get_themes());
+            HTMLEditorConfig::setThemes(SSViewer::get_themes());
         }
 
         // Assign default cms theme and replace user-specified themes
```
tractorcow: The next line below still triggers a config::set inside SSViewer::set_themes().

PR #272, opened 2017-09-22
Add browser warning

phalkunz:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/6719


client/src/components/BrowserWarning/BrowserWarning.js
```diff
@@ -0,0 +1,6 @@
+import browser from 'detect-browser';
+import $ from 'jquery';
+
+if (browser.name === 'ie' && parseInt(browser.version, 10) <= 10) {
+  $('.browser-warning-wrapper').addClass('browser-warning-wrapper--show');
+}
```
flamerohr: I think this would be more suited in the legacy folder than inter-mingled with the components, or even the lib folder.
This doesn't have any "component" exported.

client/src/styles/bundle.scss
```diff
@@ -63,6 +63,7 @@
 @import "../components/Modal/Modal";
 @import "../components/TreeDropdownField/TreeDropdownField";
 @import "../components/TinymceInlineToolbar/TinymceInlineToolbar";
+@import "../components/BrowserWarning/BrowserWarning";
```
flamerohr: should this be imported in the main bundle if it's already getting its own separate output file?
I'd suggest including the css file in the template which includes the separate javascript file as well.

webpack.config.js
```diff
@@ -59,6 +63,10 @@ const config = [
     entry: {
       bundle: `${PATHS.SRC}/styles/bundle.scss`,
       editor: `${PATHS.SRC}/styles/editor.scss`,
+      // For IE version 9 and below. These browsers doesn't handle large
+      // resource files so need to break browser detection and warning code into
+      // its own file
+      browser_warning: `${PATHS.SRC}/styles/browser-warning.scss`,
```
flamerohr: naming consistency, this should be `'browser-warning'`

templates/SilverStripe/Admin/Includes/BrowserWarning.ss
```diff
@@ -0,0 +1,26 @@
+<!--[if lte IE 9]>
+    <link rel="stylesheet" href="{$BaseURL}silverstripe-admin/client/dist/styles/browser_warning.css">
+<![endif]-->
+
+<div class="browser-warning-wrapper">
+    <div class="browser-warning vertical-align-items">
+        <div class="browser-warning__content flexbox-area-grow">
+            <p class="ss-logo">SilverStripe</p>
+            <h1>Unsupported browser</h1>
```
chillu: Make this translatable please

templates/SilverStripe/Admin/Includes/BrowserWarning.ss
```diff
@@ -0,0 +1,31 @@
+<!--[if lte IE 9]>
+    <link rel="stylesheet" href="{$BaseURL}silverstripe-admin/client/dist/styles/browser-warning.css">
```
flamerohr: I remember there was a `<% require css("cms/css/TreeSelector.css") %>` syntax you could use, similar to the javascript one below

templates/SilverStripe/Admin/Includes/BrowserWarning.ss
```diff
@@ -0,0 +1,31 @@
+<!--[if lte IE 9]>
```
flamerohr: are you sure about not including the css file for ie10? as above I saw `version <= 10`

templates/SilverStripe/Admin/Includes/BrowserWarning.ss
```diff
@@ -0,0 +1,31 @@
...
+            </ul>
+        </div>
+    </div>
+</div>
+
+<% require javascript("silverstripe-admin/client/dist/js/browserWarning.js") %>
+
+<!--[if lte IE 8]>
+    <script>
+        document.getElementById('browser-warning-wrapper').className = 'browser-warning-wrapper';
```
flamerohr: is this needed because browser-detect doesn't detect ie8 properly? almost seems counter-intuitive for this to happen...

templates/SilverStripe/Admin/Includes/BrowserWarning.ss
phalkunz: Dang! I was so focused on working on the workaround for IE 9 and below. Good catch!

templates/SilverStripe/Admin/Includes/BrowserWarning.ss
phalkunz: This is because the script in `browserWarning.js` doesn't execute on IE8 (to be honest I'm not sure why exactly).

templates/SilverStripe/Admin/Includes/BrowserWarning.ss
flamerohr: I think that's fine for this css file, it goes hand-in-hand with the javascript file

Issue #229, opened 2017-09-07
Popover and tooltip no longer have arrows pointing elements initiating them


phalkunz:
![popever_and_tooltip_no_longer_have_arrows](https://user-images.githubusercontent.com/318116/30145384-1241133c-93e5-11e7-97c1-3fde53d0826d.png)

![silverstripe_-_campaigns_](https://user-images.githubusercontent.com/318116/30145385-126aa756-93e5-11e7-946f-1ba3dfe7c4e0.png)

# Pull request(s)

- [x] https://github.com/silverstripe/silverstripe-admin/pull/279
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/650

chillu:
See https://github.com/silverstripe/silverstripe-admin/issues/275 for more use cases.

PR #279, opened 2017-09-27
Fix tooltip arrow style and position

phalkunz:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/229


dhensby:
there are some linting issues with this (and it needs rebase/rebuild)

flamerohr:
Arrow is on the wrong side when in mobile
screen shot 2017-10-02 at 11 39 39 am

client/src/components/Tooltip/Tooltip.scss
```diff
@@ -11,3 +11,77 @@
   }
 }
 
+/**
+ * Tooltip arrow style below is copied (unless stated otherwise) from Bootstrap
+ * v4 beta because there's a missmatch between classname in react-bootstrap and
+ * its counterpart in Bootstrap v4.
```
flamerohr: could this use `@extend` so it's not hardcoded-copying? use `// sass-lint:disable-block extend-rule-name` for linting with the comment above

phalkunz: This requires us to extends `.tooltip .arrow` but unfortunately we can't extend a nested selector.

PR #282, opened 2017-10-03
Add style for gridfield status badges

phalkunz:
Fixes https://github.com/silverstripe/silverstripe-framework/pull/7422


tractorcow:
colours for all of the badges are the same.

dhensby:
@tractorcow is that now resolved?

tractorcow:
Turns out that we should only show "draft" states, so it's ok for it to be the same colour.

PR #288, opened 2017-10-06
Give more space to profile name and remove "hi" before name

clarkepaul:
This PR gives the profile name more space before being truncated and gives the profile, log out and site name a background color on hover to make them more interactive. I'm not sure how some of the css files updated based on my changes so needs a check over.

To merge https://github.com/silverstripe/silverstripe-framework/pull/7453 afterwards


client/src/components/Menu/Menu.scss
```diff
@@ -224,10 +229,6 @@
 }
 
 .cms-menu__header {
-  position: relative !important; // sass-lint:disable-line no-important
-  top: auto !important; // sass-lint:disable-line no-important
-  // Required for JLayout
-  height: auto !important; // sass-lint:disable-line no-important
```
clarkepaul: Removed as no longer required as Jlayout was removed as far as I'm aware (and tested).

client/src/styles/_bootstrap-variables.scss
```diff
@@ -269,7 +269,7 @@ $font-size-sm: .923rem; // 12px
 //$font-weight-bold: bold !default;
 
 //$font-weight-base: $font-weight-normal !default;
-$line-height-base: 1.539;
+$line-height-base: 1.5384;
```
clarkepaul: Ensures line-height remains under 20px rather than over 20px

templates/SilverStripe/Admin/Includes/LeftAndMain_MenuStatus.ss
```diff
@@ -3,8 +3,7 @@
         <a href="{$AdminURL}myprofile" class="cms-login-status__profile-link">
             <i class="font-icon-torso"></i>
             <span>
-                <%t SilverStripe\\Admin\\LeftAndMain.Hello 'Hi' %>
-                <% if $FirstName %>$FirstName<% else %>$Email<% end_if %>
+                <% if $FirstName && $Surname %>$FirstName $Surname<% else_if $FirstName %>$FirstName<% else %>$Email<% end_if %>
```
flamerohr: oh, are we bringing back the full name?

clarkepaul: I know it was only removed but that was done to allow more space, I think removing the last name was one way to give more space to the first name but I thought it made more sense to actually adjust the CSS to provide more space to the name so there was less chance of using the ellipse. Removing the "Hi" also provide more space for the name.

I'm hoping this change makes sense, open to feedback but in my mind it's a good improvement :)

flamerohr: cool, I had a similar idea when I merged the surname removal :)

lets put it back in for now then, there will need to be a framework pull request made to fix the behat tests for surname displayed - I'll create that now

flamerohr:
I've rebased with the correct build files

clarkepaul:
@flamerohr I did the same thing regarding rebuilding, did I overwrite your changes? Hope I did it right if I did.

clarkepaul:
Nice one cheers

PR #291, opened 2017-10-08
Allow breadcrumbs to define HTML support or not.

wilr:
Required to add support for custom breadcrumb designs such as https://projects.invisionapp.com/share/2MDS99OTZ#/screens/256544889. This allows the breadcrumb to specify either HTML or Varchar support.

See related PRs:

- [x] https://github.com/silverstripe/silverstripe-cms/pull/1985
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7449

### DIFF
--- a/templates/SilverStripe/Admin/Includes/CMSBreadcrumbs.ss
+++ b/templates/SilverStripe/Admin/Includes/CMSBreadcrumbs.ss
@@ -2,9 +2,9 @@
 	<h2 id="page-title-heading">
 		<% loop $Breadcrumbs %>
 			<% if $Last %>
-				<span class="cms-panel-link crumb last">$Title.XML</span>
+				<span class="cms-panel-link crumb last">$Title</span>
 			<% else %>
-				<a class="cms-panel-link crumb" href="$Link">$Title.XML</a>
+				<a class="cms-panel-link crumb" href="$Link">$Title</a>
 				<span class="sep">/</span>
 			<% end_if %>
 		<% end_loop %>


wilr:
Happy to do it either way, this way seemed to involve the least changes and would be just as secure for core modules. Overriding Breadcrumbs would be a fairly niche use case.

Subtitle field would make sense as well, happy to rewrite to use a new `Subtitle` field though might cause issues if objects have a `Subtitle` - seems excessive given this is only for CMS breadcrumbs template

dhensby:
> This could open up a whole can of XSS potential issues, could we perhaps have a Subtitle data property to go alongside the Title data property?

Would it? It's cast as `Varchar` so the `.XML` from the template is implied by default, I believe? so this should be safe to render.

flamerohr:
> It's cast as `Varchar` so the `.XML` from the template is implied by default

Oh, I was not aware of the implied factor, if that's true, then that's ok then

tractorcow:
You don't need the Varchar convert; It's the default_cast. You can leave out the `.XML` entirely and it'll be safe.

code/LeftAndMain.php
```diff
@@ -1163,25 +1163,32 @@ public function Breadcrumbs($unlinked = false)
     {
         $items = new ArrayList(array(
             new ArrayData(array(
-                'Title' => $this->menu_title(),
```
tractorcow: It's safe to drop all the changes to LeftAndMain.php

Issue #290, opened 2017-10-06
Button sizes have an inconsistent height and icon alignment out 


clarkepaul:
Button heights are still slightly too big (only by a pixel for the majority but some extend to 4px bigger) on buttons with icons, the total height based on dimensions should total 32px (20px line-height, 5px padding top/bottom and 1px borders).

Before: You can see the height of the main actions are 35px when they should be 32px
![image](https://user-images.githubusercontent.com/555033/31323823-2cc552bc-ad09-11e7-8a5f-f2ef1bef04fb.png)

clarkepaul:
Seems this is caused by mixing Flex box layouts and non-flex box together—you get some weird results, putting a PR together for this... hopefully.

clarkepaul:
PR https://github.com/silverstripe/silverstripe-admin/pull/295

PR #295, opened 2017-10-09
Change buttons to use flexbox layout in flex box areas…

clarkepaul:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/290

Included change to preview switch as that had a lot of issues which
needed addressing so made some improvements to its clarity based on
user feedback

Before: (button height was 35px)
image

After: (now are 32px as they should be)
image

Tested in IE 11:
pasted_image_9_10_17__2_34_pm


clarkepaul:
Cheers

Issue #286, opened 2017-10-04
Layout issues with Page Title


NightJar:
![capture](https://user-images.githubusercontent.com/778003/31156445-dfa06996-a910-11e7-9678-5f4be92895de.PNG)

Upon default fresh latest `composer create-project tryitout 4.x-dev` there is an issue with the Title field, it looks fine at wider real-estate sizes, however the split screen by default shows it seeming to adapt to the height of it's label, which is now on two lines. However this is different to the behaviour of the field for the Navigation Label below it, which is more what I'm accustomed to.

The screenshot attached shows a trial of entering a load of data to see if it would wrap, and it doesn't. That's fine, but then the magic update of the title atop the admin looks completely naff also.

There is one caveat in this test which I would hope is not of consequence, but could be:
LastPass is apparently doing some injection to the Title field, which in itself is mildly interesting. Dunno why.

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-admin/pull/304
- [x] https://github.com/silverstripe/silverstripe-cms/pull/1991
- [x] https://github.com/silverstripe/silverstripe-reports/pull/87

chillu:
Grrrrr, we really need to test UI work with unusual data by default - too much "happy path" development.

clarkepaul:
Same issue I think but this is a screenshot from @wilr on mac. 
![image](https://user-images.githubusercontent.com/555033/31323262-df581b42-ad01-11e7-9229-51f1d34df101.png)

clarkepaul:
This PR fixes the input height https://github.com/silverstripe/silverstripe-admin/pull/296 it just needed a margin-bottom:  auto;

I took a look at the Toolbar issue and it wasn't an quick fix, now that the Pages toolbar is flexbox none of the floats will work and with the complexity of all the nesting it isn't a quick fix. Checkout this example for a possible solution to truncating the toolbar title http://jsfiddle.net/iamanubhavsaini/zWtBu/ . This would need a rebuild to get it working as intended or revert to non-flex box to return the floats as a potential quicker temporary fix.

PR #296, opened 2017-10-09
Removed gradient from accordion, includes fix for input height bug in flexbox

clarkepaul:
image
Removed the gradient from accordion (maybe the last gradient?) to match the rest of the UI.

Also includes fix for input heights https://github.com/silverstripe/silverstripe-admin/issues/286 (doesn't include fix for toolbar though)


clarkepaul:
Before:
image

How it looks now:
image
image

PR #301, opened 2017-10-11
NEW Allow individual form fields to opt out of divider lines

robbieaverill:
You can add `$field->addExtraClass('form-field--no-divider')` to remove a single form field's divider line, and can still use `form--no-dividers` on a container to remove them for a whole form.

For consideration after beta4


client/src/components/Form/Form.scss
```diff
@@ -53,9 +53,11 @@
...
 }
 
-// Remove underline and column offset
-// Used on places like tree batch actions
-.form--no-dividers .form-group {
+// Remove underline and column offset. Used on places like tree batch actions.
+// Apply form--no-dividers to a form element, or form-field--no-divider to an individal
+// form field.
+.form--no-dividers .form-group,
+.form-group.form-field--no-divider {
```
flamerohr: nice one, I think `.form-group--no-divider` is sufficient, no need to nest/combo classes :D

robbieaverill: Fair enough, updated

Issue #257, opened 2017-09-14
Breadcrumb/header alignment issue for smaller viewports


clarkepaul:
This was observed on the fromeworktest module for an example of gridfield viewed at a smaller viewport size.
<img width="162" alt="pasted_image_14_09_17__4_35_pm" src="https://user-images.githubusercontent.com/555033/30459049-603a1da4-9a02-11e7-8ecb-827f2abe580a.png">

Fixed with https://www.github.com/silverstripe/silverstripe-admin/issues/286

chillu:
It's broken in different ways now. Can we please get this section of the UI sorted properly please? This seems to break every second week :( The outer items use flexbox, but the inner ones don't. @phalkunz in case you pick this up, and you need to influence markup to make the CSS work (direct children for flexbox etc), then just talk to Chris or Damian to help you out. 

![image](https://user-images.githubusercontent.com/111025/31281019-26a9d14c-aaa6-11e7-9d7f-9b850b6af1a9.png)

![image](https://user-images.githubusercontent.com/111025/31281805-6b749896-aaa8-11e7-834c-5e067c99a994.png)

PR #304, opened 2017-10-12
BUG Fix up markup and classes to work better with flexbox

flamerohr:
Please see https://github.com/silverstripe/silverstripe-admin/issues/286 for more details


flamerohr:
Have added commits to fix https://github.com/silverstripe/silverstripe-admin/issues/257

PR #306, opened 2017-10-12
Update listbox field styles

kinglozzer:
Minor update so it doesn’t stick out like a sore thumb 😅.

Before:
screen shot 2017-10-12 at 16 55 41
screen shot 2017-10-12 at 17 00 46

After:
screen shot 2017-10-12 at 16 53 14
screen shot 2017-10-12 at 17 02 05


newleeland:
It looks great!. Great to have the old tree dropdown updated, as its still used in some areas of the CMS. Thanks you for your work @kinglozzer. We are eventually going to replace all tree dropdowns with the new react variant. SEE BELOW

screen shot 2017-10-26 at 11 16 20 am
 @kinglozzer

PR #312, opened 2017-10-18
BUG fix clipping dropdown content

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/663


newleeland:
The Clipping fix works great on:
Firefox, Chrome, Safari. However, I can't test this on IE as action bar with move button does not appear once an item is selected.

flamerohr:
Not sure why it's not working for you in IE - that's weird

here's a screenshot of the fix in IE11
screen shot 2017-10-19 at 3 58 38 pm

PR #313, opened 2017-10-18
BUG Fix issue where you need to add 'menu_icon_class = false' to a ModelAdmin that has a defined "menu_icon" or else it won't show.

SilbinaryWolf:
BUG Fix issue where you need to add 'menu_icon_class = false' to a ModelAdmin that has a defined "menu_icon" or else it won't show.


kinglozzer:
Duplicate of https://github.com/silverstripe/silverstripe-admin/pull/311?

SilbinaryWolf:
Yep

PR #316, opened 2017-10-20
BUG add bootstrap4 specific classes for readonly/disabled radio/checkbox

flamerohr:
Also, prevent handleChange trigger when disabled/readonly

Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/662

I've discussed with @newleeland and we both agree to leave the text colour the same black - bootstrap's faded grey colour does not comply with WCAG.


client/src/components/OptionsetField/OptionField.scss
```diff
@@ -0,0 +1,5 @@
+.option-field--disabled {
+  label, input[type="radio"], input[type="checkbox"] {
+    cursor: not-allowed;
+  }
+}
```
tractorcow: Instead of requiring a CSS class, can we please respect the underlying html prop?

```css
input[disabled], input[readonly] {
	cursor: not-allowed;
}
```

flamerohr: some "do not respond to clicks" behaviours only come with the disabled attributes, unfortunately.

that css is fine - but labels wouldn't get the cursor style without the class anyway

PR #317, opened 2017-10-22
FIX amend preview URL correctly, more flexibly

NightJar:
When using either split mode or preview mode while
working with content in the CMS, the preview pane
would simply tack ?CMSPreview=1 on to the end of
whatever URL it is given. This is not ideal when
working with e.g. Elemental, as there is a fragment
for jumping to the relevant content in the preview,
so we end us with some/page?Stage=stage#someBlock&CMSPreview=1
which counts as neither a CMSpreview query, nor a vaild (existing)
fragment identifier. So I've patched it to correctly amend to
e.g. some/page?Stage=stage&CMSpreview=1#someBlock


client/src/legacy/LeftAndMain.Preview.js
```diff
@@ -453,8 +453,13 @@ $.entwine('ss.preview', function($){
       }
 
       // Mark url as a preview url so it can get special treatment
-       if (url) {
-          url += ((url.indexOf('?') === -1) ? '?' : '&') + 'CMSPreview=1';
+      if (url) {
+        let urlFrag = url.split('#');
+        let urlBits = urlFrag.shift().split(/[?&]/);
+        let urlBase = urlBits.shift();
```
flamerohr: urlBits and urlBase should be using `const`
yes, we disabled eslint for legacy files but let's not let standards slip too much :D

NightJar: Neato, wasn't aware. Thanks, will fix :)

flamerohr: no, it's fine actually :)

`const` means a constant reference point, you can still mutate/change it but cannot reassign it.
if you had done `urlBits = anotherBit` then it would complain.

NightJar: Right, so `const` isn't a `constant`. Javascript ✅ Gotcha.

flamerohr: Yeah Javascript 👍 
To make a true constant, you'd need a recursive `Object.freeze()` like `deepFreeze` then go `const MY_CONSTANT = deepFreeze(anything);`

NightJar: Ironically (not actually ironically), webpack knew this already. Only thing to commit after change & build was the source file, dist unchanged! :)

PR #318, opened 2017-10-24
Add class to help entwine determine when injector becomes available

flamerohr:
I think this and the existing `.js-react-boot` class added is not a good solution... :/ but it'll work for entwine at least.

e.g. prefix with `js-react-boot` if only relying on React from SilverStripe
```
$('.js-react-boot .TreeDropdownField').entwine({
  onmatch() {
    // stuff here
  },
});
```

or prefix with `js-injector-boot` if relying on SilverStripe's injector
```
$('.js-injector-boot .MyFormContainer').entwine({
  onmatch() {
    // stuff here
  },
});
```

When I get more time, I'd like to investigate using the javascript event system to trigger these instead of using a class selector/marker.


flamerohr:
I've added a `Injector.ready` API endpoint to injector :)

This will collect all callbacks before Injector is loaded and those callbacks are called after Injector is loaded.
Callbacks added after Injector is loaded are pass-through and just called immediately.

As proof of concept for this, I've applied the `Injector.ready` API to the boot in SilverStripe's admin.

tractorcow:
Request peer review from @unclecheese please. :)

client/src/boot/index.js
```diff
@@ -59,14 +59,20 @@ function appBoot() {
...
+    routes.start(window.location.pathname);
+    if (window.jQuery) {
+      window.jQuery('body').addClass('js-injector-boot');
+    }
+  });
+
+  // Force this to the end of the execution queue to ensure it's last.
+  window.setTimeout(() => {
+    Injector.load();
   }, 0);
```
unclecheese: Since this is now just a single method call, we don't need a function literal. `window.setTimeout(Injector.load, 0)`

client/src/lib/dependency-injection/tests/InjectorContainer-test.js
```diff
@@ -16,6 +16,35 @@ describe('InjectorContainer', () => {
...
+      const callback = jest.fn(() => 'hello');
+
+      container.load();
+      expect(callback).not.toBeCalled();
+
+      container.ready(callback);
+      expect(callback).toBeCalled();
+    });
+  });
+
```
unclecheese: Let's add a test for an invalid callback.

PR #325, opened 2017-10-26
NEW Allow popover menu icon size class to be adjustable via class props

### LINKED ISSUE #323: Allow PopoverField icon size to be configurable
**SS: 4.0.0-rc1**

Currently the PopoverField component adds `btn--icon-xl` to all PopoverField instances.

It would be nice to be able to control the size of the meatballs icon it adds either via class properties or a public API method of some sort. We wanted to use `btn--icon-md` in one instance of this component, but weren't able to modify this.

Code:

https://github.com/silverstripe/silverstripe-admin/blob/1.0.0-rc1/client/src/components/PopoverField/PopoverField.js#L56

robbieaverill:
Minor enhancement to allow the icon size class to be adjustable via the injected class properties.

E.g.:

```js
return (
  <PopoverField
    id={this.props.id}
    popoverClassName="mypopover__menu"
    buttonSizeClassName="btn--icon-md"
    data={{ placement: 'bottom' }}
    container={this.props.container}
  >
    {children}
  </PopoverField>
);
```

Resolves #323


client/src/components/PopoverField/PopoverField.js
```diff
@@ -103,6 +115,7 @@ PopoverField.defaultProps = {
   data: {},
   className: '',
   popoverClassName: '',
+  buttonSizeClassName: '',
```
flamerohr: default `btn--icon-xl` here :)

Or if we go with just `buttonSize` prop, this could be just `xl`

client/src/components/PopoverField/PopoverField.js
```diff
@@ -89,6 +100,7 @@ PopoverField.propTypes = {
   container: React.PropTypes.any,
   className: React.PropTypes.string,
   popoverClassName: React.PropTypes.string,
+  buttonSizeClassName: React.PropTypes.string,
```
flamerohr: how about a `buttonSize` prop instead and a validation with `PropTypes.oneOf(['xl', 'md', 'sm', 'xs']);`

client/src/components/PopoverField/PopoverField.js
robbieaverill: Sounds good, will update

robbieaverill:
Updated, thanks for reviewing :)

client/src/components/PopoverField/PopoverField.js
```diff
@@ -16,7 +16,7 @@ class PopoverField extends Component {
   /**
    * Get popup placement direction
    *
-   * @returns {String}
+   * @return {String}
```
flamerohr: oops, missed this... valid JSDocs is `@returns` :P

robbieaverill: Oh!

Issue #309, opened 2017-10-16
Remove the circular gradient from the main CMS loading screen


clarkepaul:
We could simply fall back to the generic background color which would be #f6f7f8 (unless that changes in the meantime).
<img width="400" alt="pasted_image_16_10_17__5_30_pm" src="https://user-images.githubusercontent.com/555033/31595394-fe4a761c-b297-11e7-9257-e86d3fa6c7ca.png">

We should also not use the loading S next to the main logo but that can be looked at as part of a separate issue.

clarkepaul:
PR https://github.com/silverstripe/silverstripe-admin/pull/330

PR #330, opened 2017-10-30
Loading screen updates

clarkepaul:
* Removed gradient from the loading screen (Fixes https://github.com/silverstripe/silverstripe-admin/issues/309).
* Removed old logo with "create the web" as this is no longer used for OS and replaced with simple font-icon logo. Replaced SS loading indicator on main loading page as it looks weird having two SS logos. Was:
image
Now has animated loading text (quicker and smaller than gif below):
loading

* Added loading text to the main loading screen so that it is accessible.
* Replaced the background mask on inner loading screens to a black with opacity rather than a light grey to give a cleaner look.
* Added style to the noscript message (there were no styles). Now:
pasted_image_30_10_17__11_06_pm


dhensby:
This PR is pretty confusing as it's proporting to do one thing but is pulling in a lot of commits from the `1` branch and it's not clear if that's intentional or not or if this is targeted against the wrong branch,

I've cherry-picked the relevant commit against the 1.0 branch so hopefully I've done the right thing (I'll leave this here to be merged once we've got clarification.

clarkepaul:
Cheers, you've understood what I was trying to do (I take it I had other changes that weren't mine which you removed). Seems all my updates are still there :)

PR #335, opened 2017-11-01
BUG Do not hide all columns in the gridfield when in mobile view

flamerohr:
The style is too aggressive, perhaps it could be opt-in instead?
screen shot 2017-11-01 at 4 38 36 pm

cc @newleeland @clarkepaul


clarkepaul:
I agree they are too aggressive and assumes the first column is the most important but without having a dev at the time of adding those styles it was the best option I had and worked with what we had. We will be reducing all gridfield actions into a more options dropdown (...) so we would be able to have those actions available on mobile as well. I thought we should stay away from making the gridfields scrollable, I'm not sure that is a good option. Would be good to ba able to specify the columns wanted for mobile (or not wanted) so we can hide the rest. This could be achieved by adding a class main/secondary column classes perhaps?

The screenshot looks like content blocks, we would need to add our own mobile overriding styles there as the columns differ quite a bit (not a show stopper and relatively easy to override).

I don't think a mobile user needs all the data that's presented in the desktop view but enough to distinguish between items so they can navigate deeper.

flamerohr:
It's the Elemental reports actually :)

How about we keep this aggressive behaviour something in the middle?
We'll add generic "always show" columns such as the "Title" and "Actions" columns? We can still keep the first column in the table.

PR #338, opened 2017-11-03
BUG Fix react-select does not return the true value

flamerohr:
when the option is missing

Fixes https://github.com/silverstripe/silverstripe-cms/issues/2008


tractorcow:
This has regressed the asset-admin

https://s3.amazonaws.com/silverstripe-travis-artifacts/silverstripe/silverstripe-asset-admin/302183318/302183320/screenshots/manage-files.feature_189.png

There should be a `(Top level)` option that's no longer visible.

PR #340, opened 2017-11-06
Feature debounce change events in changetracker

flamerohr:
This is to reduce change event load during the main event

Found that changetrackewr stuck out (aside from tinymce core performance metrics) when profiling performance in Firefox/IE when making changes.

NOTE: this change will need testing to see if change detection isn't too slow, so immediate button clicks after making a change are behaving as expected.


colintucker:
Just adding my two cents. This issue is a fairly major priority for me, as typing in TinyMCE is *super* laggy for all projects currently built upon 4.x, to the point of receiving client complaints about it.  It's not just old hardware or browsers either, it's also slow for me in Firefox Quantum + Win 10, with decent hardware.

I've just merged this fix and rebuilt the admin client bundles for a test app here, and I can confirm that this fix does indeed rectify this issue; or at least, gives a major speed boost to entering text within `HTMLEditorField`.  The client (who is stuck with IE11, poor soul) has also confirmed that the fix has made a significant difference to the speed. 

👍

flamerohr:
Rebased, will test soon then push to peer review :)

flamerohr:
replies for @dhensby :)

> @flamerohr I think it may be worth having a "leading" and "trailing" execution which would allow us to immediately change button states on change as well as unset the button state if the change went back immediately (eg: adding and deleting a character from an input)

That's a neat idea, and touches on one of the testing criteria I wanted to cover before I push it to peer review... immediately change button states would mean blocking immediate button clicks (e.g. click before you blur the changed field) since the form is marked as changed until the debounce happens - which was my main testing criteria.
Do we block until the change tracker kicks in, or let this go through until the change tracker kicks in? the current implementation is the latter.

> Is lodash (or underscore) a dependency we have at the moment? I think it'd add a bit more clarity if we just used a pre-authored debounce method than writing it bespoke just for this instance.

No it is not, I think we could import just the debounce function from lodash since our webpack build does a little bit of tree shaking.
I'll need to assess how much it increases the bundle sizes with this though, potentially no increase if one our other dependencies depends on lodash already.

> Do we also need to consider to cancelling (clearing the timeout for) the debounce if (say) the save button is clicked or some other action happens that means a deferred change track event shouldn't be happening?

Yeah, it is possible that we need something like this, I'd prefer to test and see if we really need something like this - as feels like it can easily become a spaghetti code mess.

flamerohr:
I have switched to using lodash debounce, filesize went from `1967389` to `1969141`, not too bad.

I've set options `leading` and `trailing` to true, so it's guaranteed to run once immediately and once again after there's a quiet period.

Here are two images of the performance profile I get before and after debouncing
Before:
screen shot 2018-01-31 at 11 55 15 am

After:
screen shot 2018-01-31 at 11 57 06 am

@colintucker would you have time to check if this continues to give you some performance gain? :)

colintucker:
@flamerohr Sure thing, I'll check it out. 👍

tractorcow:
Pushed up an enhancement; Now that we have debounce we can capture keyup too.

PR #341, opened 2017-11-07
NEW: Remove base tag from CMS

sminnee:
This is part of work to make the base tag unnecessary.

Initial testing suggests that this will work without changes to other admin modules.


michalkleiner:
Please please :)

maxime-rainville:
I've rebased the PR which should hopefully make Travis happy.

maxime-rainville:
@sminnee Do you want to have another look at this one?

sminnee:
I'll have another look...

sminnee:
OK I've rebased and fixed the bug with the re-login form.

sminnee:
This change is moderately risky (a small change but it affects the whole admin) and I wonder if it's best to embargo it until after 4.4 is branched? It's also arguably a breaking change to remove the base tag, since it could potentially break the javascript of other admin modules, if they also had hardcoded base-relative URLs.

Perhaps I should add a config option LeftAndMain.include_base_tag, which is true by default? That way people can turn this off incrementally and we could look at removing the base tag entirely in a major release...

ScopeyNZ:
> Perhaps I should add a config option LeftAndMain.include_base_tag, which is true by default? That way people can turn this off incrementally and we could look at removing the base tag entirely in a major release...

I like this. I would even consider a comment indicating that it won't just be set to `default: false` in 5 but it will be flat out removed.

maxime-rainville:
Sounds like a solid course of action.

maxime-rainville:
I think we've missed the 4.4 deadline by a bit. I've rebased. Let's see if the behat failure is still there.

client/src/legacy/LeftAndMain.js
```diff
@@ -1013,8 +1013,9 @@ $.entwine('ss', function($) {
 
     renderModal(isOpen) {
       // Build properties
+      const baseUrl = Config.get('baseUrl').replace(/\/+$/, '');
       const tempid = $('body').data('member-tempid');
-      const url = $.path.addSearchParams(Config.get('baseUrl') + '/CMSSecurity/login', {
```
maxime-rainville: This wasn't quite working because `Config.get('baseUrl')` would be a `/` when your site is hosted in the root of your domain ... which would produce a URL of `//CMSSecurity/login` and be interpretated has `http://CMSSecurity/login` by browsers.

Cheddam:
@sminnee Keen to get this tidied up for a release in 4.7? If not, we'll close this out for now.

sminnee:
Let me see if I can get some hack time on this in the next week or so.

sminnee:
It's been so long, I can't remember. I think the main thing was that the change would require a fair bit of testing to ensure that it hadn't broken some forgotten corner of the system (or a module)...

Issue #201, opened 2017-08-21
Clean out legacy CSS


chillu:
We've got a bad habit of just piling new CSS on top of old CSS in the CMS. While BEM has brought a bit more structure to the CMS styling as a whole, there's still a lot of potentially unused CSS. 

It looks like there's no easy way to audit all CMS views in their entirety - only a single view is possible e.g. with the Chrome Dev Tools. That's not very useful. See https://unused-css.com/ for a comparison of options.

# Related

 * [Remove all legacy CSS browser hacks](https://github.com/silverstripe/silverstripe-framework/issues/6298)
 * [Use Bootstrap styling throughout the CMS](https://github.com/silverstripe/silverstripe-framework/issues/5650)
 * [Clean out ss-ui-button style use](https://github.com/silverstripe/silverstripe-framework/issues/6479)
 * [Remove jquery-ui](https://github.com/silverstripe/silverstripe-admin/issues/411)

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-admin/pull/412

phalkunz:
I couldn't find a good tool for collecting unused selector across different UI states and pages. I wrote a script in an attempt to cover do that. We still have to click on the interface and it won't cover everything. Hopefully, we can get a small enough number of unused CSS selectors that we can verify by hand.

```js

/** 
 * The script below accumulatively collects and reports unused CSS rule
 * selectors.
 * 
 * Usage: 
 *  1. Include the script in a place where it runs on all page, or create
 * a Chrome snippet in Devtool's Source tab and run it manually after each page
 * load 
 *  2. Try every possible user interaction possible and observe the number
 * of unused CSS selectors in the browser console 
 *  3. When the number of unused selectors is, hopefully, small enough (to
 *  verify manually in the codebase), run `unCssReportUnuseds()` in the console
 *  to get the list of all unused CSS selectors
 * 
 * This script uses `localStorage` to retain the list of CSS rules and unused selectors between pages. To reset the data, run `unCssReset()` in the console.
 */

/**
 * Returns a flat list of all CSS rules from all stylesheets on the page
 * 
 * @return {array} of {CSSStyleSheet}
 */
function allCssRules() {
    const styleSheets = Object.values(document.styleSheets);
    const ruleList = Object.values(styleSheets.reduce((acc, curr) => {
        return acc.concat(curr.cssRules);
    }, []));

    return ruleList.reduce((acc, curr) => {
        return acc.concat(Object.values(curr));
[...]

PR #346, opened 2017-11-07
Remove older style variables and replaced with newer ones

clarkepaul:
Replaced older variables with newer ones where possible.
Most notable visual differences are seen in the preview/edit mode toggle dropdown:
image
 
Requires asset-admin PR https://github.com/silverstripe/silverstripe-asset-admin/pull/690
Related/contributes to issue https://github.com/silverstripe/silverstripe-admin/issues/201

I've checked the main repo's (security, cms, framework, campaign-admin) to make sure they weren't being referenced but it could be that some might be used in modules? CWP modules are being re-checked with SS4 in the coming months so it would be good to see this get in before all of the main modules get re-checked with SS4.


client/src/styles/_bootstrap-variables.scss
```diff
@@ -50,16 +50,15 @@
...
-$gray-900: darken($gray-700, 20%); // old $gray-darkest
+$gray-700: #43536d !default;
+$gray-100: lighten($gray-700, 60%) !default;
+$gray-200: lighten($gray-700, 50%) !default;
+$gray-300: lighten($gray-700, 40%) !default;
+$gray-400: lighten($gray-700, 30%) !default;
+$gray-500: lighten($gray-700, 20%) !default;
+$gray-600: lighten($gray-700, 10%) !default;
+$gray-800: darken($gray-700, 10%) !default;
+$gray-900: darken($gray-700, 20%) !default;
```
flamerohr: I don't mind using `lighten()` and `darken()` in variables - could we put the calculated colour codes next to them in comment?
Makes searching for colour codes much much easier e.g. especially when we get design details like `#ced4da` and we have no idea which one that is... :/

client/src/styles/_bootstrap-variables.scss
```diff
@@ -79,12 +77,12 @@ $blue: #005a93; // custom
 $indigo:  #6610f2 !default; // bootstrap default
 $purple:  #6f42c1 !default; // bootstrap default
 $pink:    #e83e8c !default; // bootstrap default
-$red: #d40404; // custom
+$red:     #DA273B !default; // custom
```
flamerohr: mix of upper case colours and lower case colours, could we stick to one for consistency? ;)

client/src/styles/_bootstrap-variables.scss
```diff
@@ -165,7 +159,7 @@ $spacers: (
 //
 // Settings for the `<body>` element.
 
-$body-bg: #f6f7f8;
+$body-bg: lighten($gray-700, 63);
```
flamerohr: is this different from the other `gray-*` colour codes? are we able to use those instead of another variation?
(p.s. I don't mind if this is kept, but could we have the computed colour code commented here as well?)

client/src/styles/_bootstrap-variables.scss
```diff
@@ -235,14 +229,14 @@ $line-height-lg: (4 / 3);
 $line-height-sm: 1.5;
 
 //$border-width: 1px !default;
-$border-color: darken(#f6f7f8, 10);
+$border-color: lighten($gray-200, 2);
```
flamerohr: same as above, is this different from the other `gray-*` colour codes?

client/src/styles/_typography.scss
```diff
@@ -86,7 +86,7 @@ th {
   text-transform: uppercase;
   font-size: $font-size-xxs;
   font-weight: 400;
-  letter-spacing: .4px;
+  letter-spacing: .1px;
```
flamerohr: `.1px` is really small, and you could end up with rounding differences across browsers with such a small value.

I believe Firefox rounds UP for this and all other browsers rounds DOWN.

client/src/styles/_variables.scss
```diff
@@ -147,10 +144,11 @@ $z-depth-2: 0 8px 12px 0 $shadow-level-3, 0 6px 20px 0 $shadow-level-2;
 $z-depth-3: 0 12px 20px 0 $shadow-level-4, 0 17px 40px 0 $shadow-level-3;
 $z-depth-4: 0 16px 28px 0 $shadow-level-5, 0 25px 55px 0 $shadow-level-5;
 
-// Masks: typically behind modals, drag drop areas
-
-$page-mask-light: rgba(236, 239, 241, .9);
+// Menu theme colours
 
+$base-menu: #d8e4eb !default;
+$base-menu-bg: 	lighten($base-menu, 5%);
```
flamerohr: spacing is a bit wierd here

client/src/styles/_bootstrap-variables.scss
clarkepaul: I could add them. I am setting up an easier way to get all the colours through this tool which syncs with the design doc and should match the variables sheet also https://brand.ai/silverstripe/cms-library

It will all be clearer with the pattern library too—when we have it.

client/src/styles/_bootstrap-variables.scss
clarkepaul: Yup easy fix, typically all uppercase is cleaner.

client/src/styles/_bootstrap-variables.scss
phalkunz: @flamerohr Are you saying you should put a color code in a comment where there's color calculation or just this section? I can see it being really helpful for searching and debugging purposes but it'd be a pain to have to do that everywhere.

client/src/styles/_bootstrap-variables.scss
phalkunz: Also the spacing between variables and their values.

client/src/styles/_bootstrap-variables.scss
phalkunz: I can see it make sense do this on a color that is used on a large surface but with color to be used with border properly 2% is hardly any different.

client/src/styles/_typography.scss
phalkunz: I just tested this out. It quite noticeable between `0px`, `.1px` and `1px` on Chrome and Firefox.

client/src/styles/_variables.scss
```diff
@@ -9,31 +9,29 @@ $brand-primary: #29abe2;
...
-$background-main: #f6f7f8;
-$background-darker: darken($body-bg, 2);
+$body-bg-dark: darken($body-bg, 2);
 
 // Extra border colors
-$border-color-lighter: darken($body-bg, 2);
-$border-color-light: darken($body-bg, 6);
-$border-color-dark: darken($body-bg, 14);
+$border-color-light: lighten($border-color, 4);
+$border-color-dark: darken($border-color, 2);
```
phalkunz: The same above comment about border colors.

client/src/styles/_typography.scss
clarkepaul: I think letter spacing is rounding is fine for all browsers. Not sure why I reduced that as .4px looked fine actually (probably .3 looks even better).

client/src/styles/_variables.scss
clarkepaul: Cool can fix

client/src/styles/_bootstrap-variables.scss
clarkepaul: Seems Bootstrap use all lower so best to stick to that actually :)

clarkepaul:
These variables are in the legacy folder already (from 2.4!) and shouldn't really be used but they are continued to be used and updated. We currently have three variables files which are already super confusing, maybe we could leave the deleted ones and put a note not to use them so that we can delete later. The main thing I was trying to achieve is to get everything using the right variables as the code is so misaligned with designs at the moment, so not a biggy leaving the old ones there.

client/src/styles/_bootstrap-variables.scss
clarkepaul: Probably just on these grays it would be okay.

client/src/styles/_typography.scss
clarkepaul: Going with .3px. Left is Chrome and right is Firefox.
image

clarkepaul:
@flamerohr @phalkunz I think I got all recommendations and I've added back all the unused variables :( I don't have time to go through them all and link/reference the newer ones in case they do get used—there are many which don't have a newer similar variable. I still think this is making good progress as it stands. I did let @chillu know I wanted to get rid of all unused variables and I think from memory it wasn't a deal breaker but shouldn't matter anymore.

client/src/styles/_bootstrap-variables.scss
flamerohr: just on these greys would already be a big help :)

client/src/styles/_typography.scss
flamerohr: sweet :)

flamerohr:
@clarkepaul sweet, looking good :)

hmm... I guess I'd prefer it was a warning if used than outright removed... where the warning would be butt-ugly but still work. (could make them all Bright green colours and giant width/height values)

Do you know if the CC's will be importing the legacy variables (assuming we put it in a separate file)? I guess I'd ask if they could exclude that - so then the unused variables wouldn't work for them.

I can investigate if there's a better way to do this in scss

clarkepaul:
I talked to @robbieaverill who possibly had a good idea which was to lint the older variables so we get warnings when they get used but thought it might be a fair bit of work.

All the modules would be importing the legacy variables as far as I'm aware.

Let me know if you think it needs any more changes.

chillu:
As stated in other contexts, I don't think rigorously applying sermver on "frontend APIs" like SCSS variable names will do us much good in the 4.x release line. We're just not stable on this stuff yet, and need to move relatively fast in terms of getting the ReactJS and Bootstrap transition over the line. So I don't think we should keep unused SCSS variables around in core just in case someone built their own SCSS on top of it. We'll want more stability around those around a 4.3 release or even a 5.0, but until then, building CSS functionality on top of the CMS will always require more upgrading efforts.

That being said, if there's a low impact way to mark things like these as deprecated without impacting the overall CMS frontend developer experience, sure thing.

clarkepaul:
Rebased and removed bug of square corners 
pasted_image_21_11_17__12_03_pm

flamerohr:
Sweet as, retaining variables were the only contentious point I had left.

Everything else has been addressed, and great effort from @clarkepaul 

So we've shifted from "Gradient EVERYTHING" to "Round ALL THE CORNERS", such a fun world to be in :)

PR #355, opened 2017-11-14
Remove disabled colour, add not-allowed cursor

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-cms/issues/2017


phalkunz:
@unclecheese The text should also be italic

PR #368, opened 2017-11-27
FIX Issue where logging out from the CMS presents you with a login fo…

dhensby:
…rm with no BackURL

Fixes https://github.com/silverstripe/silverstripe-framework/issues/7636

cc @tractorcow @MichaelJJames

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -1818,6 +1818,18 @@ class LeftAndMain extends Controller implements PermissionProvider
     }
 
     /**
+     * Generate a logout url with BackURL to the CMS
+     *
+     * @return string
+     */
+    public function LogoutURL()
+    {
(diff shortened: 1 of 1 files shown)


tractorcow:
I thought the issue was logging INTO the CMS, not logging out of the CMS? It seems a slightly different use case.

I'm not 100% sure that the logout url should intrinsically redirect the user back into the CMS, when the intention may be for the user to leave the CMS.

tractorcow:
I re-read through the discussion and I now agree with you @dhensby.

dhensby:
> I'm not 100% sure that the logout url should intrinsically redirect the user back into the CMS, when the intention may be for the user to leave the CMS.

I somewhat agree but it seems the old behaviour was to redirect back to the referer (which would be the CMS in this case) and so this just means it works where the referer is not available.

PR #371, opened 2017-12-03
Add a `loadComponent()` helper which provides a lot of the boilerplating

flamerohr:
Reduces the code normally done when needing a component from Injector


client/src/lib/dependency-injection/tests/loadComponent-test.js
```diff
@@ -0,0 +1,95 @@
...
+    expect(testComponent.textContent).toBe('abc');
+  });
+
+  it('should return a rendered TestComponent when loading TestComponent from Injector', () => {
+    const Temp = loadComponent('TestComponent', null);
+    const temp = ReactTestUtils.renderIntoDocument(<Temp />);
+    const testComponent = ReactTestUtils.findRenderedDOMComponentWithTag(temp, 'span');
+
+    expect(testComponent.textContent).toBe('def');
+  });
```
unclecheese: Maybe add a test for a non-falsey `context` param?

unclecheese:
Looks like the tests are having a hard time with the late `require()` of React.

client/src/lib/dependency-injection/tests/loadComponent-test.js
flamerohr: that's sort of what's happening with passing `undefined` :)
> Default function parameters allow formal parameters to be initialized with default values if no value or undefined is passed.
Ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

But fair point, that is a bit magical to read - I'll update it to be explicit

flamerohr:
oh, case-sensitivity issue :( uppercase `R` broke that on travis

flamerohr:
Screenshot of the `NotFoundComponent` when it's shown
screen shot 2017-12-06 at 12 42 09 pm

unclecheese:
Thank you! I was going to ask for that :)

PR #376, opened 2017-12-07
Set .gitattributes to ignore docs and tests folder

flamerohr:
Also included some source asset files which are not needed

Parent issue: https://github.com/silverstripe/silverstripe-framework/issues/6890

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,34 @@
-docs/ export-ignore
+/tests export-ignore
+/.tx export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.travis.yml export-ignore
+/.scrutinizer.yml export-ignore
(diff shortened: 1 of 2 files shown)


.gitattributes
```diff
@@ -1,8 +1,37 @@
...
+/client/src/components export-ignore
+/client/src/containers export-ignore
+/client/src/font export-ignore
+/client/src/images export-ignore
+/client/src/legacy export-ignore
+/client/src/lib export-ignore
+/client/src/sprites export-ignore
+/client/src/state export-ignore
+/client/src/stories export-ignore
+/client/src/.modernizrrc export-ignore
```
flamerohr: I needed to list all `/client/src` sub-directories as `.gitattributes` does not allow for negation selection like in `.gitignore`

Issue #377, opened 2017-12-08
Entwine TreeDropdownField styling broken


tractorcow:
Styling of all tree dropdown fields in the pages setting form is messed up

![image](https://user-images.githubusercontent.com/936064/33747319-37182b20-dc27-11e7-9369-0c7a58713fd5.png)

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/378
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/706

The issue is caused by this code:

```
// shifts the react component to override the entire field holder element
if (this.parent('.form__field-holder').length) {
    this.unwrap('.form__field-holder');
    this.siblings().remove();
    this.unwrap('.form-group').attr('class', 'TreeDropdownField form-group');
}
```

The problem is that this incorrectly mangles the DOM classes.

Expected dom:

![image](https://user-images.githubusercontent.com/936064/33747341-4ec2ac14-dc27-11e7-8fd0-853286fa7478.png)

Actual dom:

![image](https://user-images.githubusercontent.com/936064/33747354-6244394c-dc27-11e7-95f9-98b25739b2fe.png)

Instead of flipping the dom inside out, can't you just attach the component to the outer html?

```diff
-const container = this[0];
+const container = this.parent('.form__field-holder')[0];
      if (container) {
        ReactDOM.unmountComponentAtNode(container);
      }
```

flamerohr:
> Instead of flipping the dom inside out, can't you just attach the component to the outer html?

You lose the schema data on the div inside the field holder then...

I guess I don't prefer a destructive approach if possible

PR #378, opened 2017-12-12
Fix TreeDropdownField layout

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/377


tractorcow:
Looks good, but we should disclose (maybe with a `@todo`) the fact we are aware we are double-nesting the holder.

tractorcow:
Better solution. :)

PR #384, opened 2017-12-17
FIX Ensure testLeftAndMainSubclasses test runs some assertions

robbieaverill:
The assertions in the foreach loop don't run, because `$menuItems` returns an empty ArrayList. This means the test is "risky".

### DIFF
--- a/tests/php/LeftAndMainTest.php
+++ b/tests/php/LeftAndMainTest.php
@@ -85,11 +85,12 @@ class LeftAndMainTest extends FunctionalTest
      */
     public function testLeftAndMainSubclasses()
     {
-        $adminuser = $this->objFromFixture(Member::class, 'admin');
-        $this->session()->set('loggedInAs', $adminuser->ID);
-
+        $this->logInWithPermission('ADMIN');
         $this->resetMenu();
+
         $menuItems = LeftAndMain::singleton()->MainMenu(false);
+        $this->assertGreaterThan(0, count($menuItems));
+
         foreach ($menuItems as $menuItem) {
             $link = $menuItem->Link;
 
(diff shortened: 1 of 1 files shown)


robbieaverill:
Linting failures are unrelated to this change

dhensby:
> Linting failures are unrelated to this change

can you rebase? the 1.0 branches on travis are passing :/

robbieaverill:
Apparently I'm up to date with origin/1.0 =/

robbieaverill:
I'll re-run the 1.0 branch build to check

dhensby:
weird, I guess fix the linting issues, if you don't mind?

robbieaverill:
I'll see what I can dig up. I've seen this happen before with tabs inside PHP comments on Travis, but still pass locally. Mildly disconcerting that the lint rules aren't doing the same thing for whatever reason

robbieaverill:
Yep, that's what it was

robbieaverill:
Ah ok, so the culprit is [this change](https://github.com/squizlabs/PHP_CodeSniffer/pull/1607/) between PHPCS 3.1 and 3.2 which now sniffs for tabs in comments as well. I was running 3.1 locally, Travis running 3.2.

tractorcow:
Probably needs a rebase now.

robbieaverill:
This is a patch change so won't be affected by your 1.1 work =)

tractorcow:
I was referring to the code style cleanup changes.

robbieaverill:
Gotcha, done - thanks!

robbieaverill:
Re-enabling these assertions broke the kitchen sink because us TestOnly DataObjects being accessed in the reports section (https://github.com/silverstripe/silverstripe-framework/issues/7978)

I've made a PR at #704 to make the test less "all the things" and more specific

PR #380, opened 2017-12-12
ENHANCEMENT: Major refactor of React client side validation

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-admin/issues/45

## Summary of changes

* Turn off HTML5 validation
* Move core Validator validation to middleware
* New `FormValidationManager` API for more robust management of validation state
* Uniform handling of error messages
* Translatable validation messages
* More object spreads

PR #387, opened 2017-12-18
BUGFIX: applyTransforms not using new addErrors() API

unclecheese:
Resolves errors cited in https://github.com/silverstripe/silverstripe-admin/pull/380#discussion_r157400105


tractorcow:
Testing to confirm this is fixed. ;)

PR #388, opened 2017-12-21
ENHANCEMENT Add data object status badge in gridffield detail view

phalkunz:
Issue: Issue: https://github.com/silverstripe/silverstripe-versioned/issues/51


templates/SilverStripe/Admin/Includes/CMSBreadcrumbs.ss
```diff
@@ -1,7 +1,7 @@
 <div class="breadcrumbs-wrapper flexbox-area-grow" data-pjax-fragment="Breadcrumbs">
 	<% loop $Breadcrumbs %>
 		<% if $Last %>
-			<span class="cms-panel-link crumb last">$Title</span>
+			<span class="cms-panel-link crumb last">$Title<% if $Extra %>$Extra.RAW<% end_if %></span>
```
tractorcow: Not safe to use .RAW instead we should require casting in PHP.

PR #389, opened 2017-12-21
FIX Update input-group-addon-bg variable

sachajudd:
Update variable from $gray-200 to $gray-100

e.g. Will update elemental-blocks 'Displayed' field.

cc @clarkepaul


robbieaverill:
I've pushed 0e309ef to fix the PHP linting errors, if you rebase on top and rebuild the frontend assets we should be green =)

Issue #392, opened 2018-01-14
Not providing a third parameter to inject HOC triggers errors with updated components.


bummzack:
If a component uses the `inject` HOC to inject (sub-)components as props like this:

```
export default inject(
  ['UploadFieldItem', 'AssetDropzone', 'InsertMediaModal']
)(UploadField);
```

then there will be a JS error raised at [`MiddlewareRegistry.js`](https://github.com/silverstripe/silverstripe-admin/blob/master/client/src/lib/dependency-injection/MiddlewareRegistry.js#L195), whenever a custom component was injected into one of these props. (The error being that `context` is `null`)

So if you then do this:

```
Injector.transform(
  'enhance-uploadfield',
  (updater) => {
    updater.component('UploadFieldItem', CustomUploadFieldItem);
  }
);

```
The error will occur at runtime.

The issue is caused whenever `inject` was called without the third parameter (`getContext`). So this implementation here will not break:

```
export default inject(
  ['PopoverField'],
  (PopoverField) => ({ PopoverField }),
  () => 'BulkActions' // <- This seems to be a requirement
)(BulkActionsWithState);

```
while this will:

`export default inject(['FieldGroup'])(ProportionConstraintField);`

The way the parameters are ordered and how `inject` was implemented leads me to believe that the `getContext` should actually be optional. Maybe there could also be a default implementation for it, the same as for `mapDependenciesToProps`?

As previously stated, this error will only appear, once you injected a custom component via `Injector.transform`.

flamerohr:
Have created https://github.com/silverstripe/silverstripe-admin/pull/394 to address this (with `defaultContext`
Had a few other points listed on the pull request to address improving extensibility

flamerohr:
Fixed with the above pull request and merged

PR #394, opened 2018-01-16
Injected component context awareness

flamerohr:
This issue was discovered with the discussion here https://github.com/silverstripe/silverstripe-asset-admin/issues/718#issuecomment-357731283

Also addresses the issue of the `getContext` parameter here https://github.com/silverstripe/silverstripe-admin/issues/392

I'd like to hear @unclecheese's opinion (and anyone else who is keen) on the approaches.

Enhancement allowed inject() props to be overridden by a parent Component
Enhancement added context as a prop to injected components


client/src/lib/dependency-injection/buildComponentContainer.js
```diff
@@ -23,18 +24,16 @@ const buildComponentContainer = (base = buildBaseContainer()) => ({
...
+   * @returns {function}
    */
   get(key, context, ...args) {
-    const service = base.get.call(this, key, context, ...args);
+    const Component = base.get.call(this, key, context, ...args);
 
-    if (service.displayName && service.displayName.match(/\]$/)) {
-      return service;
-    }
+    const service = (props) => <Component identifier={context} {...props} />;
```
flamerohr: The idea is to included the `context` as the `identifier` prop to the service found.
Straight forward and enables authors to tweak the identifier within the Component if they wanted (if that's even desired?)

client/src/lib/dependency-injection/inject.js
```diff
@@ -9,66 +11,74 @@ import injectorContext from './injectorContext';
...
+        const newProps = {
+          ...props,
+          ...this.props,
+        };
+        return <Component {...newProps} />;
       }
-      const newProps = {
-        ...this.props,
-        ...props,
-      };
```
flamerohr: The other option is to allow the injected component props be replaceable by the calling function/component.

An example of UploadField:
`<UploadField UploadFieldItem={CustomComponent} {...props} />`
would allow for customising the `UploadFieldItem` child component which is injected to `UploadField`

I have a separate pull request coming up which switches `UploadFieldItem` from `import` to using `inject()`

client/src/lib/dependency-injection/buildComponentContainer.js
unclecheese: I worry about conflicts with a generic name like `identifier`. Maybe `injectorContext`? Seems more appropriate.

client/src/lib/dependency-injection/inject.js
unclecheese: I still think this should be offered as a function, which does this by default. There may be cases where you want the local props to be overriden by the injector props, or better yet, create, alter, or remove props based on the diff.

flamerohr: yeah, but I struggled to find a case for that... considering you're in the parent context and should have most of the props that are being passed down to the injected component?

I also couldn't think of how to provide that function between `inject()` and the `Component`. The `inject()` has the second param `mapDependenciesToProps` - but this is instance it is inaccessible and already executed.

client/src/lib/dependency-injection/buildComponentContainer.js
flamerohr: yeah, could be something like that, cheers

client/src/lib/dependency-injection/inject.js
bummzack: I also think that if the parent enforces a subcomponent via props, it should not be overridden by `inject`. Maybe the inject HOC could check if the prop is already set and leave it untouched if it is?

This would give developers the possibility to change sub-components based on state, which could be useful with something like a sortable-uploadfield, where sorting can be enabled via a prop.

bummzack: Hm, I like that idea. If I'm interpreting this right, then it would work for all cases. Eg. to globally replace a component, replace it for a given context, or even for a context that is generated dynamically (eg. based on some props). Sounds good.

unclecheese: Yeah, I like it. I think there's a really good use case for passing the context to children, for all the reasons @bummzack has stated.

This is getting a bit meta, but rather than passing the context as a prop, what about passing it via `context`? That way you're not bound to the parent-child hierarchy.  You could in theory leave `UploadFieldItem` alone, but customise `UploadFieldItemThumbnail` (or whatever) based on the `UploadField` context.

bummzack: Passing the context in the context… down the rabbit hole we go 👍 
Is it this context we're talking about here? https://reactjs.org/docs/context.html#why-not-to-use-context
Seems like they would rather un-invent that feature :)
I'm happy with either approach. If context is already being used for SS react-components this way, we can surely expand on that.

unclecheese: Yeah, it's not an API that's often viewed in a very positive light, and that's unfortunate. The point of that article (which I find a bit too absolute and admonishing) is that the basic design pattern of passing props should not be broken without serious consideration. That doesn't mean that it's never a good idea. Sometimes it's exactly the right idea. It's useful anytime you're trying to influence how a component behaves based on _where_ it is, rather than _how_ it is. Drag-and-drop libraries, redux stores, even my [react-selectable](https://github.com/unclecheese/react-selectable) library makes good use of it.

The SS React injector itself is a context service, and I think that's the best argument for using context here.

flamerohr: I had bounced the idea of using `react.context` for this, it definitely suits the case well.

I had also wondered if it would play nice with the `context` tweaking I mentioned above, you want to change the context in the middle for a new transform - or if that's even a necessary feature?
Simplicity and I know that it'd work with props were the reasons why I've implemented it this way :)

Actually, not allowing context tweaking could work if `inject` didn't override Component props provided by parent... and simplify the API/magic a bit :)

flamerohr:
I've shifted the `injectorContext` prop to `this.context.injector.context` (context in context... so meta)
also added unit tests to cover expectations of `inject()`

docs updated and could be found here: https://github.com/silverstripe/silverstripe-framework/pull/7768

bummzack:
Looking good. Who has the final say when this is going to be merged?

flamerohr:
I believe at the moment @unclecheese is our man/cheese

unclecheese:
Yeah, this one is on me now. I will review tomorrow during hackday.

unclecheese:
Looks good. Just needed a rebase. Nice work on this.

tractorcow:
This seems to have broken all behat tests; Now all react fields are re-rendered every time a key is pressed (so only the first character of any string is entered).

client/src/lib/dependency-injection/inject.js
bummzack: Guys, is there some documentation available for this? I'm struggling with this use case:
 - `UploadField` uses `UploadFieldItem` as child component.
 - I decorate `UploadField` with my own HOC
 - Depending on a prop passed to my HOC, I want to decorate `UploadFieldItem` with a HOC as well.

Is that even possible?

flamerohr: Hey @bummzack, you would be better off to have the "decorate with HOC" logic flipped, so the HOC will "pass-through" if certain props don't meet a condition.
So your UploadFieldItem is decorated with the HOC without knowing if the decorator is active or not.

`fieldHolder` in `silverstripe-admin` is a good example of passing-through if the "noHolder" prop was turned on.

HOCs shouldn't be hot-loaded on render, as it leads to performance hits (noticeably if it has many children)
It's a good idea to define constants for the React render engine to use - since it does a shallow reference check on the class/function that it tries to render to see if it should re-render for changes.

bummzack: @flamerohr Thanks for the explanation

I actually don't need hot-loading. It's for the use-case where the `UploadField` doesn't need any sorting at all, so I don't really want my HOC and styles to be applied at all. I'm not really sure how I can achieve that, since I'm decorating the `UploadField`, but I actually only need it to be decorated when the user makes use of the matching PHP field class (`SortableUploadField` instead of `UploadField`).

That's the first issue and it could be solved by a pattern like used in `fieldHolder`, where you just render the "base" component if there was no `sortable` property being passed from the PHP field.

The second issue is what I outlined above with the `UploadFieldItem`. If I don't decorate the parent, I obviously also don't want to decorate the child. But since I'm unable to pass properties to the child, I don't know how I would solve that?

flamerohr: yeah, with the current API, it's a convoluted way of doing it...

since you're decorating the UploadField, you could try take advantage of the new-ish API where you could overwrite an injected component by passing your own component to the parent via props.

It was a sort of abuse of that API, but you can do something like:
```
constructor(props) {
  super(props);

  this.DecoratedItem = itemDecorator(props.UploadFieldItem);
}

render() {
  const Item = this.props.sortable
    ? this.DecoratedItem
    : props.UploadFieldItem;

  return <UploadField UploadFieldItem={Item} />;
}
```

I suggest constructing the decorated item in the constructor so you're not hot-loading the decorated item every render.
I think the main caveat with this approach is if another decorator before your decorator did the same thing, then it could become really confusing...
maybe we could investigate exposing a `getItemProps` prop callback, so then item props are extensible and this way we can go without the default over-exposing props :)

bummzack: Thanks, I'll give this a try! Exposing `getItemProps` sounds like a good idea now indeed… and we have gone full circle :)

PR #400, opened 2018-01-19
Overrides tinymce fullscreen z-order

DrMartinGonzo:
so that it's not hidden by CMS panels


robbieaverill:
I think we fixed this (after your PR) in https://github.com/silverstripe/silverstripe-admin/pull/512

Sorry it didn't get looked at sooner, and thank you for contributing!

Issue #390, opened 2018-01-10
reducer transform of initial state only works with dev tools extension


blueo:
I'm following the guide to set a different initial state (https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/reactjs_redux_and_graphql/#setting-a-different-initial-state) but strangely, this only seems to work if you have the redux dev tools installed/active. Without that, the transformer function never gets to set the initial state. 

Putting a console.log in the transformer shows that the at some point the transformer does get a `state` variable that is `undefined` but the resulting initial state value is never set. 

I've reproduced the bug here: https://github.com/blueo/transformerbug. 
To see it in action open the asset admin section and inspect the console. With dev tools installed you should see the following:
```
index.js:6 Initial State Runs!
index.js:6 Initial State Runs!
index.js:6 Initial State Runs!
 window.ss.store.getState().assetAdmin.testVal
"I'm here!!"
```

Then disable dev tools and it shows:
```
index.js:6 Initial State Runs!
index.js:6 Initial State Runs!
 window.ss.store.getState().assetAdmin.testVal
undefined
```

It seems to log three times with dev tools and only twice without. The code looks like this:
```
/* global document */
import Injector from 'lib/Injector';

export const testTransformer = originalReducer => () => (state, { type, payload }) => {
    if (typeof state === 'undefined') {
      console.log('Initial State Runs!')
      return {
          ...originalReducer(state, { type, payload }),
          testVal: "I'm here!!",
      }
    }
    return originalReducer(state, { type, payload });
  };

document.addEventListener('DOMContentLoaded', () => {
    Injector.transform(
        'test-transformation',
        (updater) => {
          // Reducer updates
          updater.reducer('assetAdmin', testTransformer);
        },
      );
});
```

## Pull request
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/402

chillu:
@unclecheese Can you please look into this?

tractorcow:
@unclecheese I have looked into this and can reproduce it.

I have no idea what's going on with this... I need your help sorry. =(

tractorcow:
I have a feeling it's an inclusion order issue.

flamerohr:
ok, I think I see it (will test when I have time tomorrow)

You need to wrap `window.ss.store.getState().assetAdmin.testVal` with `Injector.ready(() => { });`
because you're calling the store while it's still registering and the reducer injector stuff has finished loading yet.

[...]

PR #402, opened 2018-01-22
Fix booting and store initialisation

flamerohr:
so that initial state is not triggered too early in the process

This stems from the discussion here https://github.com/silverstripe/silverstripe-admin/issues/390


client/src/lib/dependency-injection/buildReducerContainer.js
```diff
@@ -11,8 +11,6 @@ const buildReducerContainer = (base = buildBaseContainer()) => ({
   store: null,
 
   setStore(store) {
-    this.isProtected();
```
tractorcow: @unclecheese can you double check this logic is ok? It gets my approval if so.

tractorcow:
I approve, but I'd like @unclecheese 's review too.

tractorcow:
Rebased.

client/src/lib/dependency-injection/buildReducerContainer.js
unclecheese: It's probably okay, but I'm wondering if there is truly no viable case for a user replacing the store post-boot, we should still protect it.

What if we added a second property (e.g. `frozen`) that got set when the `ready()` callbacks finished, which would allow us to protect methods less strictly?

There is a good use case for replacing the store (i.e. adding new middleware), but we want to enforce that it happens post init, pre boot. So we may need more state.

unclecheese:
Rebased.

tractorcow:
I'll rebase and test.

tractorcow:
ok, rebased and rebuilt again.

Can merge on green. ;0

PR #403, opened 2018-01-22
ENHANCEMENT Put gridfield page count in button toolbar for security admin

### LINKED ISSUE #256: Remove the gridfield item count from the top of gridfields when there is no header
The gridfield item count is not required when there is no gridfield title directly above the grid as it causes an unwanted space. By removing the view count at the top the gridfield actions will be closer to the gridfield providing a clearer UX.

We could also remove it all together (whether there is a header or not), but it doesn't seem to do any harm when there is a header present.
pasted_image_15_09_17__10_38_am

## Pull request

- [ ] ~~https://github.com/silverstripe/silverstripe-framework/pull/7646~~
- [x] https://github.com/silverstripe/silverstripe-admin/pull/403

tractorcow:
Replaces https://github.com/silverstripe/silverstripe-framework/pull/7646

Fixes #256

### DIFF
--- a/code/SecurityAdmin.php
+++ b/code/SecurityAdmin.php
@@ -5,26 +5,26 @@ namespace SilverStripe\Admin;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Convert;
+use SilverStripe\Forms\FieldList;
+use SilverStripe\Forms\Form;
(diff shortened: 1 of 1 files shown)


flamerohr:
Please rebase the changes to the `1` branch instead, it's an enhancement not a bugfix

tractorcow:
Ok done @flamerohr

PR #404, opened 2018-01-22
FIX Non-required fields failing when empty

### LINKED ISSUE #396: Unaltered date-field in asset-admin requires value (fails validation).
When I add a date-field to asset-admin (eg. by decorating `FileFormFactory`), I can no longer save the record, unless I enter a valid date.

screen shot 2018-01-17 at 21 27 20

With the empty default value, validation will always fail with the message: **&lt;Fieldname&gt; is not a proper date format.**

tractorcow:
Fixes #396


client/src/lib/Validator.js
```diff
@@ -35,6 +35,10 @@ class Validator {
   }
 
   validateValue(value, rule, config) {
+    // Empty values suppress error unless rule is required
+    if (value === '') {
+      return rule !== 'required';
+    }
```
flamerohr: neat... but check out `validateField` line 157 and line 174, wouldn't that have caught the `required` rule already?

Maybe that part could be investigated to be removed and shifted here

tractorcow: That line only shortcuts required failures (required = true, empty = true).

This fix shortcuts non-required successes (required = false, empty = true) which should bypass other rules, but doesn't. Thus all of the specicchecks.validate('') return false.

PR #405, opened 2018-01-23
Fix performance issue with context for injector

flamerohr:
Rather than have context provided automatically by injector (in `injector.component.get()`)
You'll now need to provide context with provideContext or in the component which is using injector instead.

I totally forgot that I shouldn't be composing within dynamic calls like `injector.component.get()`


tractorcow:
Do you think some of these methods would benefit from descriptions? It's a huge black box to someone trying to understand the code. (e.g. 'what is a context?")

flamerohr:
it's in the docblock for `loadComponent`
```
  * @param context - properties include the following
  *    - store for the redux store
  *    - apolloClient for the apollo client (graphql)
  *    - context for filtering/applying transformations to the obtained component
```

unclecheese:
I agree, @tractorcow .. it's particularly confusing when you run into things like `context.context`.

unclecheese:
Do we need a docs update on the new explicit use of `provideContext`?

flamerohr:
yeah, there aren't any existing docs about context, so will need to look at adding some in...

flamerohr:
like here https://github.com/silverstripe/silverstripe-framework/pull/7768/files @tractorcow @unclecheese

PR #406, opened 2018-01-23
FIX Badge component test: convert to Component and add truthy test

robbieaverill:
The original "should return null if status is empty" test was asserting that the component render result was null, but the component wasn't being executed at all, so the result was null by default.


client/src/components/Badge/tests/Badge-test.js
```diff
@@ -1,32 +1,39 @@
...
+          className="customclass"
+        />
       );
 
-      expect(badge).toBe(null);
+      expect(badge.render()).toEqual(
+        <span className="customclass badge badge-success badge-pill">
+          Hello world
+        </span>
+      );
```
flamerohr: I prefer cherry picking parts that you truly expect, than a whole dump of a big object...
so test that `badge-success` class exists
test that `Hello world` is in the `textContent`

That way unit tests don't ALWAYS break when we tweak the component

robbieaverill: Sure. I’d argue the opposite in that changes to a component should be reflected in the tests too, but I agree that partial assertions would be totally fine

robbieaverill: Updated =)

flamerohr: Yeah, I'd argue with breaking that test down even more `it()` statements :D
That way you test each feature atomically, this refactor is good to go though.

robbieaverill: Good counter! Thanks :-)

PR #418, opened 2018-02-01
MINOR: Add actions-warning class

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-versioned/issues/71


flamerohr:
I've darkened the text a little to closer match the designs
And I've disabled the `:active` and `:focus` state box-shadows

PR #419, opened 2018-02-02
Adjust colour for warning message

flamerohr:
Follow up for https://github.com/silverstripe/silverstripe-versioned/issues/71 (got a more accurate colour)


client/src/components/FormAction/FormAction.scss
```diff
@@ -306,7 +306,7 @@ button.close {
 }
 
 .actions-warning {
-  color: darken($state-modified-bg, 15);
+  color: #856404; // in bootstrap: theme-color-level(map-get($theme-colors, "yellow"), 6)
```
robbieaverill: Are we able to use a variable, out of interest?

tractorcow:
I'm not sure if I find the former or the latter colour more repulsive, but we'll follow the designs so we can't say we didn't do as we were told. :)

tractorcow:
Merge it on green.

client/src/components/FormAction/FormAction.scss
flamerohr: I had opted out of using a variable for now, if we're finding ourselves using more of bootstrap's mixin'd theme colours we can setup a block in the _variables.scss file

PR #420, opened 2018-02-02
Update LeftAndMain_MenuLogo.ss

worikgh:
"span" misspelt


worikgh:
On my copy when I straighten out that spelling then the $CMSVersion is displayed in blue over the logo!  Quite ugly really....

dhensby:
the second span probably needs the `sr-only` class too?

PR #421, opened 2018-02-04
javascript i18n : lang attribute is set on html tag, not body

DrMartinGonzo:


tractorcow:
Yes this seems correct. Example in https://www.w3.org/TR/html5/dom.html#elements-semantics shows this on the `<html>` tag also.

However this needs a rebase. I will merge this in manually.

flamerohr:
I would like to extend it to fallback on the body element if the documentElement doesn't have a lang attribute :)

That way we could merge this into the `4.0` branch and then up all the way back to master

tractorcow:
Sounds good.

tractorcow:
I've rebased this and cleaned up the file at the same time https://github.com/silverstripe/silverstripe-admin/pull/458

PR #424, opened 2018-02-06
FIX Add text-colour to status-archived, remove span.badge styles

sachajudd:
Resolves: https://github.com/silverstripe/silverstripe-cms/issues/2084

cc @clarkepaul 

Before:
image

After:
screen shot 2018-02-07 at 11 53 55 am

Sitetree:
image


clarkepaul:
Looks good :). Can you just add a screenshot of the tree with an archived item as well? (to make sure we haven't disrupted that one).

PR #425, opened 2018-02-08
Created SilverStripe TinyMCE Skin

purplespider:
Created a new “silverstripe” skin for TinyMCE based upon the existing “lightgray” skin but with adjusted colours so it fits in more with the SilverStripe UI. Set as new default skin. 

`skin.json` is included so any further adjustments can easily be made via http://skin.tinymce.com

_Resolves #83_ 

**Before:**
Before

**After:**
After


dhensby:
Test failure due to lack of postgres support in 5.x, so unrelated

purplespider:
Hmm, maybe I didn't mean to propose to merge with master then, this PR was created with v4 in mind.

dhensby:
OK - I've rebased for you

dhensby:
Thoughts:

- This shouldn't be in the `thirdparty/` dir. It's first-party code and if we upgrade tinymce then this work will be lost/overridden
- We shouldn't be using LESS, our all the admin styles are built on SASS and we should use that and the predefined vars that we already have in our core frontend code.
- This could be a good chance to pick some "silverstripe" buttons images too...

cc @clarkepaul @newleeland

purplespider:
- I just put it in the standard TinyMCE `skins/` dir, but it does look like you can specify a `skin_url` instead. Where would you like it to live?
- For ease of future adjustments, or for upgrading to support future versions of TinyMCE, the skin was generated using the official TinyMCE Skin Creator: http://skin.tinymce.com so the LESS file isn't actually used, it was just provided in the generated files as I believe that's their other official option for creating/editing TinyMCE skins. Generating the skin CSS from SilverStripe's SASS will obviously make sure the skin keeps up to date with any adjustments to the SilverStripe SASS variables, but could mean that there would be more work in upgrading the skin for future TinyMCE versions?

flamerohr:
I finally see a difference in colours! (one has less impact on my eyes less when staring at them for a set amount of time)

I think it's fine in the thirdparty skins folder, and using tinymce's tools will make it easier to future enhancements we could want.

I wouldn't be keen on converting the less file to sass for consistency in this instance - we lose the ability to use the tool provided and we'd need a fair amount more setup/work to maintain it without much benfit.

We can tweak the icons in a separate pull request :)

(PS: 5000 lines for minified files... that's some great "minifying")

tractorcow:
I was asked to tie break whether or not this is in thirdparty folder or not.

I don't think the extra file size justifies the value we get for a few colour changes.

Can we just override the colours with non-skin css?

dhensby:
I'd definitely be keen on accepting this. I wouldn't put it in `thirdparty/` as we're going to lose track of it if we do that.

purplespider:
Ok. I just tested `skin_url` instead of `skin` and it is easy to place it anywhere else. Where would you like it? `/client/dist/tinymce/`?

dhensby:
@purplespider the NZ team would be better placed to say where is good for it to go. As it doesn't go through our build process it probably shouldn't live in `/client/dist/` but let's see what they say

tractorcow:
@dhensby custom skin paths are a pain to configure. We have a standard tinymce "resources" folder which is in thirdparty, which is why I think it's ok to just leave it there.

flamerohr:
yeah, I'd prefer the `thirdparty` folder :) if putting it in the `tinymce/skin` folder is not preferred, how about `thirdparty/tinymce_silverstripe_skin`? Although it's still in thirdparty, it's directly visible.

the `dist` folder is meant to be for generated/shifted files only, if we cleared the `dist` folder and ran `yarn build` it should give the same files... ideally.

flamerohr:
@purplespider After discussing it with @chillu, we're happy to have this merged in if the `lightgray` skin could be removed in this pull request as well :)

Thanks for your patience!

chillu:
Haven't read through the backstory of this, just my 2cts that we really shouldn't add more HTTP requests in the CMS, it's already worse than in 3.x. File size is less of a concern (at least for a few kilobytes) as long as they're transferred in the same bundle in one well gzipped HTTP request

flamerohr:
to clarify what's happening :)
No extra http requests are made (except for when this is merged and users download this for the first time)
The request to delete the existing lightgray skin is to reduce the download size of this module, since it's no longer used with the new silverstripe skin.

flamerohr:
I've removed the `fonts/tinymce*.json` they didn't do anything
and the old skin is removed, can easily be re-added by the skin editor linked in OP

dhensby:
whoop

purplespider:
Ace! :) Thanks @flamerohr

purplespider:
Can I just check what release this is likely to make it into?

dhensby:
This should be in 4.1

dhensby:
Oh no, it looks like it got merged after 4.1 was branched off so this won't make it until 4.2 :(

purplespider:
Ah, that'll explain it. Cheers Dan!

dhensby:
@purplespider if you want it in your projects you can always add the requirement of `silverstripe/admin: ^1.2` to your `composer.json` but do note that it'll bring in a dev build which will be less stable and more prone to bugs/regressions than waiting for the tagged release.

PR #427, opened 2018-02-09
Feature Show usage of files in "edit file" view

flamerohr:
Added a FormField to handling displaying DataObject ownership.
screen shot 2018-02-09 at 3 31 13 pm

This works in both entwine and react sections.
The component can be used independently without the autoloader if desired.

Please see for more details https://github.com/silverstripe/silverstripe-asset-admin/issues/593


code/Forms/UsedOnTable.php
```diff
@@ -0,0 +1,158 @@
...
+
+        return null;
+    }
+
+    /**
+     * @return DataObject|RecursivePublishable
+     */
+    public function getRecord()
+    {
+        return $this->record;
```
tractorcow: I think we need a setter as well.

code/Forms/UsedOnTable.php
```diff
@@ -0,0 +1,158 @@
...
+    /**
+     * @var DataObject|RecursivePublishable
+     */
+    protected $record;
+
+    /**
+     * @param string $name
+     * @param DataObject|RecursivePublishable $record
+     */
+    public function __construct($name, $record)
```
tractorcow: Mandatory record field will make this field impossible to scaffold without a parent record. Shift this to a setter instead.

See `FileUploadReceiver::getRecord()` for lazy-loading record from a parent form. It should be inferrable from `$this->getForm()->getRecord()` instead of requiring it upfront.

code/Forms/UsedOnTable.php
```diff
@@ -0,0 +1,158 @@
...
+            'usage' => $usageData,
+        ];
+
+        $response = new HTTPResponse(json_encode($data));
+        return $response
+            ->addHeader('Content-Type', 'application/json');
+    }
+
+    /**
+     * @param $record
```
tractorcow: type on phpdoc.

code/Forms/UsedOnTable.php
tractorcow: See other note; If $this->record isn't set, you should lazy-infer this from the form record.

templates/SilverStripe/Admin/Forms/UsedOnTable.ss
```diff
@@ -0,0 +1 @@
+<div class="used-on__polyfill-holder $extraClass" $AttributesHTML('class')></div>
```
tractorcow: You need the `$get` in `$getAttributesHTML()` otherwise it won't work properly. Calling a getter as a method (instead of as a getter) requires the full method name.

flamerohr: strange, this works for me without the `$get`
otherwise my entwine polyfill wouldn't have worked...

flamerohr:
thanks @tractorcow I've added the changes in a separate commit

tractorcow:
Don't merge this just yet.

PR #430, opened 2018-02-12
BUG improve the browser warning logic show

flamerohr:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/2076

Previously used the `hide` class to HIDE the browser warning...

Sometimes the main `bundle.css` in admin is downloaded after the `browser-warning.css` file, leading to a flash of incompatibility - especially on slow connections and with "immediate render" engines like the new Firefox.

This fixes that race condition


phptek:
@flamerohr quality branch naming!

PR #431, opened 2018-02-12
BUGFIX: Force gridfield search button to show when no actions are available

unclecheese:
A complex, very hacky solution to https://github.com/silverstripe/silverstripe-framework/issues/7610

Please read the full thread, including the assessment of the issue before judging. :)


client/src/legacy/GridField.js
```diff
@@ -9,6 +9,12 @@ require('../../../thirdparty/jquery-entwine/dist/jquery.entwine-dist.js');
 
 $.entwine('ss', function($) {
   $('.grid-field').entwine({
+    onmatch: function () {
+      if (this.needsColumnFix()) {
+        this.fixColumns();
+        this.injectSearchButton(false);
```
flamerohr: for consistency, perhaps both can include the param

unclecheese: Whoops. Yeah, I had done it that way initially, but it looks like the `fixColumns` function is no longer using the `visible` param, so it should just be removed from the definition.

Issue #383, opened 2017-12-15
Tabs have a border radius which needs to be removed


clarkepaul:
Only viewable when the UI is zoomed in. The line below the tabs should be straight and not rounded.
<img width="496" alt="pasted_image_15_12_17__11_22_am" src="https://user-images.githubusercontent.com/555033/34020680-785e1c02-e19a-11e7-8100-7af43385a9cd.png">

PR #434, opened 2018-02-13
FIX Remove border-radius add hover states to non-active tabs

sachajudd:
Resolves: #383, https://github.com/silverstripe/silverstripe-asset-admin/issues/689

Remove border-bottom, border-radius:
Before:
image

After:
image

Tab hovers:
Before: 
image

After (hovered):
image

image

Files area:
Before (hovered):
image

After (hovered):
image


dhensby:
Looking issues, I'm afraid

Issue #410, opened 2018-01-25
SS4: ModelAdmin shows blank page when search with single result is clicked


axllent:
This happens in both 4.0.1 and 4.0.2 (unsure about 4.0.0)

Sorry but it is difficult to describe this bug and it has taken a while to track down a scenario where it happens and is reproducible.

Basically if you have a ModelAdmin that contains multiple results, search for something that produces a **single** search result. When you click that result you get a blank page (no JS errors). I can always get it in some instances, and others not, so I'm not sure if it is relative to the number of objects being displayed. In my case I have just 16 pages.

To provide a working scenario, take the following ModelAdmin as a test:

```php
<?php
use SilverStripe\Admin\ModelAdmin;
class TestModelAdmin extends ModelAdmin
{
    private static $menu_title = 'Broken Search';
    private static $url_segment = 'broken-search';
    private static $managed_models = [
        'SilverStripe\CMS\Model\SiteTree'
    ];
}
```
Now in the CMS, go to /admin/broken-search/ and search for "page not" (which should hopefully produce one result, eg: the `Page not found` page). Click that result = blank page / frame.

I've tested this on both Firefox and Chrome, bot with the same result, so I hope you can reproduce this too :-)

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/435

unclecheese:
Have research this bug and found that it only happens when there are spaces in the search term. On list view, spaces are properly expressed with a `+` sign (`?q%5BTitle%5D=page+not`), but in detail view, that `+` sign gets replaced with its hex equivalent `%2B`, which is too eager. Plus signs are valid in URLs. This causes the search term to be interpolated into the query incorrectly ('page+not' instead of 'page not'), and a record of the given ID cannot be found in the list due to that filter that never matches, and the request redirects back and dies.

The culprit is the `addSearchParams` method in `sspath.js`, a library from 2011. This function calls `$.param`, a native jquery method for turning objects into parameter strings. Patching the native jquery method would be too aggressive, so the `addSearchParams` will be updated to undo `+` replacination after `$.params` is called.

Issue #409, opened 2018-01-25
ModelAdmin search broken in 4.0.2


axllent:
Introduced in SilverStripe 4.0.2 - The ModelAdmin search seems to have an issue whereby if multiple words are searched, spaces after the second word are substituted with a "+".

So searching for `game of thrones` results in `game of+thrones` (which leads to no results as the + is taken as part of the search string). 

Another example: `this is my search` results in `this is+my+search`

unclecheese:
Have confirmed this is resolved by https://github.com/silverstripe/silverstripe-admin/pull/435

PR #435, opened 2018-02-14
BUGFIX: Double encoding of "+" symbol in URL params

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-admin/issues/410


flamerohr:
today I learned that `replacination` is a word...

merge on green

axllent:
Thanks for fixing this! I'm assuming this also relates to/fixes #409?

flamerohr:
@axllent good question, I don't think it does

the fix is close but not quite the same

cc @unclecheese

Issue #433, opened 2018-02-13
Bootstrap 4.0 upgrade


chillu:
# Overview

We're currently on Bootstrap 4.0.0-beta, and a [Reactstrap](https://github.com/silverstripe/silverstripe-admin/issues/415) upgrade necessitates an upgrade to Bootstrap 4.0.0 stable. 

# Acceptance Criteria

 * CMS uses Bootstrap 4.0.0
 * All CMS bootstrap uses are regression tested (with a good mix across our supported browsers)
 * Any design variations are checked with Jared or Paul
 
# PRs:

See https://github.com/silverstripe/silverstripe-admin/issues/415 for PR list

chillu:
Copying comments from @lukereative:

> At the moment SS4 is using Bootstrap 4.0.0-beta and the compatible reactstrap version is 4.8.0, if we wanted we could upgrade to Bootstrap 4.0.0 (stable) and use reactstrap 5.0.0-beta. There have been a few changes from Bootstrap 4.0.0-beta into stable that affect us:

>    Slight change in badge colour behaviour
>    A change in structure of option and checkbox form fields
>    A change in structure of input addons
>    Popover changed js libraries from tether to poppers, but this doesn't seem to change functionality overall
>    Active states colours seem to have changed as well so this will need to be checked

> I think there might be a few others that will come out of the woodwork but looking at the bootstrap docs for their listed breaking beta changes they were fairly minimal. Overall it doesn't seem like too much extra effort and this seems like a good time to do it.

@lukereative Can you please add a points estimate to this card?

robbieaverill:
Duplicate of https://github.com/silverstripe/silverstripe-admin/issues/395, closing that one in favour of this one

tractorcow:
Closing in favour of #415. Admin module has been updated to latest bootstrap.

Issue #44, opened 2017-04-27
Pressing ESC key should dismiss the modal box


phalkunz:

phalkunz:
Actually, the issue already exists https://github.com/silverstripe/silverstripe-framework/issues/6661

PR #441, opened 2018-02-19
Upgrade to Bootstrap 4.0.0-stable and change to reactstrap 5.0.0-beta

### LINKED ISSUE #415: Switch from react-bootstrap to reactstrap
In @flamerohr assessment, [react-bootstrap](https://react-bootstrap.github.io/) has lost traction, and [reactstrap](https://reactstrap.github.io/) is much more active. react-bootstrap took a long time to get a Bootstrap 4 upgrade started. Since this is a library we want to build on going forward, we want a solid choice.

Current Usage:

 * Popover
 * Buttons
 * Modal (dialogs and alerts)
 * FormControl, FormGroup
 * Tab

Areas:

 * AssetAdmin edit form (tabs and forms)
 * CampaignAdmin modals
 * Insert Links and Media

PRs:
- [x] https://github.com/silverstripe/silverstripe-admin/pull/441
- [x] https://github.com/silverstripe/silverstripe-admin/pull/449
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2101
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/737
- [x] https://github.com/silverstripe/silverstripe-campaign-admin/pull/88
- [x] https://github.com/silverstripe/webpack-config/pull/18

lukereative:
Resolves the following issues:
- #415 
- #433
- #44 

Fixes https://github.com/silverstripe/silverstripe-framework/issues/6661


client/src/components/HtmlReadonlyField/HtmlReadonlyField.js
```diff
@@ -37,7 +35,8 @@ class HtmlReadonlyField extends Component {
 
   render() {
     return (
-      <FormControl.Static
+      <Input
+        static
```
flamerohr: `static` is deprecated for `plaintext`

client/src/components/Tabs/TabItem.js
```diff
@@ -13,39 +13,32 @@ class TabItem extends Component {
...
+  }
+
   render() {
     const tabProps = this.getTabProps();
     return (
-      <Tab.Pane {...tabProps}>
-        {this.props.children}
-      </Tab.Pane>
+      <TabPane {...tabProps}>
+        <Fade in={this.isActive()} unmountOnExit>
```
flamerohr: I don't think unmountOnExit is a good idea, at least in ss4 context.

The input values and markup will be removed from the HTML, which is fine if you have the values recorded in something like redux-form... but if the field was used outside this context - e.g. an entwine section which we will need to support, then those values will not be posted back until the tab was shown which means a different set of values will be lost instead.

client/src/components/Tabs/tests/Tabs-story.js
```diff
@@ -0,0 +1,17 @@
...
+  .add('Basic', () => (
+    <Tabs defaultActiveKey="Main" id="Root">
+      <TabItem name="Main" title="Main">
+        My first tab content
+      </TabItem>
+      <TabItem name="Settings" title="Settings">
+        My settings tab here
+      </TabItem>
+    </Tabs>
+  ));
```
flamerohr: Added a story for single tab in the tabset

client/src/components/IframeDialog/IframeDialog.js
```diff
@@ -15,39 +15,39 @@ class IframeDialog extends Component {
   render() {
     return (
       <Modal
-        show={this.props.show}
-        onHide={this.props.onHide}
+        isOpen={this.props.isOpen}
+        onExit={this.props.onExit}
```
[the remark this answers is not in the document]
flamerohr: Found that there is a valid `onExit` API on reactstrap, but I do not see why `IframeDialog` was treated differently...

client/src/components/OptionsetField/OptionField.js
```diff
@@ -76,25 +70,14 @@ class OptionField extends Component {
...
-    const label = (typeof labelText === 'string')
-      ? { react: <span>{labelText}</span> }
-      : labelText;
-
-    return castStringToElement(Option, label, this.getInputProps());
+    return (
+      <FormGroup check>
+        <Label check>
+          <Input {...this.getInputProps()} />
+          {labelText}
```
flamerohr: just realised the `castStringToElement()` call missing for this label, which is important for if the PHP returns an html object.

client/src/components/CheckboxField/CheckboxField.js
```diff
@@ -1,13 +1,8 @@
...
-import OptionField from '../OptionsetField/OptionField';
-import fieldHolder from 'components/FieldHolder/FieldHolder';
+import OptionField from 'components/OptionsetField/OptionField';
 
-const CheckboxField = (props) => {
-  // Build standard checkbox with fieldholder
-  const FieldHolder = fieldHolder(OptionField);
-
-  // set to not show field holder labels, as checkbox already generates a label
-  return <FieldHolder {...props} type="checkbox" hideLabels />;
```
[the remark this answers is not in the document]
tractorcow: Shouldn't we export both? E.g.

```
export { CheckboxField as Component };

export default fieldHolder(CheckboxField);

```

client/src/components/DateField/DateField.js
```diff
@@ -21,24 +21,24 @@ class DateField extends TextField {
       i18n._t('Admin.FormatExample', 'Example: {format}'),
       { format: moment().endOf('month').format(localFormat) }
     );
-    const props = {};
 
-    let val = this.props.value;
-
-    if (!this.props.data.html5 || (this.hasNativeSupport() && this.props.data.html5)) {
```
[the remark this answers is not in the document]
tractorcow: I've fixed up a patch with this.

client/src/components/CheckboxField/CheckboxField.js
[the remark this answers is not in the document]
tractorcow: Pushed up a fix.

tractorcow:
Approved pending behat green.

tractorcow:
Labels in form field holders have lost their `for` property. This breaks accessibility (and thus behat tests).

tractorcow:
Need to replicate the `for` logic in https://github.com/open-sausages/react-bootstrap/blob/master/src/ControlLabel.js but with the reactstrap `Label` element.

client/src/components/CheckboxField/CheckboxField.js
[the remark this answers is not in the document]
flamerohr: strange, I remember adding `fieldHolder(Checkbox)` to the export...

tractorcow:
OptionsetField broken.

image

client/src/components/DateField/DateField.js
```diff
@@ -63,21 +73,20 @@ class DateField extends TextField {
...
     if (typeof this.props.onChange === 'function') {
-      this.triggerChange(isoValue);
+      this.triggerChange(event, isoValue);
     }
   }
 
-  triggerChange(value) {
-    this.props.onChange(value);
+  triggerChange(event, value) {
+    this.props.onChange(event, { id: this.props.id, value });
```
tractorcow: Apparently all of these onChange() events fail if you don't pass in `value`. Will need to review each of these tomorrow.

Reverting them fixes it.

tractorcow: See redux forms docs on onChange here https://redux-form.com/7.1.2/docs/api/field.md/#-code-input-onchange-eventorvalue-function-code-

flamerohr: hmm, something else is wrong then, because it should be `this.props.onChange(event, { id: this.props.id, value });`
other form fields follow this pattern and return two objects.

PR #444, opened 2018-02-20
Add warning for unpublishing owned records

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-versioned/issues/26


client/lang/src/en.json
```diff
@@ -31,5 +31,8 @@
     "Admin.VALIDATOR_MESSAGE_DATE": "{name} is not a proper date format.",
     "Admin.VALIDATOR_MESSAGE_ALPHANUMERIC": "{name} is not an alphanumeric value.",
     "Admin.VALIDATOR_MESSAGE_ALPHA": "{name} is not only letters.",
-    "Admin.VALIDATOR_MESSAGE_DEFAULT": "{name} is not a valid value."
```
tractorcow: Can we update the `client/lang/en.js` as well?

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -433,6 +433,39 @@ $.entwine('ss', function($){
...
+  // Hack: silverstripe/versioned coupling
+  $('.cms-edit-form .btn-toolbar button[name=action_doUnpublish].btn.action').entwine({
+    onclick: function(e) {
+      var owners = this.data('owners');
+      if (owners && parseInt(owners) > 0) {
+        var message = [
+          i18n.inject(
+            i18n._t(
+              'Admin.OWNED_WARNING_1',
+              'You are unpublishing content that is being used in {count} other published section(s).',
```
tractorcow: Trailing `,` is out of place here I think.

flamerohr: it's a valid trailing comma, in fact a future eslint rule ENFORCES trailing commas for multi-line params

tractorcow: So weird.

PR #445, opened 2018-02-25
HTML typo in cms template

elliot-sawyer:
The span tag that displays the application name and version contains a typo


elliot-sawyer:
Just noticed this is a duplicate of https://github.com/silverstripe/silverstripe-admin/pull/420

dhensby:
I thought this looked familiar :P

dhensby:
@elliot-sawyer can you double check there's no regression with `$CMSVersion` being visible?

PR #449, opened 2018-02-27
ENHANCEMENT Decouple preview from campaign admin

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-admin/issues/415


client/src/components/Preview/Preview.js
```diff
@@ -108,6 +134,12 @@ Preview.propTypes = {
   itemId: React.PropTypes.number,
   onBack: React.PropTypes.func,
   moreActions: React.PropTypes.arrayOf(React.PropTypes.element),
+  moreActionsPopoverId: React.PropTypes.string,
+};
+
+Preview.defaults = {
+  className: 'campaign-admin__campaign-preview', // Legacy defaults
+  moreActionsPopoverId: 'campaign-preview-popover', // Legacy defaults
```
flamerohr: Could these defaults be shifted over to `campaign-admin` entirely? Doesn't look like anything is using the Preview yet and would mean decoupling is complete

tractorcow: done.

PR #458, opened 2018-03-07
BUG Ensure lang is detected from html tag

tractorcow:
Replaces https://github.com/silverstripe/silverstripe-admin/pull/421

I cleaned up all the linting in this file because it was horrible.


tractorcow:
Yeah indentation was legacy.

Issue #287, opened 2017-10-04
Grid field actions to move into a more options dropdown


clarkepaul:
# Overview

Now that grid field items have more actions (including publish, unpublish, archive, and potentially more like Add to campaign) we need a more flexible approach to showing available actions on the grid field holder itself and not just in the detailed view of items. We also give too much prominence to the delete action on grid field items—we should be encouraging constructive actions over deconstructive actions. 

This is the proposed functionality for grid items and how content blocks (WIP) are shaping up to use the same functionality.
![pl_gridfield_options](https://user-images.githubusercontent.com/555033/31205067-9e99911e-a9cb-11e7-9cf6-d9e182953836.png)

Side note: I was sure there was an issue for this already but can't find it so creating a this as a new one (might have been part of another issue but excluded from AC's)?

## Acceptance Criteria

 * Developers can use the GridField PHP API to add options
 * Options can be global or row specific
 * Options can be grouped (with a visual separator)
 * Existing GridFields (ModelAdmin, Security, etc) use this new component
 * The dev documentation for setting up GridFields reflects this change
 * GridField still supports multiple buttons in parallel to this new UI component

## Notes

 * Excludes list in AssetAdmin
 * This information will need to come through as structured data for React-based GridFields soon
 * Auto-adding tabs from detail forms is out of scope
 * Should be built as a React component embedded in a Entwine-based GridField (so we don't have to rewrite it later)
 * Track under ORB-175

## Pull requests
- [x] https://github.com/silverstripe/silverstripe-framework/pull/8083
- [x] https://github.com/silverstripe/silverstripe-admin/pull/530
---
- (OLD REFERENCE ONLY) https://github.com/symbiote/silverstripe-gridfieldextensions/pull/225

chillu:
We're past the point where we can add new features in 4.0, sorry - removing milestone.

chillu:
@clarkepaul Does this make it clear enough that clicking on the whole row triggers the edit view on a standard GridField? Do you expect mixed cases where you have primary actions shown alongside a "more actions" button?

clarkepaul:
[...]

PR #459, opened 2018-03-08
Fix "container" logic for FormBuilder and PopoverField, improves accessibility

flamerohr:
Popover markup will attach to the Button which invoked it, this means pressing "tab" on the keyboard will shift focus on the buttons in the popover instead of skipping the popover contents altogether

Fixed while investigating https://github.com/silverstripe/silverstripe-admin/issues/287


tractorcow:
Cannot nest <button /> tags. It breaks lots of things.

See https://stackoverflow.com/questions/39386497/can-i-nest-button-inside-another-button

Issue raised at https://github.com/silverstripe/silverstripe-asset-admin/issues/749

PR #461, opened 2018-03-12
NEW Update table border colour and allow badge-pill to be overridden in Badge

robbieaverill:
Changes the table border colour from gray-300 to gray-200 (not used anywhere as far as I could see with a quick search), and allows the `badge-pill` prop in Badge component to be overridden, since not all badges should automatically assume this stylistic modifier.

Parent story: https://github.com/silverstripe/silverstripe-versioned/issues/37


dhensby:
I rebased this to get it passing

PR #466, opened 2018-03-14
Make webpack work nicely when bin-links are off

dhensby:
When running on a shared drive through a VM, I have to use `--no-bin-links` when running `yarn`; these changes help with that.

Also, webpack isn't a direct dependency of the package, even though we need it, so this declares it correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,8 @@
     "@silverstripe/webpack-config": "^0.5.0",
     "babel-jest": "^20.0.3",
     "copy-webpack-plugin": "^4.2.0",
-    "jest-cli": "^19.0.2"
+    "jest-cli": "^19.0.2",
+    "webpack": "^2"
   },
   "resolutions": {
     "eslint": "^4.6.1"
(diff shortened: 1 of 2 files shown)


dhensby:
gah, the location of the webpack bin seems to be different locally to on travis, which is a pain. I've reverted those changes; we should still explicitly declare our dependency on webpack, though.

robbieaverill:
Does this need to be added to all of our modules?

dhensby:
if webpack is a dep of all of them, then probably

robbieaverill:
Last question - should it be a dev dependency?

dhensby:
> Last question - should it be a dev dependency?

I thought about that myself... I guess the question is do we need this dependency *only* during dev or do we need it during prod deploys too? IE: Do we build dists on deploy?

As we ship the built dist files with the application, it may well be better suited to dev, though I'm flexible on this point...

What's your feeling @robbieaverill ?

robbieaverill:
I think that should be up to the developer and how it best suits their deployment workflow. Personally I’m not a fan of building frontend assets during a deployment, and would probably lean towards making it a dev dependency by default. Either way, wouldn’t devs who want to build the assets on deployment be doing it for their project rather than the core modules?

dhensby:
yep - most likely

dhensby:
@robbieaverill moved it to a dev dep

PR #469, opened 2018-03-26
GraphQL query customisation

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/13


client/src/lib/dependency-injection/InjectorProvider.js
```diff
@@ -0,0 +1,4 @@
+import React from 'react';
+import provideInjector from './provideInjector';
+
+export default provideInjector((props, context) => <div>{children}</div>);
```
flamerohr: where's `children` defined? ;)

unclecheese: Not sure how this got in here. Should have been deleted.

client/src/lib/dependency-injection/buildApolloGraphqlContainer.js
```diff
@@ -0,0 +1,162 @@
...
+  /**
+   * A register of templates registered with this container
+   */
+  templates: {
+    [CREATE]: buildCreateMutation(captureTag),
+    [READ]: buildReadQuery(captureTag),
+    [READ_ONE]: buildReadOneQuery(captureTag),
+    [UPDATE]: buildUpdateMutation(captureTag),
+    [DELETE]: buildDeleteMutation(captureTag),
+    [QUERY]: buildBaseQuery(captureTag),
```
unclecheese: We should initialise with these empty and inflate them at boot time. Would help to decouple Injector from SilverStripe. Each consumer of Injector should in theory be supplying its own set of templates.

client/src/boot/buildApolloClient.js
```diff
@@ -16,8 +16,8 @@ function buildApolloClient(baseUrl) {
     shouldBatch: true,
     addTypename: true,
     dataIdFromObject: (o) => {
-      if (o.id >= 0 && o.__typename) {
-        return `${o.__typename}:${o.id}`;
+      if ((o.id >= 0 || o.ID >= 0) && o.__typename) {
```
tractorcow: Can you explain the case issue here? Should we check that at least one of id or ID are not undefined first?

Maybe shift the ID into a var instead for clarity.

client/src/boot/buildApolloClient.js
```diff
@@ -16,8 +16,8 @@ function buildApolloClient(baseUrl) {
...
+        return `${o.__typename}:${(o.id >= 0) ? o.id : o.ID}`;
```
tractorcow: Maybe we should shift id into a variable here for readability.

client/src/lib/dependency-injection/ApolloGraphqlManager.js
```diff
@@ -0,0 +1,312 @@
...
+    this.fragments = { ...fragments } || {};
+
+    this.reduceApolloConfig = this.reduceApolloConfig.bind(this);
+  }
+
+  setConfig(name, value) {
+    // prevent overriding functionally important configs
+    if (protectedConfig.includes(name)) {
+      throw new Error(`
+Tried to set protected config values: '${name}', which is discouraged.
```
tractorcow: Weird newline here.

client/src/lib/dependency-injection/ApolloGraphqlManager.js
```diff
@@ -0,0 +1,312 @@
...
+    }
+    this.config = {
+      ...this.config,
+      [name]: value,
+    };
+
+    return this;
+  }
+
+  transformApolloConfig(config, callback) {
```
tractorcow: jsdocs please :D

client/src/lib/dependency-injection/ApolloGraphqlManager.js
```diff
@@ -0,0 +1,312 @@
...
+      [config]: [
+        ...transformList,
+        callback,
+      ],
+    };
+
+    return this;
+  }
+
+  addParam(param) {
```
tractorcow: jsdocs please :D

client/src/lib/dependency-injection/buildApolloGraphqlContainer.js
tractorcow: Can you implement this or should we leave it for this merge @unclecheese ?

tractorcow:
I think taking the asset-admin changes https://github.com/silverstripe/silverstripe-asset-admin/pull/753/files and constructing an equivalent unit test from this is sufficient. :) but with an arbitrary dataobject type.

client/src/lib/dependency-injection/buildApolloGraphqlContainer.js
unclecheese: done

client/src/lib/dependency-injection/ApolloGraphqlManager.js
flamerohr: I think all error messages should use something like this https://github.com/airbnb/deline to get rid of coding indents when it appears in the console

Those lengthy error messages plus indents is quite an eyesore now :D 

I can raise a pull request later on.

PR #470, opened 2018-03-28
BUGFIX: Don't allow button to be removed from DOM before click handler fires

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/749


tractorcow:
When I unpublish a file I get this dialog.

image

When I click "OK" nothing happens (no ajax handler etc).

tractorcow:
I've done some research, and this isn't the correct fix.

We STILL need to un-nest the <button /> tag surrounding all nested popover tags.

See https://stackoverflow.com/questions/39386497/can-i-nest-button-inside-another-button

tractorcow:
Re-raising at https://github.com/silverstripe/silverstripe-asset-admin/issues/774

Issue #472, opened 2018-04-04
TreeDropdownField renders no items after navigation + remount


scott1702:
After navigating to children inside the TreeDropdownField, when the component is mounted to the DOM again (see example scenarios below) then there are no items rendered into the dropdown. This seems to be caused by the `visible` prop on TreeDropdownField, ie. on mount the tree is loaded again and the tree prop is reset for children at the root but the visible prop is still set to the child previously navigated so no items are rendered.

Example scenarios:
- Navigate to child and attempt to save page with failing validation
- Navigate to child items in TreedropdownField, then navigate to other area of CMS with a TreedropdownField of the same ID (e.g. a gridfield of items containing TreeDropddownFields).

![treedropdownfield](https://user-images.githubusercontent.com/10215604/38286359-1246bbb0-3819-11e8-9944-cd8176bcf447.gif)

robbieaverill:
@scott1702 is this the same issue as https://github.com/silverstripe/silverstripe-cms/issues/2112?

scott1702:
@robbieaverill Yep, sure is

PR #475, opened 2018-04-04
Ensure visible prop is reset even if value doesn't exist

scott1702:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/472 and fixes https://github.com/silverstripe/silverstripe-cms/issues/2112


robbieaverill:
@nglasl would you like to check to see if this fixes https://github.com/silverstripe/silverstripe-cms/issues/2112 for you?

nglasl:
@robbieaverill, don't have time to look at this right now, but I did play around with this change and it fixed the problem while causing other quirks (from what I remember). Hence why my temporary solution was to remove that code block entirely.

scott1702:
Closing in favour of https://github.com/silverstripe/silverstripe-admin/pull/476

PR #476, opened 2018-04-05
BUG Fix issue with selected values in large trees breaking initialisation

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/2112


tractorcow:
Fix is:

 - ensure that empty values on re-loaded fields reset the path to []
 - ensure that we forceValue = thevalue when loading the tree for the first time
 - switch out initialise() into a promise-returning method so it is now testable

scott1702:
Can confirm this fixed my issue at https://github.com/silverstripe/silverstripe-admin/issues/472 👍

PR #477, opened 2018-04-09
BUG Don't add redundant href="#" to tabs

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/761


robbieaverill:
This probably also means I can remove this workaround: https://github.com/silverstripe/silverstripe-versioned-admin/pull/5/files#diff-03b5d7d0650bd01a8b44f44467639d11R36

tractorcow:
Oh yes, that's a problem there. :D

PR #485, opened 2018-04-14
FIX Make 'id' a required prop

raissanorth:
The `PopoverField` renders a `Popover` component, which is imported from the `reactstrap` node module . The latter requires a target prop.

The `PopoverField` assigns its own `id`, which is passed down as a prop, to the `Popover`'s `target`. Therefore the `PopoverField`'s `id` should be a required prop also.

I wonder if it should be provided as a default prop or if the change in this PR is enough.


dhensby:
this needs rebasing + building

robbieaverill:
@raissanorth if you want to rebase and build this again I'm happy to merge it before it gets stale again

raissanorth:
@robbieaverill Yes, please.

PR #490, opened 2018-04-17
NEW Add Loading indicator component, implement into FormBuilderLoader

robbieaverill:
This PR adds a general purpose Loading indicator component that will grow to its container:

image

It also removes the custom loading indicator from FormBuilderLoader and uses this component instead. This was patched recently in https://github.com/silverstripe/silverstripe-admin/pull/487, but makes more sense to default to centering itself in the area it has. This also means that containers with fill-height won't jump around during loading.

### Example from versioned-admin

Before:

image

After:

image

### Example from asset-admin

Note:
* asset-admin PR is separate
* the left box loading indicator alignment is pre-existing, I didn't resolve it in this PR

Before:

image

After:

image

---

## Follow up pull requests

- [x] https://github.com/silverstripe/webpack-config/pull/19
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/771
- [ ] https://github.com/silverstripe/silverstripe-campaign-admin/pull/94
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/778


unclecheese:
Have resolved the double loader in asset-admin here: https://github.com/silverstripe/silverstripe-asset-admin/pull/778

PR #491, opened 2018-04-17
MINOR: Stop alerts stretching in flex modal contexts

lukereative:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/2136


robbieaverill:
Related (but different): https://github.com/silverstripe/silverstripe-admin/issues/492

PR #499, opened 2018-04-23
Revert button container changes, use new transparent container div

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/774


client/src/components/PopoverField/PopoverField.js
```diff
@@ -62,27 +48,26 @@ class PopoverField extends Component {
...
+        <Button {...buttonProps}>{this.props.title}</Button>
         <Popover
           id={`${this.props.id}_Popover`}
           placement={placement}
           isOpen={this.state.isOpen}
           target={this.props.id}
           toggle={this.toggle}
           className={this.props.popoverClassName}
-          container={this.getContainer()}
+          container={this.props.container}
```
tractorcow: The fix was to make the `div` the container not the parent form. it messes up accessibility; We're back to where we were before the original broken fix was applied.

tractorcow: You need innerRef applied to the div, and retain the getContainer() method.

tractorcow: ok, div doesn't support innerRef... can be solved :D

unclecheese: Yeah, `innerRef` is a special reactstrap thing. I guess I didn't understand the goal of the fix. I was just looking at it like fixing the invalid HTML, and by association, the broken submit buttons.

Rendering the popover in that container div causes issues, too, because it's not tall enough to fit the popover and it gets squashed.

tractorcow: `ref`. Follow up at https://github.com/silverstripe/silverstripe-admin/pull/501

PR #501, opened 2018-04-26
ENHANCEMENT Ensure that popover has correct container assigned

tractorcow:
Fixes regressions in https://github.com/silverstripe/silverstripe-admin/pull/499

Fixes https://github.com/silverstripe/silverstripe-asset-admin/issues/774


client/src/components/PopoverField/PopoverField.scss
```diff
@@ -1,24 +1,4 @@
...
-    &:hover,
-    &:focus,
-    &:active {
-      background-color: transparent;
-      border: 0;
-    }
-  }
-}
-
 .popover {
```
unclecheese: This was added based on @flamerohr 's feedback:

> you're going to break styling things like button groups with that container logic... Another option is to move the styling to the container div and make that div look like a button (with the generic btn class from bootstrap)
Then have a transparent button cover the entire container div.

What has changed that we now don't have to do that?

tractorcow: For now it looks and works appropriately; It's admittedly a less than perfect solution, until `<React.Fragment>` is available in framework. Once it is, we should set the container to the button group holding the actions, and we can stop using the wrapping div.

tractorcow: Ok, so we're on react 15.3.1, and we need at least react 16.2 to use fragments. I guess we wait until then to implement it?

tractorcow: Also I had to remove the extra styling because the button appeared twice.

flamerohr: transparent overlay is what I described for the "real" button should be styled as :) a simple `visibility: hidden`

maxime-rainville:
To summarize the delightful round-about conversation we've been having...

The root cause of the problem is that we're using the the Popover component to achieve a purpose that should have been implemented by the [Dropdown Component](https://reactstrap.github.io/components/dropdowns/).

This was meant to be a temporary solution because reacstrap hadn't implemented the drop down at the time.

As part of https://github.com/silverstripe/silverstripe-asset-admin/issues/779 we will implement the DropDown component properly, at which time we can apply it where we are misusing the Popover. 

The issue raised @flamerohr is relevant, but only occurs if we use the popover in a ButtonGroup, which we are not doing anywhere right now.

I will hereby merge this pull request with the understanding that it's a temporary fix and that we shall never speak of it again.

PR #505, opened 2018-05-01
NEW Make Preview component injectable

robbieaverill:
Parent story: https://github.com/silverstripe/silverstripe-versioned/issues/95


robbieaverill:
We lost FormAlert here during the rebase (it was a separate commit). I'll make another pull request.

PR #507, opened 2018-05-01
Include missing bugfixes from 1.0 into 1.1

stephenmcm:
Currently 1.1 is missing:
https://github.com/silverstripe/silverstripe-admin/pull/476 which impacts https://github.com/silverstripe/silverstripe-cms/issues/2112

The other two missing commits ( 79dba01 and c632ff2 ) appear less critical but none the less feature drift is bad.


robbieaverill:
Thanks, we will be merging these branches up shortly

PR #508, opened 2018-05-02
Allow the site tree sidebar when editing a page to be resized

mikenz:
A bit hacky but does what we needed, might be of use to someone else.

https://silverstripe.uservoice.com/forums/251266-new-features/suggestions/6187340-allow-the-site-tree-sidebar-when-editing-a-page-to


dhensby:
the dist files need rebuilding :)

mikenz:
@dhensby you say that like I didn't use the github editor to create this branch and PR...

chillu:
Thanks @mikenz! Have you tested this on all supported browsers for the CMS? https://docs.silverstripe.org/en/4/getting_started/server_requirements/#client-side-cms-requirements

maxime-rainville:
I tested it in IE11, Firefox and Chrome. It's kind of cool, but they are some rough edges. There probably needs to be a min-width/max-width beyond which you can not go. I've made a video of the end results https://youtu.be/xPSIbtutY38

Before asking @mikenz to make any changes, we probably need to decide upfront if this is something we think is worthwhile. 

image

image

maxime-rainville:
I think we should let this one go. We're looking at rebuilding the CMS UI in react. Allowing resizing of the SiteTree just seems like it might open up an extra can of worms we don't need.

@clarkepaul Do you have views on the usefulness of this?

clarkepaul:
Although I'd love to see this functionality for all panels in the CMS, I think unless we look at this from a layout rebuild and do it properly I think its just going to do a lot of UI breaking.
If someone wants to continue with this for use in their own CMS I'd suggest putting a min-max width on panels so there isn't as much UI breakage and snap/close the preview if the tree and page edit panel take too much space.

maxime-rainville:
I'll close it. This needs to have a more overarching approach.

Issue #510, opened 2018-05-03
Refactor fieldHolder to be applied via transforms instead of being the default export


NightJar:
This allows developers to decorate the component proper as opposed to the higher order component (fieldHolder).

As an example:
Recently trying to change `FieldGroup` via `Injector.transform` with another higher order component that returned a subclass of the FieldGroup (as is a reasonably common case with PHP Injector), I was met with an error for "function getClassName is undefined" - because instead of extending the FeildGroup as I was expecting, I was instead extending the HigherOrderComponent (fieldHolder).

This description is probably a bit convoluted, so I'll update the issue with the workaround addressing the above as a PR soon (as worked on with @unclecheese ).

NightJar:
#511 solved the issue for a singular field, as a proof of concept for the PR (doing only what was necessary at the time instead of changing too much at once).
There are still [quite a few other fields](https://github.com/silverstripe/silverstripe-admin/blob/1/client/src/boot/registerComponents.js#L44-L61) that this separation of concerns can be applied to.

raissanorth:
Yes please we are running into a similar issue as the one described above whilst working on https://github.com/dnadesign/silverstripe-elemental/issues/486

PR #511, opened 2018-05-03
ENHANCEMENT Register fieldHolder HOCs with injector

NightJar:
rather than having their default export be a higher order component of the field proper. This allows for more intuitive use of the Injector API, particularly around the use of transforms.

A _partial_ solution for #510 

This is a proof of concept PR, showing the approach taken for #510, and also provides the ability to garnish the `Injector` higher order component with extra _context_ as to provide a way to _group_ components, rather than just specificity in order to target _singular_ components.

Without this extra functionality one is left with no way to achieve a common use case, such as I had in attempting to `transform` a _set_ of `FieldGroup` components - in first attempts I was inadvertently overriding _all_ `FieldGroups`. This was rather in-ideal in the test of a Page type with Elemental applied featuring both a native FieldGroup and elements via an ElementalArea on the view ([history](https://github.com/dnadesign/silverstripe-elemental/pull/224))


dhensby:
dist needs rebuilding :( I assume a rebase is needed too

NightJar:
Sure did!

client/src/boot/applyTransforms.js
```diff
@@ -31,6 +32,15 @@ const togglePristineState = (field, isPristine = false) => {
 };
 
 const applyTransforms = () => {
+  Injector.transform(
+    'field-holders',
+    (updater) => {
+      const fields = [
+        'FieldGroup',
+      ];
+      fields.forEach((field) => updater.component('FieldGroup', fieldHolder, `${field}Holder`));
```
robbieaverill: This feels like a premature optimisation - was there a justification for making this an array off the bat?

robbieaverill:
@flamerohr @unclecheese it would be cool to get your insight on this change when you have some time =)

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -113,9 +113,16 @@ class FormBuilder extends Component {
     delete componentProps.input;
     const { identifier } = this.props;
     const { name } = componentProps;
+    const context = [
+      identifier,
+      componentProps.data.extraContext,
```
flamerohr: this seems like a bad practice and also not mentioned in the pull request, why isn't the context in the identifier where it is easily discoverable?

So you're looking to infer an extra dynamic context right?
I remember we removed dynamic contexts due to performance hits it potentially brings to the table (especially because props can and will change dynamically with transformers).

client/src/boot/applyTransforms.js
flamerohr: agreed, add a few more fields to it to justify :)

client/src/components/FormBuilder/FormBuilder.js
NightJar: Hmm, you're completely right. Let me edit the PR context instead of elaborating here.

client/src/boot/applyTransforms.js
NightJar: Hmm. Originally this body of work included _all_ affected form fields, however that required far more than a simple pull request as a proof of concept (such as complete regression testing of _all_ components in the CMS). I worked on this along side @unclecheese and had (foolishly) omitted context from the PR thinking it was pre-gained knowledge (plus was in a rush at the time). That's bad for open source, allow me to rectify in the opening comment.

It _is_ premature optimisation... but it is with the intent to revisit in the near future to complete a much more overarching (and completely separate - c.f. #510) story. This is the minimal to get feedback on the approach, and complete the task at hand (https://github.com/dnadesign/silverstripe-elemental/issues/181)

flamerohr: add a `// @todo: we need more!` comment then, I'm fine with this

NightJar:
Thanks team, pushed the comment and modified the commit message to expand a little :)

client/src/components/FormBuilder/FormBuilder.js
flamerohr: hmm, I don't think `extraContext` is the proper name for this property described then.

I can understand the need for something like the usecase but why can't the transformer check for other data that you've provided such as `ElementIcon`, `ElementID` and `ElementID`? And if those data points don't match then pass-through the original FieldGroup.

I see that you've already got a filter for the transformer, so you're not too far off from that, I think?

flamerohr: don't get me wrong, I'm looking for a justification for this potential performance hit :)

robbieaverill: bump @NightJar

NightJar: Naming things is hard. The intention here was such that the `contextGroup` (aka extraContext) does not need to be dynamic. The data array was readily available but may not be the correct place for it I can agree. However I think having some form of grouping elements via context should something provided via the core. Ideally I guess grouping would be a part of the FormField's `getSchemaState` so that it comes out in `this.props` as opposed to `componentProps`, but that would require a third dependency PR to framework. Is this something you'd see as a way forward, or do you think the thing you mentioned with the transformer filter that I didn't really understand (please elaborate) is better?

NightJar: Yeah cool, 'short circuiting' the render function is what I thought you meant, thanks for clarifying/confirming :)

I'm open to the idea of structuring the other PR another way, the subclassing was more the immediate way I could see that would allow me to alter the output of the render function. I had originally tried specifying a different component to use (via schema definition), however that didn't seem to work, and using a transform seemed better in terms of approach anyway.

I'll fiddle around a bit and see if I can maintain functionality whilst reducing this PR to only cover the fieldHolder alterations (i.e. drop the context alterations).

NightJar: Cool, I just got rid of all this stuff :)
Thanks for helping me through it @flamerohr

Issue #423, opened 2018-02-06
Update ApolloClient from 0.5 to 2.x


chillu:
# Overview

Our version of ApolloClient is 1.5 years old. While its still performing it's current duties well, a version that old won't be a good base to build further features on.

# Tasks

 * [ ] ~~Switch to `apollo-boost` package~~
 * [ ] Rewrite query manager to links

# Notes

 * Already have a fetch polyfill (isomorphic-fetch)
 * Need to review changelog for 0.5 to 1.0 breaking changes

# Related

 * [Apollo: Migrating from 1.0](https://www.apollographql.com/docs/react/2.0-migration.html)
 * [CHANGELOG](https://github.com/apollographql/apollo-client/blob/master/packages/apollo-client/CHANGELOG.md)
 * [UPGRADE.md](https://github.com/apollographql/apollo-client/blob/master/Upgrade.md)
 * [Apollo Client 2.0 Blog Post](https://dev-blog.apollodata.com/apollo-client-2-0-beyond-graphql-apis-888807b53afe)
 * [Customise GraphQL queries and mutations](https://github.com/silverstripe/silverstripe-graphql/issues/13)

## Pull Requests
- [x] https://github.com/silverstripe/silverstripe-admin/pull/514
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/780
- [ ] https://github.com/silverstripe/silverstripe-graphql/pull/154

lukereative:
In order to upgrade to Apollo 2.x we need to use the new cache system called [IntrospectionFragmentMatcher](https://www.apollographql.com/docs/react/advanced/fragments.html#fragment-matcher) which requires providing JSON to Apollo that lists all the fragments in the GraphQL schema.

## Proposed solutions

1. Generate a JSON file on `flush` and store it in the `ASSETS_DIR`, since that is the only guaranteed writable directory, this file can then be read when creating the Apollo client while loading the CMS.
   - The main benefit of this is avoiding adding a dynamic request upon loading the CMS (which in testing added ~500ms to initial CMS load)
   - The downside being that this adds another step for developers making changes to PHP GraphQL interfaces, if their changes are to YAML then they would need to `flush` anyway.
2. Add a dynamic request to the building of the Apollo client, which happens upon CMS load. The pros and cons are opposite to the previous solution:
   - The main benefit of this is not having to `flush` when making changes to PHP GraphQL interfaces.
[...]

PR #514, opened 2018-05-07
Apollo 2.5 upgrade

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-admin/issues/423


dhensby:
needs rebasing and building it seems :(

client/src/boot/buildApolloClient.js
```diff
@@ -1,53 +1,87 @@
-import ApolloClient, { createNetworkInterface } from 'apollo-client';
-import { printRequest } from 'apollo-client/transport/networkInterface';
+/* global window */
+import fetch from 'isomorphic-fetch';
+import ApolloClient, { InMemoryCache, IntrospectionFragmentMatcher, printAST, ApolloLink } from 'apollo-boost';
+import { BatchHttpLink } from 'apollo-link-batch-http';
 import qs from 'qs';
 
 function buildApolloClient(baseUrl) {
```
maxime-rainville: It might be worthwhile splitting some of that logic up and cutting down a bit on the inline variables/function. It would make it easier to figure out what's going on and allow us lesser mortals to follow along.

`buildApolloClient` could become something like:
```JS
function buildApolloClient(baseUrl) {
    return fetch(`${baseUrl}graphql`, httpOptions)
       .then(doSomething)
       .then(doSomethingElse)
       .then(doFinalThing)
}
```
Each method applied to the promise should have proper doc.

On another note, should there be some error handling on `fetch`? It's an HTTP call, so there's a chance it will fail. Not sure what the best practice is here.

unclecheese: Kind of a coincidence, but maybe not so much -- this is exactly the thing I noticed when I was reviewing it, and refactored this into smaller bits as you suggested. I just haven't pushed up the branch yet.

maxime-rainville:
I'm happy. @unclecheese asked me to merge on Slack.

PR #518, opened 2018-05-11
FIX: Fix tooltip/toggleable descriptions (fixes #517)

### LINKED ISSUE #517: CMS field description tooltips/toggles are broken
I had no idea this functionality even existed until it was raised on Slack... looks like we broke this: https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/how_tos/cms_formfield_help_text/#how-to-show-help-text-on-cms-form-fields

This looks like it should be fairly easy to fix (\*gulp\*):

- https://github.com/silverstripe/silverstripe-admin/blob/1.1.0/client/src/legacy/LeftAndMain.FieldDescriptionToggle.js
- https://github.com/silverstripe/silverstripe-admin/blob/1.1.0/client/src/legacy/LeftAndMain.FieldHelp.js

kinglozzer:
Replacing hardcoded classes with hardcoded classes sucks, but until there’s a proper refactor it gets things working 🤷‍♂️ 

What it looks like:

screen shot 2018-05-11 at 11 46 02

Clicking the _i_, or “My custom icon”, will show the description beneath the field with the default description appearance


client/src/legacy/LeftAndMain.FieldHelp.js
```diff
@@ -10,28 +10,40 @@ $.entwine('ss', function($) {
...
         descriptionEl.remove();
       }
     },
   });
 
   $(".cms .field.cms-description-tooltip :input").entwine({
     onfocusin: function(e) {
-      this.closest('.field').tooltip('open');
+      var holder = this.closest('.form__field-holder');
+      if (holder.data('tooltip')) {
```
kinglozzer: Prevents “calling open prior to initialization” errors

client/src/legacy/LeftAndMain.FieldHelp.js
```diff
@@ -10,28 +10,40 @@ $.entwine('ss', function($) {
...
       if(descriptionEl.length) {
+        var offset = this.css('padding-left');
         this
           // TODO Remove title setting, shouldn't be necessary
           .attr('title', descriptionEl.text())
-          .tooltip({content: descriptionEl.html()});
+          .tooltip({
+            content: descriptionEl.html(),
+            position: {
+              my: "left+" + offset + " top"
```
kinglozzer: Default is `left top+15`, which appears below and to the left of the input - this aligns it against the input as per screenshot above

client/src/legacy/LeftAndMain.FieldHelp.js
```diff
@@ -10,28 +10,40 @@ $.entwine('ss', function($) {
    * input fields (e.g. GridField), and aren't compatible
    * with showing tooltips.
    */
-  $(".cms .field.cms-description-tooltip").entwine({
+  $(".cms .field.cms-description-tooltip .form__field-holder").entwine({
     onmatch: function() {
       this._super();
 
-      var descriptionEl = this.find('.description'), inputEl, tooltipEl;
+      var descriptionEl = this.find('.form__field-description');
```
dhensby: I'm not 💯 on top of our conventions, but shouldn't classes that trigger JS behaviour not be BEM classes (ie: not style classes)?

I'd be using `.js-tooltip`

client/src/legacy/LeftAndMain.FieldHelp.js
dhensby: Couldn't we had a class to make it visible instead so it's managed in CSS rather than JS? Maybe not a big deal but I'd probably be using `.is-visible` on the `.form__field-description` elements...

client/src/legacy/LeftAndMain.FieldHelp.js
flamerohr: it's a debatable point, I think we still allow for (and use) BEM classes for JS behaviour all over the codebase.

No convention/practice has really stuck - e.g. trying to prefix `js-` for function/js classes got cumbersome.
I think in the end, as long as the class communicates the purpose it has (the whole point of BEM), it's fine 👍

client/src/legacy/LeftAndMain.FieldHelp.js
flamerohr: I thought it was this way to check if it has been initialised, since it's jQuery UI and not jQuery core
```
if($element.data('ui-tooltip')) {
    $element.tooltip('destroy');
}
```

flamerohr: oh, maybe it's because of how old the jQuery UI version is :/

robbieaverill:
cc @newleeland @clarkepaul

clarkepaul:
If it isn't a [Bootstrap based](https://getbootstrap.com/docs/4.1/components/tooltips/) tooltip then I'd rather kill/deprecate it otherwise it will get used more. I probably need to see it in action, @maxime-rainville can we have a catchup about this to make a call.

maxime-rainville:
@clarkepaul I've recorded a demo of what this looks like https://youtu.be/GXW4eZho_cc

maxime-rainville:
I had a chat with @clarkepaul. We've concluded that the UI is not inline with the rest of the CMS and has accessibility issues. We might end up doing something similar in the future, but we'll need to have a design process upfront and a better API.

I'll do a separate PR to remove the article from the doc and I'll close this PR.

maxime-rainville:
I've got a PR to ditch the buggy examples https://github.com/silverstripe/silverstripe-framework/pull/8510

PR #521, opened 2018-05-16
Add graceful fallback when static graphql schema not present

unclecheese:
We should only throw a hard error if neither the static nor dynamic options work.


client/src/boot/apollo/getGraphqlFragments.js
```diff
@@ -1,22 +1,56 @@
...
+  const fetchSchema = async (url) => (
+    fetch(url, fetchConfig)
+      .then(handleError)
+      .then(result => result.json())
+      .then(parseResponse)
+  );
+
+  let response;
+  try {
+    response = await fetchSchema(staticURL);
```
maxime-rainville: Any chance we could tell `getGraphqlFragments` not to do this call when the cache is disabled.

We're going to pollute people's error logs with meaningless 404 otherwise.

maxime-rainville: On a related note, I ran the code for real this time. My `types.graphql` file gets generated as expected when I enable the cache. However, when I disable the cache and run a dev/build and a flush, the file stays there.

Since we call staticURL first, the old staled type definition will be loaded. This might be better address in silverstripe/graphql.

unclecheese: I was trying to avoid having to pollute the config with concerns like this, but I think that's going to be the most satisfactory option. HTTP isn't the best way to read a config setting. :)

New plan:

* Pass `isUsingStaticCache` into the `getGraphQLFragments` function.
* This will determine the sequence of URLs to try
* 404 only happens when your project is not set up to support the chosen setting (missing `types.graphql` when on, or missing `IntrospectionProvider` extension when off).
* Either way, you still get a fallback

Good catch on the garbage collection. That task needs to delete the file as well.

maxime-rainville: @unclecheese I'm still getting a 404.

I pull down the latest version of admin, asset-admin and graphql. I've put the follow code in my config.

```YML
SilverStripe\GraphQL\Controller:
  cache_types_in_filesystem: false
  extensions:
    - SilverStripe\GraphQL\Extensions\IntrospectionProvider
```

Am I missing something?

PR #523, opened 2018-05-19
Add fallback for graphql types not existing

unclecheese:
When a graphql schema isn't present (e.g. a framework test, or any application not using asset-admin/versioned-admin), the `/graphql/types` endpoint errors, as it should because it's running a query against a non-existent schema.

This fix ensures a graceful fallback by catching the error, setting the response to null, and bypassing  the instantiation of `IntrospectionFragmentMatcher`.

A second fix to make this more robust should happen on the graphql side. We should probably detect non-existent schemas and return 200, but what that response should contain requires more thought. Mocking up an empty JSON response feels a bit hacky. Will create a separate story.
 
Resolves https://github.com/silverstripe/silverstripe-framework/issues/8029


unclecheese:
GraphQL issue follow-up https://github.com/silverstripe/silverstripe-graphql/issues/157

client/src/boot/apollo/buildClient.js
```diff
@@ -10,7 +10,12 @@ import Config from 'lib/Config';
 const buildClient = async (baseUrl) => {
     const graphQLConfig = Config.getSection('SilverStripe\\Admin\\LeftAndMain').graphql;
     const cachedTypenames = graphQLConfig && graphQLConfig.cachedTypenames;
-    const fragmentData = await getGraphqlFragments(baseUrl, cachedTypenames);
+    let fragmentData;
+    try {
+      fragmentData = await getGraphqlFragments(baseUrl, cachedTypenames);
+    } catch (e) {
```
maxime-rainville: Would it be practical to use a typed exception here? No biggy if it isn't.

If this behavior is only meant to be triggered during the behat test, it probably would not hurt to have comment explaining briefly why were are doing this. Otherwise someone will scratch their head on this one 2 years from now when looking at this.

unclecheese: Yeah, could do. It's not just a behat fix, though. Anytime you're using admin without `asset-admin`, you'll get this error. Still a super edge case, but more reality-based than a testing anomaly. :)

Issue #498, opened 2018-04-22
When contents of a panel is too big the North Toolbar breaks


clarkepaul:
In the following case the image within the history of a page should be limited to max-width 100%, as it is not constrained the North Toolbar breaks into the preview.
![image](https://user-images.githubusercontent.com/555033/39093670-62061134-4677-11e8-813d-146ccf45c236.png)

![image](https://user-images.githubusercontent.com/555033/39093665-5411479c-4677-11e8-8de0-2caf75408cab.png)

I would think that we need to fix the over flowing file but also it would be good to see if we can fix it even if there was overflowing content within the body of the panel. Another case of this happening is within the comments module https://github.com/silverstripe/silverstripe-comments/issues/243.

PR #527, opened 2018-05-21
Fix overflowing menu bar on admin sections

ScopeyNZ:
This addresses an issue raised in silverstripe/silverstripe-comments#243 and silverstripe/silverstripe-admin#498 where the menu at the top of the admin screen extends into the preview when in split mode.

This is caused by the whole content section having content that is larger than the container. The blocks do not flex-shrink in some browsers due to a quirk where they need a width or min-width set low enough to indicate the block width can be shrunk to less than the width of the content in it. This happens when there's an inflexible block (the page tree in this case). Here's some relevant SO questions I dug up trying to diagnose the issue:

https://stackoverflow.com/questions/38723559/flex-item-exceeds-its-container
https://stackoverflow.com/questions/33790219/firefox-not-respecting-flex-shrink

The change affects all blocks with the class `flexbox-area-grow` but it should only change any block that has overflowing content. In testing I couldn't identify any other block with this class that was negatively affected.


robbieaverill:
cc @clarkepaul

clarkepaul:
Cool, took a look and seems to fix the tab issue. I don't fully understand how this works or if there is a better fix but I think this fix does what we need it to. In other words "Looks good".

This fix wont stop content from extending the panel which is below the tabs when in Split Mode (for example whats happening on this issue https://github.com/silverstripe/silverstripe-comments/issues/250 ), but does ensure the tabs are usable.

robbieaverill:
The change looks good to me, but seeing as `flexbox-area-grow` is used in so much of the CMS I'd like to get someone with a bit more CSS knowledge to have a look first. @lukereative would you mind having a look when you have a minute?

PR #532, opened 2018-05-29
BUGFIX: Move AnchorSelectorField to admin

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-framework/issues/7996


tractorcow:
See my comment at https://app.zenhub.com/workspace/o/silverstripe/silverstripe-framework/issues/7996#issuecomment-395295048. I dislike this approach.

unclecheese:
The CMS bundle included in `CMSMain`, so if you go directly to a ModelAdmin page, I don't see any reason why that file would be in the document.

I wonder if we need to make `registerComponents` a special file that gets loaded via `LeftAndMain.extra_requirements_javascript`?

tractorcow:
> The CMS bundle included in CMSMain, so if you go directly to a ModelAdmin page, I don't see any reason why that file would be in the document.

The reason is that your modeladmin might use components provided by CMS module.

> I wonder if we need to make registerComponents a special file that gets loaded via LeftAndMain.extra_requirements_javascript?

Sure you can do that, if you want to split "shared" and "exclusive" scripts.

robbieaverill:
The original issue was closed with an alternative solution

PR #540, opened 2018-06-12
NEW: Add behat tests from framework

unclecheese:
Resolves: https://github.com/silverstripe/silverstripe-framework/issues/8112


behat.yml
```diff
@@ -0,0 +1,34 @@
...
+    SilverStripe\BehatExtension\MinkExtension:
+      default_session: facebook_web_driver
+      javascript_session: facebook_web_driver
+      facebook_web_driver:
+        browser: chrome
+        wd_host: "http://127.0.0.1:9515" #chromedriver port
+
+    SilverStripe\BehatExtension\Extension:
+      screenshot_path: %paths.base%/artifacts/screenshots
+      bootstrap_file: vendor/silverstripe/framework/tests/behat/serve-bootstrap.php
```
robbieaverill: This file has been removed in the corresponding PR - do we need to move it to this module instead?

.travis.yml
```diff
@@ -41,18 +59,33 @@ before_script:
...
+  - if [[ $BEHAT_TEST == admin ]]; then (vendor/bin/serve --bootstrap-file vendor/silverstripe/framework/tests/behat/serve-bootstrap.php &> artifacts/serve.log &); fi
+  - if [[ $BEHAT_TEST == cms ]]; then (vendor/bin/serve --bootstrap-file vendor/silverstripe/cms/tests/behat/serve-bootstrap.php &> artifacts/serve.log &); fi
+
 script:
   - if [[ $PHPUNIT_TEST ]]; then vendor/bin/phpunit tests/php; fi
   - if [[ $NPM_TEST ]]; then yarn run build; fi
   - if [[ $NPM_TEST ]]; then git diff-files --quiet -w --relative=client; fi
   - if [[ $NPM_TEST ]]; then git diff --name-status --relative=client; fi
   - if [[ $PHPCS_TEST ]]; then composer run-script lint; fi
+  - if [[ $BEHAT_TEST == admin ]]; then vendor/bin/behat @admin; fi
```
maxime-rainville: It would be nice if could do something like `if [[ $BEHAT_TEST ]]; then vendor/bin/behat $BEHAT_TEST; fi`

But that would probably complexify the `vendor/bin/serve` call above.

Can't think of an elegant way of doing this.

unclecheese: Agreed, but I think there's value in not being too cute here. Readability is important in these configuration files.

PR #545, opened 2018-06-17
Allow overflow-x on scrollable panels

lukereative:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/3742


maxime-rainville:
Maybe I'm misunderstanding what this thing is meant to do, but my gridfield with a large number of summary fields looks exactly the same with or without this fix. Plus with the fix on, I get a weird border in the left nav.

# With the fix on Firefox
image

# Without the fix on Firefox
image

lukereative:
@maxime-rainville It's not an issue in model admin, but for some other scrollable views (main example being edit forms), so if you view a gridfield on an edit page form that is too wide, before this fix you couldn't scroll horizontally but with the fix you can.

As for that weird display in the menu, it's because I used `scroll` rather than `auto`. `scroll` always shows bars but only on OS setups that don't have the newer 'overlay' style scrollbars. So I've changed it to use `auto`.

maxime-rainville:
Ha ... that makes more sense. Thank for screwing my head straight. I'll have another crack at it.

Issue #528, opened 2018-05-21
Grid field headers misaligned if there is a mix of sortable and non-sortable 


clarkepaul:
The table header does not align if there is a mix of sortable and non-sortable headers
![image](https://user-images.githubusercontent.com/555033/40291111-e8f07562-5d16-11e8-80fa-afbc65ad1afa.png)

PR #549, opened 2018-06-20
FIX - Grid field headers misaligned

hdpero:
as per #528


dhensby:
looks like this was built against a stale commit; can you rebase and rebuild, please?

hdpero:
@dhensby I've built a new assets with latest changes on "1" branch, is it ok now?

And I will squash my commits, is that ok?

hdpero:
@dhensby I'm sorry I *ucked this up, that is why I closed and deleted the branch, I'll start a new one, I think that is simpler.

PR #550, opened 2018-06-22
FIX - Grid field headers misaligned

hdpero:
as per https://github.com/silverstripe/silverstripe-admin/issues/528


hdpero:
@dhensby this a new PR that replaces #549 and fixes #528

PR #555, opened 2018-06-27
FIX Stop sslink from hijacking anchor plugin

wilr:
This was raised by a CWP client with the Anchor plugin enabled in TinyMCE.

Clicking an existing page anchor would trigger the custom sslink popup dialog which could not be used to modify the anchor. This adds the ability to ignore anchors (links without a href) from the dialog. Users can then remove / edit using the standard toolbar.

PR #560, opened 2018-07-04
NEW Add ViewModeToggle component

raissanorth:
This React component allows users to toggle between split, edit, and preview mode. 

It utilises Reactstrap's ~PopoverField~ Dropdown, connects to a Redux store.

_ViewModeToggle component_:

image

_ViewModeToggle component on hover_:

image

_Edit view mode chosen and split view mode enabled_:

image

_Edit view mode chosen and split view mode disabled_:

image

_Preview view mode chosen and hover on disabled split mode_:

image

_Sample of ViewModeToggle rendered within Preview component (split mode view chosen)_

image

_Sample of ViewModeToggle rendered within CampaignAdminList component_

image

-----
Important: The test cases introduce a dependency on the JS testing utility Enzyme (recommended by React for shallow rendering, see https://reactjs.org/docs/test-utils.html#overview)

A couple of other things to note:

- ~The value for the minimum screen width to enable split view mode is derived from the CSS variable `media-breakpoint-down(md)` and currently hard coded in `viewModeReducer.js`~ The value for the minimum screen width is based on the prior CSS variable `media-breakpoint-down(md)` used in the media query, minus the width of the expended cms-panel. It is stored in the newly added `constants.js` file and is used in the `viewModeReducer.js`

- Rather than using Reactstrap's `ToolTip` component, this implementation using the same `div` and styling as the current PHP implementation of the toggle. As a consequence the height for of 37px for this div is hard coded to 'disable' the split view option when the screen width is too narrow. While the first approach is preferable, i.e. using the button's disable property in combination with Reactstrap's ToolTip, no events are fired on the disabled button which is why the latter approach was chosen.

- Within `ViewModeToggle-test.js` the creation of the shallow rendered `wrapper` is repeated for each test group. To  avoid unnecessary repetition, , the wrapper could be defined outside of the test groups. In this case the props need to be defined before the wrapper is created. As the tests use rely on different prop values, this would need to be changed within the inidividual tests.


robbieaverill:
Added `feedback-required/core-team` (@silverstripe/core-team) since we're introducing a new React testing pattern with enzyme. This has been discussed already with @lukereative @chillu and @raissanorth internally at SilverStripe.

client/src/components/ViewModeToggle/ViewModeToggle.js
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  renderEditButton() {
+    const { activeState } = this.props;
+
+    // Highlight if chosen view mode
+    let classNames = 'btn btn-secondary icon-view last font-icon-edit-write';
+    if (activeState === 'edit') {
+      classNames += ' viewMode-selected';
+    }
```
robbieaverill: We can use the `classnames` library for this, it's already part of admin

client/src/components/Preview/Preview.js
```diff
@@ -123,6 +125,7 @@ class Preview extends Component {
           <div className="btn-toolbar">
             {this.renderBackButton()}
             {this.buildToolbarButtons()}
+            <ViewModeToggle id="view-mode-toggle-in-preview-nb" area={'preview'} />
```
robbieaverill: Ideally we'd use the injector to inject a ViewModeToggle prop into Preview which would allow devs to customise this component in the Preview implementation if they wanted to. For example:

```js
import { inject } from 'lib/Injector';

// ...

render() {
  const { ViewModeComponent } = this.props;
  return (
    // ..
    <ViewModeComponent { ...props } />

// ...

Preview.propTypes = {
  ViewModeComponent: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
};

// ...

export default inject(
  ['ViewModeToggle'],
  (ViewModeToggle) => ({
    ViewModeComponent: ViewModeToggle
  }),
  'Admin.Preview'
);
```

client/src/styles/bundle.scss
```diff
@@ -59,6 +59,8 @@
 @import "../components/Toolbar/Toolbar";
 @import "../components/TreeDropdownField/TreeDropdownField";
 @import "../components/UsedOnTable/UsedOnTable";
+@import "../components/IframeDialog/IframeDialog";
```
robbieaverill: Is this referenced somewhere else?

client/src/components/Preview/Preview.js
robbieaverill: You'll also need to register the component with Injector in `src/boot/registerComponents.js`

client/src/components/ViewModeToggle/ViewModeToggle.scss
```diff
@@ -0,0 +1,67 @@
...
+.disabled-tooltip {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  height: 37px;
+  background: rgba(255, 255, 255, 0);
+
+  &::before {
+    content: "Screen size too small";
```
robbieaverill: Is there any way of adding this with javascript or PHP? We can't make this string translatable if it's in CSS

client/src/components/ViewModeToggle/ViewModeToggle.js
```diff
@@ -0,0 +1,224 @@
+import React, { PropTypes, Component } from 'react';
+import { Dropdown, DropdownToggle, DropdownMenu, DropdownItem } from 'reactstrap';
+import i18n from 'i18n';
+import { connect } from 'react-redux';
+import { compose } from 'redux';
+import { selectEditMode, selectPreviewMode, selectSplitMode } from 'state/viewMode/viewModeActions';
+import classNames from 'classnames';
+
+/* global window */
```
flamerohr: is there a need to declare the window global anymore?

client/src/components/ViewModeToggle/ViewModeToggle.js
```diff
@@ -0,0 +1,224 @@
...
+  getIconClass() {
+    const { activeState } = this.props;
+
+    switch (activeState) {
+          case 'edit':
+              return 'font-icon-edit-write';
+          case 'preview':
+              return 'font-icon-eye';
+          default:
+              return 'font-icon-columns';
```
flamerohr: the indent for these lines seem off, they're 4 spaces instead of 2.
Could you have a map of the classes somewhere? possibly provided through props, or another extensible means. It would make customising the icons by an external author possible.

client/src/components/ViewModeToggle/ViewModeToggle.js
```diff
@@ -0,0 +1,224 @@
...
+  getTitle() {
+    const { activeState } = this.props;
+
+    switch (activeState) {
+          case 'edit':
+              return i18n._t('Admin.EDIT_MODE', 'Edit mode');
+          case 'preview':
+              return i18n._t('Admin.PREVIEW_MODE', 'Preview mode');
+          default:
+              return i18n._t('Admin.SPLIT_MODE', 'Split mode');
```
flamerohr: Similar with the above comment

client/src/components/ViewModeToggle/ViewModeToggle.scss
flamerohr: Rather than having a `:before` pseudo element, you could have a span styled the same and injected to the `.disabled-tooltip` div element, this will enable translatable strings and not require any hacks.

If you want the hacked version, use the `attr()` css function and pick an arbitrary attribute to get the translation string from.

client/src/components/ViewModeToggle/ViewModeToggle.scss
```diff
@@ -0,0 +1,67 @@
...
+  &::after {
+    content: "";
+    position: absolute;
+    width: 10px;
+    height: 10px;
+    transform: rotate(45deg);
+    display: none;
+    background: #555;
+    left: 47%;
+    top: -6px;
```
flamerohr: if you implement the `span` suggestion above, you can then use borders to implement the "speech bubble" effect which will reduce a lot of this juggling around with css3 rotate and other specific pixel placements.

client/src/state/viewMode/viewModeActions.js
```diff
@@ -0,0 +1,29 @@
+import ACTION_TYPES from './viewModeActionTypes';
+
+export function selectEditMode() {
+    return {
+        type: ACTION_TYPES.SELECT_EDIT,
+        payload: null,
```
flamerohr: can just omit the payload for all these actions...
Also indent is all over the place for this file

client/src/state/viewMode/viewModeActions.js
```diff
@@ -0,0 +1,29 @@
...
+    return {
+        type: ACTION_TYPES.SELECT_SPLIT,
+        payload: null,
+    };
+}
+
+export function enableOrDisableSplitMode(panelWidth) {
+  return {
+    type: ACTION_TYPES.SPLIT_AVAILABLE,
+    panelWidth
```
flamerohr: put panelWidth in a payload object:
```
{
  type: ACTION_SOMETHING,
  payload: { panelWidth },
};
```

flamerohr:
I remember silverstripe-admin had a click simulator somewhere... maybe it was removed because it didn't prove useful.
If Enzyme helps, then got for it. However, I think it's important that you keep it in `devDependencies` instead of `dependencies`

raissanorth:
Hi @flamerohr, thanks a lot for your feedback. I have addressed most things you mentioned - as well as changed the component to use Dropdown instead of PopoverField @lukereative. Unfortunately, yarn throws an error when I try to add enzyme to the dev-dependencies:

image

raissanorth:
If approved I suggest we implement this in https://github.com/silverstripe/silverstripe-versioned-admin. A PR was made to https://github.com/silverstripe/silverstripe-campaign-admin/pull/109 already.

flamerohr:
> If approved I suggest we implement this in https://github.com/silverstripe/silverstripe-versioned-admin. 

That would become a weird circular dependency with the `Preview` component becoming dependent on `ViewModeToggle` which belongs to a module which depends on this module.

Does injector return `null` if a component isn't registered? I remember seeing an error thrown.

> Unfortunately, yarn throws an error when I try to add enzyme to the dev-dependencies.

That's a shame, you can add `// eslint-disable import/no-extraneous-dependencies` to the top of your test file so you can leave it in dev-dependencies...
I think there was a way to config eslint to ignore import dependencies for test files, but the above comment works as a temporary measure to get this through

robbieaverill:
I don't think that means move the component to versioned-admin, but add the view mode toggle to history viewer in versioned-admin when merged here

flamerohr:
> I don't think that means move the component to versioned-admin, but add the view mode toggle to history viewer in versioned-admin when merged here

Sweet, that makes more sense

client/src/components/ViewModeToggle/ViewModeToggle.js
```diff
@@ -0,0 +1,255 @@
...
+    render() {
+      const {
+        activeState,
+        area,
+        splitAvailable,
+        dropdownToggleProps,
+        dropdownMenuProps
+      } = this.props;
+      let { className } = this.props;
+      className = classNames(className);
```
robbieaverill: I think this is a little confusing - firstly we dereference a prop then we reassign it, unclear as to whether this reassigns the prop which we shouldn't really be doing, I assume it doesn't but only the local variable. Could you instead create a `const dropdownClassName = classNames(className);`  maybe?

client/src/components/ViewModeToggle/ViewModeToggle.scss
```diff
@@ -0,0 +1,78 @@
...
+  }
+}
+
+.disabled-tooltip {
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  height: 37px;
+  background: rgba(255, 255, 255, 0);
```
robbieaverill: I think you could probably use some variables for this, and for the `#555` lower. See `$gray-darker` etc in variables.scss and maybe use `darken($gray-dark, 10%)` if you need to

client/src/lib/constants.js
```diff
@@ -0,0 +1 @@
+export const SPLITMODE_BREAKPOINT = 800;
```
robbieaverill: It'd be good to add a comment describing what this is for too :)

client/src/state/viewMode/viewModeActionTypes.js
```diff
@@ -0,0 +1,6 @@
+export default {
+    SELECT_EDIT: 'SELECT_EDIT',
+    SELECT_PREVIEW: 'SELECT_PREVIEW',
+    SELECT_SPLIT: 'SELECT_SPLIT',
+    SPLIT_AVAILABLE: 'SPLIT_AVAILABLE',
```
robbieaverill: Indentation is out here

client/src/lib/constants.js
flamerohr: I'd love to see this configurable somehow (maybe stored in the redux store instead), but out of scope for this pull request I think

raissanorth:
Changed the status to WIP since there are some minor design changes, discussed with @clarkepaul yesterday, still to be made.

client/src/components/ViewModeToggle/ViewModeToggle.scss
```diff
@@ -0,0 +1,122 @@
+.viewmode-dropdown-toggle {
+  margin-left: auto;
+  font-size: 13px;
```
robbieaverill: You could use `$font-size-base` here

client/src/components/ViewModeToggle/ViewModeToggle.scss
```diff
@@ -0,0 +1,122 @@
...
+
+  #splitModeButton {
+    &:disabled {
+      pointer-events: all;
+    }
+  }
+
+  .viewModeToggle-chosenView-title {
+    font-size: 10px;
```
robbieaverill: `$font-size-xxs`

client/src/components/ViewModeToggle/ViewModeToggle.scss
```diff
@@ -0,0 +1,122 @@
...
+    line-height: 10px;
```
robbieaverill: If there's an appropriate line-height variable in _variables.scss or _bootstrap-variables.scss in admin that'd be ideal to use

client/src/components/ViewModeToggle/ViewModeToggle.scss
```diff
@@ -0,0 +1,122 @@
...
+
+.dropdown {
+
+  &.viewModeToggle {
+    position: absolute;
+    right: 10px;
+
+    .viewmode-dropdown-toggle {
+      &:before {
+        font-size: 18px;
```
robbieaverill: `$h3-font-size: 17px` or `$h2-font-size: 19px` - if you could use one of those we'd be reducing customisation away from core

client/src/components/ViewModeToggle/ViewModeToggle.scss
```diff
@@ -0,0 +1,122 @@
...
+// When using the caret property, ReactStrap adds a .dropdown-toggle class to the DropdownToggle component, which in
+// turn changes the background colour. This behaviour is overriden here.
+.viewModeToggle {
+  &.dropdown {
+    &.show {
+      .btn-secondary.dropdown-toggle {
+        box-shadow: inset 0 3px 5px rgba(0, 0, 0, 0.125), 0 0 0 0.2rem rgba(86, 107, 141, 0.5);
+        border-color: #e7ebf0;
+        background-color: transparent;
+        color: #566b8d;
```
robbieaverill: Is there any way we can swap these out for variables, or even `darken()` or `lighten()` calls on variables that are close to these values already?

client/src/components/ViewModeToggle/ViewModeToggle.scss
```diff
@@ -0,0 +1,122 @@
...
+      }
+    }
+  }
+}
+
+
+
```
robbieaverill: Just one empty line at the end would be cool

PR #562, opened 2018-07-06
Don't include font source in export-ignore

kinglozzer:
The font directory includes `icons-reference.html`, which is _really_ useful for picking icons.


dhensby:
I assume the behat failure is not related?

kinglozzer:
> I assume the behat failure is not related?

I’d be very surprised if it is 😛

PR #565, opened 2018-07-10
NEW Update PopoverField styles to match DSM designs

robbieaverill:
Note that an existing use case for the old styles in asset-admin has been temporarily
fixed by moving the existing styles to LegacyPopoverField.scss in asset-admin.

Asset-admin PR: https://github.com/silverstripe/silverstripe-asset-admin/pull/804

Resolves https://github.com/silverstripe/silverstripe-asset-admin/issues/779

Updated design examples:

image

image

DSM: https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/components/5ae267cb98330300118a7851

cc @clarkepaul


lukereative:
The styles aren't being applied to a 'Contained' popover (which is in the pattern lib) as it isn't the child of a div with `.popover-field` class. Otherwise looks good.

screen shot 2018-07-11 at 2 31 36 pm

robbieaverill:
Just rebased to get behat green

Issue #568, opened 2018-07-10
~4 - jQuery Notice boxes not aligned - Preventing access to settings / history


caffeineinc:
I'm finding the alignment of the jquery box is preventing navigation through to a page history, or settings as the `position: fixed` element has a high z-index, and covers the navigation elements. 

It's populated from the header `x-loginattemptnotifications` and contains the content: ```:In the last 18 secs a successful login attempt to your account was registered. The attempt was made from 14.1.35.58.```

The actual div is located 300px to the right of the container, which means the container is present, but never showing the actual content. Perhaps the previous version contained an animation to the left too. 

Changing the CSS property to be shown (by assigning `.notice-item` with `left:0` resolves the issue as the dialog can then be seen, and closed if persistent. 

This can be reproduced by executing the following in the console: 

```jQuery.noticeAdd({text: "<p>myspecial text</p>", stay: false, type: 'good'});``` 

You'll notice it isn't shown until the property for left is added. (perhaps the animation was changed, without updating the css property?)

`vendor/silverstripe/admin/client/src/styles/legacy/_style.scss L:893`

affects 1.1

PR #569, opened 2018-07-10
568 - jQuery Notice boxes not aligned - Preventing access to settings

caffeineinc:
- Change alignment back to left, so that divs can be exposed and closed by the user
- see issue #568 https://github.com/silverstripe/silverstripe-admin/issues/568


maxime-rainville:
@lukereative Can you have a quick look? @caffeineinc's CSS change look good, but it may have unattended side-effects that I wouldn't know off the top of my head.

FYI I've changed the target for the PR to the 1 branch.

PR #581, opened 2018-07-26
FIX #579 BUG Ambiguous column RecordID when doing batch actions

### LINKED ISSUE #579: BUG Ambiguous column "RecordID" when doing batch actions
SilverStripe 4.x (all versions)

When doing batch actions on sitetree objects, queries can sometimes fail due to ambiguous RecordID column.

The bug is in CMSBatchActionHandler::getPages() method.

```php
$pages = Versioned::get_including_deleted($recordClass, array(
                '"RecordID" IN ('.DB::placeholders($ids).')' => $ids
            ));
```

Fix by

```php
$pages = Versioned::get_including_deleted($recordClass)->byIDs($ids);
```

And let versioned do the ID rewriting. :)

edlinklater:


tractorcow:
Nice job thanks @edlinklater

PR #582, opened 2018-07-27
ENH Improve loading screen and indicator

### LINKED ISSUE #577: when loading the CMS, an M shows - what is the M for?
When you load the CMS, you first see an "M" ... Looks like the M of Magento. Why?

image

Here are some visuals:

image

shows:

image

Once the font-family is loaded:

image

You see this:

image

I kept seeing the "M" in a client presentation - and it looked weird... 

Maybe it is better to use "L" for Logo or "S" for Silverstripe or a non-visible / meaningfull character (e.g. an underscore or a dot)

lukereative:
- Use SVG for logo on main loading screen (which fixes #577) and add an animation if the page is loading for a while
- Use SVG and CSS animation for the Loading component and use the component more widely

**Related PR** https://github.com/silverstripe/silverstripe-asset-admin/pull/815


dhensby:
NPM test is not happy and cms test is failing (may not be related)

client/src/styles/legacy/_style.scss
```diff
@@ -708,7 +708,46 @@ html, body {
...
+    padding: .5rem;
+
+    background-color: $white;
+    border-radius: $border-radius-sm;
+
+    &__animation {
+      width: 100%;
+      height: 100%;
+
+      &__empty {
```
robbieaverill: Should these be `--` modifiers rather than `__` elements? Seems a bit odd to have `.spinner__animation__empty` but it might be me and my rudimentary understanding of BEM conventions

robbieaverill: Bump @lukereative, we could get this PR into the next release

lukereative: It's a bit ambiguous for this, because these are seperate elements that represent different parts of the logo – they're not 2 states of the same element, so I didn't make them modifiers `--`.

PR #594, opened 2018-08-08
NEW Search component for SiteTree, GridFields and ModelAdmin

lukereative:
Implements https://github.com/silverstripe/silverstripe-cms/issues/1777


client/src/legacy/GridField.js
```diff
@@ -495,6 +501,100 @@ $.entwine('ss', function($) {
...
+      );
+    },
+  });
+
+  /**
+   * Stop enter from submitting the whole 'page form' and
+   * allow the search component to handle it
+   */
+  $('.js-injector-boot .grid-field .search-box__content-field').entwine({
+    onkeydown: function(e) {
```
maxime-rainville: That can probably be fixed on the Search/SearchBox component directly. Otherwise, we'll have to fix it everywhere else.

client/src/components/Search/SearchBox.js
```diff
@@ -27,7 +27,7 @@ class SearchBox extends Component {
    * @param {Object} event
    */
   handleKeyUp(event) {
-    if (event.keyCode === 13) {
+    if (event.key === 'Enter') {
```
maxime-rainville: Somehow I find 13 to be more intuitive than "Enter". Does that make me weird? :wink:

lukereative: Yes, yes it does 😜 The main reason I changed it was keyCode is deprecated and it also made it easier to test as I could write a step with and `I press the "Enter" key`

maxime-rainville:
There's some border/padding issues here and there. I'm thinking, maybe we need to add a prop to `<Search/>` so we can enable/disable borders the component.

*Doubled-right border in SiteTree search*
image

*Hackward right hand padding on ModelAdmin search*
image

*Missing top border when Gridfield are stacked*
image

maxime-rainville:
The advanced filters get cut off on the _Comment_ admin in the CWP kitchen.
image

ModelAdmin warning clash with search
image

Search showing in _Archive_ even if it's been disable
image

code/ModelAdmin.php
```diff
@@ -190,10 +185,15 @@ public function getEditForm($id = null, $fields = null)
...
             $fieldConfig = GridFieldConfig_RecordEditor::create($this->config()->get('page_length'))
                 ->addComponent($exportButton)
-                ->removeComponentsByType(GridFieldFilterHeader::class)
                 ->addComponents(new GridFieldPrintButton('buttons-before-left'))
         );
 
+        if (!$this->showSearchForm ||
+            (is_array($this->showSearchForm) && !in_array($this->modelClass, $this->showSearchForm))
+        ) {
+            $listField->removeComponentByType(GridFieldFilterHeader::class);
```
maxime-rainville: This line's throwing an error for me when accessing Fluent's `LocaleAdmin`.

PR #597, opened 2018-08-10
Update all nodes under current node to update state

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/1161


chillu:
Can you please outline what the performance tradeoffs are here? Is it reloading the current page and it's children? It shouldn't reload ancestors, siblings and unrelated tree elements, right? My assumption is that's easy to implement with the current Entwine/PJAX capabilities, but maybe I'm remembering this wrong? :D

This morning you talked about making this non-blocking to the user, at least in terms of form editing. Updating tree state should be able to happen in the background, right?

Issue #429, opened 2018-02-11
Keyboard shortcuts for "insert link"


chillu:
## Overview

Virtually all modern web apps allow insertion of links via keyboard controls. Since it's part of the flow of writing web-focused content, it's an important measure for more advanced authors. It's also a common accessibility improvement, e.g. for people without the fine motor skills (or input devices) to click on a tiny icon in the TinyMCE toolbar.

Examples:
 * Wordpress
 * Gmail
 * Confluence
 * Any OSX rich text view (TextEdit, Mail.app)

## Acceptance Criteria

 * Open "Insert Link" TinyMCE dropdown through cmd+k on OSX, and ctrl+k on Win
 * Allow users to navigate through options ("Page on this site", "External URL") via keyboard arrows (already works)
 * Close "Insert Link" dialog through ESC key (and clear out input)
 * Focus keyboard in first input (should work without keyboard shortcut as well)
 * Insert the link by pressing ENTER key if focused on any input field in modal
 * Show keyboard shortcuts in dropdown of TinyMCE toolbar
 * Document shortcut in userhelp

## Notes

 * Excludes creating a keyboard shortcuts overview page in userhelp (although that'd be nice)

## PR
- [ ] silverstripe/silverstripe-admin#599
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/8328 (Fixes test)

chillu:
@clarkepaul @newleeland  In hindsight, I think it might've been a mistake to force the user to choose between "insert external URL" and "page on this site", it adds a bit too much UX friction for a common task. If you compare it to WP, they just have the external URL screen on the same modal as "page on this site", so you can trigger the same modal via a keyboard and then tab to the field you want. Likely too late to change this, just thinking out loud.

![image](https://user-images.githubusercontent.com/111025/36078812-6798ed06-0fe0-11e8-9083-c137d0380ba6.png)

clarkepaul:
I liked the way we had it with the local pages dropdown first and then under it a "or add external link" field or link to swap out the select. I just think internal links should be more of a priority for our typical audience (different from blogs).

PR #599, opened 2018-08-13
NEW Insert link shortcut for HTMLEditorField

lukereative:
Resolves https://github.com/silverstripe/silverstripe-admin/issues/429

Tests passing relies on https://github.com/silverstripe/silverstripe-framework/pull/8328


unclecheese:
Looks good, works great. Only issue I encountered is that when you `Cmd + K` the dropdown open, you can't close it with `ESC`. You can only close it with your mouse.

Do you think that would be easy to add, or is that getting too deep into the internals of TinyMCE?

lukereative:
I looked into it and can't find a way to make 'Esc' possible. As it was the opening of the menu doesn't have a good way apart from simulating a click on the button.

PR #602, opened 2018-08-14
FIX GridField delete button to offer archive action if possible

lukereative:
Fixes silverstripe/silverstripe-versioned#78

Tests won't pass until other PRs are merged, tested failing step locally and it passes

## Related PRs
- https://github.com/silverstripe/silverstripe-framework/pull/8325
- https://github.com/silverstripe/silverstripe-versioned/pull/171
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/42


client/src/styles/legacy/GridField.scss
```diff
@@ -98,6 +98,11 @@
       margin-right: 0;
       display: inline-block;
     }
+
+    .action-menu__toggle {
+      padding-top: 0.25rem;
+      padding-bottom: 0.25rem;
```
robbieaverill: I think the linter will require `.25rem`

lukereative: We don't actually enforce a rule either way and from a quick search we seem to prefer a leading zero

client/src/legacy/GridField.js
```diff
@@ -400,10 +400,14 @@ $.entwine('ss', function($) {
...
+  $('.grid-field .grid-field__col-compact .action--delete, .grid-field .grid-field__col-compact .action--archive, .cms-edit-form .btn-toolbar .action--delete, .cms-edit-form .btn-toolbar .action--archive').entwine({
     onclick: function(e){
-      if(!confirm(i18n._t('Admin.DELETECONFIRMMESSAGE'))) {
-        e.preventDefault();
+      const confirmMessage = $(this).hasClass('action--archive')
+        ? i18n._t('Admin.ARCHIVECONFIRMMESSAGE')
+        : i18n._t('Admin.DELETECONFIRMMESSAGE');
+
+      if (!confirm(confirmMessage)) {
+        e.preventDefault(confirmMessage);
```
unclecheese: I don't believe this method accepts any parameters. Is there a reason you're passing `confirmMessage`?

PR #604, opened 2018-08-17
FIX side scroll overflow of cms-preview

scott1702:
In SS4.2 there's an issue when in split/preview which allows the user to side scroll, this fixes that issue.

image


scott1702:
Updated @robbieaverill, seemed like the root cause was the chosen dropdowns when not active

PR #605, opened 2018-08-20
FIX Ambigious RecordID columns is now specified against owner's base table verisons

robbieaverill:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/2233


dhensby:
I think this was already fixed by https://github.com/silverstripe/silverstripe-admin/pull/581

robbieaverill:
Oh true, that’s a better fix anyway. The comment above it should be removed now though

Issue #574, opened 2018-07-19
SiteTree page icons update and accessibility fixes


hdpero:
Current SiteTree page icons 
![sitetree-page-icons](https://user-images.githubusercontent.com/13778690/42943368-d76bd12c-8b62-11e8-90ee-1133a75ed27f.png)
 are kind of ugly (IMHO) and doesn't match current UI of SilverStripe 4 Admin Interface, further more they are PNG, and not SVG sprite. 

Is there a new version of icons?

# Related PRs
* [ ] https://github.com/silverstripe/silverstripe-admin/pull/803
* [ ] https://github.com/silverstripe/silverstripe-cms/pull/2364
* [ ] https://github.com/silverstripe/silverstripe-errorpage/pull/29

hdpero:
@robbieaverill @clarkepaul  Based on I didn't find suitable icons in current [SilverStripe Font Icons Reference](https://github.com/silverstripe/silverstripe-admin/blob/master/client/src/font/icons-reference.html) I've created an SVG Sprite from [IcoMoon aplication](https://icomoon.io/app) and they are from [_IcoMoon - Free set_](https://icomoon.io/#icons-icomoon) with CC BY 4.0 licence.

Icons are as follows:
![sitetree-pagetype-icons](https://user-images.githubusercontent.com/13778690/43220361-27ae2612-904a-11e8-882d-0f47d9146c89.png)

- Page
- BlogPost (SilverStripe Blog module)
- Blog (SilverStripe Blog module)
- Profile (SilverShop module)
- Home _(HomePage class or URLSegment="home")_
- VirtualPage
- ElementalPage (usage with DNADesigns Elemental module - if Page Type exists)
- CartPage (SilverShop module)
- CheckoutPage (SilverShop module)
- CalendarPage (TitleDk SilverStripe Calendar module)
- ErrorPage
- RedirectorPage

Here is how they look like in fresh SS4 install:
![sitetree-add-page-form](https://user-images.githubusercontent.com/13778690/43220620-c80c151a-904a-11e8-89e9-c6b004d3928f.png)

DorsetDigital:
I'd agree the current icons aren't that pretty, but I'd also put a slightly different perspective (for what it's worth):

I like a clean site tree, without the noise of different icons.  The names of the pages ought to provide at least some hints as to what they are.  (I don't even like the icons you get with the blog module)

[...]

PR #609, opened 2018-08-21
Update fonts from Fontastic

sachajudd:
This PR adds the new CMS site tree icons and also updates any removed/added icons.

screen shot 2018-08-22 at 10 11 43 am

Reference: https://github.com/silverstripe/silverstripe-admin/issues/574


sachajudd:
@clarkepaul I've just noticed in the screenshot from the icons-reference.html that a couple of corners are missing in the new site tree icons. Fontastic is not showing this though so I'll have another look at it.

sachajudd:
Updated screenshot above @clarkepaul.

sachajudd:
Found the removed icon being used as a fall back for menu icons so I've added it back in 🙂

PR #610, opened 2018-08-22
ENH Add dropdownToggleClassNames prop and toggleCallback to ActionMenu

raissanorth:
Adding the additional prop `dropdownToggleClassName` allows the replacement of predefined class names without breaking semver. 

The current use case is the work on the ElementalEditor, in particular https://github.com/dnadesign/silverstripe-elemental/issues/322, which makes use of the `ActionMenu` component, but needs to render the button icon in a size other than large. 

Currently there is no way to override the predefined class name `btn--icon-xl` that is applied to the `DropdownToggle` which is rendered as part of the `ActionMenu`.

-----

This PR also adds an callback to the toggle function that passes through the event to allow, e.g. event propagation.


client/src/components/ActionMenu/ActionMenu.js
```diff
@@ -53,6 +53,13 @@ class ActionMenu extends PureComponent {
 
 ActionMenu.defaultProps = {
   className: '',
+  dropdownToggleClassName: classnames(
```
robbieaverill: I'd suggest we make this an array of strings rather than calling classnames twice

client/src/components/ActionMenu/ActionMenu.js
```diff
@@ -53,6 +53,13 @@ class ActionMenu extends PureComponent {
 
 ActionMenu.defaultProps = {
   className: '',
+  dropdownToggleClassName: [
```
robbieaverill: Last thing is we should declare the proptype for this e.g. string or array of strings

PR #619, opened 2018-08-28
FIX Console error on attribute being added to div

ScopeyNZ:
This is related to #610. Props get blindly unpacked onto a div so they must be valid HTML attributes (or be removed from the props that are unpacked)


client/src/components/ActionMenu/ActionMenu.js
```diff
@@ -39,6 +39,9 @@ class ActionMenu extends PureComponent {
     );
     const menuClassName = classnames('action-menu__dropdown', dropdownMenuProps.className);
 
+    // Prevent a "toggleCallback" event being applied to a div.
+    delete restProps.toggleCallback;
```
robbieaverill: Are you able to exclude this via destructuring rather than mutating it?

ScopeyNZ: I can - but it seemed odd to do that when you're not using it. In this case `restProps` is not actually `this.props` so mutating it isn't a concern I thought?

robbieaverill: Right so the cause of the problem is that we're blindly passing all props through to the Dropdown component. There is an [eslint rule](https://eslint.org/docs/rules/no-delete-var) preventing the use of `delete`, not sure if we use it in our ruleset though.

This component's proptypes aren't all that descriptive about what it's expecting here. What other props are actually in this object at this point? Can you make it explicit rather than inheriting?

robbieaverill:
Oh I forgot that I reviewed this 😆I've replaced it with #641 which doesn't mutate the props. Sorry for doubling up on this!

PR #620, opened 2018-08-28
BUGFIX: Fix menu toggle not rendering above cms content

bergice:
Note that this is replicated on smaller resolutions.


robbieaverill:
FYI this has caused a regression in https://github.com/silverstripe/silverstripe-admin/pull/620 / https://github.com/silverstripe/cwp/issues/78

robbieaverill:
@bergice I'm not sure what the original bug was here, but the CMS menu is working fine without this change for me with my window reduced to mobile size. Could you please re-test and see if you could report what the problem was and/or fix it in a different way? I'd like to revert this to fix the above linked issue with fullscreen TinyMCE views

robbieaverill:
Ok, shall we revert then @bergice?

bergice:
@robbieaverill yes, the code doesn't seem to be necessary anymore.

robbieaverill:
Cool, I'll revert by hand - thank you

PR #624, opened 2018-08-30
FIX Remove "more" action icon size, add btn-sm and fix icon alignment in gridfield

### LINKED ISSUE #622: Grid field "more" action icon should use the small btn/icon
We should use a smaller icon/button on the "More options" (three dots icon) action for Grid Field items than what is used next to the main page actions. The reason for this is that they look too large when presented in a long row, they should have more of a secondary appearance.

Requires style guidance so I/we can provide designs/classes which can be used.

sachajudd:
This commit also adds the class btn--icon-lg which will in future replace btn--icon-large across the CMS.

Before:
image

After: 
image

Fixes #622

cc @clarkepaul


client/src/legacy/GridField.js
```diff
@@ -132,11 +132,16 @@ $.entwine('ss', function($) {
...
+<<<<<<< HEAD
         content = '<button type="button" title="Open search and filter" name="showFilter" class="btn btn-secondary font-icon-search btn--no-text btn--icon-large grid-field__filter-open"></button>';
         this.removeClass('show-filter').find('.grid-field__filter-header, .grid-field__search-holder').addClass('grid-field__search-holder--hidden');
       }
       if (hasLegacyFilterHeader) {
         this.find('.sortable-header th:last').html(content);
+=======
+        content = '<button type="button" title="Open search and filter" name="showFilter" class="btn btn-secondary font-icon-search btn--no-text btn--icon-lg grid-field__filter-open"></button>';
+        this.removeClass('show-filter').find('.grid-field__filter-header').hide();
+>>>>>>> FIX Remove "more" action icon size, add btn-sm and fix icon alignment in gridfield
```
robbieaverill: Rebasing mistake?

robbieaverill:
I think it looks better without the -1px margin and top properties:

With negative margins:
image

Without negative margins:
image

It might just be the `top: -1px` on `::before` that needs to be removed, the margin-top looks OK actually

OSX + Chrome

sachajudd:
Yeah, agreed. Sometime @clarkepaul and I will take a look at the variables because quite a few things are out eg. line height.  In the meantime, I'll push up another PR to remove those. 🙂

PR #625, opened 2018-08-30
Fix over eager form onchange in Chrome

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-framework/issues/7327


lukereative:
@unclecheese I hate to say it but this has a legitimate test/functionality breakage

PR #628, opened 2018-09-04
NEW HtmlEditorField component for react rich text

NightJar:
This allows one to use a Rich Text field from within a react form, where
as previously it was only possible from server side PHP rendering
templates, and Entwine.

The downside to this is that although `@tinymce/tinymce-react` is the
official package for this, it is in turn only a very loose wrapper
around the window.TinyMCE global. It also posed problems with the
configuration generated by the PHP config class, and altering that would
destroy backwards compatiblity - which for now is critical for adoption.

So while there is now a React component for rendering TinyMCE, it is
only a bare bone component that still relies on Entwine to boot the
actual editor via the `data-config` attribute on the React output
`textarea` element.

In order to facilitate what is essentially inserting a form into a form (in terms of HTML semantics), it was necessary to alter `components/Form/Form.js` to render with a particular tag (as opposed to `<form>` always).

Requires https://github.com/silverstripe/silverstripe-framework/pull/8359

As part of work for https://github.com/dnadesign/silverstripe-elemental/issues/327


client/src/components/Form/Form.js
```diff
@@ -102,10 +104,12 @@ Form.propTypes = {
     value: PropTypes.any,
     type: PropTypes.string,
   })),
+  formTag: PropTypes.node,
```
robbieaverill: `node` is pretty broad. Is that deliberate? I guess it means you could inject a component right? If that is the intention then all good

client/src/components/HtmlEditorField/HtmlEditorField.js
```diff
@@ -0,0 +1,46 @@
...
+      window.TinyMCE = window.tinymce;
+    }
+    this.setState({ isReady: true });
+  }
+
+  renderDependencyScript() {
+    return <Script url={this.props.data.editorjs} onLoad={this.handleReady} />;
+  }
+
+  renderRichTextEditor() {
```
robbieaverill: These two methods could probably use a minimal doc block

robbieaverill:
Also needs a rebase =)

client/src/components/Form/Form.js
NightJar: Yeah, I'm not 100% if it's the correct thing, it just seemed the closest thing to me. The idea is to be able to specify an HTML tag, _or_ supply a custom component, yes :)

flamerohr: I would recommend using `formTag: PropTypes.oneOfType([PropTypes.func, PropTypes.string]),` nowadays...
using `PropTypes.node` normally denotes that you'll output it in the render straight away: `<div>{this.props.formStuff}</div>`, which is perfect for `children`

client/src/components/HtmlEditorField/HtmlEditorField.js
```diff
@@ -0,0 +1,46 @@
...
+    const config = JSON.parse(this.props.data.config['data-config']);
```
flamerohr: this `data.config['data-config']` is verbosely repetitive, what's the reason for the extra `'data-config'` layer, and could it be moved up?

client/src/components/HtmlEditorField/HtmlEditorField.js
```diff
@@ -0,0 +1,46 @@
...
+      ...super.getInputProps(),
+      ...this.props.data.config,
+    };
+  }
+
+  handleReady() {
+    if (!window.TinyMCE && window.tinymce) {
+      window.TinyMCE = window.tinymce;
+    }
+    this.setState({ isReady: true });
```
flamerohr: I recommend having `handleReady()` called in `componentWillMount()` or even set state ready in the constructor with a check

client/src/components/HtmlEditorField/HtmlEditorField.js
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  renderDependencyScript() {
+    return <Script url={this.props.data.editorjs} onLoad={this.handleReady} />;
```
flamerohr: this is interesting, because when I had a go at implementing this, I ran into issues of the script loading multiple times if I had multiple instances of this rendered and on a slow connection.
Have you tested this out? Do you know if load-script manages that?

client/src/components/HtmlEditorField/HtmlEditorField.js
```diff
@@ -0,0 +1,46 @@
...
+    this.state = {
+        isReady: false
+    };
+    this.handleReady = this.handleReady.bind(this);
+  }
+
+  getInputProps() {
+    return {
+      ...super.getInputProps(),
+      ...this.props.data.config,
```
flamerohr: what kind of config props are spread here?
this adds a bit of ambiguity of what is expected to happen, could `HtmlEditorField` contain these props instead of getting them from the `data.config` prop?

client/src/components/HtmlEditorField/HtmlEditorField.js
NightJar: This is contrary to the intention. `handleReady` is called only after the dependency script has been loaded into the browser (i.e. finished loading, not just rendered).

As Robbie suggested, I'll add a docblock to make this clearer.

client/src/components/HtmlEditorField/HtmlEditorField.js
NightJar: I will investigate.

client/src/components/HtmlEditorField/HtmlEditorField.js
NightJar: `data-config`, `data-editor` - both of which are required for TinyMCE to bootstrap.

client/src/components/Form/Form.js
NightJar: Neat thanks, I'll update :)

client/src/components/HtmlEditorField/HtmlEditorField.js
NightJar: This is the data as it's exported from `HTMLEditorConfig`(.php) - it consists of the config, and the editor reference (c.f. `window.ss.editorWrappers`), both of which are required to load the editor, and are applied directly to the element as data-attributes.

client/src/components/HtmlEditorField/HtmlEditorField.js
flamerohr: That sounds like an anti-pattern because the script should only load once, which is fine if you only have one instance of this component rendered.
It shouldn't need to call the script thereafter which is where the `componentWillMount()`/constructor should come in.

client/src/components/HtmlEditorField/HtmlEditorField.js
NightJar: As per https://github.com/silverstripe/silverstripe-framework/pull/8359#discussion_r215117940 I've renamed the prop to `data.attributes`

client/src/components/HtmlEditorField/HtmlEditorField.js
NightJar: as above, data.config is now data.attributes - along with added docblocks, I hope confusing is reduced to any new reader :)

client/src/components/HtmlEditorField/HtmlEditorField.js
NightJar: No, it is safe :)

I tested in browser and it seemed fine.
Investigation into why revealed:
https://github.com/blueberryapps/react-load-script/blob/master/src/index.jsx#L46

client/src/components/HtmlEditorField/HtmlEditorField.js
NightJar: I checked, the script is only loaded once. The local state ensures this is checked, but the operation I feel is safe.

NightJar:
D: I don't understand why these Behat tests are failing :/
Seems to me that nothing I've done here should affect e.g. https://travis-ci.org/silverstripe/silverstripe-admin/jobs/424642152#L1563

robbieaverill:
The latest 1 branch breaks Behat tests, it's not just this PR: https://travis-ci.org/silverstripe/silverstripe-admin/builds/424174909

client/src/legacy/HtmlEditorField.js
```diff
@@ -288,7 +288,7 @@ jQuery.entwine('ss', function($) {
     /**
      * Constructor: onmatch
      */
-    onadd: function() {
+    onmatch: function() {
```
robbieaverill: This change is throwing errors when I try and add new content blocks now (testing https://github.com/dnadesign/silverstripe-elemental/pull/372)

client/src/legacy/HtmlEditorField.js
```diff
@@ -298,14 +298,22 @@ jQuery.entwine('ss', function($) {
       this._super();
     },
 
+    onadd: function() {
+      this.onmatch();
```
robbieaverill: This change is throwing errors when I try and add new content blocks now (testing https://github.com/dnadesign/silverstripe-elemental/pull/372)

NightJar: Hmm, I only added it as a backwards compatibility thing because I wasn't sure about the differences between `onadd` and `onmatch`. My impression is that the former was deprecated for the latter, but there may be subtle differences. I'll dig in to investigate my suspicions... ideally we can just stick with the original change which was simply renaming `onadd` to `onmatch` (and no BC alias).

flamerohr: No, they're intended for different proposes, onadd is specific to creating a dom element into that selector while onmatch covers a few more scenarios including onadd's
Either or can be used for most scenarios, but just don't call one with the other (let entwine decide that)

NightJar: oh, thanks for the response @flamerohr - I didn't see it before writing the above^

robbieaverill:
Behat is green again in 1 - rebase to take advantage of this

robbieaverill:
Behat failures seem legitimate. 1 branch is passing, I've re-run this PR's builds a few times and they're still failing.

NightJar:
Hmm, must be some phantom (read: hard to recreate) JS error occurring somehow. I can't reproduce in my environment as it stands, I'll have to take a closer look in the morning.
I can't think of anything other than an execution halt due to an error that would cause e.g. "There is no Publish button" and "switching preview from live to draft has no affect".

robbieaverill:
Ok I found the bug - TinyMCE isn't reinitialising after publishing a page (window reloads with PJAX). This is probably because of the entwine match changes in this PR. Demo: https://screencast-o-matic.com/watch/cFQ6IIqEAl

NightJar:
YUSSSSS 💚

Issue #632, opened 2018-09-10
Make 'name' key for client config easier to (not) override


Cheddam:
We recently ran into a situation where a subclass of `CMSPageEditController` was required, and creating this had the side-effect of breaking the Internal Link modal in TinyMCE instances.

The cause was a combination of `LeftAndMain::getClientConfig()` hardcoding the section name as `static::class` (which resolved to `OurSubclassController`), and `TinyMCE_sslink-internal.js` hardcoding its `sectionConfigKey` as `CMSPageEditController`.

It _seems_ like it would be safer (albeit less clear) to set the section name in PHP config, and force developers to explicitly modify this if required. That way, minor subclassing situations like the above don't require the developer to override the `getClientConfig` method just to revert the `name` key to the parent class.

Happy to put together a PoC change for this, but keen to get some feedback first - I don't have a ton of experience with the new architecture and there may be a very good reason not to change this!

robbieaverill:
Sounds good. You could also change `static::class` to `self::class` since it's the base controller that really initiates the config, but subclasses can augment it with `getSchemaData` etc

Cheddam:
Excellent point - PR inbound 👍

robbieaverill:
Fixed with #638, thanks @Cheddam

PR #638, opened 2018-09-12
Make section name in Client Config configurable.

Cheddam:
Implements #632. I went with an approach that avoids API changes, whilst being as flexible as possible.

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -264,6 +264,18 @@ class LeftAndMain extends Controller implements PermissionProvider
     private static $frame_options = 'SAMEORIGIN';
 
     /**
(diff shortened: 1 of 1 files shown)


code/LeftAndMain.php
```diff
@@ -316,10 +328,15 @@ public function getCombinedClientConfig()
      */
     public function getClientConfig()
     {
+        // Allows the section name to be overridden in descendant classes
+        if (!$name = $this->config()->get('section_name')) {
```
robbieaverill: It's pretty minor, but I think I'd prefer to see the assignment on its own line

Cheddam: Changed 👍

PR #641, opened 2018-09-13
FIX ActionsMenu does not pass toggleCallback prop to Dropdown

robbieaverill:
This fixes a bug from https://github.com/silverstripe/silverstripe-admin/pull/610 where the `toggleCallback` prop is passed through to Dropdown, and React gives a warning because that component renders a div.

`toggleCallback` is a callback that runs before the logic in ActionMenu's `toggle` method, but it's not designed to be passed to Dropdown.


unclecheese:
Have retargeted to `1.3`

robbieaverill:
Rebased

client/src/components/ActionMenu/ActionMenu.js
```diff
@@ -31,6 +31,8 @@ class ActionMenu extends PureComponent {
       dropdownMenuProps,
       dropdownToggleClassNames,
       className,
+      // remove props that we don't want to pass through to Dropdown
+      toggleCallback, // eslint-disable-line no-unused-vars
```
unclecheese: Why not just `delete restProps.toggleCallback`? It's already a clone, so there's no risk of breaking references.

PR #642, opened 2018-09-13
currentPageID() does not work in ModelAdmin

lpostiglione:
Added functionality to retain current page id while in ModelAdmin

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -1654,7 +1654,7 @@ class LeftAndMain extends Controller implements PermissionProvider
             return $this->urlParams['ID'];
         }
         
-        if (is_numeric($this->getRequest()->param('ID'))){
+        if (is_numeric($this->getRequest()->param('ID'))) {
             return $this->getRequest()->param('ID');
         }
         
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -1653,7 +1653,11 @@ class LeftAndMain extends Controller implements PermissionProvider
         if (isset($this->urlParams['ID']) && is_numeric($this->urlParams['ID'])) {
             return $this->urlParams['ID'];
         }
-
+        
+        if (isset($this->getRequest()->param('ID')) && is_numeric($this->getRequest()->param('ID')) {
+            return $this->getRequest()->param('ID');
+        }
+        
         /** @deprecated */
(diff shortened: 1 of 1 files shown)


code/LeftAndMain.php
```diff
@@ -1660,7 +1660,11 @@ public function currentPageID()
         if (isset($this->urlParams['ID']) && is_numeric($this->urlParams['ID'])) {
             return $this->urlParams['ID'];
         }
-
+        
+        if (isset($this->getRequest()->param('ID')) && is_numeric($this->getRequest()->param('ID')) {
```
robbieaverill: You could remove the isset check here, because it'll return null if the key doesn't exist. There's also a syntax error in the if statement

robbieaverill:
@lpostiglione would you like to update the PR to fix the syntax error and my other request above?

We could also target this at a SilverStripe 4 release branch as a patch change if you'd like to (I can do this for you). It's currently targeting SilverStripe 5.x

lpostiglione:
Fixed syntax

code/LeftAndMain.php
dhensby: How about we don't do an `is_numeric` check?

maxime-rainville:
@robbieaverill Looks like your feedback has been addressed. Do you have outstanding grievances?

robbieaverill:
Yep, @dhensby's suggestion could be treated as outstanding still. We also should decide whether to target a 4.x release or just merge this into master (5.x)

maxime-rainville:
What's the problem with the `is_numeric` check? Is it that we're considering allowing non-numeric IDs in SS5?

robbieaverill:
I suppose so. If we do move to non numeric IDs I’m sure unit tests will catch this. Merge!

code/LeftAndMain.php
```diff
@@ -1660,7 +1660,11 @@ public function currentPageID()
         if (isset($this->urlParams['ID']) && is_numeric($this->urlParams['ID'])) {
             return $this->urlParams['ID'];
         }
-
+        
+        if (is_numeric($this->getRequest()->param('ID'))){
```
bergice, line 1664:
```suggestion
        if (is_numeric($this->getRequest()->param('ID'))) {
```

PR #647, opened 2018-09-17
NEW Add hideNav flag to allow hiding of navigation in Tabs

raissanorth:
The changes in this PR render the navigation of the Tabs component dependent on the value of a newly introduced prop `data.hideNav`. This is required for the use case of the React driven inline editing for Elemental blocks for instance, see https://github.com/dnadesign/silverstripe-elemental/issues/324


client/src/components/Tabs/tests/Tabs-test.js
```diff
@@ -0,0 +1,40 @@
...
+
+import React from 'react';
+import Tabs from '../Tabs';
+import Enzyme, { shallow } from 'enzyme';
+import Adapter from 'enzyme-adapter-react-15.4/build/index';
+
+Enzyme.configure({ adapter: new Adapter() });
+
+describe('Tab', () => {
+  describe('Test rendering of navigation', () => {
```
robbieaverill: Let's describe the `render()` method

client/src/components/Tabs/Tabs.js
```diff
@@ -137,11 +140,17 @@ Tabs.propTypes = {
   id: React.PropTypes.string.isRequired,
   defaultActiveKey: React.PropTypes.string,
   extraClass: React.PropTypes.string,
+  data: React.PropTypes.shape({
+    hideNav: React.PropTypes.bool,
```
robbieaverill: I think making this its own boolean prop would be good. As discussed in Slack we should be able to add `hideNav` to the schema data defaults then use `setSchemaState(['hideNav' => true])`, since "Any passed keys that are not defined in {\@link getSchemaDataDefaults()} are ignored."

PR #651, opened 2018-09-18
WIP Manage Tabs' activeTab with Redux store

raissanorth:


client/src/components/Tabs/Tabs.js
```diff
@@ -141,12 +149,36 @@ Tabs.propTypes = {
...
 
-export default Tabs;
+function mapStateToProps(state) {
+  const {
+    activeTab,
+  } = state.tabs;
+
+  return {
+    activeTab,
+  };
```
robbieaverill: Out of interest, is this going to mean that multiple TabSets on a page will all have their active tabs set when you change the (relatively global) `stats.tabs.activeTab` value?

Perhaps it might be better to be stored in the Redux form state, which is already isolated for each form?

raissanorth: Yes, that is indeed gonna happen. Thanks for pointing this out, I hadn't considered it.

raissanorth: Hi @robbieaverill, do you have any pointers for me on how you see this work out? The thing that isn't quite clear to me is how elemental's Redux form state is gonna communicate to admin's `Tabs` component that it should update the activeTab. Do you think it should continue to handle state internally by default or do you improve of hooking it up to a Redux store? I wonder how this work out given that we won't always use `Tabs` in the context of the ElementalEditor. Does think make sense? If so, I hope you share your 'vision' with me.

NightJar: We've just had a look at this, and to my surprise the non-data form field is present (`TabSet` is essentially a `<div>`). So that's neat.
The question comes now in the form of:
- How do we update that externally (i.e. from an `AbstractAction`)
- How do we let ReduxForm know what to do with a state change like that, since this isn't a straight forward form field?

If you have an idea, or an idea on where to look, please advise.
It is the end of our day and we can of course look into this in the morning, however if you're aware of what to do a few pointers of where to look in docs would be awesome, and we can proceed from there (in a learn as we go kind of thing).

**tl;dr** Have you go some pointers (to docs) on how to achieve updating ReduxForm, and having ReduxForm update an arbitrary element (within it's control)?

robbieaverill: I know very little about the redux-form implementation, sorry. I do think we can assume that since TabSet is a form field and exists within the same form as other redux-form enabled form fields, what we probably need to do is connect the value of the field in the state to the visible tab in the set.

If you look at text field for example, redux-form tells it to bind the redux state entry for the field to the text field's value so it updates automatically. We essentially need to do something similar for the active tab in the set.

I'm sorry I can't help much more than that at the moment without trying it out =(

PR #661, opened 2018-09-25
NEW Add update extension hooks for LinkFormFactory subclasses

robbieaverill:
Issue: https://github.com/silverstripe/silverstripe-cms/issues/2273

### DIFF
--- a/code/Forms/EditorEmailLinkFormFactory.php
+++ b/code/Forms/EditorEmailLinkFormFactory.php
@@ -29,6 +29,8 @@ class EditorEmailLinkFormFactory extends LinkFormFactory
             $fields->insertAfter('Link', TextField::create('Text', _t(__CLASS__.'.LINKTEXT', 'Link text')));
         }
 
+        $this->extend('updateFormFields', $fields, $controller, $name, $context);
+
         return $fields;
     }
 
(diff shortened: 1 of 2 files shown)


code/Forms/EditorEmailLinkFormFactory.php
```diff
@@ -29,6 +29,8 @@ protected function getFormFields($controller, $name, $context)
             $fields->insertAfter('Link', TextField::create('Text', _t(__CLASS__.'.LINKTEXT', 'Link text')));
         }
 
+        $this->extend('updateFormFields', $actions, $controller, $name, $context);
```
ScopeyNZ: Was this meant to be `$fields` instead of `$actions`?

michelsteege:
Is there any news of a release date for this?

robbieaverill:
Hi @michelsteege, we tend to release new minor versions every 3 months or so. The last was 4.3 (which this missed the cutoff for) so the next might be in late Feb/early March, although this message shouldn't be taken as a guarantee of that happening at that time

michelsteege:
Thanks @robbieaverill  I wait patiently :-)

robbieaverill:
@michelsteege if you really need to modify these classes, you can always extend them with Injector and modify the bits of logic affected in this patch (in the meantime)

PR #664, opened 2018-09-27
FIX Avoid having search fields with the same name as form elements

ScopeyNZ:
Related to silverstripe/silverstripe-framework#8414

This resolves an issue where search form field names could conflict other field names (mostly in CMS). See the related PR for more context/discussion.

Issue: https://github.com/silverstripe/silverstripe-cms/issues/2251

CMS Behat tests are fixed in https://github.com/silverstripe/silverstripe-cms/pull/2279


ScopeyNZ:
I'll fix Behat tomorrow.

robbieaverill:
I've pushed a fix for behat and opened a PR for the CMS behat tests at https://github.com/silverstripe/silverstripe-cms/pull/2279

robbieaverill:
Thanks @maxime-rainville - I've validated that as a regression in this PR, will look into it

robbieaverill:
Yeah OK so the form state is still being saved without the Search__ prefix on the form field names. We had the same problem with elemental recently, will revisit what we did there

robbieaverill:
Oh there's a bigger issue here - the search renders a form, which is nested inside the default EditForm. You can't nest forms in HTML

robbieaverill:
I've had a bit of a look at this and have had some success by removing the javascript logic and handling it solely in PHP, but it was still playing up in some situations. I think @maxime-rainville would be best to pick this up if you have time

ScopeyNZ:
Ok I've added a commit which has addressed a bunch of issues.

I also removed the optimistic loading of search forms. Some page types were having ~8 search forms load immediately which was pretty overkill.

There's a full list of things I changed on the latest commit message.

ScopeyNZ:
Behat's broke because the PR relies on the framework PR. At least they pass locally given I have that framework PR checked out.

ScopeyNZ:
You're pretty thorough @maxime-rainville! Thanks :)

I've addressed that issue now.

robbieaverill:
Nevermind, that's a pre-existing issue. I'll log it separately.

robbieaverill:
The behat tests pass locally in combination with this and https://github.com/silverstripe/silverstripe-cms/pull/2279, but neither will pass without the other so I'll merge them both and monitor for green after merge

Issue #662, opened 2018-09-26
SearchForm has been removed from ModelAdmin


ScopeyNZ:
There's a commit in 4.3 (1.3) that has removed `SearchForm` from model admin:

https://github.com/silverstripe/silverstripe-admin/commit/7aa0fad69447f8820324dd7671ddbedfb8269b16#diff-fdb2c6ab0eacaf1803e1d9c64a3cb318L278

Although I understand the sentiment with the new search implementation I think this is surely too much of a BC break to be released as a minor?

## PRs
- [x] https://github.com/silverstripe/silverstripe-admin/pull/665
- [x] https://github.com/silverstripe/silverstripe-framework/pull/8420
- [ ] silverstripe/silverstripe-framework#8441

chillu:
Cringe. @lukereative @maxime-rainville  Can you please get across this? Customising `MyModelAdmin->SearchForm()` is extremely common. My understanding from scoping this work was that we're using form schema to represent existing search forms as JSON, not that we're completely removing the method. Maybe it's been moved to an extension in another commit? Would that still allow this type of cusotmisation?



Also, why isn't the search documented in the 4.3 release notes? https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/04_Changelogs/4.3.0.md

chillu:
Note that customising most public methods in ModelAdmin is common. Please go through the API surface of this class and see where expectations would be broken in a minor release, and how we can fix this. We also need to make `showSearchForm` work, for example.

sminnee:
Note that the customisable search forms are included in the down-arrow, as shown below. But they don't appear to work.

![image](https://user-images.githubusercontent.com/59968/46442469-3b786c80-c7c6-11e8-8613-b0c3464f9ac6.png)

maxime-rainville:
That looks like a different issue. I'm getting weird pagination results as well. I'll split it up.

maxime-rainville:
Raised as a separate issue in #672

sminnee:
Has this issue been resolved, or did you close it accidentally?

maxime-rainville:
@sminnee The original problem this issue was tracking has been resolved. I think the behavior you are seeing is a different problem, so I raised a distinct issue to track it #672.

PR #665, opened 2018-09-28
Add ModelAdmin search API back to maintain semver

lukereative:
Resolves silverstripe/silverstripe-admin#662

**Dependent PR** silverstripe/silverstripe-framework/pull/8420


code/ModelAdmin.php
```diff
@@ -238,6 +250,77 @@ public function getExportFields()
         return singleton($this->modelClass)->summaryFields();
     }
 
+    /**
+     * @deprecated 5.0
+     * @return \SilverStripe\ORM\Search\SearchContext
+     */
+    public function getSearchContext($context = null)
```
ScopeyNZ: I don't think you need `$context` as a param? It wasn't in the API that was removed. If we do need it, we should add a `@param`.

code/ModelAdmin.php
```diff
@@ -238,6 +250,77 @@ public function getExportFields()
...
+        return $context->getSummary();
+    }
+
+    /**
+     * Returns the search form
+     *
+     * @deprecated 5.0
+     * @return Form|bool
+     */
+    public function SearchForm($form = null)
```
ScopeyNZ: Same comment here as for `getSearchContext`.

code/ModelAdmin.php
```diff
@@ -188,6 +189,17 @@ public function getEditForm($id = null, $fields = null)
                 ->addComponents(new GridFieldPrintButton('buttons-before-left'))
         );
 
+        $listField->getConfig()->removeComponentsByType(GridFieldFilterHeader::class);
```
ScopeyNZ: I think it would be prudent to comment why we're doing this and mention it's for BC until 5.0.

code/ModelAdmin.php
lukereative: I think it was a leftover from a previous solution. Removed.

code/ModelAdmin.php
lukereative: I think it was a leftover from a previous solution. Removed.

code/ModelAdmin.php
lukereative: Yep, done.

code/ModelAdmin.php
```diff
@@ -238,6 +252,69 @@ public function getExportFields()
...
 
+    /**
+     * @deprecated 5.0
+     * @return \SilverStripe\ORM\Search\SearchContext
+     */
+    public function getSearchContext()
+    {
+        Deprecation::notice('5.0', 'Will be removed in favor of GridFieldFilterHeader in 5.0');
+
+        $gridField = $this->getEditForm()->fields()
```
maxime-rainville: Not sure about all the `getEditForm` calls. Sounds kind of heavy just so we can get a single field.

code/ModelAdmin.php
```diff
@@ -238,6 +252,69 @@ public function getExportFields()
...
+    {
+        Deprecation::notice('5.0', 'Will be removed in favor of GridFieldFilterHeader  in 5.0');
+
+        if (!$this->showSearchForm
+            || (is_array($this->showSearchForm) && !in_array($this->modelClass, $this->showSearchForm))
+        ) {
+            return false;
+        }
+
+        $gridField = $this->getEditForm()->fields()
```
maxime-rainville: Not sure about all the `getEditForm` calls. Sounds kind of heavy just so we can get a single field.

maxime-rainville:
Is the expectation that using the legacy search, we can only use the regular gridfield search in Model Admin?

# 4.2 ModelAdmin Search
image

# 4.3 ModelAdmin Legacy Search
image

code/ModelAdmin.php
maxime-rainville: Good enough.

code/ModelAdmin.php
```diff
@@ -238,6 +252,69 @@ public function getExportFields()
...
+        return $context->getSummary();
+    }
+
+    /**
+     * Returns the search form
+     *
+     * @deprecated 5.0
+     * @return Form|bool
+     */
+    public function SearchForm()
```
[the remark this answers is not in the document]
maxime-rainville: Following our discussion this morning ... there's no point whitelisting this in this context because the UI that would be using this has been removed.

maxime-rainville:
Just summarizing our discussion in the stand up this morning.

Our semantic versioning commitment does not extend to the UI. The API for ModelAdmin will be functionaly equivalent, even if some of it is no longer needed or actually use by our UI. A note to this affect should be added to the 4.3 change logs.

lukereative:
`getList` changes because all the search filtering that it used to do is now handled by the grid field and the filter header as we pass it into the gridfield in `getEditForm`. 
Additionally we don't use search params on the model admin any more so we can get them in this context because the gridfield gets search params via ajax (and is wiped if we reload the model admin).
We maintained the `updateList` extension point though so we can filter it still before we pass it to the gridfield.

robbieaverill:
I've raised a PR at #841 to update the deprecation notices - they need to be _from_ the version the notice should be raised at. I've changed them to 4.3.x instead of 5.0, since that's where you'll be removing these APIs rathe than when the deprecations should be raised.

Issue #636, opened 2018-09-11
Page search field value encoding issue


mfendeksilverstripe:
Version: `"silverstripe/recipe-cms": "^4.2.0" (115bbbb5f98f98ea1f18f0d3ddb260a9328359be)`

When I pass a search value via a text field that contains `+` sign, it gets replaced by space.

Input value:

`L + J`

value passed to get param:

`L+++J`

expected value to be passed to get param:

`L+%2B+J`

value shown in the form after search:

`L<space><space><space>J` where `<space>` represents one space character

extected value to be present in the form after search:

`L + J`

We found this changeset which is very suspicious.

https://github.com/silverstripe/silverstripe-admin/commit/050f63c4cb11caada9cca617f35718fba1cb67fc

It's intentionally replacing the correct encoded value of `+` with an unencoded version of the `+`. See code here:

`paramStr = $.param( newParams ).replace(/\%2B/g, '+')`

It's also worth noting that there are two different places where encoding is handled which is again very questionable.

https://github.com/silverstripe/silverstripe-admin/blob/ad18c29866b21fc244822e317175f6c0bfdc9cb9/client/src/legacy/LeftAndMain.js#L1257

https://github.com/silverstripe/silverstripe-admin/blob/master/client/src/legacy/sspath.js#L168

The encoding should be applied only once and properly, not patched on two different places.

# Pull Request
* [ ] https://github.com/silverstripe/silverstripe-admin/pull/667

mfendeksilverstripe:
Added related issue:

https://github.com/silverstripe/silverstripe-admin/issues/410

robbieaverill:
I've read through #410 which introduced the issue originally. [The justification](https://github.com/silverstripe/silverstripe-admin/issues/410#issuecomment-365437664) for replacing encoded + symbols with regular + symbols is that sspath.js is encoding + symbols too early, so this fix was designed to undo that behaviour (instead of patching the jQuery library that provides `$.params`).

The problem is that we have also taken out the ability to use legitimate + symbols in the search string by doing this.

I think perhaps we need to undo the change from #410 and replace the code with something else that will allow + symbols in the URL and encode them at the correct point.

robbieaverill:
This is also reproducible with the new search interface in SS 4.3, which doesn't use the legacy entwine code (AFAIK)

robbieaverill:
PR at #667 for consideration

PR #667, opened 2018-10-01
FIX Searching now allows + symbols, use own method over jQuery serialisation

robbieaverill:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/636

Changes:

* Remove unused legacy entwine code for handling CMS search form submissions - I've added breakpoints and couldn't find it being used any more with the new search interface
* Replace use of jQuery .param (and space related workarounds for a bug in it) with a method to URI encode the search params


dhensby:
@robbieaverill if you've tested it and it works... then I'm happy for it to go in; it's tricky to know if it'll cause regressions just from looking!

PR #678, opened 2018-10-08
MINOR Upgrade Bootstrap to 4.1.2

robbieaverill:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/8392


robbieaverill:
Behat failures look real, investigating

robbieaverill:
Ok the search form wasn't opening. This was caused by a change in Bootstrap: https://github.com/twbs/bootstrap/pull/25806. This means that `.collapse.show` no longer indicates `display: block` in Bootstrap 4.1.x like it did in 4.0.x.

The version of Reactstrap (5.0.x) that we're using specifies these classes: https://github.com/reactstrap/reactstrap/blob/5.0.0/src/Collapse.js#L32, so the only real option for B/C support in SilverStripe 4.x is to provide a polyfill, which I'm going to push up now.

robbieaverill:
Sweet, good result all round

PR #679, opened 2018-10-09
Implement logic to lazy load GridField in the frontend

maxime-rainville:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/2276

Depends on https://github.com/silverstripe/silverstripe-framework/pull/8460

This PR updates the GridField legacy JS logic to allow GridField data to be loaded once it's parent tab is activated. It also has the side benefit of updating the loading animation of GridField to be contained to the GridField table.

# Known compatibity issue with  `colymba/gridfield-bulk-editing-tools`
There's a bug when the lazy loading is combined with `colymba/gridfield-bulk-editing-tools`. If you start on a page that doesn't have a GridField and access a page with GridField that has the lazy loading and bulk editing component enable, the results are not properly fetch. (The Comments in the CWP kitchen are a good example of this). I think it's a bug with with `gridfield-bulk-editing-tools`, so we can raise a separate issue. If it's this is a blocker we can add an exception to disable the lazy loading if we detect `gridfield-bulk-editing-tools` is present.


client/src/legacy/GridField.js
```diff
@@ -41,14 +67,21 @@ $.entwine('ss', function($) {
...
+      tbody.append($('<tr />').append(loadingCell));
 
-      $.ajax($.extend({}, {
+      var request = $.ajax($.extend({}, {
         headers: {"X-Pjax" : 'CurrentField'},
         type: "POST",
         url: this.data('url'),
         dataType: 'html',
-        success: function(data) {
+        async:true,
```
robbieaverill: Space after `:`

client/src/legacy/GridField.js
```diff
@@ -41,14 +67,21 @@ $.entwine('ss', function($) {
...
+        success: function (data) {
+          console.dir(data)
```
robbieaverill: Remove debug code

client/src/legacy/TabSet.js
```diff
@@ -77,7 +84,23 @@ $.entwine('ss', function($){
 				if(!matches) return;
 				$(this).attr('href', document.location.href.replace(/#.*/, '') + matches[0]);
 			});
-		}
+		},
+
+    /**
+     * @func lazyLoadGridFields
+     * @desc Find all the lazy loadable gridfield in the panel and trigger their reload.
+     * @param panel
```
robbieaverill: missing type

client/src/legacy/GridField.js
maxime-rainville: `async:true` should not have been there ... that was a remnant of me knocking my head against the wall and trying stuff.

client/src/legacy/GridField.js
maxime-rainville: Gone.

client/src/legacy/TabSet.js
maxime-rainville: Fixed.

maxime-rainville:
All ready to go.

client/src/legacy/GridField.js
```diff
@@ -19,16 +19,42 @@ $.entwine('ss', function($) {
         this.fixColumns();
         this.injectSearchButton(false);
       }
+
+      if (this.is('.grid-field-lazy--loadable') && (
```
robbieaverill: The class names in the framework PR are

```
            'grid-field--lazy-loadable' :
            'grid-field--lazy-loaded');
```

robbieaverill:
@maxime-rainville re "Known compatibity issue with colymba/gridfield-bulk-editing-tools" can you please raise that issue?

PR #680, opened 2018-10-09
BUG ModelAdmin pagination with a filter

lukereative:
Fix silverstripe/silverstripe-framework#8454

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -20,6 +20,7 @@ use SilverStripe\Forms\GridField\GridFieldDetailForm;
 use SilverStripe\Forms\GridField\GridFieldExportButton;
 use SilverStripe\Forms\GridField\GridFieldFilterHeader;
 use SilverStripe\Forms\GridField\GridFieldImportButton;
+use SilverStripe\Forms\GridField\GridFieldPaginator;
 use SilverStripe\Forms\GridField\GridFieldPrintButton;
 use SilverStripe\Forms\HiddenField;
 use SilverStripe\Forms\LiteralField;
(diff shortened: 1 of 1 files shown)


code/ModelAdmin.php
```diff
@@ -202,6 +203,9 @@ function ($form) {
                 $this->extend('updateSearchForm', $form);
             }
         ));
+        // GridFieldPaginator has to be added after filter header for it to function correctly
+        $listField->getConfig()->removeComponentsByType(GridFieldPaginator::class);
+        $listField->getConfig()->addComponent(new GridFieldPaginator());
```
robbieaverill: I guess we need some APIs to let people control GridField component priority, eh?

robbieaverill: Can you please use Injector to resolve this component?

lukereative: Done

lukereative: Yeah…

code/ModelAdmin.php
```diff
@@ -202,6 +204,9 @@ function ($form) {
                 $this->extend('updateSearchForm', $form);
             }
         ));
+        // GridFieldPaginator has to be added after filter header for it to function correctly
+        $listField->getConfig()->removeComponentsByType(GridFieldPaginator::class);
+        $listField->getConfig()->addComponent(Injector::inst()->get(GridFieldPaginator::class));
```
tractorcow: This ignores the `page_length` config for modeladmin. You should be using `create` for items that have member properties, and `get` for global services / singletons.

PR #681, opened 2018-10-09
FIX Only hide overflow from inactive chosen fields

### LINKED ISSUE #660: Selecting a value in DropdownField makes the field jump upward
**Version: 4.3.x-dev**

With tractorcow/silverstripe-fluent 4.2.x-dev

* Go to locales
* Add a new locale
* Choose a value from the Locale dropdown
* Notice the field jumps upwards

image

image

robbieaverill:
https://github.com/silverstripe/silverstripe-admin/pull/604 introduced a change which hides overflow from all chosen inputs other than with a certain class. This causes https://github.com/silverstripe/silverstripe-admin/issues/660.

This patch adjusts the style to only hide overflow from inactive chosen select fields.

Resolves #660

cc @scott1702


robbieaverill:
Travis looks stuck =/

PR #684, opened 2018-10-11
NEW Adding a component for a generic popover filled with buttons

ScopeyNZ:
Adding a component that will contain a list of buttons with a search field


ScopeyNZ:
This is technically ready for merge now. There's a couple of follow-ups I want to do, and might add them to this PR if it's not merged before I get around to it:

- [ ] Add to story book
- [ ] Enhancement to add support for using arrow keys

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,213 @@
...
+PopoverOptionSet.propTypes = {
+  buttons: PropTypes.arrayOf(buttonType),
+  // Accepts a function that takes a search term as a first parameter and a set
+  // of buttons to match against that returns a filtered set of buttons
+  // Default search handler assumes button content to be plain text and performs
+  // a simple string.contains check.
+  onSearch: PropTypes.func,
+  // function that accepts a button object & returns an event handler
+  // e.g. (button) => (event) => event.preventDefault() && togglePopover();
+  onButtonClick: PropTypes.func.isRequired,
```
ScopeyNZ: I'm all ears for a better name for this - or perhaps a more accepted way of doing something like this?

brynwhyman:
@chillu we're looking to get this into the 4.3 release so the component can be used with elemental and the Wātea+ colour picker. Let me know if you see an issue with this.
cc @ScopeyNZ @robbieaverill

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,238 @@
...
+    this.handleSearchValueChange = this.handleSearchValueChange.bind(this);
+    this.handleKeyDown = this.handleKeyDown.bind(this);
+
+    this.state = {
+      searchValue: ''
+    };
+  }
+
+  /**
+   * Pass toggle to parent and clear the search input
```
robbieaverill: It's not immediately clear where it's passing the callback to the parent

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,238 @@
...
+    const { searchValue } = this.state;
+
+    if (searchValue.length === 0) {
+      return null;
+    }
+
+    return (
+      <InputGroupAddon addonType="append">
+        <button
+          className="popover-option-set__search-clear btn-link"
```
robbieaverill: I think we should make any class names in these components extensible via props

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,238 @@
...
+          onClick={this.handleSearchValueClear}
+        >
+          {i18n._t('PopoverOptionSet.CLEAR', 'Clear')}
+        </button>
+      </InputGroupAddon>
+    );
+  }
+
+  /**
+   * Render the search value intput box (which is in turn used to filter the buttons)
```
robbieaverill: Typo

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  /**
+   * Render either all buttons available, buttons matching the search term, or a message that
+   * there are no matching buttons
+   * @returns {DOMElement}
+   */
+  renderOptionButtons() {
+    const { onSearch, onButtonClick, ButtonComponent } = this.props;
+    let { buttons } = this.props;
```
robbieaverill: We shouldn't be mutating props. Perhaps you could do something like this instead:

```js
const { buttons } = this.props;

let buttonsToRender = buttons;
if (...) {
  buttonsToRender = 'something else';
```

client/src/components/PopoverOptionSet/PopoverOptionSet.js
robbieaverill: Right - if it's not an event handler itself then we shouldn't call it `onXYZClick`. Maybe something like `provideButtonClickHandler` or something?

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,238 @@
...
+PopoverOptionSet.defaultProps = {
+  searchPlaceholder: i18n._t('PopoverOptionSet.SEARCH_PLACEHOLDER', 'Search'),
+  onSearch: (query, buttons) => buttons.filter(
+    ({ content }) => content.toLowerCase().includes(query.toLowerCase())
+  ),
+  disableSearch: false,
+  ButtonComponent: Button,
+};
+
+export { buttonType };
```
robbieaverill: We should have a separate place for types - if we don't already it might be a good chance to start

client/src/components/PopoverOptionSet/PopoverOptionSet.scss
```diff
@@ -0,0 +1,94 @@
...
+      }
+    }
+
+    &-input {
+      padding-left: $panel-padding-x;
+    }
+
+    // Nesting needed for specificity, adds padding to clear link
+    .input-group-append {
+
```
robbieaverill: I definitely feel like this should be a lint error =/ I've seen issues where our SCSS linting it just not working at all before, could you make test to make sure this is valid?

client/src/components/PopoverOptionSet/PopoverOptionSetToggle.js
```diff
@@ -0,0 +1,66 @@
...
+        <Button {...buttonProps} >
+          Toggle
+        </Button>
+        <PopoverOptionSet {...popoverProps} />
+      </div>
+    );
+  }
+}
+
+PopoverOptionSetToggle.proptypes = {
```
robbieaverill: propTypes*

client/src/components/PopoverOptionSet/PopoverOptionSet.js
ScopeyNZ: Sure. `buttons` is an array and will not be taken by reference so it won't mutate. I can add the extra line but it would only be for clarity I guess.

robbieaverill: I think it'd be worth it, otherwise someone might see that buttonsToRender is mutable and do it

client/src/components/PopoverOptionSet/PopoverOptionSet.js
ScopeyNZ: Considering that we have lots of components in this module, and types are _usually_ only used by the one component I feel like it makes more sense to keep them here. I can make a type file in the folder for this component if that's nicer?

I really don't think it makes sense to create a new types folder for all components.

robbieaverill: Ok fair enough. Do you see the number of types for this component expanding? Leaving it where it is will introduce an expectation that it stays where it is (in terms of `import { typeName } from 'path/to/file';`)

ScopeyNZ: Ok in the end I just removed the export. Thinking about it, it didn't really make sense as it's not a type that's common across more than just this component. I've in-lined the shape into the proptype definitions now.

robbieaverill:
I've loaded up the story book examples and get this stack trace:

```
provideButtonClickHandler is not a function
TypeError: provideButtonClickHandler is not a function
    at http://localhost:6006/static/preview.bundle.js:8449:24
    at Array.map (<anonymous>)
    at PopoverOptionSet.renderOptionButtons (http://localhost:6006/static/preview.bundle.js:8443:25)
    at PopoverOptionSet.render (http://localhost:6006/static/preview.bundle.js:8480:14)
    at ReactCompositeComponentWrapper._renderValidatedComponentWithoutOwnerOrContext (http://localhost:6006/static/preview.bundle.js:111072:34)
    at ReactCompositeComponentWrapper._renderValidatedComponent (http://localhost:6006/static/preview.bundle.js:111099:34)
    at ReactCompositeComponentWrapper.performInitialMount (http://localhost:6006/static/preview.bundle.js:110636:30)
    at ReactCompositeComponentWrapper.mountComponent (http://localhost:6006/static/preview.bundle.js:110524:21)
    at Object.mountComponent (http://localhost:6006/static/preview.bundle.js:118451:35)
    at ReactDOMComponent.mountChildren (http://localhost:6006/static/preview.bundle.js:117141:44)
```

robbieaverill:
I've given this a quick test, some comments:

### Documentation

Usually each React component in admin comes with a short readme file explaining what it is and which are the important props. Whether or not this practice is obsolete since proptypes are pretty self explanatory, we should still follow it.

### Using it

As noted above, the handler callback is throwing an error. If you don't make it required and provide a default then this goes away, e.g. `onClick={provideButtonClickHandler ? provideButtonClickHandler(button) : () => {}}`

### Clear button

The "Clear" button doesn't have a pointer cursor to indicate that it's clickable:

image

This button already has the `btn-link` class. Adding `btn` as well fixes this issue.

### Searching

Do we want a "No results found" type message?

image

---

It looks pretty slick, nice work =)

ScopeyNZ:
Thanks Robbie. Most of those were due to refactor I did but I missed spots 🙁 .

I've also added a README. I mostly copied the proptypes but commented them a lot more thoroughly. Perhaps I should copy that into the actual component 🤔

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,256 @@
...
+          {i18n._t('PopoverOptionSet.CLEAR', 'Clear')}
+        </button>
+      </InputGroupAddon>
+    );
+  }
+
+  /**
+   * Render the search value input box (which is in turn used to filter the buttons)
+   *
+   * @return {DOMElement}
```
robbieaverill: Could be:
```suggestion
   * @return {InputGroup|null}
```

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,256 @@
...
+      case 'Escape':
+        this.handleToggle();
+        break;
+      default:
+    }
+  }
+
+  /**
+   * Render a link to clear the search field if user entered input
+   * @returns {DOMElement}
```
robbieaverill:
```suggestion
   * @returns {InputGroupAddon|null}
```

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,256 @@
...
+            {button.content}
+          </ButtonComponent>
+        ))}
+      </div>
+    );
+  }
+
+  /**
+   * Render the option set popover
+   * @returns {DOMElement}
```
robbieaverill:
```suggestion
   * @returns {Popover}
```

client/src/components/PopoverOptionSet/PopoverOptionSetToggle.js
```diff
@@ -0,0 +1,66 @@
...
+      id,
+      onClick: this.handleToggle,
+    };
+
+    return (
+      <div>
+        <Button {...buttonProps} >
+          Toggle
+        </Button>
+        <PopoverOptionSet {...popoverProps} />
```
robbieaverill: First I find it a little bit strange that the toggle component is also rendering the option set itself. It could just be a case of updating the component's doc block to clarify, at the moment it sounded to me like it should just be a toggle button of some sort.

Second, the component would ideally be injected.

robbieaverill:
For the record, I love this component. Nice job!

client/src/components/PopoverOptionSet/PopoverOptionSetToggle.js
ScopeyNZ: This was only created because there was a lot of boilerplate I was writing in the story for this component - mostly around tracking the state of `isOpen`. That's why this is rendering the component itself. I will refactor to make the `PopoverOptionSet` injected and I'll update any documentation to indicate it's a simple example - using the `PopoverOptionSet` component on its own may be more fit for purpose.

ScopeyNZ: I guess it also makes sense that the word "Toggle" is configurable and the default is translated 🤦‍♂️

dhensby:
rebase / rebuild needed

ScopeyNZ:
I've rebased. The builds don't appear to be running though - GitHub is still being weird.

client/src/components/PopoverOptionSet/PopoverOptionSetToggle.js
ScopeyNZ: Done 🙂

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,261 @@
...
+      return;
+    }
+
+    switch (event.key) {
+      case 'Escape':
+        this.handleToggle();
+        break;
+      default:
+    }
+  }
```
unclecheese: Couldn't this be rewritten:

```js
handleKeyDown(event) {
  event.key === 'Escape' && this.handleToggle();
}
```

ScopeyNZ: Ah yep this method is a little overcomplicated - I was going to add a arrow key support but never got around to it. I'll update it.

unclecheese: Yeah, we need to fix that. That's a really annoying rule.

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -0,0 +1,248 @@
...
+    buttonProps: PropTypes.object,
+  })).isRequired,
+  // Accepts a function that takes a search term as a first parameter and a set
+  // of buttons to match against that returns a filtered set of buttons
+  // Default search handler assumes button content to be plain text and performs
+  // a simple string.contains check.
+  onSearch: PropTypes.func,
+  // function that accepts a button object & returns an event handler
+  // e.g. (button) => (event) => event.preventDefault() && togglePopover();
+  provideButtonClickHandler: PropTypes.func.isRequired,
```
unclecheese: I'm trying to figure out why this needs to be a factory. In doing so, you're guaranteeing that the component is never pure, because that function is dereferenced every time. It's not a huge thing, but I'm wondering:

* Add `onclick` to the `button` shape?
* Have an all-inclusive button handler that passes the `key` to a single function in which the consumer would be expected to step through a `switch` statement.

ScopeyNZ: Updated 🙂

PR #685, opened 2018-10-12
MINOR Auto-publish pattern-lib to gh-pages

maxime-rainville:
This is a hackday project. This updates our travis config to auto-publish the pattern-lib to gh-pages whenever a new commit is added to the `1` branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -121,7 +121,7 @@ deploy:
 - provider: pages
   skip_cleanup: true
   keep-history: false
-  github-token: $GITHUB_API_TOKEN
+  github-token: $GITHUB_AUTO_PUBLISH_API_TOKEN
   local-dir: storybook-static
   on:
     branch: 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,9 @@ matrix:
     - php: 7.1
       env: DB=MYSQL PHPUNIT_TEST=1 PDO=1
     - php: 5.6
-      env: NPM_TEST=1
+      env:
+        - NPM_TEST=1
+        - RELEASE=1
     - php: 7.0
       env:
        - DB=MYSQL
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -89,3 +91,39 @@ script:
...
 after_failure:
   - php ./vendor/silverstripe/framework/tests/behat/travis-upload-artifacts.php --if-env BEHAT_TEST,ARTIFACTS_BUCKET,ARTIFACTS_KEY,ARTIFACTS_SECRET --target-path $TRAVIS_REPO_SLUG/$TRAVIS_BUILD_ID/$TRAVIS_JOB_ID --artifacts-base-url https://s3.amazonaws.com/$ARTIFACTS_BUCKET/ --artifacts-path ./artifacts/
+
+before_deploy:
+# Get asset-admin
+  - git clone https://github.com/silverstripe/silverstripe-asset-admin.git
+  - cd silverstripe-asset-admin
+  - yarn install
+  - cd ..
+  - mv silverstripe-asset-admin ../asset-admin
```
robbieaverill: You can wrap most of these commands in brackets for them to be executed in isolation from the current path, e.g. `(cd silverstripe-asset-admin && yarn install) && mv silverstripe-asset-admin ../asset-admin`

robbieaverill:
This won't work in for the master branch in Travis since you're cloning the default branch. We really need to build with the module in place as installed with Composer

maxime-rainville:
Right now, the pattern lib won't build if `silverstripe/cms`, `silverstripe/asset-admin` and `silverstripe/campaign-admin` are not installed.

That's because of this bit of code in our storybook config that explicitly reference those folders:
https://github.com/silverstripe/silverstripe-admin/blob/b6c24cf18c6de315a8bbad1be63e04b9f6c7335f/.storybook/config.js#L5-L11

`cms` and `campaign-admin` don't even add anything to the pattern-lib currently. So it would be cool to tweak that config so the pattern lib can be build if some of those module are missing.

This only gets triggered on the `1` branch right now because of this statement:
https://github.com/silverstripe/silverstripe-admin/blob/b6c24cf18c6de315a8bbad1be63e04b9f6c7335f/.travis.yml#L126-L127

It might be worthwhile adding some logic so different branches get build in different folders so you can see what exactly is available in each release.

maxime-rainville:
Right now, the pattern lib won't build if `silverstripe/cms`, `silverstripe/asset-admin` and `silverstripe/campaign-admin` are not installed.

That's because of this bit of code in our storybook config that explicitly reference those folders:
https://github.com/silverstripe/silverstripe-admin/blob/b6c24cf18c6de315a8bbad1be63e04b9f6c7335f/.storybook/config.js#L5-L11

`cms` and `campaign-admin` don't even add anything to the pattern-lib currently. So it would be cool to tweak that config so the pattern lib can be build if some of those module are missing.

This only gets triggered on the `1` branch right now because of this statement:
https://github.com/silverstripe/silverstripe-admin/blob/b6c24cf18c6de315a8bbad1be63e04b9f6c7335f/.travis.yml#L126-L127

It might be worthwhile adding some logic so different branches get build in different folders so you can see what exactly is available in each release.

robbieaverill:
Ah, ok that's good then (single branch build). That'll stop this from becoming an issue immediately anyway! I'd love to see the pattern library auto discover modules to include, and even ignore ones that haven't had their `yarn install` commands run yet, but that's both another topic and possibly wishful thinking

PR #687, opened 2018-10-15
Fix search filtering and clearing

### LINKED ISSUE #668: Cannot clear advanced filters from search form
**Version: 4.3.x-dev**

## Steps

* Go to Security
* Add a new member: Test Test
* Open search form in members GridField
* Set advanced filter: First Name to "test" and search
* Open advanced filters and clear the First Name, then search

## Expected

GridField should reset to an unfiltered state

## Actual

"test" remains in the First Name field

## Workaround

Click on the "X" to clear the search

## PRs
- [x] https://github.com/silverstripe/silverstripe-admin/pull/687
- [x] https://github.com/silverstripe/silverstripe-framework/pull/8477

### LINKED ISSUE #672: Search Form Schema doesn't handle more complex relational search field
If you try to use a more complex search that references relations, they don't appear to affect your search results.

I've confirmed in the CWP kitchen with frameworktest running. We probably want to do a PR against frameworktest so we have more fancy search field to test against going forward.

From @sminnee at https://github.com/silverstripe/silverstripe-admin/issues/662#issuecomment-426821181
> Note that the customisable search forms are included in the down-arrow, as shown below. But they don't appear to work.
> 
> Although this form was created by searchable_fields modifications rather than SearchForm customisation:
> 
> ```
>     private static $searchable_fields = array(
>         'Name',
>         'Company.Name' => array( 'title' => 'Company' ),
>         'Company.Employees.Name' => array( 'title' => 'Colleague' ),
>         'Company.Offices.City' => array( 'title' => 'Company city' ),
>     );
> ```
> 
> image

## PRs
- [x] https://github.com/silverstripe/silverstripe-admin/pull/687
- [x] https://github.com/silverstripe/silverstripe-framework/pull/8477

lukereative:
Fixes #668 and fixes #672 

**Requires https://github.com/silverstripe/silverstripe-framework/pull/8477**


robbieaverill:
Here are my test findings (with frameworktest in the Employees model admin):

* Searching for Moses shows two records 🍏 
* Removing Moses from the first name advanced filter field and clicking Search does not remove Moses from the search field 🔴 
* Pressing Clear and then Search clears all of the search filters 🔴(it doesn't clear the DOB field value) 
* Pressing Clear and then Search clears the filters, and also clears the search results 🔴 
* Filtering by advanced filter "Company Name starts with" and value "Auchan" should filter results by that company 🔴 
* Filtering by colleague name "Moses" should show people who are colleagues of Moses, it shows all 🔴 
* Filtering by DOB 18/04/1970 should show one result (Ursa) 🍏

robbieaverill:
Oh I see, this requires https://github.com/silverstripe/silverstripe-framework/pull/8477. I've added that to the description of that PR for you, and also please note that that framework PR could use a rebase now to pull in the fix for filtering on "Any" in the dropdown field =)

Here are my new test findings (with frameworktest in the Employees model admin):

* Searching for Moses shows two records 🍏 
* Removing Moses from the first name advanced filter field and clicking Search does not remove Moses from the search field 🔴 
* Pressing Clear and then Search clears all of the search filters 💛 (it doesn't clear the value of the DOB field immediately, but when you search it does)
* Pressing Clear and then Search clears the filters, and also clears the search results 🍏 
* Filtering by advanced filter "Company Name starts with" and value "Auchan" should filter results by that company 🍏 
* Filtering by colleague name "Moses" should show people who are colleagues of Moses, it shows all 🍏 
* Pressing the "X" to close the search should clear the search state and reset all fields 🔴(colleague name is still set to Moses from earlier)
* Filtering by DOB 18/04/1970 should show one result (Ursa) 🍏

client/src/components/Tag/Tag.js
```diff
@@ -61,7 +61,7 @@ const Tag = ({
       role="button"
       className={classnames('tag', 'btn-sm', { tag__deletable: deletable })}
       onClick={(e) => { e.preventDefault(); onClick(dataKey); }}
-      tabIndex={focusable ? undefined : -1}
+      tabIndex={focusable ? -1 : undefined}
```
maxime-rainville: You sure about this? Setting the focus to -1 prevents it for getting focus.

lukereative: Yea, they used to be `button`s but are now `li`s which aren't focusable by `.focus()`, adding `-1` allows them to be targeted by `.focus()`. We handle the focus programmatically anyway rather than relying on the default tab order. More info: https://snook.ca/archives/accessibility_and_usability/elements_focusable_with_tabindex

lukereative:
That SearchForm key problem is fixed now – tests won't pass thill the framework PR is merged – tested locally and the tests pass.

client/src/components/Tag/Tag.js
maxime-rainville: But people won't be able to access the tags with the tab key. I'm not against adding extra ways to navigate the CMS, but I'm kind of concern with subverting standard ways.

If you're using a screen reader, this will be very confusing.

lukereative: Ah, it was a regression before this change because we changed from `button` to `li` and we needed to make `li` focusable not just through backspace/js – I've changed it to 0 which allows keyboard focus

unclecheese:
Confirmed this fixes the bug. Merging.

PR #695, opened 2018-10-17
BUG: Fixes #674 TinyMCE width

### LINKED ISSUE #674: TinyMCE width in latest release has a smaller width
This issue has been picked up by a community member where the width of TinyMCE has reduced between 4.1 and 4.2. It would be good to get someone to look into this issue to see if the change was intentional or not. I don't see any reason why we need to reduce the width. Its only on very large screen that we wouldn't want it full screen width.

Original feedback here on the forum https://forum.silverstripe.org/t/cms-editor-width/1028/3

4.1
image

4.2
image

#### PR's

- [x] https://github.com/silverstripe/silverstripe-admin/pull/695

bergice:
This should match form field widths at lower width resolutions but expand up to the max width on wider resolutions.


client/src/styles/legacy/_forms.scss
```diff
@@ -169,6 +169,7 @@ form.small .field, .field.small {
 /* Chosen override */
 .chosen-container {
   width: 100%;
+  //max-width: $input-max-width !important;
```
robbieaverill: Is this required or can it be removed?

robbieaverill:
Needs a rebase =)

PR #697, opened 2018-10-17
Fix body overflow causing scroll bars

kinglozzer:
Regression from https://github.com/silverstripe/silverstripe-admin/pull/690. In my apparent tunnel vision, I failed to notice I’d broken the toggle for the SiteTree panel:

screenshot 2018-10-17 at 11 39 35

That toggle overflowing its container caused a vertical scrollbar on `<body>`, this change fixes that. I’ve tested in IE11 and checked that this is the only other use of `.cms-panel-toggle` in core.

screenshot 2018-10-17 at 11 39 15


client/src/styles/legacy/_style.scss
```diff
@@ -1354,13 +1354,17 @@ form.member-profile-form {
 
 .cms-panel {
   .cms-panel-toggle {
+    display: flex;
+    justify-content: flex-end;
+    padding: 0;
 
     .toggle-collapse,
     .toggle-expand {
-      height: 54px;
```
kinglozzer: As the parent `.cms-panel-toggle` is now always `display: flex`, these no longer need a hard-coded height as they’ll stretch automatically.

client/src/styles/legacy/_style.scss
```diff
@@ -1354,13 +1354,17 @@ form.member-profile-form {
...
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
```
kinglozzer: To vertically centre the `<span>&raquo;</span>`, without relying on `line-height` hacks

PR #701, opened 2018-10-18
FIX Change one tab not all tabs

raissanorth:
We need to provide the Tabs reducer with a unique id, otherwise all the tabs in a form will be changed in unison. By providing a unique id, we can change the active tab individually for each tab set.


client/src/components/Tabs/Tabs.js
```diff
@@ -154,20 +154,21 @@ Tabs.defaultProps = {
 export { Tabs as Component };
 
 function mapStateToProps(state, ownProps) {
-  const id = ownProps.id;
-  const field = (state.tabs.fields[id])
-    ? state.tabs.fields[id]
+  const fieldID = `${ownProps.formid}__${ownProps.id}`;
```
unclecheese: Let's move this to a helper function `createFieldID(ownProps)` to reduce repetition.

unclecheese:
Reabse please

PR #704, opened 2018-10-19
Only test specific parts of the CMS menu items

robbieaverill:
If Reports (for example) is run, it may be referencing TestOnly DataObjects that are not loaded into the current unit test class. This change means that only specific items will be checked.

Previously this test wasn't asserting anything at all, I changed that in https://github.com/silverstripe/silverstripe-admin/pull/384, and now that the test is running it's causing failures in the CWP Kitchen Sink Recipe because it's testing every menu item it can. Ideally it would be more mocked and isolated, but this is a start towards that by testing specific parts that should be safe.


tests/php/LeftAndMainTest.php
```diff
@@ -91,7 +91,15 @@ public function testLeftAndMainSubclasses()
         $menuItems = LeftAndMain::singleton()->MainMenu(false);
         $this->assertGreaterThan(0, count($menuItems));
 
+        // We don't want to test every menu item, some may have relations to TestOnly DataObjects
+        // that are not loaded into the scope of this unit test class
+        $itemsToTest = ['Pages', 'Files', 'Security'];
+        $tested = 0;
         foreach ($menuItems as $menuItem) {
+            if (!in_array($menuItem->Title, $itemsToTest)) {
```
dhensby: Is `Title` sufficiently static (non-translated) to rely on? Is there something less prone to change that we could use (URL?)?

tests/php/LeftAndMainTest.php
```diff
@@ -91,7 +91,15 @@ public function testLeftAndMainSubclasses()
...
+                continue;
+            }
+            $tested++;
```
dhensby: `++$tested` - come on, bro!

tests/php/LeftAndMainTest.php
```diff
@@ -108,6 +116,7 @@ public function testLeftAndMainSubclasses()
             $this->assertRegExp('/<head[^>]*>/i', $response->getBody(), "$link should contain <head> tag");
             $this->assertRegExp('/<body[^>]*>/i', $response->getBody(), "$link should contain <body> tag");
         }
+        $this->assertSame(count($itemsToTest), $tested, 'All of the menu items we want to test were tested');
```
dhensby: Hmmm - this is kind of gross, but maybe you should do:

`$this->assertCount($tested, $itemsToTest)`? Also, *the message is for printing when there's been an error* so  something like "Not all menu items were successfully tested"

tests/php/LeftAndMainTest.php
robbieaverill: True, this could be translated

tests/php/LeftAndMainTest.php
robbieaverill: Not sure - the expected value should come first

dhensby: Yeah, true....

robbieaverill:
Lol - the build matrix doesn't have CMS in it, brb

robbieaverill:
"Let's test every possible subclass of a class provided in this module" is bad test logic. I'll stub a LeftAndMain class and test that one specifically instead.

robbieaverill:
I can't stub a LeftAndMain class because the CMS menu items filter out TestOnly classes. I've changed the logic to test for the Security controller instead, since it's provided by framework so we know it will exist.

PR #705, opened 2018-10-19
FIX Remove deprecated help_link definition in testGetHelpLinks

robbieaverill:
Otherwise the assertion fails


robbieaverill:
It's a little unclear, but it's a different config prop which gets set by CWP project code. I'm looking at that separately now

dhensby:
There are no 7.2 or 7.3 test suites on travis.... do we want them?

robbieaverill:
I'll make a new PR to add them

robbieaverill:
Travis PR #707

PR #708, opened 2018-10-22
Fix decimal search filter

bergice:
Fixes #8494


client/src/components/NumberField/tests/NumberField-story.js
```diff
@@ -0,0 +1,28 @@
...
+storiesOf('Admin/NumberField', module)
+  .addDecorator(story => (
+    <div>
+      {story()}
+    </div>
+  ))
+  .addDecorator(withKnobs)
+  .addWithJSX('NumberField', () => (
+    <NumberField
+      title={text('title', '')}
```
maxime-rainville: Maybe add some sensible default example values to your knobs.

All form fields support a `message` props as well to display validation errors.
```js
    message={{
        type: select('message type', ['', 'error'], 'error'),
        value: text('message', 'A message about the field.')
      }}
```

PR #711, opened 2018-10-26
NEW Adding an HOC to provide DragDropContext for consumers of ReactDND

ScopeyNZ:
This is a simple file that exports the DragDropContext HOC from ReactDND. We can only have one source of this HOC otherwise ReactDND kicks up a fuss.

## Dependencies
* https://github.com/silverstripe/webpack-config/pull/31
* https://github.com/silverstripe/silverstripe-asset-admin/pull/857


unclecheese:
Have tested and added the other necessary PRs. This is a nice tidy solution.

ScopeyNZ:
Rebased. Hopefully we can get this into 4.3? 🙏

Issue #635, opened 2018-09-11
TinyMCE is re-fetched when multiple HTMLEditorConfigs are used


Cheddam:
Affected version: 1.2.1

Our project has two different `HTMLEditorConfig` instances - the default `cms` one, and a custom one used on a couple of fields for specific behaviour. When navigating from a UI in the CMS that uses `cms` to one that uses our custom config, a new (identical) version of TinyMCE is generated and fetched, with the name of the custom config in the filename. This doesn't break the interface, but it _does_ generate duplicate items in the Insert Link menu.

To reproduce, clone the default `cms` config to a new custom HTMLEditorConfig in `_config.php` and apply it to a field. Then navigate between the UI containing that field and one containing a field with the default config. Observe two copies of tinymce fetched in the Network tab of your dev tools, and duplicated items in the Insert Link menu.

(Worth checking - is this resolved in the new React-ified version of the field? #628)

## PRs

- [x] https://github.com/silverstripe/silverstripe-admin/pull/721
- [x] https://github.com/silverstripe/silverstripe-framework/pull/8559
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2307
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/861

kinglozzer:
Just noting one workaround for the duplicate link menu items - disabling and then re-enabling the plugins without the paths to the client-side files seems to work:

```php
// In getCMSFields() - assumes the original config is still active
$editorConfig = clone $field->getEditorConfig();
$editorConfig->disablePlugins(['sslinkinternal', 'sslinkanchor', 'sslinkfile']);
$editorConfig->enablePlugins(['sslinkinternal', 'sslinkanchor', 'sslinkfile']);
```

robbieaverill:
We _could_ block all non-current editorconfig `getScriptURL()` calls, but since it's already in the browser because we're switching pages via PJAX I don't think that will help

robbieaverill:
I think we can identify different editor configs with the active identifier key e.g. `cms` or `custom`, then only initialise our custom plugins if they're meant to be in the current identifier's config. Looking into it.

robbieaverill:
So it looks like because both bundles are running and registering the plugins with `TinyMCEActionRegistrar` (our code). The registration of our plugins only happens once per bundle, which is why we see a duplicated set of links but they don't compound to three or four duplicates (they would if you had that many different TinyMCE configs and switched to pages using them).

[...]

PR #721, opened 2018-11-01
BUG: Fix duplicate plugins on HTML editor fields

lukereative:
Fixes silverstripe/silverstripe-admin#635


webpack.config.js
```diff
@@ -47,13 +44,40 @@ const config = [
...
     }
   },
+  {
+    name: 'tinymce',
+    entry: {
+      TinyMCE_sslink: `${PATHS.LEGACY_SRC}/TinyMCE_sslink.js`,
+      'TinyMCE_sslink-external': `${PATHS.LEGACY_SRC}/TinyMCE_sslink-external.js`,
+      'TinyMCE_sslink-email': `${PATHS.LEGACY_SRC}/TinyMCE_sslink-email.js`,
+    },
+    output: {
```
unclecheese: Let's reduce the duplication here. Everything from this point down can be in a common object. Then, for both configs, use `...commonConfig`. For the vendor bundle, use

```js
...commonConfig,
plugins: [
  ...commonConfig.plugins,
  new webpack.optimise.CommonChunks(....
]
```

webpack.config.js
```diff
@@ -47,13 +44,40 @@ const config = [
       // bundles aren't duplicating libs.
       new webpack.optimize.CommonsChunkPlugin({
         name: 'vendor',
-        minChunks: module => module.context && module.context.indexOf('/node_modules/') > -1,
+        minChunks(module) {
```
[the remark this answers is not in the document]
lukereative: Ah, I thought I'd removed it.

unclecheese:
@lukereative Can you confirm that these API changes to `TinyMCEActionRegistrar` are backward compatible? e.g. If I have a custom action I've registered in my 4.2 project, is it going to work in 4.3 without intervention?

lukereative:
@unclecheese I might actually make a change so if you had a plugin that doesn't have an config identifier, `getActions` will return the actions for the specific config and grab any that were registered without a config identifier, so they aren't completely ignored

client/src/lib/TinyMCEActionRegistrar.js
```diff
@@ -18,26 +28,60 @@ class TinyMCEActionRegistrar {
...
+    return actions;
   }
 
-  getSortedActions(menu) {
-    const actions = this.getActions(menu);
+  /**
+   * Get list of actions for this menu, sorted by first by priority and then text
+   *
+   * @param {String} menu
+   * @param {Bool} configId Identifier of which config
```
robbieaverill: This isn't a bool. Also would be good to be consistent with the types (Bool vs Boolean), and also note that I think JSDoc generally uses lowercase for scalar types, other than Object and Array which seem to be used in both uppercased and lowercase in examples

PR #744, opened 2018-11-05
Add tests and styling for betterbuttons port

### LINKED ISSUE #436: Prev, Next and Create buttons on detail GridField forms
# Overview

It's often cumbersome to move through a list of items by clicking into the detail view, hitting the "back" button, and clicking into the next item. The [betterbuttons](https://github.com/unclecheese/silverstripe-gridfield-betterbuttons) module solved this for 3.x, and there's a good case for making it a core feature. @sachajudd and @clarkepaul have designed a solution as part of a [Save and add new button](https://github.com/silverstripe/silverstripe-cms/issues/2047#issuecomment-365811496) design.

# Acceptance Criteria

 - [x] I can add a new record from a record detail form
 - [x] If choosing to add a new record or navigate to the previous/next record on a record detail form with unsaved edits, I get a prompt and stay on the same screen until I've saved the record explicitly. 
 - [x] I can navigate to the previous and next records of a list when viewing a record detail form
 - [x] If there is no more previous or next records, the button is disabled
 - [x] The new buttons don't compromise more important buttons on small screens
 - [x] The functionality works in ModelAdmin and any other GridField
 - [x] Sorting and filtering is retained
 - [x] Does not double up when betterbuttons is installed (might need a patch to that module?) (UPDATE: BetterButtons already takes priority and overrides all the actions)
 - [x] Should work in default configurations, but can be opted out of
 - [x] Behat Tests written
 - [x] Update documentation
 - [x] Excludes choosing the type of record to create (that's a thing in some GridField modules)
 - [x] Excludes any "save and close" or "publish and close" logic

# Designs

[Styleguide](https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5b5e610563c7ad0012bc21fb)

# Notes

 * Some historical context and discussion in a [duplicate card](https://github.com/silverstripe/silverstripe-cms/issues/2047)

# Pull Requests
- [x] https://github.com/silverstripe/silverstripe-framework/pull/8569
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/744
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/8619

bergice:
Fixes #436


client/src/legacy/LeftAndMain.js
```diff
@@ -1086,6 +1086,14 @@ $.entwine('ss', function($) {
...
   });
 
+  $('.cms button.cms-panel-link').entwine({
+    onclick: function(e) {
+      if (!$('.cms-container').checkCanNavigate()) {
+        e.preventDefault();
+      }
+    }
+  });
+
```
unclecheese: I'm quite certain you don't need these click handlers. Just remove `cms-panel-link` from the buttons. They're form actions, so they already have ajax submit handlers attached to them. I've done a really quick test, and it works fine if you remove that class and remove these changes.

unclecheese: Gotcha. I still don't see why it needs to be `cms-panel-link`, though. That class is used for a very specific purpose, and adding it to your form actions adds complexity and potential breakages. What is `cms-panel-link` doing for you here?

bergice: It is necessary to distinguish from actions that don't need the discard confirmation (such as the save action).

I have updated this with a more relevant class name.

ScopeyNZ:
How do we feel about getting some instant tooltips on these buttons? Not everyone's familiar with `betterbuttons` 🙂

sachajudd:
Hey, I've had a quick look at this and noticed it doesn't quite match the designs.

Current:
image

A few things:
* Missing background behind the arrows.
A few options for the icon: 
* The icon could be done in CSS since the icon is currently not in the icon set.
* We could use the close icon and rotate it 45 degrees. 
* Use font-icon-plus-circled and change the icon background to green.
* Leave as is and replace the icon when we update the icon set.

Styleguide (click the preview link): https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5b5e610563c7ad0012bc21fb

sachajudd:
Great idea @ScopeyNZ 🙂 see updated [Styleguide](https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5b5e610563c7ad0012bc21fb) for tooltips.

sachajudd:
We have designed better buttons to work within the CMS with these new custom styles. We haven’t added any buttons without text yet in the CMS so because it's a new pattern we'd rather it look more cleaner rather than consistent. We hope to start using these styles throughout the CMS more often in the future.

client/src/components/Toolbar/Toolbar.scss
```diff
@@ -14,6 +14,41 @@
   .btn {
     margin-bottom: 0;
   }
+
+  .right {
+    margin-left: auto;
+
+    .btn-group:not(last-child) {
+      margin-right: 10px !important;
```
ScopeyNZ: There's a lot of use of `!important` here. Can this be avoided?

client/src/components/Toolbar/Toolbar.scss
```diff
@@ -14,6 +14,41 @@
...
+    }
+  }
+
+  .rounded {
+
+    &.btn,
+    >.btn {
+      width: 32px;
+      padding-left: 7px;
```
ScopeyNZ: There's some specific styling going on here. It probably makes sense to define these "betterbuttons" styles in BEM format? Perhaps there's some existing variables could be used for the padding here. There's some like `btn-spacing` and `panel-padding`.

ScopeyNZ: This could also allow adding a background colour on the previous and next buttons that the designs had and keep it specific to these "betterbuttons"

client/src/styles/_fonts.scss
```diff
@@ -58,8 +58,10 @@
   content: "\6f";
 }
 
-.font-icon-plus:before {
-  content: "\6a";
+.font-icon-plus-thin:before {
+  content: "\44";
+  transform: rotate(45deg);
+  display: inline-block;
```
sachajudd: Because this is one of the automated files for how we [update fonts](https://github.com/silverstripe/silverstripe-admin/pull/633) to admin it's probably best to add these styles to the button directly.

Issue #648, opened 2018-09-17
Upgrade to React 16 and latest node.js version


lukereative:
As React is the integral front-end library for the CMS, it is important we keep it up to date so we can take advantage of improvements made to it and also so we may use other packages which depend on newer versions. @unclecheese and I have already done the majority of the upgrade effort and have moved prop-types to the 'new' external library (#644) and there are branches 'dont-over-react' on each of the modules that use react with initial upgrades.

---

Also, from @torleif in #673:

node.js 6 maintenance ends in April 2019. 

From what I can tell this might be a bit of a challenge, as silverstripe/webpack-config seems to have issues with the eslint module and proptypes in the latest version of react. 

## Note before merging
Need to bump the version number and npm release webpack-config https://github.com/silverstripe/webpack-config/pull/32 and update the other PRs before merging. Use `yarn link` to test the other branches with the new webpack-config before merging

## PRs

- [x] https://github.com/silverstripe/silverstripe-admin/pull/749
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2319
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/875
- [x] https://github.com/silverstripe/silverstripe-versioned-admin/pull/80
- [x] https://github.com/silverstripe/silverstripe-campaign-admin/pull/125
- [x] https://github.com/dnadesign/silverstripe-elemental/pull/494
- [x] https://github.com/silverstripe/webpack-config/pull/32

robbieaverill:
@lukereative @unclecheese I've just changed the scope of this issue - in case that affects the estimate you've put on it already

robbieaverill:
Hey @unclecheese, quick question regarding module compatibility with this in SilverStripe 4.4:

If `silverstripe/somemodule` requires SilverStripe ^4.0 and has some React stuff that uses components from admin in it, how likely is that the code will no longer be compatible with SilverStripe 4.4 with React 16?

If it's likely to be incompatible then obviously we can release new a new minor version of `silverstripe/somemodule` which requires ^4.4, but constraining the previous releases (which have ^4.0 as the requirement) is not possible unless people `composer update silverstripe/somemodule` to pull in a new patch release with more restrictive constraints.

What should we make as the de-facto way of doing this? Patch release the stable release line to change the constraint to `^4.0 <4.4` so future updates will mark it as incompatible, then release a new minor version that requires `^4.4` ?

PR #749, opened 2018-11-09
Upgrade to React 16 and latest node.js version

lukereative:
As a part of silverstripe/silverstripe-admin#648


unclecheese:
Due to the interdependencies, this has to be merged in spite of failing behat tests. Have tested locally and have gotten solid green on all updated modules.

PR #751, opened 2018-11-12
API Add a hook to allow third party JS to detect page change (EXPERIMENTAL)

maxime-rainville:
This PR is a bit experimental and adds the ability to detect when we navigate to a different page using _Page.JS_.

It's probably not the best way of doing this, but I needed something quick and dirty for my CMS Analytics hack day project. I added a dependency to RxJS _Observables_ ... we probably could do it without it, but it's quite a useful lib to have around in some context.

To listen for a _Page.js_ navigation event, all you need to do is this:
```js
window.ss.router.subscribe(({path}) => {
  // ... do something
});
```

This doesn't with Reacto Router navgiation, only the Page.js navigation.


maxime-rainville:
Probably not needed anymore. I'll close it for now.

PR #754, opened 2018-11-15
MINOR Change Tag class to avoid conflict with taxanomie

### LINKED ISSUE #747: Tag React component styles are too broad
**Version: SS 4.3.x-dev**

https://github.com/silverstripe/silverstripe-admin/blob/1.3.0-rc1/client/src/components/Tag/Tag.scss#L1-L12

These styles are too broad. They're applying themselves to a TagField instance in the blog module and breaking the layout of it (https://github.com/silverstripe/silverstripe-blog/issues/558).

How many places would be affected if we changed the `tag` class to something less generic?

# Pull Request
* [x] #754

maxime-rainville:
Fixes #747


ScopeyNZ:
Hmmmn - `MINOR` makes sense for a commit prefix in this case, but it isn't recognised by cow 😞 : https://github.com/silverstripe/cow/blob/master/src/Model/Changelog/ChangelogItem.php#L59-L72

Even though changing classes would be considered breaking, this could just be classified as a `FIX` I'd say because this code hasn't actually been released yet. Mind changing the prefix?

client/src/components/Tag/Tag.js
```diff
@@ -59,7 +59,7 @@ const Tag = ({
     <Button
       {...props}
       role="button"
-      className={classnames('tag', 'btn-sm', { tag__deletable: deletable })}
+      className={classnames('tag-component', 'btn-sm', { 'tag-component__deletable': deletable })}
```
ScopeyNZ: Deletable is a modifier in this case so technically it should be `tag-component--deletable`.

PR #757, opened 2018-11-21
NEW Allow state to be passed back on grid field actions

ScopeyNZ:
This PR is a benign addition to GridField actions to pass through an "action-state" data attribute (if available) into the request that performs an action. This is targetting 1.3 as it's part of a fix to prevent cramming the session with unnessecary data which doesn't scale well.

See related PR silverstripe/silverstripe-framework#8627


ScopeyNZ:
I've just noticed there's an issue with `GridFieldFilterHeader`. Looking into it 😪

bergice:
Tests need to be fixed up.

ScopeyNZ:
Looks like PECL was having some problems yesterday. I've restarted the builds

client/src/legacy/GridField.js
```diff
@@ -366,21 +366,33 @@ $.entwine('ss', function($) {
...
+      var data = [
         {
-          data: [{
-            name: this.attr('name'),
-            value: this.val(),
-            filter: filterState,
-            triggerChange
-          }],
+          name: this.attr('name'),
+          state: 'abc',
```
maxime-rainville: This looks like a debug attribute.

ScopeyNZ: Yep. I'll remove this.

ScopeyNZ: Removed.

ScopeyNZ:
I've just rebased. I don't expect that a rebase will fix the failing builds that aren't related. 

I don't have time to address the failing Behat tests within this PR.

ScopeyNZ:
I think I found an issue so please don't merge yet.

maxime-rainville:
@guy Have you found/resolve your this new mysterious issue?

ScopeyNZ:
Yeah. The issue was the same that is fixed by #765 . Sorry we're good to merge given that's fixed 🙂

maxime-rainville:
Merging everything before anyone has a chance to say otherwise.

robbieaverill:
But....

robbieaverill:
Jokes

PR #759, opened 2018-11-23
FIX Removing unnecessary overflow on fieldsets

ScopeyNZ:
Potentially controversial fix here. Currently there's an overflow auto style on fieldsets. By default there's no reason content should overflow a fieldset _except_ when displaying things like a TreeDropdownField. In that case the field is actually cut off - see silverstripe/silverstripe-asset-admin#880 and https://github.com/dnadesign/silverstripe-elemental/issues/514#issuecomment-440128821 .

I argue that an overflow is unneseccary on a fieldset. It should be something that is applied per fieldset if required - and it is required in asset-admin. See PR silverstripe/silverstripe-asset-admin#885 .

Fixes silverstripe/silverstripe-asset-admin#880


unclecheese:
So what exactly does this fix? I see that the asset-admin PR fixes a display issue by applying the overflow property, but what is the issue that this is trying to solve?

ScopeyNZ:
It fixes the cut off react-select dropdowns as shown in screenshots in the elemental/asset-admin issues. The dropdown DOM exists in the fieldset and is therefore affected by the global overflow rule.

It's a bit of a "best practice" change I think because I don't think we should be applying an overflow rule so globally.

ScopeyNZ:
Rebased...

dnsl48:
We should revert or remove `b5d38b6b5ce773cb4a433e548a08bac2443e6d7e`, as it's breaking the tests now.

ScopeyNZ:
Done.

Issue #753, opened 2018-11-14
Empty confirmation message when unlinking a record from a GridField


robbieaverill:
**Version: SS 4.3.x-dev**

Use this example Page:

```php
<?php

use SilverStripe\CMS\Model\SiteTree;
use SilverStripe\Forms\GridField\GridField;
use SilverStripe\Forms\GridField\GridFieldAddNewButton;
use SilverStripe\Forms\GridField\GridFieldConfig_RelationEditor;
use SilverStripe\Forms\GridField\GridFieldEditButton;
use SilverStripe\Forms\GridField\GridFieldFilterHeader;

class Page extends SiteTree
{
    private static $many_many = [
        'RelatedPages' => Page::class,
    ];

    public function getCMSFields()
    {
        $fields = parent::getCMSFields();

        $components = GridFieldConfig_RelationEditor::create();
        $components->removeComponentsByType(GridFieldAddNewButton::class);
        $components->removeComponentsByType(GridFieldEditButton::class);
        $components->removeComponentsByType(GridFieldFilterHeader::class);

        $fields->addFieldToTab(
            'Root.RelatedPages',
            GridField::create(
                'RelatedPages',
                _t(__CLASS__ . '.RelatedPages', 'Related pages'),
                $this->RelatedPages(),
                $components
            )
        );

        return $fields;
    }
}
```

* Go into the admin
* Select a page
* Go to "Related Pages"
* Use the link existing autocompleter to link another existing page
* Click on "Archive" after it has been added

The confirmation message is empty:

![image](https://user-images.githubusercontent.com/5170590/48483352-dfb8ee80-e81b-11e8-9cbc-4a5fe0dc5fe4.png)

# Pull request
* [ ] https://github.com/silverstripe/silverstripe-admin/pull/760

maxime-rainville:
Looks like the message was added in the compiled `client/lang/en.js` instead of `src`. So it probably got overridden. https://github.com/silverstripe/silverstripe-admin/commit/2b16e2afbaa322ef4ab2fce6c2add9b8f5596ba4#diff-1cfb8615fb105f3d175fb84aaa7a3fc0

PR #760, opened 2018-11-27
FIX Readding ARCHIVECONFIRMMESSAGE

maxime-rainville:
The translation key for the archive confirmation message is missing from admin.

# Parent issue
* #753


robbieaverill:
The behat failures look unrelated to this change

PR #762, opened 2018-11-28
FIX Ensure TinyMCE is correctly instantiated and removed when dom matching changes

robbieaverill:
Also improves the logic a little so it is easier to read, and fixed a warning in a jest test

Fixes https://github.com/silverstripe/silverstripe-cms/issues/2341

cc @nightjar who worked on this recently


robbieaverill:
Yep we're getting a couple of failures at the moment, look unrelated in other PRs too

Issue #558, opened 2018-07-03
1.2.x - private static $menu_icon has no effect


kinglozzer:
Looks like a regression from https://github.com/silverstripe/silverstripe-admin/pull/412.

[`LeftAndMain::menu_icon_for_class()`](https://github.com/silverstripe/silverstripe-admin/blob/1.2.0-beta1/code/LeftAndMain.php#L913-L929) returns CSS for the menu icons, but it wasn’t updated as part of the above PR.

The classes (in `LeftAndMain_MenuList.ss`) have changed from:

`menu__icon icon icon-16 icon-{$Icon}`
to
`menu__icon font-icon-circle-star font-icon-{$Icon}`

I guess that means that the generated CSS won’t work even if the classes are updated, as the icon font uses pseudo-elements

lekoala:
I can confirm this is indeed not working in 4.2

Either old 16px should be removed entirely, or the MenuList template should allow them to be used.

dajeinc:
I am running into this problem. Any solution?

lekoala:
The solution is overriding the template SilverStripe/Admin/Includes/LeftAndMain_MenuList.ss with the content of version 4.1 and ensure you have the proper css loaded by the admin

```
<ul class="cms-menu__list">
	<% loop $MainMenu %>
		<li class="$LinkingMode $FirstLast <% if $LinkingMode == 'link' %><% else %>opened<% end_if %>" id="Menu-$Code" title="$Title.ATT">
			<a href="$Link" $AttributesHTML>
				<% if IconClass %>
					<span class="menu__icon $IconClass"></span>
				<% else %>
					<span class="menu__icon icon icon-16 icon-{$Icon}">&nbsp;</span>
				<% end_if %>
				<span class="text">$Title</span>
			</a>
		</li>
	<% end_loop %>
</ul>
```

```
.menu__icon.icon-16 {
    left: 14px;
    background-repeat: no-repeat;
}

@media (min-width: 768px) {
    .collapsed .menu__icon.icon-16 {
        left: 22px;
    }
}

```

baukezwaan:
The solution above would not be complete, as LeftAndMain classes without any icons set ($menu_icon or $menu_icon_class) wouldn't have a fallback. These should show the circle-star image...

Another work-around would be changing the menu_icon_for_class function to:

[...]

PR #763, opened 2018-11-29
Fixing custom menu_icon on LeftAndMain classes

3Dgoo:
This fixes the problem as described by #558 

These are some simple css and template tweaks to allow the custom icon images to work. If no custom icon is set, this will continue to work as it previously did. If a menu icon class is set, this will work as previously.


templates/SilverStripe/Admin/Includes/LeftAndMain_MenuList.ss
```diff
@@ -2,10 +2,10 @@
 	<% loop $MainMenu %>
 		<li class="$LinkingMode $FirstLast <% if $LinkingMode == 'link' %><% else %>opened<% end_if %>" id="Menu-$Code" title="$Title.ATT">
 			<a href="$Link" $AttributesHTML>
-				<% if IconClass %>
+				<% if $IconClass %>
 					<span class="menu__icon $IconClass"></span>
 				<% else %>
-					<span class="menu__icon font-icon-circle-star font-icon-{$Icon}">&nbsp;</span>
+					<span class="menu__icon icon icon-16 icon-{$Icon}">&nbsp;</span>
```
kinglozzer: This means that if no custom icon is set for a new `LeftAndMain` section, it’s blank instead of the fallback star icon. I have an idea I’ll post below...

client/src/components/Menu/Menu.scss
```diff
@@ -402,6 +402,8 @@
   margin-top: -1px;
   left: 13px;
   font-size: $h3-font-size;
+  min-width: 16px;
+  background-repeat: no-repeat;
```
kinglozzer: The background position seems to be vertically off-center when testing this, shouldn’t we set it to `50% 50%`?

kinglozzer:
To ensure we don’t break the fallback star icon, how about:

```php
// in LeftAndMain::MainMenu()
// ...
$iconClass = '';
$hasCSSIcon = false;
if ($menuItem->controller) {
    $menuIcon = LeftAndMain::menu_icon_for_class($menuItem->controller);

    if (!empty($menuIcon)) {
        $menuIconStyling .= $menuIcon;
        $hasCSSIcon = true;
    } else {
        $iconClass = LeftAndMain::menu_icon_class_for_class($menuItem->controller);
    }
} else {
    $iconClass = $menuItem->iconClass;
}

$menu->push(new ArrayData(array(
    "MenuItem" => $menuItem,
    "AttributesHTML" => $menuItem->getAttributesHTML(),
    "Title" => $title,
    "Code" => $code,
    "Icon" => strtolower($code),
    "IconClass" => $iconClass,
    "HasCSSIcon" => $hasCSSIcon,
    "Link" => $menuItem->url,
    "LinkingMode" => $linkingmode
)));
```

Then in the template:
```
<% if $IconClass %>
	<span class="menu__icon $IconClass"></span>
<% else_if $HasCSSIcon %>
	<span class="menu__icon icon icon-16 icon-{$Icon}">&nbsp;</span>
<% else %>
	<span class="menu__icon font-icon-circle-star">&nbsp;</span>
<% end_if %>
```

templates/SilverStripe/Admin/Includes/LeftAndMain_MenuList.ss
3Dgoo: Thanks Loz. I have added your suggested change.

client/src/components/Menu/Menu.scss
3Dgoo: Thanks Loz. I have added `background-position: 50% 2px;`. I tried 50% 50% but that dropped the icon way too low. Have a look and let me know what you think.

3Dgoo:
I tried to rebase my branch, but it wasn't playing nicely. So instead I created a new branch off of 1.2 and have submitted this as a new pull request #770 

Sorry for replacing this pull request, but it was just much easier for me to do it this way. I hope that's OK.

PR #769, opened 2018-12-04
BUG: Always continue loading scripts even if one script fails.

bergice:
This is especially important so that the deferred callbacks are triggered in the end.

Also show error messages of JS scripts that failed loading, rather than failing silently.


robbieaverill:
Please reference the GitHub issues in the description - not everyone has ZenHub =)

PR #770, opened 2018-12-04
Updating MenuList.ss to allow menu_icon images

3Dgoo:
This fixes the problem as described by #558

This is a replacement for pull request #763 targeting the 1.2 branch instead of the 1 branch.


ScopeyNZ:
I've merged that up for you so the fix is available in the `1.3` and `1` branches.

3Dgoo:
Thanks Guy and everyone else.

PR #772, opened 2018-12-05
fix 1.3 tests

dnsl48:
Fixes tests broken since [FIX Correct the entwine match for the loading animation](https://travis-ci.org/silverstripe/silverstripe-admin/builds/457683596).

## Problem

1. `.cms-edit-form` gets `changetracker` [set up](https://github.com/silverstripe/silverstripe-admin/blob/5d45c7deb81aea23f31f9f0289ea3a8837284cf9/client/src/legacy/LeftAndMain.EditForm.js#L84) in `onadd`, which happens once the form gets added to the DOM

2. Other `onadd` handlers start spinning, one of which triggers the `changetracker` to traverse the form fields and get the initial form values state

3. `onmatch` events go on, at this point `SecurityAdmin` (and `PermissionCheckboxSetField`) onmatch happen and change the form values (toggling the checkboxes) ([here](https://github.com/silverstripe/silverstripe-admin/blob/5d45c7deb81aea23f31f9f0289ea3a8837284cf9/client/src/legacy/SecurityAdmin.js#L45) and [here](https://github.com/silverstripe/silverstripe-admin/blob/5d45c7deb81aea23f31f9f0289ea3a8837284cf9/client/src/legacy/PermissionCheckboxSetField.js#L58) accordingly)

4. When we press `Unlink` button, the field triggers `changetracker` and it appends `.changed` class to the form as the initial form values and the current ones don't match

5. The `.changed` class forces the `onbeforeunload` handler to ask user a confirmation for leaving the page

## The solution

If we perform `toggleCheckboxes` in `onadd` it will groom the form into the final state before `changetracker` gets its `initialState`


.travis.yml
```diff
@@ -70,13 +70,14 @@ before_script:
   - if [[ $BEHAT_TEST == cms ]]; then composer require silverstripe/recipe-cms:4.3.x-dev --no-update; fi
 
   - composer install --prefer-source --no-interaction --no-progress --no-suggest --optimize-autoloader --verbose --profile
+  - composer show
```
robbieaverill: Is the output from `composer install` not sufficient?

dnsl48: well, technically it is, but `show` gives it ordered by alphabet and easier to read by humans (IMHO)

chillu:
@ScopeyNZ Do you have enough entwine experience to assess this? We're a bit short of that in open sourcerers (sicne Aaaron is busy)

tests/behat/features/reauthenticate.feature
```diff
@@ -14,24 +14,26 @@ Feature: Reauthenticate
...
       And I switch to the "login-dialog-iframe" iframe
-    Then I should see "Your session has timed out due to inactivity" in the ".cms-security__container" element
+    Then I should see "You must be logged in" in the ".cms-security__container" element
     When I fill in "Password" with "Secret!123"
       And I press the "Let me back in" button
       And I am not in an iframe
       And I go to "/admin/security"
-      When I press the "Add Member" button
+    When I press the "Add Member" button
+      And I wait for 5 seconds until I see the "#Form_ItemEditForm_action_doSave" element
```
chillu: This stuff isn't ideal, we have generic "after" steps which check for the presence of a loading indicator (or pending ajax request), and delay processing the next step until then. Maybe track that down? I'm happy merging as-is to unblock further work, but these "and I wait" steps are noise that should already be accounted for elsewhere.

ScopeyNZ: To be fair this one is a polling step. So `And I wait x until` is just "up to" x seconds. The actual step might only be a few milliseconds of waiting.

ScopeyNZ: Oh that makes a lot of sense, the class name has changed: https://github.com/silverstripe/silverstripe-admin/pull/765/files#diff-19f74f986482f632702ec13c64f6b019R1060
It should look for `.cms-loading-container` now. It'd be good to do a follow up on this.

ScopeyNZ: It also may mean we can revert this PR. Probably better not to change things that might not be broken if we can just wait for loading indicators correctly.

dnsl48: here we go: https://github.com/silverstripe/silverstripe-framework/pull/8663

this particular scenarios are marked as `@modal` though, so we'll have to keep waiting manually, or rewrite some other stuff in `BasicContext` in `silverstripe-behat-extension`. With just removing `@modal` it's not working at this point. (not sure why exactly at this point)

PR #773, opened 2018-12-05
FIX Firefox sticky toggle misalignment and rotate help menu caret

sachajudd:
Sticky toggle:
Before:
image

After:
image


robbieaverill:
Behat failure is unrelated

PR #776, opened 2018-12-07
NEW Allow descriptions to be added as titles on SingleSelectField

ScopeyNZ:
As titled. `title` is one of the global attributes on DOM elements so it makes sense to be able to set it. And there's the very occasional case when having a title on an option in a select makes sense (eg. Why is this option disabled?).


client/src/components/SingleSelectField/SingleSelectField.js
```diff
@@ -50,9 +50,10 @@ class SingleSelectField extends Component {
       <Input type="select" {...this.getInputProps()}>
         { options.map((item, index) => {
           const key = `${this.props.name}-${item.value || `empty${index}`}`;
+          const description = item.description || null;
 
           return (
-            <option key={key} value={item.value} disabled={item.disabled}>
+            <option key={key} value={item.value} disabled={item.disabled} title={description}>
```
robbieaverill: Do you think it's worth documenting the description prop in the source shape?

robbieaverill:
Ping @ScopeyNZ

ScopeyNZ:
I'm wondering if this is even worth it. Titles on options seems to be badly supported in a lot of browsers.

Anyway. I added the PropTypes definition and I added a knob for the description in the storybook. I'll leave it to you.

For context on why I made this PR - we had some select options that were disabled, but it's not immediately obvious why some options are disabled - having a "title" attribute seemed like a quick win.

PR #781, opened 2018-12-12
Revisions to datefield to have better HTML5 fallbacks

bergice:


maxime-rainville:
How is this different from https://github.com/silverstripe/silverstripe-admin/pull/630?

Looks like it's doing exactly the same thing.

PR #791, opened 2018-12-19
Revisions to `DatetimeField` to have better HTML5 fallbacks

bergice:


bergice:
@maxime-rainville I think this could be happening because of the member locale: 
image

Just tested on my computer and changing it seemed to fix the issue.

maxime-rainville:
@bergice Tried siwtchign to English UK and French. I'm always getting the date in en_US on this end.

bergice:
@maxime-rainville could you try overriding the locale in your `_config.php`?

maxime-rainville:
There was a problem with MomentJS locale's not being loaded. They were included when the library was being compile without the `-p` flag. However, when compiled with it the Locale would be strip out to save space. So at run time, MomentJS would only know about `en_US`.

I added a bit of code to the Webpack config that copies all the local files to the dist folder. I added a Requirements call to LeftAndMain to load the Moment Loclaes based on the current default Locale for the project and the Member's Locale. This fix relies on https://github.com/silverstripe/silverstripe-framework/pull/8726

bergice:
@maxime-rainville I've made some further fixes on top of your fixes.

PR #793, opened 2019-01-05
Update fr.yml

intwebg:
The french translation doesn't match to english at all. I have translated all the better I can. You can check if my translation is correct.


robbieaverill:
Hi @intwebg, we manage our non-English translations through Transifex. They're automatically overwritten whenever we do a release, so you'll need to update the French language translations there: https://www.transifex.com/silverstripe/silverstripe-admin/translate/#fr

I won't merge this since it'll just get overwritten, but thanks for taking the time to make a pull request.

PR #801, opened 2019-01-11
FIX Forwarding action state in another situation where it was not being forwarded

ScopeyNZ:
This is related to #757 . There's a relatively uncommon way that the new `AttributeStore` method of passing around GridField action state could fail if you used the React GridFieldActions component but you had one action that was _not_ edit. In this case the dropdown was not rendered, instead there is a single button. When that button was edit it uses a different part of entwine and works correctly but if it's any other action (eg. delete) then the action state is not present and the action fails.

Fixes silverstripe/silverstripe-framework#8675


robbieaverill:
Behat is broken upstream already

PR #806, opened 2019-01-15
FIX: Use lighter colour for placeholders (fixes #805)

### LINKED ISSUE #805: Form field placeholders are too hard to distinguish from actual data
One of these is a placeholder, the other is a real value:

screenshot 2019-01-15 at 15 40 28

Observed in Chrome 72 on OS X

# Related PRs
* [x] #806

kinglozzer:
Fixes #805. Looks like:

screenshot 2019-01-15 at 15 58 24

For reference, the contrast ratio of `$gray-400` on white is 2.68:1. I appreciate that fails most contrast checks, but anything darker is still too hard to distinguish from a value. IMO it’s acceptable to fail a contrast check here as a placeholder is more akin to progressive enhancement than something essential to the form, happy to listen to others’ views on that though.

I’m also happy to consider alternative colours if anyone can suggest any that look different enough without looking totally wild 😁


clarkepaul:
Commented on the original issue, can we bump that a tiny bit up to 3:1 (darken by x%) and happy to italicise too. I would hope there is still enough difference between them.

kinglozzer:
Thanks, will update 👍

clarkepaul:
I just did a quick check and I think it'll be another 4% darker, hopefully that still looks different enough.

PR #807, opened 2019-01-16
NEW Convert drag handle and dropdown caret to use font-icons in site tree

sachajudd:
Relies on: https://github.com/silverstripe/silverstripe-cms/pull/2367

image


client/src/styles/legacy/_tree.scss
```diff
@@ -493,6 +493,22 @@
     ins {
       background-color: transparent;
       background-image: url(../images/sitetree_ss_default_icons.png);
+
+      &.font-icon-drag-handle {
+        background: none;
+        font-size: 1.15385rem; // 15px
```
maxime-rainville: Is there value here in using `rem` for `.font-icon-drag-handle` and `.font-icon-right-dir`?

Looks like we are clearly targeting a pixel size. And that those size are meant to line up with margins else where in the file.

client/src/styles/legacy/_tree.scss
```diff
@@ -493,6 +493,22 @@
...
+
+        &::before {
+          margin-left: -1px;
+        }
+      }
+
+      &.font-icon-right-dir {
+        visibility: hidden;
+        background: none;
+        font-size: 1.53846rem; // 20px
```
kinglozzer: Comment above says `font-size: 1.15385rem; // 15px`, is one of the two px values wrong?

kinglozzer: Ha, I missed that one is 1.**1**5 and the other is 1.5 🤦‍♂️

maxime-rainville:
@robbieaverill You need the CMS PR as well. silverstripe/silverstripe-cms#2367

robbieaverill:
I like that it visually showed me that I was 50% of the way to having all the code

sachajudd:
Let me know what you think of the new alignment 🙂 
image

Issue #812, opened 2019-01-30
Add a getter for ModelAdmin::$modelClass


JoshuaCarter:
Changes to ModelAdmin.php has caused a break in silverstripe-multisites

See https://github.com/symbiote/silverstripe-multisites/issues/96 for an outline of the issue.

This could be easily fixed in `silverstripe-multisites` by exposing `ModelAdmin::$modelClass` with a getter.

I'm happy to PR that in if someone can confirm that it's a suitable change?

Cheers.

ScopeyNZ:
I think that's a reasonable enhancement. It would be a minor change though and would be released in version 1.4 of this module.

I've renamed the title of this issue to align with the suggestion :)

chillu:
Yep, confirmed that's a good idea - but I'll close the enhancement ticket for it. PR welcome :)

JoshuaCarter:
@chillu Cheers, PR here: https://github.com/silverstripe/silverstripe-admin/pull/813

PR #813, opened 2019-01-30
NEW: Add getter for ModelAdmin::$modelClass

JoshuaCarter:
Solves: https://github.com/silverstripe/silverstripe-admin/issues/812

As per @ScopeyNZ comment in the above issue, this wouldn't be part of a release until 1.4, so I've assumed that master is the correct branch, but correct me if I'm wrong.

Ideally this could go out in a patch release on 1.3 so that `silverstripe-multisites` can be made compatible with SS 4.3, but perhaps 4.4 isn't far off?

Cheers.

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -320,7 +320,6 @@ abstract class ModelAdmin extends LeftAndMain
         return $form;
     }
 
-
     /**
      * You can override how ModelAdmin returns DataObjects by either overloading this method, or defining an extension
      * to ModelAdmin that implements the `updateList` method (and takes a {@link \SilverStripe\ORM\DataList} as the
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
Hey Josh. Thanks for the contribution. The correct branch is actually `4`. Basically think about your target `4.x` and remove the `.x` and that's your target. So a target of `4.3.x` will go against `4.3`. If there was no `4` branch then `master` is the fallback.

Semver doesn't allow us to introduce these APIs in patch releases unless we're specifically fixing a bug within the codebase. To be honest I'm not sure how that translates to dependant packages but my gut feel is that they fall out of scope.

Personally I think it's low impact enough to go in a patch release but I'm not willing to make a call on that on behalf of the whole core contributor team 🙂

JoshuaCarter:
Hey @ScopeyNZ, I'm confused. The only branches I see are `1`, `1.0`, `1.1`, `1.2`, `1.3`, `gh-pages`, and `master`, so I can't target `4.x` or `1.4` and thus assumed `master` would eventually become `1.4`, etc. What am I missing?

And who do I annoy to have this getter put in a patch release for `1.3`? It's *totally* a fix. That's my story and I'm sticking to it :P

JoshuaCarter:
@ScopeyNZ Ah, gotcha. I'll retarget the PR.

And as for who to bug?

JoshuaCarter:
Closing as replaced by https://github.com/silverstripe/silverstripe-admin/pull/820

PR #817, opened 2019-02-01
NEW Add transparent background state for Badge component

robbieaverill:
Allows React Badge component to be used for the designs in https://github.com/silverstripe/silverstripe-campaign-admin/issues/77 for campaign admin


clarkepaul:
Thanks for getting this PR so quick @robbieaverill. 

Mentioned in original issue... I presume that by default its square with slightly rounded corners, applying the class .badge-pill its rounded, and by applying .badge-transparent it removes the background (the three work against each other). 
Can I request that the transparent version has no padding left and right so that it can eventually be used to replace the status-flags throughout the CMS?
And is it going to be additional work to get that preview-able in Storybook?

robbieaverill:
Ok updated- now has `stateBadge` modifier which applies a transparent background with no left or right padding

ScopeyNZ:
Didn't check it out and play this time but the diff is mostly re-naming etc.

client/src/components/Badge/README.md
```diff
@@ -18,3 +20,4 @@ const props = {
  * `message` (string): The string to display in the badge.
  * `className` (string): Any extra classes to apply for the badge.
  * `inverted` (boolean): If the colours should be inverted.
+ * `stateBadge` (boolean): For use in the context of versioning states (use with inverted).
```
maxime-rainville: It's nice to know in what context this is used, but I would rather know what it actually does.

robbieaverill: Yeah fair call. I'm not super happy about the way it's ending up either - would love a suggestion!

Basically `inverted` flips the background and text colours, and `stateBadge` or whatever you want to call it will remove the left and right padding from it so it can be used for versioning state badges e.g. "LIVE", "DRAFT" etc.

maxime-rainville: What about `unboxed` or `no-box`?

clarkepaul: I try to avoid giving things names based on appearance as that might change over time, what happens if we designers add a background colour and padding back as we often want to evolve the design over time.

clarkepaul: I think we are better off keeping the name "state" as that shouldn't change over time were as its appearance will. I don't think we need to mention "Badge" again in "stateBadge" (maybe its required so it doesn't conflict with another class).

maxime-rainville: `state` is very ambiguous in the context of a react component because you might be talking to the component's internal state or a redux state.

`status` or `isStatusBadge` might be better. Or maybe we create a `<StateBade />` that wrap around the `<Badge/>` component.

robbieaverill: Perhaps that's the way to go here. `<VersionedStateBadge />` could live in silverstripe/versioned-admin

robbieaverill:
I'm going to move this to versioned-admin

client/src/components/Badge/README.md
maxime-rainville: Is versioned-admin a dependency of campaign and CMS? If not you might be introducing some additional coupling.

robbieaverill: Fair point... which would auto enable history viewer. Maybe I’ll just put it in admin but as a new component

PR #818, opened 2019-02-01
FIX Scrolling out of auto-selected edit mode not switches back to split mode

robbieaverill:
Previously resizing the window in split mode from a size which supported split mode into one that does not would switch to edit mode, but it would not switch back to split mode when resizing back up again. This patch fixes that by storing an extra "state is locked" value to ensure that when you manually select "Edit" or "Preview" mode, it will not switch back to split mode when you resize, but also ensures that it can switch between split and edit on resize automatically otherwise.

Issue: https://github.com/silverstripe/silverstripe-campaign-admin/issues/87


maxime-rainville:
Maybe I'm misreading this, but I've installed the patch on the kitchen sink and I don't see any difference.

When I resize my split-mod campaign view, nothing happens ... it stick in split mod all the way.

When I reszie my split-mod CMS view, it switch to edit below 1500 and doesn't switch back afterwards.

robbieaverill:
The issue is definitely visible in campaign admin - it's the "mobile view" toggle which kicks in and hides the preview panel, you need to click on the item in mobile view to open the preview

maxime-rainville:
It doesn't work in Firefox, but it does work in Chrome.

robbieaverill:
Steps to reproduce:

* Create a new campaign
* Open the campaign
* At this point you should see a split view of items and a "Getting started" splash message
* Resize your window down to a mobile viewport
* At this point you should only see a list of the campaign items
* Resize your window back to a desktop viewport

At this point you will still only see a list of the campaign items. This is the bug. It should go back to split mode.

Tested on:

* IE 11 (Windows + Mac)
* Chrome 73 (Windows + Mac)
* Safari 11 (Mac)

Not working (window resize monitoring) on Firefox 66 on Windows or Mac.

robbieaverill:
Looks like you've added a polyfill to fix this elsewhere. [The issue you commented on](https://github.com/FezVrasta/react-resize-aware/issues/15) upstream says it's fixed in v2.7.2, so I'll try removing your polyfill and installing the latest version instead.

robbieaverill:
PRs all merged 🎉

PR #821, opened 2019-02-04
Remove Entwine inspector (fixes #768)

### LINKED ISSUE #768: Hot keys for jquery.entwine.inspector are the same as the emoji shortcuts on OSX
Probably not the highest of importance, file is added through [here](https://github.com/silverstripe/silverstripe-admin/blob/86282b10235358c019355fa5984ef681b2a3eb7d/code/LeftAndMain.php#L697). 
The short cut is `control + cmd + space`

Would be good to change to something that's not as likely to be taken 👍

Reproduce by going into any admin page and entering the mentioned keys

# Related PRs
* [ ] #821

chillu:
It never was of great help, and it's quite a hidden feature:
Only available via a keyboard shortcut, and not even
documented in https://github.com/hafriedlander/jquery.entwine.
So I doubt many devs will miss it.

The keyboard shortcut conflicts with emojis on OSX.

Also removing leaktools.js, there's much better ways
to detect javascript memory leaks in browsers these days.


chillu:
Nah, that's part of Entwine - we can do that on the joyous day when we get rid of `entwine.js` ;)

ScopeyNZ:
Looks like there's some reference to leak tools in a bundle somewhere from the CI build.

chillu:
Sweet, removed the reference from `webpack.config.js` and rebuilt. It also updated a patch release of bootstrap (no changes?), and removed `default-require-extensions` as a dependency from `yarn.lock`. No idea why that happened, I've run `yarn && yarn run build`. Poked around the CMS a bit, seemed benign.

chillu:
I really don't understand why the builds are still failing in https://travis-ci.org/silverstripe/silverstripe-admin/jobs/500150383. I'm running the local setup with `yarn && yarn run build`. When I call `git diff-files --quiet -w --relative=client`, the return code is `0` (based on `echo $?`). Just kicked them off again, let's see what happens.

```
node -v
v10.15.1
yarn -v
1.3.2
```

maxime-rainville:
@ingo Travis implicitly merges your PR branch with the target branch when running a build for a PR. Occasionally, you end up in weird scenario where the dist files don't have a merge conflicts, so the build goes ahead. However, when it checks dist files against the ones generated by travis's `yarn build` there's a difference so the build fails. (It took me a couple hours a few months ago to figure that one out)

I just rebased your PR against `1`. That should resolve the issue :crossed_fingers:

maxime-rainville:
And I forgot to rebuild the files locally, thereby replicating the exact same issue locally. Well done me.

maxime-rainville:
Sorry ... I meant to hit the approve button before the merged button.

chillu:
Oh weird, I thought I had rebased on `1`. Thanks for the explanation! 

> Travis CI builds a pull request when it is first opened, and whenever commits are added to the pull request. Rather than build the commits that have been pushed to the branch the pull request is from, we build the merge between the source branch and the upstream branch.

https://docs.travis-ci.com/user/pull-requests/

That's a bit of a footgun for contributions, right? Can we improve the travis.yml error output to show a message when failing the `git diff` step? Something like "you might need to rebase".

maxime-rainville:
It's not just a question of rebasing. You need to rebase and rebuild.

Normally there's going to be a conflict when rebasing, which will force you to rebuild to complete your rebase. However, if there's no merge conflict, it's easy to forget to rebuild.

chillu:
https://github.com/silverstripe/silverstripe-admin/issues/844

PR #822, opened 2019-02-08
FIX Deferring events on mount and ensuring events are triggered on change in HtmlEditorField

ScopeyNZ:
In some cases TinyMCE was beating redux when initialising from redux content. Redux would initialise and then state would change (effectively) immediately, but TinyMCE would usually mount between that. This adds a defer to the event that triggers TinyMCE to remount

This also adds some better event dispatching when changes happen to attempt to poke the change tracker


ScopeyNZ:
> when you insert an external URL link onto some text, it causes the page level change tracker to detect changes and switch to a dirty state

Isn't this the correct behaviour? The content has changed - you've added a link... TinyMCE on a page level does update the changetracker in a similar way.

PR #828, opened 2019-02-15
Adding TS support to admin

maxime-rainville:
Experimental hack day project that adds support for typescript.

* Need to get the prod build script working.
* Need to test with storybook.

Linting contingent on https://github.com/silverstripe/eslint-config/pull/6


client/src/boot/BootRoutes.tsx
```diff
@@ -23,10 +26,7 @@ class BootRoutes {
    * @param {Object} store Redux store
    * @param {Object} client The Apollo client
    */
-  constructor(store, client) {
-    this.store = store;
-    this.client = client;
-
+  constructor(private store: Store, private client: ApolloClient<any>) {
```
maxime-rainville: Typescript allows you to define class properties directly in the constructor.

client/src/boot/BootRoutes.tsx
```diff
@@ -23,24 +26,18 @@ class BootRoutes {
...
     // pageRouter must be initialised, regardless of whether we are
     // using page.js routing for this request.
     const base = Config.get('absoluteBaseUrl');
     pageRouter.setAbsoluteBase(base);
-
-    this.handleBeforeRoute = this.handleBeforeRoute.bind(this);
```
maxime-rainville: You don't need to bind your class method to this is TS.

client/src/boot/BootRoutes.tsx
```diff
@@ -23,24 +26,18 @@ class BootRoutes {
...
-    this.handleBeforeUnload = this.handleBeforeUnload.bind(this);
   }
 
-  setStore(store) {
+  public setStore(store: Store) {
```
maxime-rainville: Your class method in TS can be defined as private, public or protected. This allows you to make clear what is internal logic to your class and what is meant for public consumption by other classes.

client/src/boot/BootRoutes.tsx
```diff
@@ -166,13 +162,13 @@ class BootRoutes {
     // Legacy support – make sure we don't overwrite the `onbeforeunload`
     // handler in `LeftAndMain.EditForm.js` but run it after the new handler.
     const currBeforeUnload = window.onbeforeunload;
-    window.onbeforeunload = () => {
+    window.onbeforeunload = (event) => {
```
maxime-rainville: This is a good example of a somewhat annoying and somewhat useful behaviour. Typescript is really good at auto detecting the types of your variables. In this case, TS knows what the default signature of `window.onbeforeunload` is and guess that `currBeforeUnload` will have the same signature. A bit lower, TS was throwing an error because we we calling a method with invalid parameters.

In this context this error is not really applicable because of he nature of the method, but in other scenarios it might identify an actual error.

client/src/boot/BootRoutes.tsx
```diff
@@ -187,15 +183,15 @@ class BootRoutes {
...
     const state = this.store.getState();
     const forms = state.unsavedForms || [];
     const schemas = state.form.formSchemas;
 
     const changedForms = forms.filter((form) => {
-      const schema = Object.values(schemas).find(item => item.name === form.name);
+      const schema = Object.keys(schemas).find(item => item['name'] === form.name);
 
-      const notify = schema && schema.state && schema.state.notifyUnsavedChanges;
+      const notify = schema && schema['state'] && schema['state'].notifyUnsavedChanges;
```
maxime-rainville: This is a somewhat annoying aspect of TS. Sometimes you are pass an object with an undefined structure. If you want to access some of those object's properties, you have to either cast the object has any, or use a PHP-style square bracket syntax.

client/src/boot/apollo/getGraphqlFragments.ts
```diff
@@ -30,12 +30,11 @@ const getGraphqlFragments = async (baseUrl, preferStatic = true) => {
   }
   const [primaryURL, fallbackURL] = urls;
 
-  const fetchConfig = {
+  const fetchConfig: RequestInit = {
     method: 'GET',
     headers: {
       'Content-Type': 'application/json'
     },
-    uri: `${baseUrl}`,
```
maxime-rainville: This is a good example of a potential bug that TS can highlight. In this case `uri` is a useless value. The `fetch` method will just ignore it, but TS will throw an error. You could update `uri` and scratch your head to understand why you are not getting the write info. TS saves your from this potential problem.

maxime-rainville:
Got a better one comming

PR #833, opened 2019-02-22
FIX Move first table body row border-top to last header row border-bottom

robbieaverill:
If table rows are dynamically hidden with JavaScript, this ensures that the first row will always have a top border because we put the border-bottom on the last header row instead of a border-top on the first body row

Fixes https://github.com/silverstripe/silverstripe-sitewidecontent-report/issues/35


robbieaverill:
All table rows have a bottom border. The first row has a top border. This PR changes it so that the last header row has a bottom border, which fixes the issue of hiding the first row dynamically, but this doesn't fix cases like the CwpStatsReport where there's no table header at all, which now doesn't have a top border on the first row. Will think about this...

robbieaverill:
Yep closing, better the way it is.

PR #835, opened 2019-02-25
FIX Update font-icon hover for site tree 'Add new page here'

sachajudd:
Relies on: https://github.com/silverstripe/silverstripe-cms/pull/2391


client/src/styles/legacy/_tree.scss
```diff
@@ -312,6 +312,9 @@
           background-image: linear-gradient(to bottom, #3875d7 20%, #2a62bc 90%);
           color: #FFF;
           border: none;
+          .jstree-pageicon {
```
ScopeyNZ: I assume this is a specificity override as the colour is defined a couple of lines up. Do you want to add a comment about that?

PR #845, opened 2019-03-13
Upgrade merge and bootstrap JS dependencies

maxime-rainville:
We're just bumping a few dependencies.


robbieaverill:
@maxime-rainville thoughts on updating the constraint?

maxime-rainville:
Created https://github.com/silverstripe/silverstripe-admin/pull/849 to explicitly increment bootstrap dep.

PR #861, opened 2019-04-17
NEW Add VersionedBadge for displaying versioning states in the CMS

robbieaverill:
Replacement PR for https://github.com/silverstripe/silverstripe-admin/pull/817 which adds a new wrapper component for versioning state badges, instead of adding more props to the Badge component.

Originally raised in https://github.com/silverstripe/silverstripe-campaign-admin/issues/77 since we're displaying the wrong badges for versioning in campaign admin.

# Parent issue
* https://github.com/silverstripe/silverstripe-assets/issues/286

# Related PRs
* https://github.com/silverstripe/webpack-config/pull/37
* https://github.com/silverstripe/silverstripe-asset-admin/pull/980


maxime-rainville:
I've rebased, remove the message (since this can be inferred from the status), added an archived status and functionalise the component.

Here's what the pattern lib looks like.
image

@silverstripeux Can you let us know if you are happy with those badges? For context, this component will eventually be use everywhere we display a stage label.

sachajudd:
I think you might be using the incorrect variable sheet. I see you're using _bootstrap-variables rather than using the "Extra state colours" in _variables.scss. Which include the correct state-draft and state-modified colours. Not sure if this is the same with the live and archived colours but you might want to check incase 🙂

client/src/components/VersionedBadge/VersionedBadge.scss
```diff
@@ -0,0 +1,13 @@
+.versioned-badge {
+  padding-left: 0;
+  padding-right: 0;
+
+  &--draft,
+  &--modified {
+    color: $orange;
```
robbieaverill: @sachajudd is this the bit you were talking about? We have `$state-draft` and `$state-modified`, do we need to add `$state-live` and `$state-archived` as well (with their respective background equivalents too)?

maxime-rainville:
> The only thing with removing the message and inferring it from the status is we'll need to still allow people to add their own states I guess, Fluent for example adds its own to the site tree

Shouldn't other modules be using the Badge component directly instead?

client/src/components/VersionedBadge/VersionedBadge.scss
sachajudd: Yeah sounds good!

sachajudd:
Just a side note from Max's screenshot the type face almost looks a little off and the letter spacing. Not sure what's going on there (would be good to compare) but I understand once it's in the pattern library we can tweak it.

Demo:
image

maxime-rainville:
> E.g the 'D' in MODIFIED looks quite spaced in the first screenshot. What browser were you using

I've got the same gap in the CMS tree. It might just be that I'm looking at it on Ubuntu and you are looking at it on a Mac. Mine is falling back to Arial because I don't have Helvetica Neue or Helvetica.

maxime-rainville:
@sachajudd Are we happy with the live and archived colour in the screen shot? 
* Live uses #008a00
* Archived default to #43536d which is the default for the Badge component

I'll create them matching `$state-live` and `$state-archived` variables in `_variables.scss`

client/src/components/VersionedBadge/VersionedBadge.scss
maxime-rainville: @sachajudd Are we happy with the live and archived colour in the screen shot? 
* Live uses #008a00
* Archived default to #43536d which is the default for the Badge component

I'll create them matching `$state-live` and `$state-archived` variables in `_variables.scss`

sachajudd: Yeah so by the looks of it we can use these variables for the same colours.
$state-archived: $gray-700;
$state-live: $green;

maxime-rainville: Done.

client/src/components/VersionedBadge/VersionedBadge.scss
```diff
@@ -0,0 +1,16 @@
...
+  &--draft {
+    color: $state-draft;
+  }
+
+  &--modified {
+    color: $state-modified;
+  }
+
+  &--live {
+    color: $green;
```
robbieaverill: Did you forget to push something? I'm expecting this to be `$state-live` or `$state-published` or something, with that and the archived variable being added to variables.scss

client/src/components/VersionedBadge/VersionedBadge.scss
sachajudd: @maxime-rainville have you pushed up your changes I can't see the variable $state-live being added or used in VersionedBadge.scss

sachajudd: There’s also a badge.scss file in admin/client/components/badge that has status-archived styles which could be replaced to use the new variable you've created ($status-archived).

client/src/components/VersionedBadge/VersionedBadge.scss
maxime-rainville: How embarrassing ... should be fixed now.

robbieaverill:
We need to follow up on @sachajudd's comments first, and remove the use of the old variable in Badge.scss (doing this now)

robbieaverill:
@sachajudd this PR makes `$state-archived: $gray-700`. Originally `.status-archived { color: $text-muted; }` evaluates `$text-muted` to `$gray-600`. Which should be used?

sachajudd:
Yeah lets leave this to a separate PR as I've noticed all of the colours are a bit out. Stick with $gray-600 for now and I'll take a look next week. Happy for the PR to be merged 👍

PR #865, opened 2019-04-24
BUGFIX cms container can scroll beyond the height of the window

micmania1:
I tried making this PR against the 1 branch but I was running into npm version issues.

Running `yarn` with version 6 told me I needed version 10. Running yarn with version 10 told me I needed version 6.


blueo:
do you know if core has move to version 10 of node yet?

robbieaverill:
@blueo we're using Node 6 from SilverStripe 4.0 to 4.3, and Node 10 from SilverStripe 4.4 and onwards

micmania1:
@blueo The version we're using (1.3) is node 6, the latest (1 & 1.5) is node 10. 

I can build fine on the 1.3 branch with node 6 (although `yarn css` didn't seem to work).

On the 1 branch, if I use node 6 it tells me I need node 10. If I use node 10 it tells me I need node 6. Its as though one of its dependent packages requires a different node version.

I spent way too long trying to figure it out to do a simple 1 line fix so just made the PR against 1.3 :man_shrugging:

robbieaverill:
By the way, we can merge this fix up once it's merged

robbieaverill:
Out of interest @micmania1 do you have a way to reproduce this bug?

micmania1:
@robbieaverill install `dnadesign/silverstripe-elemental-list` and look at the "Layout" view in the CMS. I can't be certain that something in the module isn't causing it (or custom code for that matter), however the scroll should still be limited anyway as its full screen view.

Cheddam:
@micmania1 Sorry this got buried! I've tried replicating the layout issue (by disabling the manual patch on the site it was originally affecting), and I can't seem to make it occur. I believe we shipped some related fixes late last year, so this may have been resolved by one of those. Additionally, applying `overflow: hidden` means that if there were UI elements below the fold that were causing the scroll context, they'd no longer be accessible at all, which is arguably a worse UX outcome.

On that basis, I'm going to close this for now - if you or anyone else have a clear method to reproduce this issue in the current version of CMS, please let me know and we can take another look at getting this merged.

PR #874, opened 2019-05-06
Behat test complies with accessible buttons

dnsl48:
Broken test: https://travis-ci.org/silverstripe/silverstripe-admin/jobs/527589331
This patch follows up the semantic change made in this patch (buttons became links) - https://github.com/silverstripe/silverstripe-framework/commit/5337e6d04847c66d0a293e9c6c53a58d97aea3a1

### DIFF
--- a/tests/behat/features/manage-users.feature
+++ b/tests/behat/features/manage-users.feature
@@ -99,13 +99,13 @@ Feature: Manage users
     And I should see "staffmember@example.org" in the "#Root_Users" element
 
     When I click "othermember@example.org" in the "#Root_Users" element
-    And I press the "Form_ItemEditForm_action_doNext" button
+    And I follow "Go to next record"
     Then the "Email" field should contain "staffmember@example.org"
-    When I press the "Form_ItemEditForm_action_doPrevious" button
-    Then the "Email" field should contain "othermember@example.org"
 
+    When I follow "Go to previous record"
+    Then the "Email" field should contain "othermember@example.org"
     When I fill in "FirstName" with "Staff Renamed"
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
It feels like an issue that these tests fail here rather than on framework. Perhaps it makes sense to run the behat tests (or a subset of them) on framework too?

How does `follow` differ from `click`? Semantically - as they still appear as buttons - it'd be nicer if we could use "click" still.

dnsl48:
> Perhaps it makes sense to run the behat tests (or a subset of them) on framework too?

I'm thinking about an RFC to actually trigger `silverstripe/installer` build on framework repository, so we can catch regressions in other modules. Currently, we don't run behat on framework because it does not have its own stand-alone UI to test (well, with the exception for `/dev` perhaps).

> How does follow differ from click?

`follow` follows links, `press ... button` looks for buttons. Here we have a links styled as buttons, however semantically they are still links.

dnsl48:
Oh, I think I misread. You're suggesting to use something with `click` rather than `follow` here to keep test semantics?

ScopeyNZ:
> Oh, I think I misread. You're suggesting to use something with `click` rather than `follow` here to keep test semantics?

Yeah. I know that follow is for links and click is for buttons - I was just wondering if the step "under the hood" was much different. It makes sense for someone who is writing the test to write "and I click the 'next' button" and for it to work 😅 . If it's going to require changes to the step then I don't think it's worth the time right now though

dnsl48:
Yeah, sorry, I cannot find anything suiting better. We might need to patch behat-extension to add new context methods for that, which is out of scope for this one (as it's just a follow up on already merged patch).

PR #889, opened 2019-06-01
NEW: Allow export of injected GraphQL AST alongside HOC

unclecheese:
Required for: https://github.com/silverstripe/silverstripe-asset-admin/pull/951

This PR makes a slight change to the `apolloGraphQLContainer` where `getFactory` returns  a proxy object that allows the user to consume either the HOC or the AST (both injected).

Before:
```
const graphqlHOC = injector.query(myKey, myContext);
const MyComponent = graphQLHOC(Component);
```

After
```
const graphqlProxy = injector.query(myKey, myContext);
const graphQLHOC = graphQLProxy.getApolloHOC();
const MyComponent = graphQLHOC(Component);
```

Additionally, the proxy provides `getGraphqlAST()`, which is needed in any context where apollo needs to read or write from a known AST (e.g. an optimistic update). This way, the user can get an AST in its fully injected form.

```
store.writeQuery({ query: myProxy.getGraphQLAST(), variables });
```

## Todo:

- [x] Add some tests for `ApolloGraphQLProxy`
- [ ] ~~Separate PR for the 1.4 branch, using node 10~~


unclecheese:
Have added tests. For the separate build, we can do that on merge up. There will be conflicts anyway.

ScopeyNZ:
Isn't this a breaking change for existing `injector.query` usage? Can we provide this as an alternate API or as an alternate return type based on a function argument?

PR #890, opened 2019-06-05
FIX Update modal designs to match design pattern library

ScopeyNZ:
This update alters our modal styling to match what is provided by the pattern library that the design team use (eg. https://projects.invisionapp.com/share/3PNSKZQYBJZ#/screens/350729097).

This is the new look:

image

/cc @clarkepaul


ScopeyNZ:
FYI the current state:

image

clarkepaul:
Sound :), thanks for picking this up @ScopeyNZ. We should do one component each week hint hint :)

sachajudd:
Could you please back port this into 4.3 @ScopeyNZ ?

PR #895, opened 2019-06-17
NEW Add ability to inject JSX into strings

ScopeyNZ:
This uses Reacts ability to render arrays like a concatenated string


client/src/i18n.js
```diff
@@ -169,6 +169,30 @@ class i18n {
...
+   */
+  injectJSX(s, map) {
+    // Split the string by the regex for matches
+    const parts = s.split(this.matchRegex);
+
+    // The capturing group is put in where matches were found - this means all matches are at odd
+    // indexes in the "parts" array. We can replace those items (or leave the "{match}" syntax.
+    for (let i = 1, length = parts.length; i < length; i += 2) {
+      parts[i] = map[parts[i]] || `{${parts[i]}}`;
+    }
```
unclecheese: Can we do this in an immutable way? I feel like this is screaming to be a `reduce`.

ScopeyNZ: It doesn't need to be immutable because the array is generated above. I don't think we really need that complexity?

unclecheese: I guess it's a matter of preference, but I think aiming to use functional programming whenever possible is a good thing.

ScopeyNZ: I considered the reduce tbh. The awkwardness was that I only care about every `2n+1`th record though, there's not really any easy way to reduce like that. I can alter it if you feel strongly - I'm not too fussed.

ScopeyNZ:
I will add a unit test. There are currently no unit tests for i18n so I'll add a bit of new stuff for other methods. Not going to have time to do this within this week though.

client/src/i18n.js
unclecheese: All good. 👍

ScopeyNZ:
I'm not convinced this is a good idea any longer. If someone wants to take over I'm happy to reopen though :)

PR #896, opened 2019-06-17
NEW Hide iframe while loading in preview component

ScopeyNZ:
This is an arguable "fix". This allows components that render a preview to show an underlay or loading component while the iframe is loading.


client/src/components/Preview/Preview.js
```diff
@@ -98,7 +114,12 @@ class Preview extends Component {
     }
 
     // Show iframe preview
-    return <iframe className="flexbox-area-grow preview__iframe" src={previewUrl} />;
+    return (<iframe
+      style={{ visibility: this.state.frameLoaded ? 'visible' : 'hidden' }}
+      className="flexbox-area-grow preview__iframe"
+      src={previewUrl}
+      onLoad={() => this.setState({ frameLoaded: true })}
```
robbieaverill: Do you want to make a method like `setFrameLoaded(loaded = true)` and use that here and line 29?

robbieaverill:
I agree it's arguable, maybe retarget as `1` to be safe

client/src/components/Preview/Preview.js
ScopeyNZ: Done

client/src/components/Preview/Preview.js
```diff
@@ -12,9 +12,33 @@ class Preview extends Component {
...
+
+    if (prevProps.previewUrl !== this.props.previewUrl) {
+      this.setFrameLoaded(false);
+    }
+  }
+
+  /**
+   * Update state to indicate that the preview page has completed loading
+   *
+   * @param loaded
```
robbieaverill: Needs a type

client/src/components/Preview/Preview.js
```diff
@@ -12,9 +12,33 @@ class Preview extends Component {
...
+   */
+  setFrameLoaded(loaded = true) {
```
robbieaverill: Do you need to bind this to the instance?

PR #901, opened 2019-06-26
FIX/NEW: Stories are now fetched dynamically from related modules

### LINKED ISSUE #785: The pattern library should auto discover components to include, rather than hard coding them
Currently the pattern library [has hard coded modules that it looks in](https://github.com/silverstripe/silverstripe-admin/blob/1.3.0/.storybook/config.js#L5-L11) to determine which React components to include.

Ideally it would be able to look through all modules for compatible filenames and include all of them.

For example, we have reusable components in some of the other supported modules such as tagfield, sharedraftcontent, CKAN registry (in development), versioned-admin (maybe) etc. We can't add these to the pattern library without introducing coupling into this module right now, but in a perfect world they'd be included if you had them installed.

ScopeyNZ:
Modules that stories need to be gathered from should have a .ss-storybook.js configuration file in their root directory

I'll add some documentation in a framework PR, and then I'll add the `.ss-storybook.js` to asset-admin too.

I can rebase this against the minor branch if required but as this is dev changes I thought it'd be nice to have this storybook enhancement in place for fixing bugs or developing features in modules based on admin@1.4

Fixes #785 

### Related PRs:

- silverstripe/silverstripe-asset-admin#963

### DIFF
--- a/.storybook/lib/storyTeller.js
+++ b/.storybook/lib/storyTeller.js
@@ -7,21 +7,15 @@ const exec = promisify(require('child_process').exec);
 
 const configName = '.ss-storybook.js';
 
-/**
- * Locates modules in the given root (matching the base path and any composer style vendor folder
- * within that root) that define a .ss-storybook.js config file.
- *
- * Note that this function will also generate a .stories.js file that declares module roots in a way
- * that can be static analyzed for webpack to build a valid bundle for storybook. This script should
- * run before or during the webpack bundle for storybook.
- *
- * Returns an array of module configuration
- *
- * @param {string} root Root location to find stories
- * @param {bool} generate
- * @return {Array<Object>}
- */
 module.exports = {
+  /**
+   * "Collect" modules within the vendor folder OR the project root to show stories from. This will
+   * find any `.ss-storybook.js` file that provides configuration for locating stories (and their
+   * dependencies)
+   *
+   * @param {string} root
+   * @return {Array<Object>} Discovered configuration
+   */
   collectStoryRoots(root) {
     const modules = glob.sync(
       [`${root}/${configName}`, `${root}/vendor/**/${configName}`],
@@ -51,6 +45,13 @@ module.exports = {
     return output;
   },
 
+  /**
+   * Trigger a `yarn` in modules that have been "collected" but there doesn't appear to be a
+   * node_modules folder. This is intended for use with `collectStoryRoots`
+   *
+   * @param {Array<Object>} storyRoots
+   * @return {Promise<void>}
+   */
   async assertYarn(storyRoots) {
     const promises = [];
 
@@ -66,9 +67,15 @@ module.exports = {
     await Promise.all(promises);
   },
 
+  /**
+   * Generate a JavaScript file to the given output location that declares what contexts should be
+   * included when storybook builds bundles with webpack. This is generated as webpack cannot handle
+   * compiling dynamic routes - it will need to statically analyze file paths.
+   *
+   * @param {Array<Object>} storyRoots
+   * @param {string} output The output filename
+   */
   generateStoriesLoader(storyRoots, output = `${__dirname}/../.stories.js`) {
-    const realOutput = fs.realpathSync(output);
-
     const script = `export default [\n${storyRoots.reduce((acc, { src, fileMatcher }) => {
       const matcher = fileMatcher instanceof RegExp
         ? fileMatcher.toString()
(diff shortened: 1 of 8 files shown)


robbieaverill:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/785?

ScopeyNZ:
> Fixes #785?

With one slight caveat - modules still need to opt-in with a file in the root of the module.

robbieaverill:
Good enough =) nice one

.storybook/lib/storyTeller.js
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * Note that this function will also generate a .stories.js file that declares module roots in a way
+ * that can be static analyzed for webpack to build a valid bundle for storybook. This script should
+ * run before or during the webpack bundle for storybook.
+ *
+ * Returns an array of module configuration
+ *
+ * @param {string} root Root location to find stories
+ * @param {bool} generate
+ * @reutn {Array<Object>}
```
robbieaverill:
```suggestion
 * @return {Array<Object>}
```

.travis.yml
```diff
@@ -95,29 +95,10 @@ after_failure:
...
-
-# Get CMS
-  - git clone https://github.com/silverstripe/silverstripe-cms.git
-  - cd silverstripe-cms
-  - yarn install
-  - cd ..
-  - mv silverstripe-cms ../cms
-
+# Use installer to fill in any possible dependencies
+  - composer require silverstripe/installer:4.4.x-dev
```
robbieaverill: There's another part further up that uses `4.4.x-dev` - it might be worth using a variable for both of these and defining it at the top so we don't miss it, since this will have to be changed on each new minor release

.storybook/lib/storyTeller.js
```diff
@@ -0,0 +1,101 @@
...
+  },
+
+  getDefaultRoot() {
+    // Check if a `vendor` folder exists in the project root.
+    let projectRoot = path.resolve(`${__dirname}/../../`);
+
+    try {
+      fs.accessSync(`${projectRoot}/vendor/`);
+    } catch (err) {
+      projectRoot = path.resolve(`${projectRoot}/../../../`);
```
Cheddam: Should this be verifying that the alternate path is going to work? What context(s) do you expect a further level to cover?

package.json
```diff
@@ -20,7 +20,9 @@
     "coverage": "jest --coverage",
     "lint": "eslint client/src && sass-lint -v",
     "js-fix": "eslint client/src --fix",
-    "pattern-lib": "yarn && start-storybook -p 6006"
+    "pattern-lib": "yarn && node .storybook/scripts/runStoryTeller.js && start-storybook -p 6006",
+    "storybook": "yarn pattern-lib",
+    "static-storybook": "yarn && node .storybook/scripts/runStoryTeller.js && build-storybook"
```
Cheddam: Would this make more sense as `pattern-lib` and `static-pattern-lib`, to avoid duplication / inconsistency? It seems like `pattern-lib` in this context is being kept for backwards compatibility, and `storybook` is the new standard, but to me the underlying library shouldn't dictate the action.

.storybook/lib/storyTeller.js
ScopeyNZ: Possibly. Is either expects running this script with a vendor folder in the admin root (the first check) or that `admin` is itself an installed composer dependency - meaning the root is three folders up. I'm not sure if there's a valid alternate situation, but it could probably do with some error handling to assert it's correct.

package.json
ScopeyNZ: Yeah adding `storybook` was sort of my bias tbh. I kept writing `yarn storybook` and being mad when it didn't work. Happy to change it.

.storybook/lib/storyTeller.js
Cheddam: Okay, that makes sense - I think at least document what the two scenarios are for clarity, and then maybe consider breaking if a vendor folder still can't be found using the fallback.

.travis.yml
ScopeyNZ: I suppose I should do this. I didn't really consider the other `4.4.x-dev` requirement. I guess the other one should be the minimum compatible version where this one should be all the latest features though? 🤔

.storybook/lib/storyTeller.js
ScopeyNZ: I've overhauled the docs on this file

ScopeyNZ:
I should re-target this at `1`.

.travis.yml
ScopeyNZ: I've done this now.

package.json
ScopeyNZ: In the interest of removing any blockers I've removed the addition of `storybook` and changed to `static-pattern-lib`.

ScopeyNZ:
Ok that's retargetted now. I've addressed the feedback too. Hopefully we can get this merged for modules to use

PR #906, opened 2019-07-10
Update _style.scss

alessandromarotta:
Please, remove the property "height:100%;" from the class .cms-content-header-info because it causes problems at sub tabs (ex. Admin/Security -> Users, Groups, Roles tabs)


bergice:
@alessandromarotta  What is the problem you're seeing?

alessandromarotta:
hi @bergice, look the cms from the mobile, this is with the height: 100% in the class "cms-content-header-info"

image

and this is without: 
image

I fixed it with a Requirements:CSS wich cointains 

`.cms-content-header-info {
    height: initial !important;
}`

kinglozzer:
Nice coincidence - I just stumbled upon this exact issue today. It actually affects all screen sizes, it’s when the header (title + tab list) wraps. E.g:

Screenshot 2020-05-01 at 15 11 40

Removing the `height: 100%` does indeed fix it:
Screenshot 2020-05-01 at 15 11 32

Given that the parent element of `.cms-content-header-info` is `display: flex`, we should be able to safely remove this. We could add `.cms-content-header { align-items: stretch; }`, but the default value of `align-items` is equivalent to `stretch` in our case anyway.

kinglozzer:
It’s fine in Chrome, Safari, Firefox and IE11. I think we can take this in, it just needs:

- Rebasing against 1.5 (preferably)
- Dist needs building

dhensby:
Want to take it on @kinglozzer ?

michalkleiner:
Related issue that this can possibly fix - https://github.com/silverstripe/silverstripe-cms/issues/2405

sachajudd:
@kinglozzer, I've rebased to 1.5 and rebuilt the dist files if you want to take another look before merge.

kinglozzer:
@sachajudd legend!

PR #907, opened 2019-07-10
Update Menu.scss

alessandromarotta:
Sidenav animation performance improvement


bergice:
@alessandromarotta there are a couple of problems with this new look:
* When the menu is already open and you change the resolution, the menu transitions in even though it already was open
* The title of the site gets blocked by the menu button on the top left.

It would be good to get some PUX feedback on this too, @sachajudd.

### For comparison:

Before: 
image

After: 
image

client/src/components/Menu/Menu.scss
```diff
@@ -249,19 +247,18 @@
...
 }
 
-.cms-menu + .cms-menu-mobile-overlay {
-  display: none;
-}
-
 .cms-menu.cms-menu--open + .cms-menu-mobile-overlay {
   @include media-breakpoint-down(sm) {
-    display: block;
+    opacity: 1;transform: translateX(0);
```
sachajudd, line 261: This should be split onto another line, you can run yarn build and this will also run the linter indicating these errors/warnings. As @bergice mentioned you'll need to update your PR to include the rebuilt bundle files.

image

client/src/components/Menu/Menu.scss
```diff
@@ -18,14 +18,12 @@
...
-
+    left: -260px;
+    width: 250px;
+    box-shadow: 1px 0 10px rgba(0, 0, 0, 0.2);
+    transition: transform $transition-speed-fast;
     &--open {
-      left: 0;
-      height: 100% !important; // sass-lint:disable-line no-important
-      box-shadow: 1px 0 10px rgba(0, 0, 0, 0.2);
+      transform: translateX(260px) !important;
```
sachajudd, line 26: I can see there is the use of !important in the old version of the code but we try to avoid using it where possible. Can this be removed and achieved by increasing the specificity of your selector?

client/src/components/Menu/Menu.scss
```diff
@@ -249,19 +247,18 @@
 }
 
 .cms-menu-mobile-overlay {
+  display: block;
   position: absolute;
-  background-color: rgba(0, 0, 0, 0.05);
+  transition: opacity .3s;
+  transform: translateX(-100%);
+  background-color: rgba(0, 0, 0, 0.25);
```
sachajudd, line 254: Just wondering the reasoning behind the tweaked background-color?

client/src/components/Menu/Menu.scss
alessandromarotta: Yes, of course, this !important can be removed 
```suggestion
      transform: translateX(260px);
```

client/src/components/Menu/Menu.scss
alessandromarotta: because an alpha value of 0.05 is to small, the overlay is almost invisible, so 0.25 could be a good value

client/src/components/Menu/Menu.scss
alessandromarotta: I did some updates, I think that now it's better.

client/src/components/Menu/Menu.scss
alessandromarotta: Sure, I was wrong the past year, now fixed with the commit [88c4ba7](https://github.com/alessandromarotta/silverstripe-admin/commit/88c4ba70f6bf2dcaf77ee18c70e3f2ca1fb0e30f)

sachajudd: image
Looks like you still have one error which is why the builds are failing. 
Once you fix this and run yarn build you'll see the `bundle.css` file which you'll need to push up too. You can see more about how to build the bundle files here: https://docs.silverstripe.org/en/4/contributing/build_tooling/#build-commands

sachajudd:
### UX review
After building your files there are a few issues I ran into:

Bug: Looks like the overlay is always showing on any screen size whether the menu is open or not:
image
image
image

Bug: In mobile this option has reappeared and should be removed: 
image

### Other considerations
These could be split into separate issues and would be nice to fix but are not blockers for this PR.

I can see the icons slowly move to the right as I go from desktop to mobile. Looks like it's coming from how it originally functioned. Just wondering if the menu can now stay the same width as desktop at any screen size if these changes are merged.

In tablet the User help menu can now function the same way as it would on desktop:
PR:
image
Looks a bit broken now on open:
image
Desktop version (should look like this):
image
image

> When the menu is already open and you change the resolution, the menu transitions in even though it already was open

Yeah this seems to happen if the collapse panel is open or not on desktop. Not currently a blocker for the PR to go forward.
(Collapse panel):
image

alessandromarotta:
Hi! this is very strange: the overlay should appear only when the menu panel has the "cms-menu--open" class applied and only when resolution is max-width 768px. On my blank installation test this doesn't happen (but I haven't tried to build with Yarn, I'm patching directly the final stylesheet on my extension)

Have you tried my last commit 88c4ba7?
I will try to build with Yarn this night but I see that I need too to add a transition state in the event listener that add the "cms-menu--open" class (because the "transition: transform" property should be active only after the user click on the Hamburger Menu icon)

sachajudd:
Yes, I am testing with you latest commit 🙂yeah I'd suggest building with yarn as this is the correct build tool to use. I'll wait until you've pushed up those changes with the build files and take another look.

dhensby:
@sachajudd I've rebased and re-built this for you

sachajudd:
Cheers, missed this initially @dhensby, I'll take another look next week.

sachajudd:
@alessandromarotta do you still have time to work on this PR? I've left some peer review https://github.com/silverstripe/silverstripe-admin/pull/907#issuecomment-623804112 that I think should be addressed before this can be merged.

alessandromarotta:
> @alessandromarotta do you still have time to work on this PR? I've left some peer review [#907 (comment)](https://github.com/silverstripe/silverstripe-admin/pull/907#issuecomment-623804112) that I think should be addressed before this can be merged.

@sachajudd I will check this patch this night (UTC+1)

emteknetnz:

emteknetnz:

PR #908, opened 2019-07-17
FIX DateField is now controlled

ScopeyNZ:
See the [redux form docs](https://redux-form.com/7.4.2/docs/api/field.md/#-code-input-value-any-code-) for input values.

Currently DateField is uncontrolled, meaning that changes to the value of the field in _redux_ after the field is initialised is not actually reflected in the field. Changing this to be controlled (like most of our other redux fields) means that we can update the redux state and have that reflected in our field.

Fixes dnadesign/silverstripe-elemental#649


robbieaverill:
The changes look OK to me, but it seems like it was originally designed to behave in the way you've changed it from - any thoughts on why that might have been the case?

ScopeyNZ:
I'm not sure why the field was made "uncontrolled". It doesn't really make any sense to me that it remains uncontrolled when pretty much all our fields are controlled fields. I considered that this change was made on the second commit ever on this file, so it was probably early in development - and never really noticed until now.

As far as I understand there's not much difference in `defaultValue` vs `value` aside from the controlled/uncontrolled toggle of course.

PR #910, opened 2019-07-17
FIX Inline toolbar placement now works in HTMLEditorFields with less than 6 rows

robbieaverill:
This fixes a bug where TinyMCE inline toolbars (when inserting links) were being positioned off the top of the screen when the editor has less than 6 rows (e.g. `$content->setRows(5)`). The bug is a little obscure in that it is reproducible when:

* The editor has less than 6 rows
* A link is added to the first paragraph, but not in the first line (e.g. paragraph must be split over multiple lines)

The placement isn't perfect, but I think a slight (a few pixels) misposition of the popup is much lower impact than it being off the screen entirely.

This code is copied from a WordPress plugin, and the condition I've changed was originally irrelevant since the variable is never undefined, but it can be zero. I assume the upstream plugin never has small editors.

Fixes https://github.com/dnadesign/silverstripe-elemental/issues/655

We may want to open a new issue to tweak the positioning of the popups for small editors.

### Positioning screenshot examples

When clicking each link on each line from top to bottom.

image
image
image
image
image


ScopeyNZ:
Would it be better to set the default value to `null`? Is there a problem if the top is actually calculated to be `0`? I assume in the end it doesn't really matter?

ScopeyNZ:
Ok I've had a bit of a play and I've had a bit of a think about the logic. The likelyhood that zero is calculated is really low and the popup placement should still make sense in that case.

It's definitely a little weird that it covers the link but 🤷‍♂ .

Issue #639, opened 2018-09-12
Form fields with HTML array syntax in the name don't get mapped properly into redux-form state


robbieaverill:
**Version: 4.3.x-dev**

If I have a field with HTML array syntax in its name, the form schema maps it as a key value (like all other fields) into the redux state, but redux-form doesn't match the value from the state into the form field when it renders it.

When I update the form field I notice that the redux state is updated with a nested data structure instead.

I assume that our implementation of redux-form state population needs to be tweaked to detect fields with HTML array fields and break them up into nested data structures before they get written to the state, so it complies with the schema that redux-form uses to write back to the state when a field name is changed.

Related: https://github.com/dnadesign/silverstripe-elemental/pull/386

PR #912, opened 2019-07-22
FIX Support array field names when resolving values

ScopeyNZ:
Currently we can't have array style field names (eg. `name[]` - very much a PHPism) with our FormBuilder implementation. State cannot be determined as the field name in redux strips the square brackets, but the state provided by the schema does not strip the square brackets. This patch updates the function that supplies form state to redux to pass through values if it matches either the given name, or the given name with the square brackets (`[]`) added.

Related issue: silverstripe/silverstripe-tagfield#107


robbieaverill:
Does this fix https://github.com/silverstripe/silverstripe-admin/issues/639?

ScopeyNZ:
> Does this fix #639?

Unfortunately no. It only fixes instances where `[]` are on the end of a field. I think that other issues is a little more challenging.

Issue #699, opened 2018-10-17
React breadcrumb goes backwards


maxime-rainville:
The react breadcrumb parent levels are displayed backward ... the root level is displayed on the far right while the direct parent is displayed to the left. You need at least 3 level or more to see it. This can be replicated in the pattern library and in asset-admin.

I've confirmed with @clarkepaul and @sachajudd that this is not the intended behavior.

![image 1](https://user-images.githubusercontent.com/1168676/47119701-64215b80-d2c8-11e8-9764-26b950bfeb83.png)

https://www.youtube.com/watch?v=P8XxXvAd5d0&feature=youtu.be

#614 is opened to address inconsistencies between the react breadcrumb and legacy breadcrumb. We might want to look at these two issues together.

# Pull Request
* [ ] https://github.com/silverstripe/silverstripe-admin/pull/925

maxime-rainville:
Setting impact to medium as it's really confusing, but doesn't prevent you from doing anything.

ScopeyNZ:
Closed in #925

PR #925, opened 2019-08-09
FIX Display breadcrumb element from left to right

maxime-rainville:
Breadcrumb was going backwards.

I think the point of the reverse order was to try to cut off the higher level when there's not enough space to display the entire bread crumb.

This is what the new PR looks like when the breadcrumb gets cut off.
image

# Parent issue
* https://github.com/silverstripe/silverstripe-admin/issues/699


maxime-rainville:
@clarkepaul This should fix our favourite issue.

ScopeyNZ:
I think the overflow on the left hand side was intended as you're much more likely to want to know the highest level of the tree rather than the lowest when you're in a deeply nested folder.

ScopeyNZ:
New issue raised here: https://github.com/silverstripe/silverstripe-admin/issues/930

robbieaverill:
This change breaks the breadcrumbs in the "add new" screen of the CMS:

image

ScopeyNZ:
That's not even React rendered right? So this is styles bleeding across different breadcrumb implementations? 😬

robbieaverill:
Maybe. That happened with the search tag implementations in 4.2 as well, we need to ensure we name classes carefully!

maxime-rainville:
https://github.com/silverstripe/silverstripe-admin/pull/942 should fix this

Issue #914, opened 2019-07-29
The order of the insert links in the editor to be presented in a logical order


clarkepaul:
As inserting links into the editor is one of the most used features of the CMS it makes sense that they are logically ordered so the most used link type is placed at the top. This issue is also covered in the Elemental link component design consistency issue but the order of the links presented makes more sense being sorted globally.

I suggest the order should be as per designs (maybe external link could be swapped with File link?). Note that "Anchor" links are to be integrated with internal/external links as part of other issues (see https://github.com/dnadesign/silverstripe-elemental/issues/589).

<img width="122" alt="Pasted_Image_29_07_19__12_43_PM" src="https://user-images.githubusercontent.com/555033/62015298-74267e00-b1fe-11e9-9b8d-cfe0d393bed1.png">

Design reference of link component for blocks but is similar https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/folder/components/5c68f96ffb26bb0018525d2b

# Pull Requests
* [x] silverstripe/silverstripe-asset-admin#975
* [x] silverstripe/silverstripe-admin#926
* [ ] https://github.com/silverstripe/silverstripe-cms/pull/2464

maxime-rainville:
That's what's currently there.
![image](https://user-images.githubusercontent.com/1168676/62846755-9c49cd00-bd26-11e9-89dd-24f8256660ae.png)

Add link to "Anchor on another page" is not in the original mock up. I'll put it under external link for now. Let me know if you want a different order.

# Controlling the order of link option

The way those links are added is by using this bit of code.
```js
TinyMCEActionRegistrar.addAction(
  'sslink',
  {
    text: i18n._t('Admin.LINKLABEL_EMAIL', 'Link to email address'),
    onclick: (editorInst) => editorInst.execCommand(commandName),
    priority: 51,
  },
  editorIdentifier,
).addCommandWithUrlTest(commandName, /^mailto:/);
```

`priority` controls the order ... items with an higher priority gets displayed first.

I'll set the priority so there's room in-between each linking option in case a mad-scientist somewhere decides to add their own linking option and wants to put it at a specific place.

clarkepaul:
The Anchor link was meant to be an extra field on the "Link to page" but never happened.

The thing that makes me the happiest is that "link to page" is at the top finally :) 
The order looks way better.

maxime-rainville:
Got PRs for this. 

Here's what it looks like with the PRs.

[...]

PR #926, opened 2019-08-12
FIX Set better order for insert link menu

maxime-rainville:
Reorder the link options in the WYSWIYG.

# Parent issue
* https://github.com/silverstripe/silverstripe-admin/issues/914

# Depends on
* https://github.com/silverstripe/silverstripe-cms/pull/2464
* https://github.com/silverstripe/silverstripe-asset-admin/pull/975


ScopeyNZ:
Hmmmn. The changes make sense to me but I can't get it to display correctly. I'm getting: 

image

I've definitely checked out all the branches. I'm using a standard installer with `4.4.x-dev`. I've `yarn dev`ed in all the branches to make sure it wasn't anything weird happening with dist files. The source looks correct in the browser - the priority numbers have updated. Not really sure what's going on tbh.

client/src/lib/TinyMCEActionRegistrar.js
```diff
@@ -83,11 +84,13 @@ class TinyMCEActionRegistrar {
   getSortedActions(menu, configId, includingGlobal = true) {
     const actions = this.getActions(menu, configId, includingGlobal);
     return actions.sort((a, b) => {
-      if (a.priority !== b.priority) {
-        return a.priority < b.priority;
+      // Sort by descending priority
```
maxime-rainville: `sort` expects a callback that will return a numeric value. Firefox is OK with a boolean but Chrome isn't.

maxime-rainville: https://stackoverflow.com/questions/1969145/sorting-javascript-array-with-chrome

PR #929, opened 2019-08-13
FIX re-enable ability to insert anchors

NightJar:
SilverStripe 3 had the ability to insert anchors into content areas
within the CMS. SilverStripe 4 however has separated the ideas of the
admin area and the CMS - however the ability for an editor to insert
an anchor is not directly related to pages. To this end the anchor
plugin for TinyMCE should be enabled here, as opposed to globally
for all rich text editors, e.g. enabling for unauthenticated users
adding blog comments shouldn't be done.

This has the knock on affect for the CMS module, where the "link to
an anchor on another page" plugin is also enabled. Since the release
of SilverStripe 4 it has been an option to create such links, but an
editor could not also create the anchors to link to. Now they will be
able to do such.

Partial address of https://github.com/silverstripe/silverstripe-cms/issues/2268
See also https://github.com/silverstripe/silverstripe-framework/pull/9175

image

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -9,14 +9,25 @@ use SilverStripe\Admin\CMSProfileController;
 use SilverStripe\Core\Manifest\ModuleLoader;
 use SilverStripe\Forms\HTMLEditor\TinyMCEConfig;
 
-// Use module_priority config to prioritise module assignment
-TinyMCEConfig::get('cms')
+// Enable SilverStripe insert link dialog on the rich text editor
+$module = ModuleLoader::inst()->getManifest()->getModule('silverstripe/admin');
+/** @var TinyMCEConfig $editorConfig */
+$editorConfig = TinyMCEConfig::get('cms');
+$editorConfig
+    ->enablePlugins([
+        'contextmenu' => null,
+        'image' => null,
+        'anchor' => null,
+        'sslink' => $module->getResource('client/dist/js/TinyMCE_sslink.js'),
+        'sslinkexternal' => $module->getResource('client/dist/js/TinyMCE_sslink-external.js'),
+        'sslinkemail' => $module->getResource('client/dist/js/TinyMCE_sslink-email.js'),
+    ])
     ->setOptions([
         'friendly_name' => 'Default CMS',
         'priority' => '50',
         'skin' => 'silverstripe',
         'body_class' => 'typography',
-        'contextmenu' => "sslink ssmedia ssembed inserttable | cell row column deletetable",
+        'contextmenu' => "sslink anchor ssmedia ssembed inserttable | cell row column deletetable",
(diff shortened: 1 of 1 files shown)


_config.php
```diff
@@ -35,17 +46,9 @@
...
-        'contextmenu' => null,
-        'image' => null,
-        'sslink' => $module->getResource('client/dist/js/TinyMCE_sslink.js'),
-        'sslinkexternal' => $module->getResource('client/dist/js/TinyMCE_sslink-external.js'),
-        'sslinkemail' => $module->getResource('client/dist/js/TinyMCE_sslink-email.js'),
-    ])
-    ->setOption('contextmenu', 'sslink ssmedia ssembed inserttable | cell row column deletetable');
+// enable ability to insert anchors
+$editorToolbar = $editorConfig->getButtons();
+array_splice($editorToolbar[2], array_search('sslink', $editorToolbar[2]), 0, ['anchor']);
```
robbieaverill: Is there way you can do this without the magic numbers?

NightJar: ok

robbieaverill:
image

Presumably this is out of the box TinyMCE? Would be nice if it could say "ID" instead of "Id", maybe with an example or even an auto-generated equivalent of whatever you're anchoring, but I realise that's probably scope creep for this change.

NightJar:
> Presumably this is out of the box TinyMCE?

Yes.

>that's probably scope creep for this change.

Yes ;)

_config.php
```diff
@@ -35,17 +46,10 @@
...
-        'sslink' => $module->getResource('client/dist/js/TinyMCE_sslink.js'),
-        'sslinkexternal' => $module->getResource('client/dist/js/TinyMCE_sslink-external.js'),
-        'sslinkemail' => $module->getResource('client/dist/js/TinyMCE_sslink-email.js'),
-    ])
-    ->setOption('contextmenu', 'sslink ssmedia ssembed inserttable | cell row column deletetable');
+// enable ability to insert anchors
+$editorToolbar = $editorConfig->getButtons();
+const SECOND_ROW = 2;
+array_splice($editorToolbar[SECOND_ROW], array_search('sslink', $editorToolbar[SECOND_ROW]), 0, ['anchor']);
+$editorConfig->setButtonsForLine(SECOND_ROW, $editorToolbar[SECOND_ROW]);
```
robbieaverill: How about this?
```php
$editorConfig->insertButtonsBefore('sslink', 'anchor');
```

NightJar: 😮 how did I miss that one :S

NightJar:
Updated. Checked with @clarkepaul.

image

robbieaverill:
I've cherry-picked this back into the 1.4 branch for you

PR #932, opened 2019-08-15
FIX jstree state when saving a page by retaining the open/closed state and selected node state

bergice:
See: https://github.com/silverstripe/silverstripe-cms/issues/2355


maxime-rainville:
> but removing the call to `self.jstree('reopen');` results in the children of the current page being collapsed on save.

I was actually seeing both this behaviour and the reverse. It looks like the open state of the tree node gets sticky ... after a save it keeps reverting to being open or close. I'm not 100% sure what's causing it, but it was doing it with and without the PR for me.

bergice:
> I've taken this for a whirl, and it seems to achieve the goal - but removing the call to `self.jstree('reopen');` results in the children of the current page being collapsed on save. I suggest reinstating this before we merge.

I've reinstated that and it's looking good. :)

PR #935, opened 2019-08-19
FIX Ensure IE renders the CMS UI at full width consistently

Cheddam:
Using percentage units all the way down the tree seems to cause IE to struggle when calculating the correct width for elements. By changing the main cms-container to use Viewport Width units (vw), we provide a more concrete value for IE to base its calculations on.


ScopeyNZ:
Tested in various browsers and seems good. As discussed in person, there's still an issue where there's extra scrolling at the bottom of the page in IE11 at certain resolutions but it might be an existing issue and this is at least better than previous. Still going to merge this.

PR #937, opened 2019-08-20
Add a `canMove` callback which makes it possible to add conditional logic before moving nodes around the tree

bergice:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/2154


client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -88,7 +88,13 @@ $.entwine('ss.tree', function($){
               }
             }
           })
-          .bind('move_node.jstree', function(e, data) {
+          .bind('move_node.jstree', async function(e, data) {
+            if (!await self.canMove(data)) {
```
unclecheese: Where is this promise defined?

bergice: I promise it is extended in https://github.com/silverstripe/silverstripe-cms/pull/2470/files#diff-126d72786d20ac82ea0f3445505130a7R146

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -162,6 +168,10 @@ $.entwine('ss.tree', function($){
       }
     },
 
+    canMove: function(data) {
```
ScopeyNZ: The default implementation should be a promise so that it can be `await`ed above:
```
    canMove: Promise.resolve(true);
```

robbieaverill:
I don't think this is a patch change - could you rebase onto the `1` branch and add `NEW` as a commit prefix?

bergice:
> I don't think this is a patch change - could you rebase onto the `1` branch and add `NEW` as a commit prefix?

All done, ready for another look.

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -162,6 +168,15 @@ $.entwine('ss.tree', function($){
       }
     },
 
+    /**
+     * Validates the moving of a node
+     * @param data data provided by the move_node.jstree event
```
robbieaverill: The type is incorrect - should it be `{object}` ?

PR #940, opened 2019-08-22
BUGFIX: react-select clears input on search. Monkey patch, needs upgrade

unclecheese:
## Yes, this is horrible.

Resolves: https://github.com/silverstripe/silverstripe-framework/issues/9078

There is a bug in `react-select` relating to how state is derived comparing two arrays. The reliance on referential inequality is tripping up the component and causing the behaviour described in the above issue.

The long-term fix for this problem is to upgrade `react-select`, but the API has changed drastically  since version `1.3` (which is installed on `silverstripe/admin:1.x-dev`), and requires a major rewrite. Biggest change is around the new component injection that replaces the renderers like `menuRenderer`.

This should be considered technical debt to get us through the next couple releases while we work on upgrading `react-select`. A separate issue will be created for that upgrade.


robbieaverill:
image

Did you have a young kiwi guy moaning to you about this problem?

unclecheese:
> Did you have a young kiwi guy moaning to you about this problem?

"eeeek", as in ugly fix, but it works on multiple levels. ;)

robbieaverill:
Ah - in my head I heard a horrible New Zealand accent

PR #941, opened 2019-08-23
FIX enter to submit form not working on `Add new page`

bergice:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/2338


client/src/legacy/LeftAndMain.js
```diff
@@ -904,39 +904,45 @@ $.entwine('ss', function($) {
...
+              }
+            });
+          }
+          if(index !== null){
+            tabset.tabs('option', 'active', index);
+            self.trigger('tabstaterestored');
+          }
+        });
+      } else {
+          $('#Form_AddForm_action_doAdd').focus();
```
bergice: This is the main change.

PR #943, opened 2019-08-25
 ENHANCEMENT: Search component persists state in URL

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-framework/issues/9080

Depends on https://github.com/silverstripe/silverstripe-framework/pull/9199

Not the most elegant thing in the world, but this is actually a really tricky thing to solve. Lots of legacy fixtures to cooporate with.

Could probably use a test or two?


client/src/components/Search/SearchStatefulURL.js
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
+import url from 'url';
+
+const makeStateful = (Search) => class StatefulURLSearch extends React.Component {
```
maxime-rainville: Maybe add a comment describing the purpose of this thing.

emteknetnz:

emteknetnz:

PR #944, opened 2019-08-25
TreeDropdownField renders when being loaded without data through form loader 🐝

adrhumphreys:
This fixes an issue where the field is loaded through the FormLoader react component and therefore is initially passed props as `{ props: {}}` where the actual props are hidden inside so to access them you'd have to go `this.props.props`.

The easiest way to reproduce the issue is to create a page with the field and then use silverstripe-snapshot-admin to show a snapshotted version of the field which is the read only version


robbieaverill:
Perhaps we need to fix the nested props object issue instead of disabling it while it loads?

adrhumphreys:
Yeah this was very much a quick fix from our side, I think it still applies though as if this component is created without props then it will throw an exception and then nothing from the form loader will display

robbieaverill:
I've rebased and built the dist files for you. I've also removed the emoji from your commit message (sorry!) - they are used in our changelogs, which may not necessarily support emojis, so we usually avoid using them

robbieaverill:
I'm not sure, the Behat tests might be valid - any ideas?

adrhumphreys:
Haven't a clue sorry, any way to view the travis artifacts?

robbieaverill:
Not for pull request builds, unfortunately. I'll give this PR an actual test at some point

robbieaverill:
The latest `1` branch build is green, so it could be a relevant failure

PR #945, opened 2019-08-26
Bug IE11 fallback fix for HTMLEditorField

bergice:
Without this, editing content blocks on IE11 will not work.

Resolves https://github.com/dnadesign/silverstripe-elemental/issues/700
Resolves https://github.com/silverstripe/silverstripe-cms/issues/2459


client/src/components/HtmlEditorField/HtmlEditorField.js
```diff
@@ -80,9 +80,16 @@ class HtmlEditorField extends TextField {
     const { value } = this.props;
 
     if (value !== prevProps.value) {
-      const event = new Event('change', { bubbles: true });
+      let event;
+      if (typeof (Event) === 'function') {
```
robbieaverill: So IE11 doesn't support this as an interface? Is there a polyfill we can use instead of this, or is this the best approach?

NightJar: Event is supported.
image

NightJar: hahahah, omg, so @Cheddam just pointed out to me that although `Event` is supported, the _constructor_ isn't 😂 
https://developer.mozilla.org/en-US/docs/Web/API/Event#Browser_compatibility

maxime-rainville:
I think this PR got superseded by https://github.com/silverstripe/silverstripe-admin/pull/951

@bergice Let me know if I'm telling porkies.

PR #947, opened 2019-08-26
API Add a `extra_requirements_i18n_javascript` config option to allow additional client side translations to be included

maxime-rainville:
This adds a convenience config option to allow developers to override our own translation with their own.

# Parent issue
* https://github.com/silverstripe/silverstripe-framework/issues/9137

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -216,9 +216,9 @@ class LeftAndMain extends Controller implements PermissionProvider
      *
      * YAML configuration example:
      * <code>
-     * LeftAndMain:
+     * SilverStripe\Admin\LeftAndMain:
      *   extra_requirements_javascript:
-     *     - mysite/javascript/myscript.js
+     *     - app/javascript/myscript.js
      * </code>
      *
      * @config
@@ -234,9 +234,9 @@ class LeftAndMain extends Controller implements PermissionProvider
      *
      * YAML configuration example:
      * <code>
-     * LeftAndMain:
+     * SilverStripe\Admin\LeftAndMain:
      *   $extra_requirements_i18n_javascript:
-     *     - mysite/client/lang
+     *     - app/client/lang
      *     - my-company/my-module:client/lang
      * </code>
      *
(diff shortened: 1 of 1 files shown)


code/LeftAndMain.php
```diff
@@ -693,9 +712,34 @@ protected function init()
...
+        Versioned::set_stage(Versioned::DRAFT);
+
+        // Set default reading mode to suppress ?stage=Stage querystring params in CMS
+        Versioned::set_default_reading_mode(Versioned::get_reading_mode());
+    }
 
+    /**
+     * Go through the `extra_requirements` config option and require them.
+     */
+    private function extraRequirements(): void
```
maxime-rainville: I only added this method for readability.

code/LeftAndMain.php
```diff
@@ -226,6 +226,25 @@ class LeftAndMain extends Controller implements PermissionProvider
...
 
+    /**
+     * Register additional translation sources to be included in the CMS client side code.
+     * This can be use to override SilverStripe translations. This configuration option
+     * expects a list of browser accessible directories that can be pass to
+     * {@link Requirements::add_i18n_javascript()}.
+     *
+     * YAML configuration example:
+     * <code>
+     * LeftAndMain:
```
robbieaverill:
```suggestion
     * SilverStripe\Admin\LeftAndMain:
```

code/LeftAndMain.php
```diff
@@ -226,6 +226,25 @@ class LeftAndMain extends Controller implements PermissionProvider
...
+     *   $extra_requirements_i18n_javascript:
+     *     - mysite/client/lang
```
robbieaverill:
```suggestion
     *     - app/client/lang
```

We haven't used mysite for a few minor releases now, may as well not keep documenting it in examples

maxime-rainville:
I agree it doesn't 100% fix the original issue. Since we already have config option for adding extra CSS and extra JS, I think it makes some sense to have one for adding extra translations.

emteknetnz:

maxime-rainville:
Where do we actually use JS translation string in the front end? The only place I could think off is the login-form module.

I still think this PR is worth merging as-is, because it basically extend the existing `extra_requirements_css`/`extra_requirements_javascript` logic to a similar use case.

robbieaverill:
React components in the CMS use them, and userforms on the frontend does also

maxime-rainville:
Humm ... OK back to the drawing board.

PR #948, opened 2019-08-27
FIX implement tinyMCE repaint

NightJar:
When a TinyMCE instance is moved in the DOM, most especially if it is
moved down the DOM tree, the content will disappear and many of the
interactions (toolbar buttons) will cease functioning as intended.

The only way to 'refresh' an instance seems to be to remove it then
re-add it, so this commit introduces this functionality to the repaint
method in entwine, meaning the editor will continue to use the same
config it was first instantiated with.

Ref: https://github.com/dnadesign/silverstripe-elemental/issues/692


robbieaverill:
Do you think we still need this patch @NightJar? Since we've delayed fixing the issue this was for, perhaps we should leave the method the way it is until it's looked at again?

NightJar:
Sure, no impact so not worth merging.

PR #949, opened 2019-08-28
BUG When using the esc key to exist a popover, refocus to the target element

maxime-rainville:
When using the Esc key to dismiss the popover, you probably want the focus to go back to element that was used the toggle the popover in the first place.

This spun out of a discussion on https://github.com/dnadesign/silverstripe-elemental/pull/724

# Related PR
* https://github.com/dnadesign/silverstripe-elemental/pull/727


client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -16,20 +16,47 @@ class PopoverOptionSet extends Component {
...
     this.state = {
       searchValue: ''
     };
   }
 
   /**
-   * Pass toggle to parent (props requires a toggle function) and clear the search input
+   * Handle the toggle from the underlying react strap component.
    */
   handleToggle() {
```
maxime-rainville: React strap sends a event object to this method, so we couldn't just reuse it as is..

client/src/components/PopoverOptionSet/PopoverOptionSet.js
```diff
@@ -16,20 +16,47 @@ class PopoverOptionSet extends Component {
...
+    }
+  }
+
+  /**
+   * Move the focus back to the popover target
+   */
+  focusOnTarget() {
+    const { target } = this.props;
+    if (target) {
+      const el = reactstrapUtil.getTarget(target);
```
maxime-rainville: This [getTarget](https://github.com/reactstrap/reactstrap/blob/3df7c031830b92c5b9e0e8dc53b1ec80bec3dfa9/src/utils.js#L258) method is used internally by react strap to find the target.

It's explicitly exported by the library, so I think it's safe to use it.

client/src/components/PopoverOptionSet/PopoverOptionSet.js
maxime-rainville: I didn't want to alter the focus when the user clicks out of the popover, because that might cause some other element the user wants to interact with to loose their focus.

PR #951, opened 2019-08-29
FIX content block editing breaking when editing using IE11

bergice:
Fixes it by adding a Event constructor polyfill.

Resolves dnadesign/silverstripe-elemental#700
Resolves silverstripe/silverstripe-cms#2459


maxime-rainville:
This sounds like this is something that would be worth fixing in all supported version. Do we want to back port it all the way back to 4.2 or 4.3?

NightJar:
4.2 is out of support isn't it?

maxime-rainville:
We still have a branch for it ... or have we decided to keep those around?

robbieaverill:
I think it’s still in support. I was excluding it from Travis matrixes because it will probably get deleted soon

PR #956, opened 2019-09-24
BUG Search for both Save and Apply change when runnign behat test

maxime-rainville:
When versioned is installed, it changes the "save" button label which confuses behat.

This PR tests both possible labels when saving members.

# Depends on 
* https://github.com/silverstripe/silverstripe-behat-extension/pull/189


maxime-rainville:
In 4.4.3, the label was "Apply Changes" for non-versioned objects and "Save" for versioned objects. In 4.4.4 the label was changed to Save for everything.

Not sure which module this came from or if it was done on purpose. Admin requires versioned `^1@dev`. So in theory, the tests should be green whatever you install a release of versioned that produces an "Apply changes" button or a "Save" button.

robbieaverill:
That’s the background I needed :-) thanks

PR #957, opened 2019-09-25
Add tests for UsedOnTable

bergice:


bergice:
> Your new test doesn't appear to be passing.
> 
> I don't we should have this test on admin, because it will rely on elemental, which may or may not be installed.

How does this rely on elemental?

tests/php/UsedOnTableTest.php
```diff
@@ -0,0 +1,47 @@
...
+
+namespace SilverStripe\Admin\Tests;
+
+use SilverStripe\Admin\Forms\UsedOnTable;
+use SilverStripe\Assets\File;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\Dev\TestOnly;
+
+class UsedOnTableTest extends SapphireTest implements TestOnly
```
robbieaverill: You don't need to implement TestOnly in test classes (SapphireTest already does), only in stubs used in tests

maxime-rainville:
From the discussion we had earlier, I thought this was meant to test https://github.com/dnadesign/silverstripe-elemental/pull/734

If this test is meant to stand by itself and test the `updateUsage` hook on `UsedOnTable`, then it's fine, but it's not a substitute for writing unit tests for the other elemental PR.

tests/php/UsedOnTableTest.php
```diff
@@ -0,0 +1,47 @@
...
+        $usedOnTable = UsedOnTable::create("Used On");
+
+        $file1 = $this->objFromFixture(File::class, 'file1');
+        $usedOnTable->setRecord($file1);
+
+        $request = new HTTPRequest('GET', '/');
+        $response = $usedOnTable->usage($request);
+
+        $this->assertEquals(<<<JSON
+{"usage":[{"id":"SilverStripe\Assets\Shortcodes\FileLink\/1","title":"#1","type":"File Link","state":null,"link":null}]}
```
maxime-rainville: JSONDecoding the body and comparing it to a PHP array will probably give you more consistent results.

With a plain string, whatever you test passes depends on things like formatting and the order of keys.

robbieaverill:
Any chance of getting the tests green?

maxime-rainville:
Andre is off for the next fortnight. We've got a related issue with campaign admin https://github.com/silverstripe/silverstripe-campaign-admin/issues/128

We're thinking of introducing some fix that would allow DataObject to be marked as internal so they don't show up in places like campaign list and "used on" table.

ScopeyNZ:
Sounds like a job for https://github.com/silverstripe/silverstripe-cms/issues/2454

code/Forms/UsedOnTable.php
```diff
@@ -72,6 +72,10 @@ public function usage(HTTPRequest $request)
         $record = $this->getRecord();
         $usage = $record->findOwners();
 
+        $usage = $usage->filterByCallback(function($item) {
+            return !$item->config()->get('hide_in_campaigns');
```
emteknetnz, line 76: this is nasty and we definitely shouldn't do it.  it's creating a coupling between campaigns and the used on table when there doesn't need to be one

the way to remove stuff from the used on table is via the updateUsage extension hook

tests/php/UsedOnTableTest/UsedOnTableExtension.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace SilverStripe\Admin\Tests\UsedOnTableTest;
+
+use SilverStripe\Assets\Shortcodes\FileLink;
+use SilverStripe\Core\Extension;
+use SilverStripe\ORM\ArrayList;
+use SilverStripe\ORM\DataObject;
+
+class UsedOnTableExtension extends Extension
```
emteknetnz, line 10: should implement TestOnly

tests/php/UsedOnTableTest/UsedOnTableExtension.php
```diff
@@ -0,0 +1,20 @@
...
+{
+    public function updateUsage(ArrayList &$usage, DataObject &$record)
+    {
+        $usage = $usage->exclude([
+            'ClassName' => [
+                FileLink::class,
```
emteknetnz, line 16: Should try and remove something other that FileLink since that's already removed by default https://github.com/silverstripe/silverstripe-asset-admin/blob/1.6/code/Extensions/UsedOnTableExtension.php

Even though it's on a different module, something may get confused when running this in the context of silverstripe/recipe-cms or kitchen-sink

code/Forms/UsedOnTable.php
```diff
@@ -72,6 +72,10 @@ public function usage(HTTPRequest $request)
         $record = $this->getRecord();
         $usage = $record->findOwners();
 
+        $usage = $usage->filterByCallback(function($item) {
```
emteknetnz, line 75: This is nasty and creates a coupling between campaigns and the used on table where there doesn't need to be one

Please remove this block.

tests/php/UsedOnTableTest/UsedOnTableExtension.php
emteknetnz, line 10:
```suggestion
class UsedOnTableExtension extends Extension implements TestOnly
```

tests/php/UsedOnTableTest/UsedOnTableExtension.php
emteknetnz, line 16: Should try and remove something other than FileLink since that's already removed by default https://github.com/silverstripe/silverstripe-asset-admin/blob/1.6/code/Extensions/UsedOnTableExtension.php

Even though it's on a different module, something may get confused when running this in the context of silverstripe/recipe-cms or kitchen-sink

Try and remove something generic such as a Page

tests/php/UsedOnTableTest.php
```diff
@@ -0,0 +1,56 @@
...
+use SilverStripe\Admin\Forms\UsedOnTable;
+use SilverStripe\Assets\File;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Dev\SapphireTest;
+
+class UsedOnTableTest extends SapphireTest
+{
+    protected static $fixture_file = 'UsedOnTableTest.yml';
+
+    public function testOK()
```
emteknetnz, line 14:
```suggestion
    public function testNotExcluded()
```

tests/php/UsedOnTableTest/UsedOnTableExtension.php
emteknetnz: @bergice still one unresolved comment here

bergice: We override the extensions in the tests which should resolve this.

emteknetnz: What I mean is update the test so that you're excluding File's on Pages from showing in the UsedOnTable, rather than Files associated with FileLink, because FileLink is removed by default anywhere whereas Pages aren't, so this way we're testing that the `updateUsage` extension is actually removing something

bergice: How come `FileLink`'s are removed by default and where does this happen?

emteknetnz: In UsedOnTableExtension.php - I linked to that a few comments above

bergice: That's what I meant by excluding having the extension be excluded in the tests: https://github.com/silverstripe/silverstripe-admin/pull/957/files#diff-1a65f9abdcc3690e7d4512d51a4ac09aR16

emteknetnz: OK gotcha

emteknetnz (approved):

PR #959, opened 2019-09-25
FIX: JSTree error if callback isn't passed (fixes #958)

### LINKED ISSUE #958: JavaScript error / SiteTree disappearing
PR: https://github.com/silverstripe/silverstripe-admin/pull/959

This happens when viewing a “lumberjacked” page - not sure if that module is the only trigger for this bug...

https://github.com/silverstripe/silverstripe-admin/blob/a3eaf95bb2e37641d5a2f476807cc9696d55fac4/client/src/legacy/LeftAndMain.Tree.js#L301

> `Uncaught TypeError: r is not a function`

The error seems to be caused by sometimes not passing a third argument when calling this function... looks like a regression from https://github.com/silverstripe/silverstripe-admin/commit/1c548cb599563997687cd1062ff2a0985c43197e

I guess it just needs to be wrapped in an `if` statement

--- 

Edit:

Actually the error doesn’t seem to cause the tree break, it’s just collapsed at the very top level (the site name)

kinglozzer:


client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -298,7 +298,9 @@ $.entwine('ss.tree', function($){
           }
           // Substitute html from request for that generated by jstree
           node.addClass(origClasses).html(newNode.html());
-          callback(node);
+          if (callback) {
+            callback(node);
+          }
```
kinglozzer: Not strictly necessary now I’ve re-added a callback, but may as well be defensive

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -391,7 +393,7 @@ $.entwine('ss.tree', function($){
 
       if(blocking && this.getIsUpdatingTree()) return;
 
-      var self = this, i, includesNewNode = false;
```
kinglozzer: Dead code

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -391,11 +393,12 @@ $.entwine('ss.tree', function($){
...
+      var self = this;
       this.setIsUpdatingTree(true);
       self.jstree('save_selected');
 
       // TODO 'initially_opened' config doesn't apply here
+      self.jstree('open_node', this.getNodeByID(0));
```
kinglozzer: This ensures the root node (the site name) is open, so the tree is visible while any missing nodes are AJAX loaded in

taoceanz:
Any chance of this getting merged to be available to `silverstripe/recipe-cms: 4.4.4` any time soon?

ScopeyNZ:
I've merged the PR. It won't appear in a recipe until the next release (whenever that may be). You might be able to ask someone nicely to release a stable version of `admin` that you can opt into for the time being though.

taoceanz:
Thanks @ScopeyNZ

PR #961, opened 2019-09-30
fixed creating multiple duplicate data objects

Makreig:
As the save button wasn't being locked out, if it was clicked multiple times then multiple duplicate objects would be created.


client/src/legacy/LeftAndMain.js
```diff
@@ -443,8 +443,9 @@ $.entwine('ss', function($) {
 
       // set button to "submitting" state
       $(button).addClass('btn--loading loading');
+      $(button).attr('disabled', 'disabled');
```
ScopeyNZ:
```suggestion
      $(button).prop('disabled', true);
```

client/src/legacy/LeftAndMain.js
```diff
@@ -462,6 +463,7 @@ $.entwine('ss', function($) {
 
       var clearButton = function() {
         $(button).removeClass('btn--loading loading');
+        $(button).attr('disabled', false);
```
ScopeyNZ:
```suggestion
        $(button).prop('disabled', false);
```

client/src/legacy/LeftAndMain.js
```diff
@@ -443,8 +443,9 @@ $.entwine('ss', function($) {
 
       // set button to "submitting" state
       $(button).addClass('btn--loading loading');
+      $(button).attr('disabled', 'disabled');
 
-      if($(button).is('button')) {
+      if($(button).is('button')) {
```
ScopeyNZ:
```suggestion
      if($(button).is('button')) {
```

Makreig:
> If there's a validation error, will the button be returned to its original state?

from what i can see, the buttons state is cleared whether there is an error or not, so it should be.

Makreig:
I have no idea why this test has failed. anyone else know what's wrong?

ScopeyNZ:
> I have no idea why this test has failed. anyone else know what's wrong?

There's an automated test that asserts the bundle you've included has been compiled correctly - but that's failing. You'll need to `yarn build` again from your branch and commit the new bundle.

PR #963, opened 2019-10-04
API Convert SecurityAdmin to be a ModelAdmin

maxime-rainville:
This is a bit experimental. It converts SecurityAdmin to be a ModelAdmin.

It seems to work remarkably well. If we think this is worthwhile, we can take it forward and finalise it.

Couple things to address:
* Rename Members to Users ... that is if we prefer that label.
* Bring back the "Caution: Removing members from this list will remove them from all groups and the database"
* I think there's a bit of custom import logic that I didn't copy over. I need to dig deeper and see if differs from the boiler plate ModelAdmin import logic.
* Clean up any redundant JS and SS files

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1350

### DIFF
--- a/code/SecurityAdmin.php
+++ b/code/SecurityAdmin.php
@@ -29,39 +29,29 @@ use SilverStripe\View\Requirements;
 /**
  * Security section of the CMS
  */
-class SecurityAdmin extends LeftAndMain implements PermissionProvider
+class SecurityAdmin extends ModelAdmin implements PermissionProvider
 {
+
+    private static $managed_models = [
+        Member::class,
+        Group::class,
+        PermissionRole::class
+    ];
 
     private static $url_segment = 'security';
 
-    private static $url_rule = '/$Action/$ID/$OtherID';
-
     private static $menu_title = 'Security';
-
-    private static $tree_class = Group::class;
-
-    private static $subitem_class = Member::class;
 
     private static $required_permission_codes = 'CMS_ACCESS_SecurityAdmin';
 
     private static $menu_icon_class = 'font-icon-torsos-all';
-
-    private static $allowed_actions = array(
-        'EditForm',
-        'MemberImportForm',
-        'memberimport',
-        'GroupImportForm',
-        'groupimport',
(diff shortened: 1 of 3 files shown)


bergice (commented):
I agree with renaming from `Users` to `Members`.

The warning message is no longer displayed and needs to be re-added:
image

Some reasons why we should consider leaving the changes the way they are:
* The tabs now load when opened, making for a slower user experience.
* We will likely introduce a couple of bugs with these changes, requiring further work
* What are we gaining with making this change?

maxime-rainville:
I agree that it's not definitely clear that this is something worth doing and that there is some risk of regression. The big plus I see for doing this are that it reduces/simplifies the code base we have to maintain in the long run and any future bug fix we apply to ModelAdmin will automatically be applied to SecurityAdmin.

maxime-rainville:
Might be worth waiting until we have some sizeable chunk of work to do the SecurityAdmin

emteknetnz:

maxime-rainville:
FYI I had a look at applying the Versioned extension to Member. The resulting Security section wasn't working great ... presumably because Security Admin is not a ModelAdmin.

https://github.com/silverstripe/silverstripe-framework/pull/9671

emteknetnz:
I'm not sure what that means in relation to this pull request?  Is this pull-request still useful and you'll do further work on this or is it something we should now close?

maxime-rainville:
I think the pull request is still useful. Maybe I should have created a parent issue with the overarching context.

The main problem is that SecurityAdmin reimplement all the features of ModelAdmin without being a ModelAdmin. This increase the amount of code we have to maintain and increase the risk that we'll introduce new bugs because we have two classes to maintain that do exactly the same thing.

maxime-rainville:
Current assumption is that there's nothing controversial about this. It would be a good way to reduce technical debt.

emteknetnz:
Closed in favor of https://github.com/silverstripe/silverstripe-admin/pull/1359

PR #964, opened 2019-10-09
NEW: Add Tip UI to InputField

Cheddam:
This PR introduces the Tip UI displayed in the Alternative text and Title attribute fields in the [File details DSM document](https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5d70ab0cd81b4b445753d31b). It goes hand-in-hand with a [PR to framework](https://github.com/silverstripe/silverstripe-framework/pull/9281) adding support for the required schema to the `TextField` field class.

It also adds Markdown document processing to the Storybook Webpack configuration to enable expansive documentation for components in the Pattern Library, and updates the icon font to include three new icons: `lamp`, `flag`, and `ticket`.


client/src/components/InputField/InputField.js
```diff
@@ -1,12 +1,34 @@
...
+const tipImportanceColorMap = {
+  normal: 'muted',
+  high: 'danger',
+};
+
+const tipImportanceMessageMap = {
+  normal: 'Tip',
+  high: 'Important tip',
+};
+
```
maxime-rainville: Why not combine there two objects into one? The keys probably need to be const and the message need to be translatable.
```suggestion
export const TIP_IMPORTANCES = {
  NORMAL: 'normal',
  HIGH: 'danger'
}

const tipMap = {
  [TIP_IMPORTANCES.NORMAL]: {
    color: 'muted',
    message: i18n._t('Admin.TIP_IMPORTANCE_NORMAL', 'Tip')
  },
  [TIP_IMPORTANCES.HIGH]: {
    color: 'danger',
    message: i18n._t('Admin.TIP_IMPORTANCE_HIGH', 'Important tip')
};

const tipImportanceMessageMap = {
  normal: 'Tip',
  high: 'Important tip',
};
```

client/src/components/InputField/InputField.js
```diff
@@ -55,7 +77,58 @@ class InputField extends Component {
...
+  }
+
+  renderTip() {
+    const id = this.props.id || this.props.name;
+    const title = this.props.title;
+    const { icon = 'lamp', importance = 'normal', content } = this.props.tip;
+    const { tipOpen } = this.state;
+
+    const tipIconColor = tipImportanceColorMap[importance];
+    const tipType = tipImportanceMessageMap[importance];
```
maxime-rainville: Feels like those values should be customisable.

client/src/components/InputField/InputField.js
```diff
@@ -55,7 +77,58 @@ class InputField extends Component {
     }
   }
 
+  handleTipToggle() {
+    this.setState((state) => ({ tipOpen: !state.tipOpen }));
```
maxime-rainville: Right now the tip only goes away when I click on the toggle button. Feels like it should be dismissed when the main field loose focus or when the user hits the ESC key.

client/src/components/InputField/InputField.js
```diff
@@ -72,6 +145,11 @@ InputField.propTypes = {
   type: PropTypes.string,
   autoFocus: PropTypes.bool,
   attributes: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
+  tip: PropTypes.shape({
+    content: PropTypes.string.isRequired,
+    importance: PropTypes.string,
```
maxime-rainville: If the list of possible values is known up front, then important should be a oneOf condition. Although I suspect we may want to add more possible values down the road.

client/src/components/TextField/tests/TextField-story.js
```diff
@@ -41,4 +51,21 @@ storiesOf('Admin/TextField', module)
...
+    withNotes(
+      withTipNotes
+    )(
+      () => (
+        <div style={{ margin: '5em', width: '30em' }}>
+          <TextField
+            {...props}
+            value="Default text"
+            description="Field description"
+            tip={{ icon: 'lamp', importance: 'high', content: 'Additional information about the field' }}
```
maxime-rainville: Could have use some knobs here, so the user can try out different values.

client/src/components/InputField/InputField.js
Cheddam: You can customise the icon, but the colour and description are baked into the importance levels for consistency. @clarkepaul - any thoughts on whether the icon colour and accessible description of the button should be directly modifiable?

client/src/components/InputField/InputField.js
Cheddam: This was the interaction pattern agreed upon with @clarkepaul, but I could add an additional event if it makes sense (though this would make it tougher to extract into a separate component).

client/src/components/InputField/InputField.js
Cheddam: Excellent point - I've implemented this change in the new `Tip` component.

client/src/components/InputField/InputField.js
Cheddam: Definitely - I'm using the `TIP_IMPORTANCE_LEVELS` constant with oneOf to restrict this prop now.

client/src/components/TextField/tests/TextField-story.js
Cheddam: I've applied knobs across the whole `TextField` story, along with the new `Tip` story.

client/src/components/InputField/InputField.js
clarkepaul: I think we should only allow colours that we recommend for select purposes. If they are completely customisable you'll get people using all sorts of random colours which won't have much meaning for the end-user and will contradict the other system colours. The colours we provide should align with system alerts, badges... that sort of thing. Happy to add new colours as they are needed, so trying to keep them as few as possible to start with.

client/src/components/InputField/InputField.js
clarkepaul: Originally I wanted it to open when either the field or button was in focus, naturally that when either was no longer in focus the tooltip would close. After discussing it seemed it was going to complicate things so opted with the simplest form to get it in. If adding the functionality to close the tip when the focus is lost that would be great.

Cheddam: I pushed for the limited triggers to make sure the popover behaved appropriately for screenreaders - if it opened as soon as the field had focus, it would trample the announcement for "field gained focus" with the popover's contents. (If we could use passive mode, this wouldn't happen, but that results in the popover's contents being read aloud twice by VoiceOver for some reason).

I'll look into having the popover close when both the field and button have lost focus.

client/src/components/Tip/tests/Tip-story.js
```diff
@@ -0,0 +1,48 @@
...
+  title: 'Field title',
+  placeholder: 'Placeholder text',
+};
+
+storiesOf('Admin/Tip', module)
+  .addDecorator(withKnobs)
+  .addDecorator((storyFn) => (
+    <ValueTracker>{storyFn()}</ValueTracker>
+  ))
+  .addWithJSX(
```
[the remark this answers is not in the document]
maxime-rainville: e.g.:
```js
  .addWithJSX(
    'FormField example',
    withNotes(notes)(() => {
      const tip = {
        content: text('Content', 'Example tip contents'),
        icon: selectV2('Icon', ['lamp', 'flag', 'ticket']),
        importance: selectV2('Importance', importances, 'normal')
      };

      return <TextField tip={tip} />;
    }

  ));
```

client/src/components/Tip/Tip.js
```diff
@@ -0,0 +1,86 @@
...
+
+    return [
+      (
+        <Button
+          key={`${id}-tip-button`}
+          color="outline-secondary"
+          id={`${id}-tip`}
+          onClick={this.handleTipToggle}
+          className={`btn--no-text btn--last font-icon-${icon} text-${tipIconColor}`}
+          aria-label={`${tipType} for ${fieldTitle}`}
```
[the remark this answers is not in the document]
Cheddam: I'll localise it, that was an oversight, but I disagree that it should be provided by the parent - this description is specifically tied into the importance of the tip and it can already be influenced via the `fieldTitle` prop.

client/src/components/Tip/Tip.js
```diff
@@ -0,0 +1,86 @@
...
+
+  handleTipToggle() {
+    this.setState((state) => ({ open: !state.open }));
+  }
+
+  render() {
+    const { content, fieldTitle, icon, id, importance } = this.props;
+    const { open } = this.state;
+
+    const tipIconColor = tipImportanceMap[importance].color;
```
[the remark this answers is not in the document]
Cheddam: They're not super descriptive names for their purpose - `description` could refer to the message in the popover, and `color` could refer to pretty much any color in the button / popover.

That said, whilst introducing translation for the field I ended up adjusting the names and syntax a bit anyway.

client/src/components/Tip/tests/Tip-story.js
```diff
@@ -0,0 +1,48 @@
...
+        <div style={{ margin: '5em', width: '30em' }}>
+          <InputGroup>
+            <Input {...inputProps} />
+            <InputGroupAddon addonType="append">
+              <Tip
+                fieldTitle={'Tip Field'}
+                id={'tip-field'}
+                content={text('Content', 'Example tip contents')}
+                icon={text('Icon', 'lamp')}
+                importance={selectV2('Importance', Object.values(TIP_IMPORTANCE_LEVELS), 'normal')}
```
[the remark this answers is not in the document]
Cheddam: That works nicely, thanks!

client/src/components/Tip/tests/Tip-story.js
```diff
@@ -0,0 +1,48 @@
...
+      () => (
+        <div style={{ margin: '5em', width: '30em' }}>
+          <InputGroup>
+            <Input {...inputProps} />
+            <InputGroupAddon addonType="append">
+              <Tip
+                fieldTitle={'Tip Field'}
+                id={'tip-field'}
+                content={text('Content', 'Example tip contents')}
+                icon={text('Icon', 'lamp')}
```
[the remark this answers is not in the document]
Cheddam: `flag` and `ticket` only come in with this PR because they were introduced to the icon font between the last update and now and I needed to pull in `lamp`. I'll switch this to a select list anyway, and add a couple more example icons that would make sense as alternatives.

client/src/components/Tip/tests/Tip-story.js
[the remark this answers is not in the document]
Cheddam: Hmm, I'm not sure this adds value - the JSX tab will output the `FieldHolder` component and nothing else, plus it'll look strange because there are no useful props (aside from the `tip`) passed to the `TextField`. I could reintroduce the 'With Tip' example in the TextField story though?

client/src/components/InputField/InputField.js
Cheddam: I've opted not to pursue the focus-change behaviour, as it would be too time-intensive to find a good pattern for it. We can look into implementing this in the future.

Cheddam:
@maxime-rainville I've implemented the vast majority of your feedback (with some caveats in the comments) - the only outstanding item is the alternative use-case suggestion, which I've come back on.

I've also added the component to our expose loader / component registration; I'm not sure what purpose the declaration in the external webpack-config module has? Ideally we wouldn't need to manually keep the two modules in sync...

client/src/components/Tip/tests/Tip-story.js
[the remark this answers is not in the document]
maxime-rainville: It was more on the nice to have side of things.

maxime-rainville:
From memory, if you don't add the `<Tip/>` component in the webpack-config external, other module (e.g.: asset-admin) won't be able to reference the it.

All my annoying points have been properly addressed.

PR #968, opened 2019-10-16
Add some wrappers around the generic ReactStrap button

maxime-rainville:
Add SS specific Button that accepts a `noText` and `icon` props. It also adds matching pattern-lib stories and a reusable BackButton.

# Related PR
* https://github.com/silverstripe/webpack-config/pull/39


client/src/components/Button/IconHOC.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+/**
+ * Higher order component that adds a `icon` props to another component and translate
+ * it to a matching class.
+ */
+const IconHOC = (Component) => {
```
maxime-rainville: I made this an HOC component thinking we might end up reusing it on other components that need to display icons.

But it might be a bit overkill.

client/src/components/Button/BackButton.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
maxime-rainville: We have Back Button in many places. I figure that in these kind of cases it makes sense to define a generic Button that can be reuse in multiple spots. 

I've put it under the `components/Button` directory, which is a bit unusual compared to the rest of our components, but I think it makes sense in this scenario, Button and BackButton are closely related.

client/src/components/Button/IconHOC.js
Cheddam: This seems sensible. Are there any other examples where this would work currently?

client/src/components/Button/README.md
```diff
@@ -0,0 +1,53 @@
...
+  onClick: console.dir,
+  className: 'action__result',
+};
+
+<BackButton {...props} />
+```
+
+## Properties
+
+ * `children` (string): The content of the tag will be use as an `aria-label` if provided. Otherwise the localised string from 'Back' will be used.q
```
Cheddam:
```suggestion
 * `children` (string): If present, any nested content inside the button will apply to the `aria-label` attribute. Otherwise, the localised string 'Back' will be used.
```

PR #973, opened 2019-10-22
FIX Use arrow syntax to avoid unnamed function linting error

Cheddam:


Cheddam:
@robbieaverill Most likely - this wasn't intended to fix them, just to resolve a linting error. (We definitely need to get them green though!)

PR #983, opened 2019-10-23
Update chosen-js version in package.json

shoosah:
- This update fixes one of issues using `ListboxField` which would show the wrong suggestion(s) upon typing special characters like dollar sign, single quote, etc

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "apollo-link-state": "^0.4.1",
     "babel-polyfill": "^6.26.0",
     "bootstrap": "^4.3.1",
-    "chosen-js": "^1.5.1",
+    "chosen-js": "^1.8.3",
     "classnames": "^2.2.5",
     "deep-equal": "^1.0.1",
     "deep-freeze-strict": "^1.1.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -2306,9 +2306,10 @@ chokidar@^2.0.2:
   optionalDependencies:
     fsevents "^1.2.2"
 
-chosen-js@^1.5.1:
+chosen-js@^1.8.3:
   version "1.8.7"
   resolved "https://registry.yarnpkg.com/chosen-js/-/chosen-js-1.8.7.tgz#9bfa5597f5081d602ff4ae904af9aef33265bb1d"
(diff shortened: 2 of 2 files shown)


ScopeyNZ:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request).

ScopeyNZ:
The builds haven't completed yet but I see that there's no dist files bundled with this PR. We have to (unfortunately) commit our dist files into our repo, so you need to ensure that you run a `yarn build` and add the `client/dist` bundles to your commit. The CI builds will fail if the bundle is not provided.

shoosah:
Thanks @ScopeyNZ for your comments. There is no change after running `yarn build` for this PR and the CI build was successful (https://travis-ci.org/silverstripe/silverstripe-admin/builds/602077040).

ScopeyNZ:
Hmmn. This fixed an issue for you too? If the dist files didn't change then how does this affect a live site?

shoosah:
Yes, this was fixed in later versions of `chosen-js`. Since this has not been updated, I was getting errors using `ListboxField`. 
I use `$0-50` in my search filter which was using `ListboxField`. On typing `$` it shows the last character which is `0`. This behaviour was true for any text that starts with a special character.

ScopeyNZ:
I see. One last thing I just realised - you can target this at the 1.4 branch as a bugfix for the 1.4.x line if you want. Just change the base branch in GitHub and then `git rebase HEAD~1 --onto upstream/1.4 && git push --force-with-lease`.

shoosah:
Sure, I just did that @ScopeyNZ.

shoosah:
Ping!

ScopeyNZ:
I'm still really confused as to why this fixed any problem. The dist files aren't updated. But whatever, it's good to keep this stuff up to date.

PR #985, opened 2019-10-31
Fix toasts popping up forever

bergice:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/2474


bergice:
I've updated this with a new approach.

Since entwine doesn't handle asynchronous functions, the ajax call at `updateNodesFromServer` will complete immediately, setting the `IsUpdatingTree` flag back to false which is what caused the infinite loop.

I changed this so the flag remains unchanged inside the `async` method.

maxime-rainville:
Moving pages around doesn't quite seem to work all the time. Moving a page up in the hierarchy  seems to work all the time. But moving down seems to be hit and miss: the page gets updated in the hierarchy but doesn't get it's new modified state. And refreshing the page undoes the change.

maxime-rainville:
FYI here's what I tested:
* Renaming the home page segment
* Moving a page up and down the hierarchy
* Move the homepage up down the hierarchy
* Archiving a page
* Restoring a page
* publishing a page with sub pages.


I tested it in IE11

Issue #911, opened 2019-07-18
Re-enable ability to auto select a tab from a URL anchor


robbieaverill:
**SS 4**

I want to deep link to a GridField item's Settings tab (for example). Allegedly you used to be able to do this by appending `#Root_Settings` to the URL for it (maybe in SS3?), which would detect the anchor in the URL and automatically set it as the active tab.

This doesn't work in SS 4.4.x - there's a todo example [here](https://github.com/silverstripe/silverstripe-framework/blob/4/src/Security/Group.php#L231-L237) which indicates that it didn't work when that comment was written either.

At some stage this jQuery UI interface is probably going to be React driven anyway, but this might be a nice quick win addition in the meantime.

## PR's
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/1030/files
- [ ] ~https://github.com/silverstripe/silverstripe-admin/pull/986~

kinglozzer:
Confirmed this feature used to exist - it works on a 3.7.3 site

robbieaverill:
How should this work with nested tab sets? Could I do `#Root_Content-Submissions` or something to open Content > Submissions for a userforms page for example?

kinglozzer:
That’s a good question. I just checked in 3.7.3 and it doesn’t work at all...

I suppose ideally the anchor would still match the ID, so you can right click and copy a link to the nested tab. Do nested tabsets already generate IDs like your example?

robbieaverill:
The root tabset doesn't have an ID like that either as far as I can tell - it's `ui-id-3` etc, with a numeric index which is presumably specific to jQuery UI's handling of the tabs when it runs

robbieaverill:
I can't get this to work with pages (by trying `#Root_Settings`), or within a Member's GridField edit form (by trying `#Root_Permissions`) on SS 3.7.x.

In the jQuery UI library there is the following logic which should enable this functionality for us:

```js
		if ( active === null ) {
			// check the fragment identifier in the URL
			if ( locationHash ) {
				this.tabs.each(function( i, tab ) {
					if ( $( tab ).attr( "aria-controls" ) === locationHash ) {
						active = i;
						return false;
					}
				});
			}
```

In SS 4.x at least, this code does fire but at the point where it runs the `aria-controls` attributes haven't been added to the DOM yet, so it can't find a matching tab for the `locationHash`. If this logic were to be deferred until the `DOMContentLoaded` event, then it would work correctly.

[...]

PR #986, opened 2019-10-31
Allow targeting of a tab in GridFields via URL anchor

mlewis-everley:
This works towards fixing issue #911, it updates TabSet.js so that tabs inside GridFields can can be directly targeted via a URL anchor (which was possible in SS3, but lost in SS4).

I have tested this on two different installs, one that is framework only, the second is Framework + CMS. In both cases I could target A tab in a `Form` loaded from a `GridFieldDetailForm` (such as `SecurityAdmin` or `ModelAdmin`) or via `SiteConfig`.

The only thing missing is that I cannot get sub tabs inside a page to be targeted (EG the `Categorisation` tab in a `Blog` in `silverstripe-blog`).


emteknetnz:
I believe this is fixed in https://github.com/silverstripe/silverstripe-admin/pull/1030 which is pretty close to being merged.  I'll close this PR since it's a duplicate.

Issue #862, opened 2019-04-17
"Add" button is too small for loading indicator when adding a page to a campaign


robbieaverill:
**Version: 1.x-dev (1.4.x-dev)**

When adding a page to a campaign, the "Add" button is not wide enough to contain the loading indicator. It squashes together when you click on it.

Changing "Add" to "Add to campaign", or anything longer, fixes the issue.

robbieaverill:
70px min-width seems like the minimum required to keep the loading indicator on a single line, but that also means the button grows slightly in size when you click it.

robbieaverill:
70px min-width seems like the minimum required to keep the loading indicator on a single line, but that also means the button grows slightly in size when you click it.

robbieaverill:
Still an issue in 4.4.3

brynwhyman:
Just came across this too... Here's a GIF of current behaviour:
![current-loading](https://user-images.githubusercontent.com/15868440/63905008-f22cad80-ca66-11e9-86de-3ec9b64bbe57.gif)

maxime-rainville:
Still an issue in 4.5-rc1. Got a PR https://github.com/silverstripe/silverstripe-admin/pull/993

PR #993, opened 2019-11-18
BUG Set min-width on loading button to avoid having the loading indicator break over 2 lines

maxime-rainville:
This give the loading indicator button a 

It's arguably a bit weird to have the [add] button suddenly get a bit wider when you click on it. But it's less weird than having it break over two lines with a broken animation. So I'm calling this an improvement.

I don't think we can apply a 70px min-width to all buttons, because that would probably end up making some button out of place that don't actually need the loading indicator.

# Parent issue
* https://github.com/silverstripe/silverstripe-admin/issues/862


robbieaverill:
Does this work for batch action buttons?

maxime-rainville:
Which batch action are we talking about? The CMS one and Asset-Admin one don't seem to be displaying any loading animation.

robbieaverill:
Cool, yeah I meant the CMS ones

maxime-rainville:
There's a really old bug open to do add some feedback when performing a batch CMS actions. Probably one of those thing that will need to wait until we migrate the sitetree UI to react. https://github.com/silverstripe/silverstripe-cms/issues/441

PR #995, opened 2019-11-19
API Break up ModelAdmin::getEditForm into getGridField and getGridFieldConfig

maxime-rainville:
This is a sort of RFC/PR.

Many times over the years, I wanted to do a minor tweak to a ModelAdmin gridfield. However, to actually get to the instance of your GridiField, you got to override getEditForm and then fetch your gridfield instance, which will have a different name base on what model the user is currently viewing.

Basically it's unruly and not a very nice developer experience.

My thinking is let's split the ModelAdmin `getEditForm` method into smaller protected methods which will allow developers to override the specific bits they want to customise:
* `getGridField` for the actual field ;
* `getGridFieldConfig` for the config part.

As an added benefit, I also think this makes the code more readable.

We might also want to add a few extra extension points for Extensions to hook into.

Following our wonderful discussion about "public vs private vs protected", I think those new methods should be considered part of our official API. I haven't written unit tests for them.

The point of this PR is to demonstrate that this change is pretty trivial and compatible with Semver. If people think this is worthwhile, I can add some unit tests to for the new protected methods.

# Related PR
* https://github.com/silverstripe/silverstripe-framework/pull/9334 (adds relevant doc)

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -198,25 +198,31 @@ abstract class ModelAdmin extends LeftAndMain
     /**
      * Generate the GridField field that will be used for this ModelAdmin.
      *
-     * Developers may override this method in their ModelAdmin class to customise their GridField.
+     * Developers may override this method in their ModelAdmin class to customise their GridField. Extensions can use
+     * the `updateGridField` hook for the same purpose.
      *
      * @see {@link getGridFieldConfig()}
      * @return GridField
      */
     protected function getGridField(): GridField
     {
-        return GridField::create(
+        $field = GridField::create(
             $this->sanitiseClassName($this->modelClass),
             false,
             $this->getList(),
             $this->getGridFieldConfig()
         );
+
+        $this->extend('updateGridField', $field);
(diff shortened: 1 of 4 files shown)


robbieaverill:
Yes please! Modifying GridFields in ModelAdmin could definitely be made easier!

maxime-rainville:
I added unit tests and some Extension hooks. If people like the approach, I'll write some docs about how to use this.

stevie-mayhew:
This is a great idea, would save a lot of headaches where you end up implementing your own `getEditForm` in place of something like an extension like this. I'd write the tests, I like the approach.

michalkleiner:
Yes, please! ❤️

maxime-rainville:
I added some docs to explain how to use the new changes https://github.com/silverstripe/silverstripe-framework/pull/9334

That's all the changes I can think off. Please go ahead and review the finer details.

stevie-mayhew:
Thanks @maxime-rainville - looks good!

PR #997, opened 2019-11-26
BUG case error in SelectionGroup template

MasonD:
The SelectionGroup template uses `$Fieldholder` which does get the result of the `FieldHolder` method but doesn't get automatically cast to `HTMLFragment` because casting is case-sensitive. This fixes that.


robbieaverill:
The Behat test is already failing in master, so merging. Thanks for the patch @MasonD

robbieaverill:
I've noticed that you've created this pull request from one of the "core branches" rather than a branch specifically intended for this pull request. This can cause problems if we ask you to target this PR at a different branch. No action is needed at the moment, this is for your information.

In the future please create a new branch for each individual pull request. See our [guide about contributing PRs](https://docs.silverstripe.org/en/4/contributing/code/#step-by-step-from-forking-to-sending-the-pull-request)

PR #999, opened 2019-12-09
NEW Export url lib in admin vendor bundle

maxime-rainville:
Exports the `url` lib to the vendor bundle.

Depends on https://github.com/silverstripe/webpack-config/pull/43. Once the webpack-config PR is merged and release to NPM, we'll need to increment to recompile this one.

# Parent issue
* https://github.com/silverstripe/silverstripe-versioned-admin/issues/134


robbieaverill:
Is this necessary? I would've thought that modules could install this package if they need it. Does it have SilverStripe context in it somewhere that needs to be centralized?

maxime-rainville:
If we don't expose those libraries out of admin, we risk bundling them in all the other libraries ... which mean we'll end up shipping them multiple times. So the point is to reduce our total bundle size.

robbieaverill:
I think that's kind of expected, until a time where SilverStripe's frontend bundles are pre-compiled to avoid this problem. I guess for now it's OK to do this though.

PR #1001, opened 2019-12-19
BUG Upgrade webpack config to 1.4

maxime-rainville:
# Parent issue
* https://github.com/silverstripe/silverstripe-versioned-admin/issues/134


maxime-rainville:
Oups ... Sorry, didn't mean to merge that before a peer reivew. It's a simple JS dep upgrade. If someone feels like kicking a fuss, I revert it.

Issue #455, opened 2018-03-05
Add "copy" or "duplicate" icon


dhensby:
Please can we add a duplicate or copy icon to the SS font icons set as this is a pretty handy icon to have specifically for the https://github.com/unisolutions/silverstripe-copybutton module

newleeland:
I've added a clone icon to fontastic. 

This is the CSS mapping:
![Screen Shot 2019-12-17 at 4 46 31 PM](https://user-images.githubusercontent.com/8085295/70963472-68741f00-20ed-11ea-8ed9-29ef995ef525.png)

This is the character mapping:
![Screen Shot 2019-12-17 at 4 51 23 PM](https://user-images.githubusercontent.com/8085295/70963499-83469380-20ed-11ea-826f-f3dcb4bb1b65.png)

sachajudd:
Cool this will have to be updated in admin e.g https://github.com/silverstripe/silverstripe-admin/pull/633 
I'll be adding a new block icon set early next year so can do this at the same time.

sachajudd:
Clone icon added here: https://github.com/silverstripe/silverstripe-admin/pull/1003

PR #1003, opened 2020-01-15
NEW Add new block icon set for open source use

sachajudd:
This pull request adds a new set of (44) block icons for the open source community. This is due to the demand of new block types. The icons will provide more variety and help the community use the correct 'block-icons' due to the lack of icons provided for their block types. After chatting with @Cheddam although this is technically considered a 'feature', and as we likely aren't doing another minor release until June (at this point) we feel that this could be including in a patch release due to the icons limited scope, as long as we clearly state in the patch notes for 4.5.1.
Happy to discuss this further.

New block icon set:
new-block-icon-set

cc @silverstripeux 

Also:
Closes: #455


kinglozzer:
I’m personally happy for this to be merged into a patch release. Part of the motivation for new features going in minor releases in semver is to give them time to mature and discover potential bugs/regressions... there’s no risk of that with icons, so I think we’re fine to make an exception here

client/src/styles/_fonts.scss
```diff
@@ -774,7 +765,202 @@
   content: "\e05d";
 }
 
-.font-icon-ticket:before {
```
ScopeyNZ: Has `.font-icon-ticket` been removed?

sachajudd: Yes, it was accidentally added for a very short amount of time in this PR https://github.com/silverstripe/silverstripe-admin/pull/964. I doubt anyone is using it but I can add it back if it causes any concern.

ScopeyNZ: To be honest I don't like claiming that our icons are part of our API anyway. So it's fine by me

sachajudd:
If there are any other icons anyone would like to have available in the near future please let me know. Alternatively we can advise the best fit icon for a particular block you have in mind. Cheers!

robbieaverill:
Just popping in to say that I also support adding new fonts in patch releases, it helps get them to people quicker, which is useful for community module developers.

If you needed a new font, you could set your minimum silverstripe/admin requirement to `^1.5.2` (a patch release constraint) instead of a new minor version - I think that'll be less restrictive for people, and I agree with Guy that I don't think icons are part of our semver obligations.

PR #1013, opened 2020-02-18
DEPENDENCY: Replace getCMSValidator with new framework method getCompositeValidator

chrispenny:
# Dependent PRs (to be merge before this)

- [x] https://github.com/silverstripe/silverstripe-framework/pull/9406

# Change

As part of the above PR, there is now a standardised method (`getValidatorList()`) for adding/manipulating `Validators` for a `DataObject`.

This new method provides BC for the old `getCMSValidator()` method.

# Tests

Tests are expectedly failing, as I have replaced call to `getCMSValidator()` with `getValidatorList()`, which won't exist until the above dependent PR has been merged.

I have, however, run the tests locally (with the dependent PR checked out), and all tests are currently passing.

Screen Shot 2020-02-19 at 11 26 03 AM

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -1442,7 +1442,7 @@ class LeftAndMain extends Controller implements PermissionProvider
         $form->addExtraClass('fill-height');
 
         // A CompositeValidator is always available form a DataObject, but it may be empty (which is fine)
-        $form->setValidator($record->getCompositeValidator());
+        $form->setValidator($record->getCMSCompositeValidator());
 
         // Check if this form is readonly
         if (!$record->canEdit()) {
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -1441,8 +1441,8 @@ class LeftAndMain extends Controller implements PermissionProvider
         }
         $form->addExtraClass('fill-height');
 
-        // A ValidatorList is always available form a DataObject, but it may be empty (which is fine)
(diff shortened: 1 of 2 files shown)


robbieaverill:
Upstream is merged, I've restarted the builds

code/LeftAndMain.php
```diff
@@ -1441,18 +1441,8 @@ public function getEditForm($id = null, $fields = null)
...
-            // which can be evaluated as javascript, hence we need
-            // to override any global changes to the validation handler.
-            if ($validator) {
-                $form->setValidator($validator);
-            }
-        } else {
-            $form->unsetValidator();
-        }
+        // A CompositeValidator is always available form a DataObject, but it may be empty (which is fine)
+        $form->setValidator($record->getCMSCompositeValidator());
```
robbieaverill, line 1445: Whoops, I merged this too early =( we either need to check whether this method exists first, or bump the Composer constraint on framework so it always will. Would you mind doing one of those two things?

chrispenny: Ah cool. Yea, I can do this now. I'll raise a new PR.

chrispenny: PR raised here:
https://github.com/silverstripe/silverstripe-admin/pull/1080

I opted for adding BC support, rather than updating the constraint - I felt that was more forgiving if someone has pinned themselves to a particular framework version.

PR #1019, opened 2020-03-06
FIX Rename exposed url module to node-url to avoid API clash

Cheddam:
Browsers already have a `URL` key on their `window` object, so this clashes with the expose-loader entry and makes it unusable.

Pairs with an adjustment to the externals config in `@silverstripe/webpack-config`: silverstripe/webpack-config#44


robbieaverill:
You probably need to tag/release/update webpack-config in this PR as well?

maxime-rainville:
Just published a new version of webpack-config. We'll probably need to bump which version of we're using on this PR.

maxime-rainville:
Merged on Green

PR #1027, opened 2020-04-25
ENH Make CMSProfileController use required_permission_codes

mattclegg:
It appears that CMSProfileController is hardcoded to check for the permission "CMS_ACCESS", but this prevents updating permissions via _config/yaml files.

I've refactored the function to use the same permission login & incorporate the use of $required_permission_codes.

Any feedback welcome.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1537

### DIFF
--- a/code/CMSProfileController.php
+++ b/code/CMSProfileController.php
@@ -18,7 +18,7 @@ class CMSProfileController extends LeftAndMain
 
     private static $menu_title = 'My Profile';
 
-    private static $required_permission_codes = false;
+    private static $required_permission_codes = 'CMS_ACCESS';
 
     private static $tree_class = Member::class;
 
@@ -59,8 +59,10 @@ class CMSProfileController extends LeftAndMain
 
     public function canView($member = null)
     {
+        $currentUser = Security::getCurrentUser();
+
         if (!$member && $member !== false) {
-            $member = Security::getCurrentUser();
+            $member = $currentUser;
         }
 
         // cms menus only for logged-in members
@@ -68,14 +70,8 @@ class CMSProfileController extends LeftAndMain
             return false;
         }
 
-        // Check they can access the CMS and that they are trying to edit themselves
-        if (Permission::checkMember($member, "CMS_ACCESS")
-            && $member->ID === Security::getCurrentUser()->ID
-        ) {
-            return true;
-        }
-
-        return false;
+        // Check they are trying to edit themselves and they have permissions
+        return $member->ID === $currentUser->ID && parent::canView($member);
     }
 
     public function save(array $data, Form $form): HTTPResponse


dhensby:
Behat failures - probably related - can you investigate?

mattclegg:
@dhensby I think this is unrelated to this change as the tests for the `master` branch are already failing; https://travis-ci.org/github/silverstripe/silverstripe-admin/branches

Just to clarify, should PRs only be made to the default branch "1"?

dhensby:
Ah - I think I had missed this was against master

code/CMSProfileController.php
```diff
@@ -67,11 +67,21 @@ public function canView($member = null)
...
+        if ($member->ID === Security::getCurrentUser()->ID) {
+            $codes = $this->getRequiredPermissions();
+
+            // allow explicit TRUE
+            if ($codes === true) {
+                return true;
+            }
+
+            // Check they can access self::$required_permission_codes
+            foreach ((array)$codes as $code) {
```
dhensby, line 80: I'm not sure I like this style of forcing an array. I think that `getRequiredPermissions` should return a consistent API which is always an array.

dhensby (changes requested):

code/CMSProfileController.php
mattclegg: I'm not sure it's always an array. I think it could return an array, a string, or a bool; https://github.com/silverstripe/silverstripe-admin/blob/1/code/LeftAndMain.php#L578 ?

Looking at this again now; I also don't like the style of forcing an array. Would you be ok with a simple `is_array` test?

dhensby: Sure, us_array would be fine

mattclegg: Actually `Permission::checkMember()` will accept an array so I've tidied this up.

mattclegg:
@dhensby I've rebased this from the default branch. Can you review it again, please?

dhensby (commented):
OK - I think this looks good but we should have a test that makes sure if this config value is changed then a member needs the correct permission to gain access.

There should also be a note made in the changelog for this

emteknetnz:
This has a change to a `private static` so this would be an API change and would require a major release? e.g. should target `master` instead of `1`?

mattclegg:
How do you define if it's to `master` or `1`? Are there some guidelines somewhere?

dhensby:
I think we can take this change in without the change to the default value of the config value. That change should probably go into master

mattclegg:
That sounds fine to me but wouldn't splitting it into 2 seperate changes break existing functionality?

dhensby:
Looking at `LeftAndMain` the `required_permission_codes` is meant to be an array, so we should change that.

An explicit `false` value also has a meaning which is "any one with CMS access".

So actually, looking at `LeftAndMain::canView()` - I think this change should actually be a check on the current user or parent::canView`

code/CMSProfileController.php
```diff
@@ -67,11 +67,13 @@ public function canView($member = null)
...
-            && $member->ID === Security::getCurrentUser()->ID
-        ) {
-            return true;
+        // Check they are trying to edit themselves
+        if ($member->ID === Security::getCurrentUser()->ID) {
+            // Check they can access via self::$required_permission_codes
+            $codes = $this->getRequiredPermissions();
+
+            // allow explicit TRUE || default to Permission::checkMember()
+            return $codes === true || Permission::checkMember($member, $codes);
```
dhensby, line 76: $codes should be an array - but this logic is actually defined in the parent function, so I think this whole thing can be changed to:

`return $member->ID === Security::getCurrentUser()->ID && parent::canView($member);`

dhensby (changes requested):

GuySartorelli:
I've rebased and retargetted this against `2` and made the requested change. I think this has some clear if fairly low impact value.

michalkleiner (approved):

GuySartorelli (approved):
Sort-of self merging here, but given there has been a lot of review on this and another core committer has approved I think that's okay.

PR #1028, opened 2020-04-27
BUGFIX: Change event with GET parameters in Grid field reload

### LINKED ISSUE #1026: Grid field reload always triggers change event if request has GET parameters
The grid field reload function always triggers a change event if the request has any GET parameters, even if the calling function specifies `triggerChange = false` in the `ajaxOpts` parameter. This is caused by the way `ajaxOpts.data` is made into a search string in this case, in `client/src/legacy/GridField.js`, line 67:
```
      // Include any GET parameters from the current URL, as the view state might depend on it.
      // For example, a list prefiltered through external search criteria might be passed to GridField.
      if(window.location.search) {
        ajaxOpts.data = window.location.search.replace(/^\?/, '') + '&' + $.param(ajaxOpts.data);
      }
```

The implication of this bug is that if any GET parameters are used, then all grid field actions (including search and filtering) cause the changetracker to trigger, and opening any grid field item then shows the "Are you sure..."-alert dialog, which really confuses our users.

The grid state is also included as a GET parameter when opening an item in `ModelAdmin`, so this bug is triggered any time a `DataObject` subclass has a `GridField` of it's own in `getCMSFields` and the user uses any grid field actions on that `GridField`.

This could be fixed by splitting `window.location.search` into an array, and combining it into `ajaxOpts.data`.

forsdahl:
Fixes #1026


client/src/legacy/GridField.js
```diff
@@ -64,7 +64,13 @@ $.entwine('ss', function($) {
       // Include any GET parameters from the current URL, as the view state might depend on it.
       // For example, a list prefiltered through external search criteria might be passed to GridField.
       if(window.location.search) {
-        ajaxOpts.data = window.location.search.replace(/^\?/, '') + '&' + $.param(ajaxOpts.data);
+        let searchParams = window.location.search.replace('?', '').split('&');
+        for (let i = 0; i < searchParams.length; i++) {
+          let parts = searchParams[i].split('=');
+          if (parts.length == 2) {
+            ajaxOpts.data.push({name: parts[0], value: parts[1]});
```
dhensby, line 71: Don't these need url decoding or does `wiondow.location.search` come url decoded?

forsdahl: `window.location.search` doesn't seem to be url encoded, at least not in Chrome of Firefox. I thought about using `URLSearchParams` for a bit cleaner code, but that is not supported in all legacy browsers (read IE11).

dhensby: I've done a little research and it looks like you do need to url decode the data: https://stackoverflow.com/a/2091331/4978044

client/src/legacy/GridField.js
```diff
@@ -64,7 +64,13 @@ $.entwine('ss', function($) {
       // Include any GET parameters from the current URL, as the view state might depend on it.
       // For example, a list prefiltered through external search criteria might be passed to GridField.
       if(window.location.search) {
-        ajaxOpts.data = window.location.search.replace(/^\?/, '') + '&' + $.param(ajaxOpts.data);
+        let searchParams = window.location.search.replace('?', '').split('&');
```
dhensby, line 67: I think we need to keep the `/^\?/` regex because URLs can contain `?` that aren't at the beginning

forsdahl: Added in commit #2a3bf4c

client/src/legacy/GridField.js
forsdahl: Ah, you are correct. Fixed in commit #2a3bf4c.

dhensby:
OK - there are some linting issues and the dist files need building

forsdahl:
Fork rebased, and dist files built. Didn't find any linting issues, travis build fail might have been caused by outdated fork.

dhensby:
Fantastic, thanks

PR #1030, opened 2020-05-01
Fix URL anchors not selecting tabs

### LINKED ISSUE #911: Re-enable ability to auto select a tab from a URL anchor

bergice:
For example `/admin/security/EditForm/field/Members/item/1/edit#Root_Permissions` should open the Permissions tab.

Resolves #911 - https://github.com/silverstripe/silverstripe-admin/issues/911

*Note that this change does not update the hashtag of the URL when clicking on the tabs. That should be done in a separate issue.*


michalkleiner:
🙇

client/src/legacy/TabSet.js
```diff
@@ -8,100 +8,108 @@ require('../../../thirdparty/jquery-entwine/dist/jquery.entwine-dist.js');
...
+      if(this.data('tabs')) this.tabs('destroy');
+      this._super();
+    },
 
     redrawTabs: function() {
-      this.rewriteHashlinks();
-      this.tabs();
+      // If accessing a tabset from a page in the CMS module
+      // alter how the tabs are rendered
+      if ($(this).hasClass('ss-tabset') == true) {
```
emteknetnz, line 47: can remove ` == true`

client/src/legacy/TabSet.js
```diff
@@ -8,100 +8,108 @@ require('../../../thirdparty/jquery-entwine/dist/jquery.entwine-dist.js');
...
+        }
+      });
+
+      // If there's no tab, it means the hash is invalid, so do nothing.
+      if ($trigger === void 0) {
+        return;
+      }
+
+      // Switch to the correct tab when the page is loaded
+      $(document).ready(function () {
```
emteknetnz, line 78: could change to shorthand i..e `() => {` though I'm not too fussed

bergice:
> Very minor recommended changes
> 
> Failing behat tests that will need fixing

I fixed the behat test by disabling this logic for `.cms-tabset` as the tabs use unique URL's anyways.

emteknetnz:
@bergice Tested using firefox & chrome, though this does not seem to fix the original issue in either browser.  I navigated to `/admin/security/EditForm/field/Members/item/1/edit#Root_Permissions` and it did not go to the 'Permissions' tab.  Just in case, I tried running `yarn build`, though the there was no diff for bundle.js

bergice:
> @bergice Tested using firefox & chrome, though this does not seem to fix the original issue in either browser. I navigated to `/admin/security/EditForm/field/Members/item/1/edit#Root_Permissions` and it did not go to the 'Permissions' tab. Just in case, I tried running `yarn build`, though the there was no diff for bundle.js

@emteknetnz I've fixed this now. The test should also be passing.


Some URL's I've tested:

* http://localhost:8888/admin/security/groups#Root_Groups
* http://localhost:8888/admin/security/EditForm/field/Groups/item/1#Root_Roles
* http://localhost:8888/admin/pages/settings/show/10#Root_ContentReview

PR #1032, opened 2020-05-04
NEW Enable TreeDropdownField in pattern library, reorder addon tabs

### LINKED ISSUE #462: Pattern Library for TreeDropdownField
# Acceptance Criteria

>As a frontend dev and UX designer, I can use the pattern library as a complete reference for visual and behaviour variations of TreeDropdownField

# Notes

 * Track under ORB-175

# PRs

- [ ] #1032

Cheddam:
This component was listed in the pattern library, but not yet rendering, as it has some complexities (the biggest one being a pretty hard dependency on Redux.) I've wrapped it in a small store so it can keep track of its state, and made the `fetch` dependency in the component overridable via a prop so that we can mock API responses for the Storybook demo.

I've also reordered the tabs in Storybook, so that docs render first. Not many components _have_ docs yet, but I want to encourage use of the tab and make it more visible when information is available. If this is sufficiently controversial, I can back that part out.

Resolves #462.


client/src/components/TreeDropdownField/tests/TreeDropdownField-story.js
```diff
@@ -1,35 +1,50 @@
...
-/*
+import notes from '../README.md';
+import treeDropdownFieldReducer from 'state/treeDropdownField/TreeDropdownFieldReducer';
+import { ConnectedTreeDropdownField as TreeDropdownField } from 'components/TreeDropdownField/TreeDropdownField';
+import mockTree from './mockTree';
+
+// TreeDropdownField depends on having a basic Redux store present to operate
+const store = createStore(
+  combineReducers({ treeDropdownField: treeDropdownFieldReducer }),
+  window.__REDUX_DEVTOOLS_EXTENSION__ && window.__REDUX_DEVTOOLS_EXTENSION__()
```
emteknetnz, line 16: is it reasonable to assume that everyone who would view this, including front-end designers, would have `__REDUX_DEVTOOLS_EXTENSION__` installed?

would it not just make more sense to create our own "basic Redux store"?

client/src/components/TreeDropdownField/tests/TreeDropdownField-story.js
```diff
@@ -1,35 +1,50 @@
...
+);
+
+// Returning the mockTree in the fetch mock allows us to manipulate the length of the loading state
+const fetchMock = () => new Promise(resolve => {
+  setTimeout(() => resolve({ json: () => mockTree }), 500);
```
emteknetnz, line 21: only 4 functions? I think we need at least 5 :-)

emteknetnz: does it need to be 500? does it still function just as well if it's 0?

if it needs to be 500 or more, I'm just worried there may be a bit of a lottery of it working or not depending on if it loads too quickly / too slowly compared to other stuff that gets loaded in

client/src/components/TreeDropdownField/tests/TreeDropdownField-story.js
ScopeyNZ: The first part of this statement (`window.__REDUX_DEVTOOLS_EXTENSION__ &&`) ensures that it only runs if that property exists. This is fine. Also, it's a debug function that's provided by the redux devtools extension that should be run to hook up your redux store to the devtools extension, so this is still just a "basic Redux store".

This LGTM.

client/src/components/TreeDropdownField/tests/TreeDropdownField-story.js
ScopeyNZ: I agree, shouldn't this just read `resolve => resolve({ json: Promise.resolve(mockTree) })`?

client/src/components/TreeDropdownField/tests/TreeDropdownField-story.js
emteknetnz: yeah i mean it requires people viewing this to actually have that extension installed in their chrome/firefox.  I wouldn't expect a designer to have this installed

Cheddam: Seeing the state changes for a component will almost certainly be irrelevant to designers,  and developers working with the state layer in Silverstripe CMS generally have the plugin installed for other diagnostic purposes, so I don't see any reason to try and bake it in at this stage. I'll add a note to the README stating that it's driven by Redux, though.

ScopeyNZ: Sorry, I was just saying I don't think "some people might not be using redux devtools" is a good reason to say "don't initialise redux devtools for anybody".

client/src/components/TreeDropdownField/tests/TreeDropdownField-story.js
Cheddam: The idea was that seeing the loading state would be useful. There's no risk of it failing, and there's nothing special about `500` - it's just an arbitrary unit of time that seemed slow enough to visualise the loading state but fast enough not to be annoying.

In any case, it's not really that useful, and it's probably better to simulate the loading state with actual state, so I've simplified it per @ScopeyNZ's suggestion. (The nested functions are still required to replicate the behaviour of the Fetch API.)

Cheddam:
@emteknetnz I meant to investigate why that wasn't rendering correctly (based on other experiences with read-only views, my suspicion is that this doesn't work properly in the CMS either.) For now, I've dropped it.

emteknetnz:
Cool looks good now

PR #1024, opened 2020-04-09
File manager icons

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1066
assets pr: https://github.com/silverstripe/silverstripe-assets/pull/386
asset-admin pr: https://github.com/silverstripe/silverstripe-asset-admin/pull/1087

Issue https://github.com/silverstripe/silverstripe-userforms/issues/948
framework-pr: https://github.com/silverstripe/silverstripe-framework/pull/9470

wip, so dist files are probably built using `yarn dev`

PR #1033, opened 2020-05-06
File manager icons

emteknetnz:
Issue silverstripe/silverstripe-asset-admin#1066 (includes links to related PRs)

old pr https://github.com/silverstripe/silverstripe-admin/pull/1024


emteknetnz:
Been built throughout with `yarn dev` so make sure to run `yarn build` before merge

robbieaverill:
Can you run the build now so it's not forgotten?

ScopeyNZ:
I'm not sure what the intention of this being draft is, but in a large open source repo, I would really rather than we try to keep PRs unraised until they're ready for review.

Would you mind closing this and re-opening when it's ready?

If you want members of the core contributor team to assist you, you're welcome to reopen the PR with some specific questions. Otherwise, I would suggest that you collaborate on your fork with others 🙂

emteknetnz:
@ScopeyNZ
>> I'm not sure what the intention of this being draft is
- Allows other people to comment on the progress of the pull-request, while still making it clear that it's not finalised and may change
- Another core contributor asked me to create my "in progress" PRs as draft PR's
- Make it easy for myself to check all the files changed in the dozens of commits (pre-squash)

I hear where you're coming from, though my understanding of githubs "Draft PR" functionality is that for pull-requests that are not ready for review.  To take a PR out of draft you literally click a button that says "Ready for review".

image

It seems like I'm following a fairly reasonable workflow.

ScopeyNZ:
Hey @emteknetnz. Sure I understand the general intent of the draft PR feature, but not every team/project/repository has the same rules here.

Remember that this is a large open source community, and by default, everybody who "watches" this repository gets a notification every time you add a commit to this PR. AFAIK there's no way to "watch a repository but ignore draft PRs".

I'm just asking, that for my sanity (and probably the sanity of other people who watch this repository), if you're not asking for specific feedback but instead just providing the opportunity to "have a look", can you refer people to your branch on your fork, rather than having an open draft PR? You may even wish to create a draft PR on your fork if you want the PR interface for contributors to comment. 

Anyway, we don't have any specific guidelines in our contribution docs about draft PRs right now, but we do mention that we expect work to be complete before a PR is raised. I'll put together an RFC for core contributors on what our stance should be, and we can update the docs.

client/src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -25,34 +32,58 @@ class Breadcrumb extends Component {
...
+      const classNames = ['breadcrumb__icon'];
+      if (className) {
+        classNames.push(className);
+      }
+      attrs.className = classNames.join(' ');
+      attrs = { tabIndex: '0', ...attrs };
+      if (attrs.hasOwnProperty('onClick')) {
+        attrs = { role: 'button', ...attrs };
+      }
+      attrs.key = `breadcrumb-icon-${k}`;
```
ScopeyNZ, line 73: Array index (which you can get as the second argument on your map function) [isn't suitable as a key](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318), even if you prefix it with a string. You should use something else that is a unique property on the icon if possible. Otherwise, you can avoid the `let k` by using the second arg of the mapping function.

client/src/components/FileStatusIcon/README.md
```diff
@@ -0,0 +1 @@
+TODO
```
ScopeyNZ, line 1: ...?

client/src/components/FileStatusIcon/FileStatusIcon.js
```diff
@@ -0,0 +1,112 @@
...
+    const backgroundClass = includeBackground ? 'file-status-icon--background' : '';
+    const classes = [`${this.cssClass}`, backgroundClass, className];
+    const attrs = isUserFormUpload
+      ? this.buildUserFormAttrs(hasRestrictedAccess)
+      : this.buildRestrictedFileAttrs(hasRestrictedAccess);
+    const idType = isUserFormUpload ? 'userform-upload' : 'restricted';
+    const id = `FileStatusIcon-${idType}-${fileID}`;
+    const tooltip = disableTooltip ? '' : this.renderTooltip(placement, id, attrs['data-title']);
+    return (
+      <div className={classes.filter(x => x).join(' ')}>
```
ScopeyNZ, line 88: We already use `classnames` in this repo (and pretty heavily in presentational components), you might find that more helpful than custom `filter` and `join` calls.

client/src/components/FileStatusIcon/FileStatusIcon.js
```diff
@@ -0,0 +1,112 @@
...
+import i18n from 'i18n';
+import { UncontrolledTooltip } from 'reactstrap';
+
+/**
+ * File status icon (e.g. restricted access) with a tooltip
+ */
+class FileStatusIcon extends Component {
+  constructor(props) {
+    super(props);
+    this.cssClass = 'file-status-icon';
```
ScopeyNZ, line 12: Is this meant to be a prop? It might as well be, otherwise you can just hardcode it, or make it `const cssClass` declared above the class.

client/src/components/FileStatusIcon/FileStatusIcon.js
```diff
@@ -0,0 +1,112 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import i18n from 'i18n';
+import { UncontrolledTooltip } from 'reactstrap';
+
+/**
+ * File status icon (e.g. restricted access) with a tooltip
+ */
+class FileStatusIcon extends Component {
```
ScopeyNZ, line 9:
```suggestion
import React, { PureComponent } from 'react';
import PropTypes from 'prop-types';
import i18n from 'i18n';
import { UncontrolledTooltip } from 'reactstrap';

/**
 * File status icon (e.g. restricted access) with a tooltip
 */
class FileStatusIcon extends PureComponent {
```

client/src/components/Breadcrumb/Breadcrumb.js
emteknetnz: Updated to use lodash.uniqueid

client/src/components/FileStatusIcon/README.md
emteknetnz: heh. Have filled in readme.

client/src/components/FileStatusIcon/FileStatusIcon.js
emteknetnz: Updated to use classnames

client/src/components/FileStatusIcon/FileStatusIcon.js
emteknetnz: Updated to use `const cssClass`

client/src/components/FileStatusIcon/FileStatusIcon.js
emteknetnz: Updated to use `PureComponent`

client/src/components/Breadcrumb/Breadcrumb.js
ScopeyNZ: Is there no property in the `icon` object that's sufficiently unique? It doesn't have to be unique in the DOM, just in that array. It'd be nice to avoid adding the dependency if possible. Also, I'm not completely sure on the theory here, but I assume using unique ID is no better, as the key is changing on each re-render, which sort of ruins the point of it...

emteknetnz: I didn't find anything too overly unique on the icon object.  

>> It'd be nice to avoid adding the dependency if possible

Agree, though at the same time it's a tiny size import

>> as the key is changing on each re-render,

Yes that's what happens.  uniqueId() is just basically just i++.  

>> which sort of ruins the point of it

It does? The article you linked to implied the point of unique keys was to avoid collisions, which uniqueId() will avoid

client/src/components/Breadcrumb/Breadcrumb.scss
```diff
@@ -79,11 +79,16 @@
...
 .breadcrumb__icon::before {
   margin-left: 1rem;
   font-size: 1.5rem;
   cursor: pointer;
-  vertical-align: middle;
+}
+
+.breadcrumb__icon {
+  margin-top: 5px;
 }
```
Cheddam, line 94: This doesn't seem to be helping with vertical alignment:

image

client/src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -25,34 +33,57 @@ class Breadcrumb extends Component {
...
+      // reassign with let so linter won't suggest 'const' above for unmodified nodeName/className
+      let attrs = { ...other };
+      const classNames = ['breadcrumb__icon'];
+      if (className) {
+        classNames.push(className);
+      }
+      attrs = { tabIndex: '0', ...attrs };
+      if (attrs.hasOwnProperty('onClick')) {
+        attrs = { role: 'button', ...attrs };
+      }
```
Cheddam, line 71: Any element can have a click event attached to it, but only buttons need `role=button`.

```suggestion
      if (nodeName === 'button') {
        attrs = { role: 'button', ...attrs };
      }
```

client/src/components/Breadcrumb/Breadcrumb.js
Cheddam: What's the likelihood of rendering the same icon twice in a single breadcrumb? It feels like that'd be a much better piece of information to use for an ID than a generated number, and I  agree with @ScopeyNZ that pulling in a dependency to achieve effectively the same approach as you were taking before isn't a good outcome.

ScopeyNZ: The article I linked has a key point:

> a key is the only thing React uses to identify DOM elements

The problem with an array index is that re-ordering items (for example) will change which item exists at each key without React actually seeing the change. The problem will manifest as React choosing not to re-render the re-order. By using `uniqueid` you're doing the opposite, the key still doesn't relate to the DOM element, but this time it's regenerated every time, meaning React can't optimise the render as it's going to re-render each key every time. Marginally better, but removing the point of `key`.

emteknetnz: It's a slightly different outcome than before.  For a breadcrumb with a single icon, previously the key was `breadcrumb-icon-1`, then if you navigated away from the folder, to another folder with a single icon, then back to the original folder, then then the key is still `breadcrumb-icon-1`.  Now when you do the above it'll be `breadcrumb-icon-3`

If there's a quick way to convert an object to something that can be used as key (e.g. md5(icon) then cool, though I suspect that would also require a third party library

Feels like we're talking about something that doesn't really matter a whole lot though, cos it seemed like it worked fine before.  And if it's inefficient now with uniqueid, it seems like trivial amount of inefficiency

client/src/components/FileStatusIcon/FileStatusIcon.js
```diff
@@ -0,0 +1,109 @@
...
+  /**
+   * @returns {*}
+   */
+  render() {
+    const {
+      fileID, hasRestrictedAccess, isUserFormUpload, placement, extraClassName,
+      disableTooltip, includeBackground
+    } = this.props;
+    const backgroundClass = includeBackground ? 'file-status-icon--background' : '';
+    const className = classNames([cssClass, backgroundClass, extraClassName]);
```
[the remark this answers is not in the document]
emteknetnz: that's just adding more lines and characters and has worse readability imo

client/src/components/FileStatusIcon/FileStatusIcon.scss
```diff
@@ -0,0 +1,35 @@
+.file-status-icon {
+  display: inline-block;
+  margin-left: 6px;
+
+  &:last-of-type {
+    margin-right: 6px;
+  }
```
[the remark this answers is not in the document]
emteknetnz: If we did that then we'd need to add more stuff in JS to detect if an FileStatusIcon is that last.  Since we're not using a FileStatusIconList we would end up with duplicated code since this would need to happen in GalleryView.js, TableView.js, Breadcrumb.js, as well as in File/FolderFormFactory.php which doesn't use react.

Much simpler just to keep as is.

client/src/components/Breadcrumb/Breadcrumb.js
emteknetnz: True, though there's some legacy code that passes in an object with an 'onClick' though not a 'nodeName'.  I'd rather not refactor legacy code in a feature pr if I can avoid it.

client/src/components/Breadcrumb/Breadcrumb.scss
emteknetnz: Bug in js, wasn't adding the class to the icon

client/src/components/Breadcrumb/Breadcrumb.js
ScopeyNZ: Ok, then I'd prefer we go back to my original suggestion, if there's nothing sufficiently unique, use the index, but extract it from the mapping function args rather than maintaining your own increment. It seems unlikely the icons are going to be re-ordered by the UI anyway right?

We want to always do things correctly in our repos btw, pointing out inefficiencies might have little impact now, but could end up being important down the line. And also, contributors are likely to use existing code as a reference, so we should be setting good examples.

client/src/components/FileStatusIcon/FileStatusIcon.js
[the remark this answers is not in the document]
ScopeyNZ: You can do 
```suggestion
    const className = classNames([
      [cssClass, extraClassName],
      { `${cssClass}--background`: includeBackground },
    ]);
```
instead?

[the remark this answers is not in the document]
ScopeyNZ: In fact, the outer-most array probably isn't required there either.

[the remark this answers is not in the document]
emteknetnz: I don't like the mixing of an array of strings and an object with a boolean value

I don't think we need to spend any time trying to optimise 2 lines of readable code

client/src/components/FileStatusIcon/FileStatusIcon.scss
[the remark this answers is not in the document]
Cheddam: The key thing I'm asking here is that instead of overriding these values in some places, you remove them from here and instead only introduce them in the places they're needed, whether that's via a modifier class or using the approach you have in other places (wrapping the `.file-status-icon` selector in a class representing the particular area the tweak needs to take place). That means _less_ code, and more importantly less overrides. You can still make use of `:last-of-type`, and you don't need any additional JS logic to make this happen.

client/src/components/Breadcrumb/Breadcrumb.js
emteknetnz: Has been changed to use the index from map()

client/src/components/FileStatusIcon/FileStatusIcon.js
```diff
@@ -0,0 +1,109 @@
...
+    const className = classNames('icon', `${cssClass}__icon`, fontIconClass);
+    const dataTitle = !hasRestrictedAccess
+      ? i18n._t(
+        'SilverStripe\\Admin\\FileStatusIcon.USERFORM_UPLOAD_UNRESTRICTED',
+        'Form submission, unrestricted access'
+      )
+      : i18n._t(
+        'SilverStripe\\Admin\\FileStatusIcon.USERFORM_UPLOAD_RESTRICTED',
+        'Form submission'
+      );
```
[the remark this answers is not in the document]
emteknetnz: Updated

client/src/components/FileStatusIcon/FileStatusIcon.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  /**
+   * @param {boolean} hasRestrictedAccess
+   * @returns {Object}
+   */
+  buildRestrictedFileAttrs(hasRestrictedAccess) {
+    const fontIconClass = hasRestrictedAccess
+      ? 'font-icon-user-lock'
+      : ''; // no icon for unrestricted
```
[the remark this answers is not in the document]
emteknetnz: Removed the no icon code

client/src/components/FileStatusIcon/FileStatusIcon.scss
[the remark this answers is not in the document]
emteknetnz: Have removed margins and added where needed

client/src/components/FileStatusIcon/FileStatusIcon.scss
```diff
@@ -0,0 +1,35 @@
+.file-status-icon {
+  display: inline-block;
```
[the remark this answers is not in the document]
emteknetnz: Removed inline-block

client/src/components/FileStatusIcon/README.md
```diff
@@ -0,0 +1,19 @@
+# FileStatusIcon Component
+
+Generates a file status icon element with a reactstrap tooltip.
```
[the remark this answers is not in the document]
emteknetnz: Updated readme

client/src/components/FileStatusIcon/tests/FileStatusIcon-story.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { storiesOf } from '@storybook/react';
+import FileStatusIcon from 'components/FileStatusIcon/FileStatusIcon';
+
+storiesOf('Admin/FileStatusIcon', module)
```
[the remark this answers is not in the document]
emteknetnz: Have updated storybook to use notes and knobs

client/src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -25,34 +33,57 @@ class Breadcrumb extends Component {
...
+   */
+  renderIcons(icons) {
+    return icons.map(icon => {
+      const { nodeName, className, hasRestrictedAccess, ...other } = icon;
+      // reassign with let so linter won't suggest 'const' above for unmodified nodeName/className
+      let attrs = { ...other };
+      const classNames = ['breadcrumb__icon'];
+      if (className) {
+        classNames.push(className);
+      }
```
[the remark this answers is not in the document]
emteknetnz: Now using `classnames`

client/src/components/Breadcrumb/Breadcrumb.js
Cheddam: I'm a doofus, I was thinking of `type="button"` 🙃

client/src/components/FileStatusIcon/FileStatusIcon.scss
```diff
@@ -0,0 +1,35 @@
...
+  margin-left: 6px;
+
+  &:last-of-type {
+    margin-right: 6px;
+  }
+}
+
+.file-status-icon__icon {
+  display: inline-block;
```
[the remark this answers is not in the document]
emteknetnz: This one actually does need to be inline-block, looking especially weird with the background option turned on

PR #1034, opened 2020-05-07
DOCS: Tidy travis tests

mattclegg:
I've tidied up the test matrix so it's a bit more readable and reduces the amount of duplicated lines.

I've also removed old tests for PHP 7.1 and added tests for PHP 7.4. The tests for PHP 7.4 are currently failing so I've added an exception until that can be fixed.

Any feedback welcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,38 +20,32 @@ addons:
 
 env:
   global:
+    - DB=MYSQL
     - COMPOSER_ROOT_VERSION=1.x-dev
     - SS_BASE_URL="http://localhost:8080/"
     - SS_ENVIRONMENT_TYPE="dev"
     - RECIPE_VERSION=4.x-dev
+  jobs:
+    - BEHAT_TEST=admin
+    - BEHAT_TEST=cms
+    - PHPCS_TEST=1
+    - PHPUNIT_TEST=1
 
-matrix:
+
+php:
+  # Legacy Versions of PHP
+  - 7.1
+  - 7.2
+  # Currently Supported Versions of PHP; https://www.php.net/supported-versions.php
+  - 7.3
+  - 7.4
+
+jobs:
   fast_finish: true
   include:
-    - php: 7.1
-      env: DB=PGSQL PHPUNIT_TEST=1
-    - php: 7.2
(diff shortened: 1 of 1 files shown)


ScopeyNZ:
Also, we definitely do _not_ want to allow failures on Behat. We should fix the test that are broken in `master` instead.

mattclegg:
Thanks for the quick feedback @robbieaverill & @ScopeyNZ.

I've reverted the change to remove PHP 7.1. I'd also like to fix the behat issues for PHP7.4 but from what I have discovered so far it looks like the problem is caused in another repository (that also doesn't support PHP7.4). I think the test job/matrix might be better to be use the `RECIPE_VERSION` as there seems to be a recorruing problem between testing correct package/php versions. Do you both have thoughts or an opinion on this?

The tests in `master` have been failing for months; https://github.com/silverstripe/silverstripe-admin/commits/master would it be better to remove the tests for PHP7.4 until it's fixed? I'd thought it would be better to have it here so the problem is more visible but I can easily remove it if it helps.

ScopeyNZ:
Yeah I can't really speak to the failures in `master` - it was something we looking into months ago but `master` doesn't get much love right now as it's the SS5 branch, which doesn't have much traction right now.

As for the matrix/travis thoughts, I'll leave those for @robbieaverill who has his head around this stuff more than I do 😅

robbieaverill:
Unfortunately the tests in the master branches of our repositories break all the time. It's an unstable working branch that regularly has merge ups into it from a different major version (4 into 5), and we don't usually pull request those merge ups, so we don't often see the tests breaking when we do it.

I would not advise that we should disable or comment out our failing tests, rather someone needs to spend a bit of time getting them green again. I do this semi-regularly, and it usually doesn't take too long to do.

I think I'm going politely say no to this pull request (for the reasons mentioned previously), but thanks for taking the time to have a look at it. If you wanted to help get things green again you could chip away at individual test failures in either behat or unit tests. Often test failures in the master branch are caused by other modules not having been merged up, or small API changes between 4 and 5 that are used in merged up code. The master branch has been green recently, so getting it back shouldn't be a mammoth task.

PR #1035, opened 2020-05-08
BUG: Page search form layout overflow issue

mfendeksilverstripe:
# Page search form layout overflow issue

* page search form has an UI overflow issue when using the form in the context of search result (`admin/pages/`)
* same page search form but in the context of page edit form (`admin/pages/edit/show/<ID>`) doesn't have this issue as there is already a fix in place
* this change-set simply makes the existing fix more broad to also cover the other page search form instance

## Search box left panel (via page edit form)

Working fine as seen on the screenshot

search-box-panel

## Search box upper right (via pages overview page) - before fix

Date fields have an overflow issue

search-box-before

## Search box upper right (via pages overview page) - after fix

Now consistent with the left panel

search-box-after

## Related issues

https://github.com/silverstripe/silverstripe-admin/issues/1036


michalkleiner:
Should the fields and their labels stack instead of sitting narrow next to each other, obscuring part of the value? It would allow the full date to be visible and, if applied to the fields, a date picker icon visible.

sachajudd:
Could be an alternative for sure!

mfendeksilverstripe:
That is our current bespoke solution on our project as it was easier to alter the fields compared to alteration of markup.

Screen Shot 2020-05-12 at 7 15 39 AM

@michalkleiner @sachajudd

clarkepaul:
The main reasons for keeping the fields side-by-side is so that the panel has plenty of space for developers to add additional fields (I have seen 3-4 additional fields been added), but there is also a user benefit of being accustomed to seeing the fields side-by-side and being related to each other.

The date fields have a minimum width that they are usable at, enforced by the browser to allow space for field actions from what I can tell. Rather than stacking them for everyone, I think the better option is to:

-  increase the width of the panel for desktop users to 375px (bigger viewport than 768px), and for smaller viewports (less than 768px) go 100% width.
-  for viewports less than 768px remove the CSS which positions the panel 53px from the right.
-  and lastly for viewports less than 375px stack the fields rather than having them inline (change flex-direction: row to column)

Hopefully, that would provide the best experience for all users. @mfendeksilverstripe would that be a reasonable change?

mfendeksilverstripe:
@clarkepaul is right when it comes to limited space. The picture below is from one of my projects. We noticed that if this form grows too large it doesn't fit the viewport and bottom fields can't be even reached in some screens.

Screen Shot 2020-05-12 at 12 27 32 PM

With that said, the scope of this change-set is to make current display consistent across application. Any other proposed layout improvements should be split out into its own issue / PR.

Based on the feedback from clients we know that it's more preferable to not display the whole date value (when a date is already selected) compared to rendering the form with obvious field overflow. The selected values in the form rarely matter as they are displayed above search results in a better format.

Screen Shot 2020-05-12 at 12 32 16 PM

mfendeksilverstripe:
@maxime-rainville It's rebased now :)

PR #1038, opened 2020-05-11
FIX Resize address-card-warning

sachajudd:
To be merged once I've approved that the icon looks correct in the CMS.


sachajudd:
@emteknetnz

PR #1040, opened 2020-05-21
ENHANCEMENT: Split icon storybook into sections

scott1702:
This makes it really clear for people where icons should be used. 
image

Have also added an option to add a summary to allow for further description of each section in the future:

image

### DIFF
--- a/.storybook/preview-head.html
+++ b/.storybook/preview-head.html
@@ -1,9 +1,12 @@
 <style>
-	#root {
-		width: 100%;
-		height: 100%;
-		background-color: #f6f7f8;
-		padding: 15px 25px;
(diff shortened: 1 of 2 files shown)


sachajudd:
Yay! This is awesome, thanks Scott 👏

clarkepaul:
Ka-ching! 
Bit harder to maintain I imagine?

scott1702:
@clarkepaul 
> Bit harder to maintain I imagine?

Not at all! I had a chat with @sachajudd around the process to add new icons, and that remains the exact same even after these changes, just replace `icons-reference.html` file with the auto-generated one from Fontastic

PR #1041, opened 2020-05-25
NEW Add new variations of block icons and update existing

sachajudd:
Adds a few new block-icons some devs have been asking for.
Updated one existing block-icon but have made sure the naming and character mapping is the same as before.
image
(Crossed out icons were just reordered in the set.)


sachajudd:
Yup you're right, I forgot. Please hold.

PR #1042, opened 2020-05-26
Vertically align icons added to form descriptions

bergice:
Relates to https://github.com/silverstripe/silverstripe-userforms/issues/949


client/src/components/Form/Form.scss
```diff
@@ -131,6 +131,14 @@
   &:empty {
     display: none;
   }
+
+  .icon {
+    &::before {
+      vertical-align: middle;
+      margin-right: 6px;
```
Cheddam, line 138: What happens if the icon is used at the end of the description? 🤔

client/src/components/Form/Form.scss
```diff
@@ -131,6 +131,14 @@
...
+      font-size: 16px;
```
Cheddam, line 139: Is there a variable we can use here instead of hardcoding this size?

bergice: Perhaps:

```suggestion
      font-size: $font-size-lg;
```

@sachajudd any ideas?

client/src/components/Form/Form.scss
bergice: No margin is shown between the text and the icon.

Some potential solutions:
* Wrap text in `<p>` tag to be able to correctly apply margin
* Treat description blurb as single element and expect people to add whitespace `&nbsp` between icons
* Add icon classes to the description element

May be out of scope for this piece of work though.

client/src/components/Form/Form.scss
```diff
@@ -123,6 +123,8 @@
 
 .form__field-extra-label,
 .form__field-description {
+  display: flex;
```
bergice, line 126: @Cheddam I'm confused about why this was added. It seems to mis-align the content.

Cheddam: Can you provide an example of this misalignment? I introduced it (along with `align-items: center;`) in order to fix the alignment, not break it 😅

client/src/components/Form/Form.scss
Cheddam: I thought we'd run into trouble with using this variable (as it calculates out to `16.003px`) but I've just tested it again and it seems fine, so yep apply it 👍

client/src/components/Form/Form.scss
bergice: Without flex, align:

image

With flex, align:

image

sachajudd: Can you check how Firefox vs Chrome behaves?

client/src/components/Form/Form.scss
Cheddam: I think putting the icon at the start of the description is likely the most common use-case, and we can improve this later, so happy to proceed for now.

client/src/components/Form/Form.scss
Cheddam: I'm really not sure how you're hitting this @bergice. Here's what it looks like for me in Chrome 83:

image

And this is Firefox 77:

image

bergice: Okay, me neither. I just tested it again and it looks fine. 🤷

ScopeyNZ:
Hey @bergice, in the future can you make sure that you don't create merge commits on the base branch (please rebase), and you don't self-merge un-approved work so soon after adding changes?

Thanks

michalkleiner:
This caused a regression for field descriptions containing markup. We rely quite extensively on being able to provide useful guidance to content editors, so this is quite incovenient.

Screen Shot 2020-08-09 at 5 33 12 PM

Cheddam:
@michalkleiner Thanks for the heads up!

PR #1050, opened 2020-06-05
Make LeftAndMain independent of the presence of Versioned module

alessandromarotta:
LeftAndMain's init() method calls two Versioned methods and this makes it impossible to have the Silverstripe-admin module installed without the Versioned module.

In other parts of the Silverstripe-admin code we have the verification of the presence or absence of the Versioned class, instead in the init() method it is missing.

Please, could you also introduce in the init() method the verification of the presence of the Versioned module in order to install the administrative panel without the Versioned module?

Like in the file commit, for example.

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -754,12 +754,12 @@ class LeftAndMain extends Controller implements PermissionProvider
         SSViewer::set_themes(LeftAndMain::config()->uninherited('admin_themes'));
         
         // Check the presence of Silverstripe-Versioned module
-        if(class_exists(Versioned::class)) {
-           // Set the current reading mode
-           Versioned::set_stage(Versioned::DRAFT);
-
-           // Set default reading mode to suppress ?stage=Stage querystring params in CMS
-           Versioned::set_default_reading_mode(Versioned::get_reading_mode());
+        if (class_exists(Versioned::class)) {
+            // Set the current reading mode
+            Versioned::set_stage(Versioned::DRAFT);
+            
+            // Set default reading mode to suppress ?stage=Stage querystring params in CMS
+            Versioned::set_default_reading_mode(Versioned::get_reading_mode());
         }
     }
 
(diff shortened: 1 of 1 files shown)


alessandromarotta:
p.s: we should also remove the dependence of Versioned in composer.json

emteknetnz:
The versioned module in a requirement of the admin module
https://github.com/silverstripe/silverstripe-admin/blob/1/composer.json#L23 so it is assumed that it will always be there

Removing the versioned module via `"replace": { "silverstripe/versioned": "*" }` is not something we actively support

There are various other places in the within the admin module that reference the Versioned class, without first testing to see if that Versioned module exists, because it's always assume that it does

We'll close this pull-request now.  Feel free to open a new pull-request for anything else you think could be fixed or improved.

alessandromarotta:
Hi, have you read here? 

"LeftAndMain's init() method calls two Versioned methods and this makes it impossible to have the Silverstripe-admin module installed without the Versioned module.

In other parts of the Silverstripe-admin code we have the verification of the presence or absence of the Versioned class, instead in the init() method it is missing."

I think that it's possible to have the SilverStripe Admin module without the Versioned module.

Only this modification is needed: https://github.com/silverstripe/silverstripe-admin/pull/1050/files

dhensby:
For clarification: Adding and removing dependencies is not a semver issue; semver is about a guarantee of our APIs, not our dependencies. If someone is using admin and they themselves need versioned it should be in *their* `composer.json`.

However, trying to decouple `versioned` from `admin` is a good idea in principle, but it needs a lot more consideration than just wrapping one use of it in a `class_exists` call.

alessandromarotta:
@dhensby ok, but my commit does not break anything, but it would give the possibility, for those who needs, to install the SilverStripe Admin module without the Versioned module, simply adding `"replace": {"silverstripe / versioned": "*"}` in the _composer .json_ of the project.

I emphasize that in other parts of SilverStripe Admin the presence of the Versioned module is verified with the method _hasExtension()_, less than in the _Init()_ method of the LeftAndMain.php file. This really doesn't make sense.

dhensby:
as @emteknetnz has pointed out, this is an incomplete implementation as `Versioned` is used in other places in the code. If you want to go and fix that up I'd be happy to take in a change that made `Versioned` optional and removed it from `composer.json` - but we also need to make sure that the other aspects of the module (frontend aspects) aren't making assumptions about versioned either.

Issue #1053, opened 2020-06-09
JS source files should be included by default


adrhumphreys:
Currently these files are excluded by default.

We've enabled users to inject over them through the injector but it's common for users to extend these components, make small changes, then inject over them which is not possible with a default composer install

This also impacts automated build systems which will need to run `composer install --prefer-source` to build any project that has these kinds of components

The negative impacts I can think of would be increasing the size of the composer dependency by now including these JS files and increasing the chances that updating these components creates breaking changes for projects.

I'm happy to start making some PR's but I would like to first get everyone on the same page and agree to the change going forwards and which release we could make the change in (selfishly I'd like to add it to 4.5 as a minor patch but can see why we'd want to make it happen in a future minor release)

Thoughts, feelings?

blueo:
something to consider is that this pattern may be more common in class based react components. It may change as react shifts more towards functional components with hooks etc.

chillu:
Can you solve this by requiring those modules to be checked out via `type=vcs` rather than `type=dist` in composer? Alternatively, can you run some stats on how much extra network transfer a standard `composer create-project silverstripe/installer` would be with this change?

blueo:
that's a good idea - we can certainly give it a whirl.

ScopeyNZ:
> stats on how much extra network transfer a standard `composer create-project silverstripe/installer`

Should we really be concerned about this metric over actually making our intended JS API consumable?

I'm still of the opinion we need to figure out an NPM offering to properly fix this.

blueo:
just an update on using composer to set the install type. Unless you're referencing a commit and using a repository config (with type="vsc" or type="git") then composer will grab the dist version of the package. So it doesn't work well for when you want to include the src files for a package that is not a fork etc. 

I also tried the `preferred-install` config option, that will be overridden by the cli argument `--prefer-dist` which is not ideal as lots of CI configs use this.

Cheddam:
Thoughts on the NPM approach:

- We'd need to expand Cow's tooling to release NPM packages alongside module tags, else this would be a major increase in toil.
- This still wouldn't result in source JS being provided in dist bundles, and I assume we would still bake in the dist JS to avoid introducing a hard dependency on NPM.
- Developers wanting to consume these NPM packages would need to be careful to keep the versions in sync with their equivalent Composer modules, otherwise risking incompatibility between their extended components and the pre-baked core JS.

[...]

PR #1054, opened 2020-06-09
JS source files included by default

adrhumphreys:
Relates to: https://github.com/silverstripe/silverstripe-admin/issues/1053

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,24 +9,11 @@
 
 # all client-side source except styles
 /.storybook export-ignore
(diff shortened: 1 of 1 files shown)


emteknetnz (changes requested):
Seems like there's not much agreement on the parent issue https://github.com/silverstripe/silverstripe-admin/issues/1053

Is this idea going to progress further? Or should we now close this PR?

dhensby:
I'm going to close this as there's no consensus on the issue at the moment. Once there is this can be re-opened

PR #1057, opened 2020-06-16
FIX: Ensure graphql URL is inserted before admin

sminnee:
I’ve had issues getting the URL admin/graphql to work on a ModelAdmin
app that doesn’t have CMS installed. This fix ensures that the graphql
URL rule is inserted before the admin one, making it work.


chillu:
Confirmed this still works in asset admin on a standard 4.x install.

PR #1059, opened 2020-06-22
Fix external link setting text to `undefined` on text

bergice:
Passing `undefined` to html sanitiser should return an empty string

Resolves https://github.com/silverstripe/silverstripe-cms/issues/2512


client/src/legacy/TinyMCE_sslink.js
```diff
@@ -109,7 +109,7 @@ jQuery.entwine('ss', ($) => {
 
       const attributes = this.buildAttributes(data);
       const sanitise = createHTMLSanitiser();
-      const linkText = sanitise(data.Text);
+      const linkText = data.Text === undefined ? '' : sanitise(data.Text);
```
Cheddam, line 112: I think it would make more sense to apply the fix to the `createHTMLSanitiser()` method in `ShortcodeSerialiser.js` - that way it'll fix any other cases of this problem at the same time.

bergice: Cool, I've changed it.

PR #1060, opened 2020-06-23
Travis 1.6

emteknetnz:


.travis.yml
```diff
@@ -21,37 +21,29 @@ addons:
 env:
   global:
     - COMPOSER_ROOT_VERSION=1.6.x-dev
+    - VERSIONED_VERSION=1.6.x-dev
+    - DISPLAY=":99"
```
dnsl48, line 25: I don't think we need this. The `xvfb` service [already defines](https://docs.travis-ci.com/user/gui-and-headless-browsers/#using-services) that out of the box.

.travis.yml
```diff
@@ -21,37 +21,29 @@ addons:
...
+    - XVFBARGS=":99 -ac -screen 0 1024x768x16"
```
dnsl48, line 26: Same here. I'm not sure we need this variable.

PR #1061, opened 2020-06-24
Implementation of horizontal scrolling in "ui-tabs-nav"

alessandromarotta:
When a ModelAdmin has many tabs, as in the attached image, the controls for moving from one tab to another are arranged in another row.
With this patch the controls are kept horizontally and you can scroll to view the hidden ones.

Before:
image

After:
image

### DIFF
--- a/client/src/styles/legacy/_style.scss
+++ b/client/src/styles/legacy/_style.scss
@@ -148,8 +148,15 @@ html, body {
         height: $toolbar-height;
       }
     }
+    .ui-tabs-nav {
+      min-width: max-content;
+    }
 
     // CMS admin
+    .cms-content-header-tabs {
+      overflow: auto hidden;
+      white-space: nowrap;
+    }
     .cms-content-header-nav .breadcrumbs-wrapper,
     .cms-content-header-nav .cms-content-header-tabs,
       // Model admin


michalkleiner:
There's already an existing issue related to the tabs — https://github.com/silverstripe/silverstripe-framework/issues/1943

ScopeyNZ:
Yeah I have a feeling that @silverstripeux are probably not going to be too keen on the horizontal scroll... Not sure that's an improvement from the stacking they're doing currently. The issue @michalkleiner linked is the intention though if you want to re-purpose this PR 🙂

alessandromarotta:
Well, this is only a rapid solution for an old problem.
The horizonal scrolling is pretty good on mobile and can be a nice solution when you many elements on the viewport.

clarkepaul:
Very tempting as there has been no progress on a full solution. The only part that makes me hesitant is that will be times when you won't see the tabs being cut off and the user will think there isn't anything to scroll for. There are concepts which introduce indications of more tabs with a bit of JS ([example](https://www.codeply.com/p/YVSUDyMvZc)) but I'm not sure what would be required to get that integrated in a way which compliments our React components. I don't think we can introduce a pattern which isn't obvious/accessible to all users unfortunately.

emteknetnz:

Issue #504, opened 2018-05-01
Allow toast notifications to be added from React


robbieaverill:
If we add some sort of connection between the toast notification system and the Redux store, we should be able to push new toast notifications from React applications anywhere in the CMS.

Currently there are controlled via PHP and AJAX requests.

For example, when reverting to a version in history viewer, ideally we could push a toast notification via Redux. At the moment we'll be rendering an inline FormAlert in the app instead.

# Acceptance criteria
- Toast notification can be used in React
- Existing `jQuery.noticeAdd()` calls still work
- Toasts are not auto-dismissed when the user is interacting with them
- Actions can be added to individual Toast.
- A proper API and documentation is provided so third party developer can use Toast in their own app
- We have clear guidelines where it's appropriate to use toast notification and where it isn't
- Component added to pattern library

# Note
- The purpose of this card is to provide a solid API so we can start using Toast where it makes sense to use them
- We might end up converting a few alerts to toast for to demonstrate the feature, but converting all the existing alerts to to toast is not the goal of this card
- Toast could be use for a lot of other cool things (e.g. async push notification). While that's cool, it should be handle via follow up cards.

# Design
https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/asset/componentContainers/5f0fdc1f9e3d52081dbf4408/tab/design?mode=preview

# Pull requests
* [x] https://github.com/silverstripe/silverstripe-admin/pull/1068
* [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/1111
* [x] https://github.com/silverstripe/webpack-config/pull/48
* [x] https://github.com/silverstripe/silverstripe-framework/pull/9695
* [ ] ~~https://github.com/silverstripe/silverstripe-admin/pull/1125~~

robbieaverill:
I think there's still a jQuery plugin that we could use via the global window object to push new toast messages...

robbieaverill:
I'm pretty keen to work on this myself, so might pick it up at some point in the next month or two

robbieaverill:
I've had a quick tinker and have come up with a lightweight API that we can start to move towards, gradually in SilverStripe 4 and potentially entirely in SilverStripe 5. You can see the changes here: 



I tested this with a success toast whenever you insert an image into TinyMCE content via the asset-admin module: https://gist.github.com/robbieaverill/59958fdfbdf2064c48ea651c6c9ab41e

```diff
[...]

PR #1068, opened 2020-07-04
Reimplement toast in React so we can use them everywhere

maxime-rainville:
# Overview

Our toast notifications are currently implemented using a jQuery plugin. This makes it difficult to fire toast notification in our React component. This remove the existing jquery notify plugin and re-implements the functionality in react using redux.

There's also a few improvements like the additions of actions to the toasts.

# Details
- the Jquery notice plugin got remove and replace with a polyfill that fires off our redux based toast.
- Toast can now have actions attached to them. 
  - They can have 1, 2 or no actions attached to them.
  - Actions can fire a callback or be a link.
  - Actions don't have dedicated styles or icons
  - Clicking an action will dismiss its toast
- By default, toast are displayed for 3 sec before self dismissing. This was retain from the previous implementation
  - Like the previous implementation, toast can be configured to "stay" on screen until the user manually dismiss them instead of using a timeout.
  - The ability to explicitly specify a timeout has been removed. This wasn't used anywhere.
  - The auto-dismissal is cancelled when the user's mouse enter the toast area or a toast gains focus. When the toast loose focus or the mouse moves out, the auto-dismissal restarts. This behaviour is debounced to avoid firing the pause resume redux event for nothing.
- The toast reducer and actions looks a bit unusual in that they make extensive use of timeouts. This is needed to:
  - auto-dismiss toast after 3 sec
  - add a 1.2 sec fadeout after dismissal before completely removing the toast for the redux store
  - control the pause/resume feature when the toast gain and loose focus
- The toast fade out transition has been implemented using key frame animation, which should hopefully make it a tiny bit smoother.
- Four style of toasts are supported. Shorthand redux actions are provided for all of them.
  - info
  - waning
  - success
  - error
- I thought about allowing HTML in the toast message, but that would have required using `dangerouslySetInnerHTML` which is not idea. I couldn't find any pre-existing place where we provide raw HTMl in a toast. I also think it's questionable whether we should allow extra styles in the toast. It's probably better to keep the look and feel of the toast consistent.
  - There is however a lot "alerts" with links that would be good candidate to be converted to toast. Those links probably need to become actions.

# Parent issue
* https://github.com/silverstripe/silverstripe-admin/issues/504

# Depends on 
* https://github.com/silverstripe/webpack-config/pull/48 (builds won't be green until we release v1.7.0)
* https://github.com/silverstripe/silverstripe-framework/pull/9695 (for behat tests)

# Related PR
* https://github.com/silverstripe/silverstripe-asset-admin/pull/1111


client/src/components/Toasts/Toasts.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import i18n from 'i18n';
+
+
+const Toast = ({type, text}) => (
+  <div className={`toast toast__${type}`}>
+    <div className="toast--close">x</div>
```
ScopeyNZ, line 8:
```suggestion
  <div className={`toast toast--${type}`}>
    <div className="toast__close">x</div>
```

ScopeyNZ: You've got the element and modifier around the wrong way a bunch in this PR. I'm not going to suggest changes for it all 😅

maxime-rainville: Fixed.

client/src/components/Toasts/Toasts.js
```diff
@@ -0,0 +1,31 @@
...
+  const onClick = (e) => {
+    e.preventDefault();
+    onDismiss();
+  };
+
+  const dismissing = status === 'dismissing' ? 'toast--dismissing' : '';
+
+  return (
+    <div className={`toast toast--${type} ${dismissing}`}>
+      <div className="toast__close" onClick={onClick}>x</div>
```
maxime-rainville, line 16: Need to make this accessible.

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,101 @@
...
+  width: 250px;
+  z-index: 9999;
+
+  &__notice-wrap {
+    position: absolute;
+  }
+}
+
+
+@keyframes toastFadeIn {
```
maxime-rainville, line 14: I'm using keyframe animation instead of the jQuery animate method in the plug-in. This should make it smoother.

client/src/components/Toasts/tests/Toasts-test.js
```diff
@@ -0,0 +1,79 @@
+/* global jest, describe, beforeEach, afterEach, it, expect */
+
+import React from 'react';
```
maxime-rainville, line 3: These tests got copied from Tips. I haven't updated them yet.

client/src/containers/ToastsContainer/ToastsContainer.js
```diff
@@ -0,0 +1,27 @@
+import React, {useEffect} from 'react';
+import PropTypes from 'prop-types';
+import Toasts from 'components/Toasts/Toasts';
+import { connect } from 'react-redux';
+import {bindActionCreators} from 'redux';
+import * as toastsActions from 'state/toasts/ToastsActions';
+
+const ToastsContainer = ({toasts, actions}) => {
+  useEffect(() => {
```
maxime-rainville, line 9: That's just there for testing.

client/src/containers/ToastsContainer/tests/Toasts-story.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
```
maxime-rainville, line 1: Remove this file.

client/src/containers/ToastsContainer/tests/Toasts-test.js
```diff
@@ -0,0 +1,79 @@
+/* global jest, describe, beforeEach, afterEach, it, expect */
```
maxime-rainville, line 1: Refactor this file.

client/src/legacy/ToastsContainer.js
```diff
@@ -0,0 +1,16 @@
+/* global window */
```
maxime-rainville, line 1: This bit detects a `<div class="toasts-container"></div>` tag and wired it up. There might be better way of doing this.

Maybe we should just instantiate the `<ToastsContainer />` direction in some higher `<App />` component.

client/src/state/toasts/ToastsActions.js
```diff
@@ -0,0 +1,43 @@
...
+ * @param {object} config
+ */
+export function display(options) {
+  const key = 'toast-' + uuid();
+  const payload = { key, ...defaultOptions, ...options }
+
+  return (dispatch) => {
+    dispatch ({type: ACTION_TYPES.DISPLAY, payload});
+    if (!payload.stay) {
+      setTimeout(() => dismiss(key)(dispatch), payload.stayTime);
```
maxime-rainville, line 25: There's some clever use of timeout here to trigger the fade out of the toast. It's not clear that it makes a lot of sense to have this here, but it has the advantage of keeping all the setTimeout logic in one place.

client/src/state/toasts/ToastsActions.js
```diff
@@ -0,0 +1,43 @@
...
+export function dismiss(key) {
+  return (dispatch) => {
+    dispatch({
+      type: ACTION_TYPES.DISMISS,
+      payload: { key }
+    });
+
+    setTimeout(
+      () => dispatch({type: ACTION_TYPES.REMOVE, payload: { key }}),
+      1200
```
maxime-rainville, line 39: When we're done animating the fade out, tell the redux store to remove the toasts.

client/src/state/toasts/ToastsActions.js
```diff
@@ -0,0 +1,43 @@
...
+  type: 'notice' 			// could also be error, succes
+};
+
+/**
+ * Sets global config for the application.
+ *
+ * @param {object} config
+ */
+export function display(options) {
+  const key = 'toast-' + uuid();
```
maxime-rainville, line 19: I'm using uuid to generate a unique key for each toast. The method calling the display method has the option to specify it's own id as well.

client/src/state/toasts/ToastsReducer.js
```diff
@@ -0,0 +1,23 @@
...
+
+];
+
+function toastsReducer(state = initState, {type, payload}) {
+  switch (type) {
+    case ACTION_TYPES.DISPLAY:
+      return [...state, payload];
+
+    case ACTION_TYPES.DISMISS:
+      return state.map((toast) => (toast.key === payload.key ? {...toast, status: 'dismissing'} : toast));
```
maxime-rainville, line 13:
```suggestion
      return state.map((toast) => (toast.key === payload.key ? {...toast, dismissing: true} : toast));
```

This makes more sense, I think.

client/src/styles/legacy/_style.scss
```diff
@@ -959,67 +959,6 @@ html, body {
   margin: 0 0 $spacer; // gets padding from tab panel
 }
 
-.notice-item {
```
maxime-rainville, line 962: All those styles got copied in the Toasts SCSS file.

package.json
```diff
@@ -86,6 +86,7 @@
     "resize-observer-polyfill": "^1.5.0",
     "toposort": "^1.0.3",
     "url": "^0.11.0",
+    "uuid": "^8.2.0",
```
maxime-rainville, line 89: This lib makes easy to generate unique identifiers. I could probably just use a variable that increment on every call to the `display` action, but this uuid lib is very well supported, relatively small and it will probably have other uses in the long run.

client/src/styles/legacy/_style.scss
sachajudd: We created custom toasts before bootstrap released their own which we called ‘notices’. Going forward and adding this component to the pattern library I think we should adopt bootstrap’s toasts instead of our own custom ones. This means we can align with the toasts bootstrap has to offer and features which we do not yet have implemented. https://getbootstrap.com/docs/4.5/components/toasts/ cc @silverstripeux

maxime-rainville: Cool ... looks like it's in reactstrap as well. Not sure when it got added, so I might have to bump the version that we're shipping with admin 1.7.x. https://reactstrap.github.io/components/toasts/

Do you have a mock-up showing how it should look? Or is it meant to retrain the existing look and feel?

If you don't have a mock-up, I can probably just put it in with the vanilla look for now and we can go from there.

maxime-rainville: FYI this is what it looks like right now. 
image

sachajudd: This is what we have currently in [DSM]( https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5dae8013fccf7c3293fb199d). I'll make sure to do a review in the coming week or so to make sure it's up to date.

maxime-rainville:
@sachajudd I got this thing doing things. I've switched to using the Bootstap Toast. It looks kind of weird right now because I blindly  applied the existing styles to the BS toasts. Obviously this will need some refinements.

I managed to get the [existing jQuery toast showing using the react component](https://youtu.be/QV9rTalGusU). I also managed to get some of the [asset-admin alert to use the toasts](https://youtu.be/wWugspFM41M).

Couple points to consider:
- The bootstrap toast has a toast title and a toast body. The jQuery one only had a body. The close button is tied to the title. If we don't want to use the ToastTitle will have to hack a close button in there.
- The BS Toast come in with some basic fade-in and fade-out animation. My current demo uses the vanilla bootstrap transition which I don't think are as nice as the one in the jquery plugin.
- Some of the alerts include actions and links to other part of the CMS. I would need some guidance about whatever we want people to be able to interact with the toasts beyond clicking the dismiss button.

client/src/components/Toasts/Toasts.js
maxime-rainville: Bootstrap toast takes care of this.

client/src/containers/ToastsContainer/ToastsContainer.js
maxime-rainville: It's gone.

client/src/components/Toasts/Toasts.js
```diff
@@ -0,0 +1,39 @@
...
+
+  const [show, setShow] = useState(false);
+  useEffect(() => {
+    setShow(!dismissed);
+  }, [dismissed]);
+
+  const className = classnames('toast', `toast--${type}`, {'toast--dismissing': dismissed});
+
+  return (
+    <BsToast className={className} isOpen={show} aria-live="assertive" aria-atomic="true">
```
maxime-rainville, line 21: The aria attributes in this file comes from the [accessibility suggestion on the bootstrap site](https://getbootstrap.com/docs/4.3/components/toasts/#accessibility).

client/src/state/toasts/ToastsActions.js
```diff
@@ -0,0 +1,40 @@
...
+  stay: false,				// should the notice item stay or not?
+  type: 'notice' 			// could also be error, succes
+};
+
+/**
+ * Sets global config for the application.
+ *
+ * @param {object} config
+ */
+export function display(options) {
```
maxime-rainville, line 16: I'm going to add a few alias here for displaying different message type:
- `success()`
- `warning()`
- `error()`
- `notice()`

sachajudd: Would it be notice or info? 🤔

client/src/styles/legacy/_style.scss
clarkepaul: @maxime-rainville is that displaying just the title (without body text) here?

client/src/state/toasts/ToastsActions.js
clarkepaul: I think BS uses **info** as a class so we should use that.

clarkepaul:
We are going to want to retain links in the Toast. We should actually have a design which has the possibility of having 0/1/2 actions on the bottom of the Toast, for the likes of Undo, View item...

FYI some people refer to Toasts with just the heading as a **Snack**, I don't think it's an official term or component—so many food references in web design :)

client/src/styles/legacy/_style.scss
maxime-rainville: That screen shot comes from my effort to exactly replicate the jQuery notice HTML. That screen shot doesn't use the bootstrap toast. That should be exactly how they look like right now.

I can get the boostrap toasts to look like that if you want to retain the existing look and feel.

client/src/state/toasts/ToastsActions.js
maxime-rainville: `info()` it is.

maxime-rainville:
Just added some convenience action for adding success/info/warning/error toast and some unit test for the reducer.

robbieaverill:
Awesome to see this happening @maxime-rainville !

maxime-rainville:
Added the ability have some actions to the toast.

image

maxime-rainville:
Just had a catch up with @clarkepaul and @sachajudd. Here's what's next:
- Need to implement design as demonstrated here: https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/asset/componentContainers/5f0fdc1f9e3d52081dbf4408/tab/design?mode=preview
  - No Toast title
  - Height is dynamic and toast should expand to display all of its content
- Need to double check if our existing toast display HTML or if they are all plain text.
- At most, we'll have 2 actions per toast
  - Action don't have any style or icons
  - Paul will provide mouse-over styles for the actions
- Need to replicate the current transition effect: fade out followed by a slide up
- Mousing over a toast should cancel whatever timeout currently applies to this toast
- We'll get rid of the option to control the display time. Currently you can explicitly define how long the toast should be displayed, but we don't use that feature any where. Developers will still be able to specify that their toast should have no timeout and require a click on the X button to dismiss.

# Mouseover toast
Didn't envision this when I started, but it makes a lot of sense that toast should not go away if your mouse is over them.

"jQuery.notice" currently does that out of the box. But I think we can probably improve on that:
- If you have many toasts, if you mouse over one of them that should probably freeze all the other ones as well ... otherwise you might end up in the weird situation where the toast above the one that currently has the mouse over goes away.
- I'm thinking we should also freeze if the toast action or dismiss button gets focus
- I'm thinking on mouseover/focus the toast timeout is reset rather than pause. e.g.: If you had .1 sec left in the timeout, the toast doesn't go away .1 sec after you mouseout.

# Guidelines
We'll probably need to draft/revise some formal guidelines about when to use alert messages and when to use toast. It probably will look something like:
- use toast for message occurring in response to actions by users
- use alerts for validation errors and passive information that doesn't get triggered by user actions.

robbieaverill:
I think you could also use toasts for async tasks or events notifications - eg "your long running job finished successfully" or "you have a new change to review"

maxime-rainville:
@silverstripeux I've got the styles in and some transition https://youtu.be/bj7CW1YUxNE

clarkepaul:
@maxime-rainville at first look this is amazing, you've really paid attention to the design notes and how we envisioned it to be. Awesome work!

maxime-rainville:
I've add a pause on mouse over and on focus. https://youtu.be/M9CAfPBw1wU

Mouse overs will actually pause all the toasts. I was thinking I can't really only pause the active one, because if one of the toast above it gets dismissed, it will slide up which will proboably be annoying for the user.

I was wondering if we want to dismiss the toast when its action is click? That would make some degree of sense, otherwise users would be able to click them multiple times, which could lead to trouble.

clarkepaul:
Disappearing after a link is clicked makes sense to me. I see them as navigational links most of all e.g. go to page after moving it. Even with task based actions like "Undo" we would probably followup with another toast notification of successfully undoing the action—not that this action exists just yet.

client/src/boot/registerComponents.js
```diff
@@ -98,5 +100,7 @@ export default () => {
     HtmlEditorField,
     NumberField,
     PopoverOptionSet,
+    ToastsContainer,
+    FileStatusIcon
```
maxime-rainville, line 104: `FileStatusIcon` was implemented in another PR, but it didn't get registered in the Injector.

client/src/bundles/vendor.js
```diff
@@ -51,8 +51,6 @@ require('../../../thirdparty/jquery-entwine/dist/jquery.entwine-dist.js');
 require('../../../thirdparty/jquery-cookie/jquery.cookie.js');
 require('../../../thirdparty/jquery-query/jquery.query.js');
 require('../../../thirdparty/jquery-form/jquery.form.js');
-require('../../../thirdparty/jquery-notice/jquery.notice.js');
```
maxime-rainville, line 54: We no longer need this plugin, because it fully got reimplemented in react. There's a polyfill to make sure existing calls still works.

client/src/components/LookupField/tests/LookupField-test.js
```diff
@@ -1,12 +1,11 @@
 /* global jest, describe, beforeEach, it, expect, require */
 
-jest.unmock('react');
-jest.unmock('react-dom/test-utils');
-jest.unmock('../LookupField');
-
 import React from 'react';
-import ReactTestUtils from 'react-dom/test-utils';
```
maxime-rainville, line 8: ReactTestUtils got bumped to a newer version that fires off an error because of some rogue input statement in reactstrap. My guess is that we can't fix this issue without upgrading reactstrap. So I'm switching this test to enzyme instead.

package.json
```diff
@@ -79,19 +80,20 @@
...
     "validator": "^6.1.0",
     "webpack-sources": "^1.1.0"
   },
   "devDependencies": {
     "@silverstripe/eslint-config": "^0.0.5",
-    "@silverstripe/webpack-config": "^1.5.0",
+    "@silverstripe/webpack-config": "^1.6.0",
```
maxime-rainville, line 96: This will need to be incremented to 1.7 once the webpack PR is merged and release.

package.json
```diff
@@ -79,19 +80,20 @@
     "react-router-config": "^4.4.0-beta.6",
     "react-router-dom": "^4.4.0-beta.6",
     "react-select": "^1.3",
-    "reactstrap": "^6.4.0",
+    "reactstrap": "^8.5.1",
```
maxime-rainville, line 83: We need to bump the dependency here to get the toast component from Reacstrap

maxime-rainville:
Everything has been rebased. I've added a big note at the top to the state of the issue and summarise all the discussion on the PR.

@silverstripeux There's currently a slight fade transition when you switch tab. To get the toasts working I had to upgrade the version of Reactstrap we use, which removed an API we were using to get that fade-in transition working. How attached are you to that transition? I can probably find a way to replicate it with a bit of hair pulling, but only if you think it's worth it.

Here's a video that shows the transition ... the transition is then turned off after the refresh: https://youtu.be/JDIC6sm-PZs

client/src/components/Tabs/TabItem.js
```diff
@@ -1,62 +1,22 @@
...
-    return (
-      <TabPane {...tabProps}>
-        <Fade in={this.isActive()}>
-          {this.props.children}
-        </Fade>
-      </TabPane>
-    );
-  }
-}
+const TabItem = ({ name, className, extraClass, disabled, children }) => (
```
maxime-rainville, line 5: I've reformatted this component to remove the transition because it's buggy on Reactstrap 8. BUt PUX needs to sign off on this before we merge.

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,160 @@
...
+
+  &:hover {
+    box-shadow: $z-depth-2;
+  }
+  animation: $transition-speed-mid ease-out 0s 1 toastFadeIn;
+
+  &--success,
+  &--good,
+  &--green {
+    border-color: $color-good;
```
Cheddam, line 75:
```suggestion
    border-color: map_get($theme-colors, success);
```

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  &--notice,
+  &--info,
+  &--blue {
+    border-color: $color-notice;
```
Cheddam, line 81:
```suggestion
    border-color: map_get($theme-colors, info);
```

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  &--warning,
+  &--caution,
+  &--yellow {
+    border-color: $color-warning;
```
Cheddam, line 87:
```suggestion
    border-color: map_get($theme-colors, warning);
```

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  &--bad,
+  &--error,
+  &--red {
+    border-color: $color-error;
```
Cheddam, line 93:
```suggestion
    border-color: map_get($theme-colors, danger);
```

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  &--dismissing {
+    animation: $transition-speed-mid*2 ease-out 0s 1 toastFadeOut;
```
Cheddam, line 97: Is `$transition-speed-mid*2` an arbitrary value? Maybe you could introduce `$transition-speed-extra-long`? 😅

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,160 @@
...
+    color: $gray-600;
+  }
+
+  &__close {
+    flex: 0 0;
+    padding: 0;
+    margin: 0;
+    opacity: .8;
+
+
```
Cheddam, line 131:
```suggestion
```

client/src/components/Toasts/Toast.js
```diff
@@ -0,0 +1,56 @@
...
+    e.preventDefault();
+    onDismiss();
+  };
+
+  const className = classnames('toast', `toast--${type}`, { 'toast--dismissing': dismissed });
+
+  return (
+    <BsToast className={className} isOpen aria-live="assertive" aria-atomic="true">
+      <ToastBody className="toast__body">
+        <Button className="toast__close" icon="cross-mark" noText onClick={toggle}>
```
Cheddam, line 34:
```suggestion
        <Button className="toast__close" icon="cancel" noText onClick={toggle}>
```

Cheddam: Sacha has also requested that the close button not have a hover state (e.g. no background) per the designs, but to retain the focus state (outline).

client/src/state/toasts/ToastsReducer.js
```diff
@@ -0,0 +1,68 @@
...
+ * @typedef {import('./ToastsTypes').ToastOption} ToastOption
+ */
+
+/**
+ * @type {ToastOption[]}
+ */
+const initState = [];
+
+/**
+ * Start a timeout to eventually dismiss a toast if the toast is not configured to indfenitely stay.
```
Cheddam, line 14:
```suggestion
 * Start a timeout to eventually dismiss a toast if the toast is not configured to stay indefinitely.
```

Cheddam (changes requested):
Looking really good, @maxime-rainville! In addition to a ton of comments, here are some notes from my testing:

- Found an issue where if a toast is triggered while you are hovering over the toast container, it doesn't pause
- The close icon doesn't render correctly in IE 11 (as shown below)
- Toasts are announced by VoiceOver (yay!) as "Dismiss (contents of message)" (hmm) and focus is _not_ applied to them, nor does it seem possible to focus them before they disappear (eek). The Toast pattern [isn't exactly a champion of accessible design](https://adrianroselli.com/2020/01/defining-toast-messages.html) in the first place, so I'm not sure how much you'll be able to improve this situation. In particular, I'm wondering whether we should drop actions as a feature - screenreader users effectively can't use them, so it doesn't seem like a good idea to provide them to developers. Will be chatting to @silverstripe-ux about this tomorrow.

Screen Shot 2020-09-14 at 5 27 38 PM
<small><em>IE 11 issue</em></small>

client/src/boot/registerComponents.js
maxime-rainville: That bit might not be necessary. I've added it in there because I was trying to debugging an out-of-sync `@silverstripe/webpack-config` dependency. I'll try removing it and see if it makes any difference.

client/src/components/Toasts/README.md
```diff
@@ -0,0 +1,127 @@
...
+
+Three other redux actions are available. Those are used internally by the `<ToastContainer />` to manage its action, but can be use by other components as well:
+
+- `dismiss()` will remove a toast notification. It expects the `id` of the toast to remove.
+- `pause()` prevents the auto-dismissal of all the toast. Dismissal can still be manully triggered by the user.
+- `resume()` resets the timeout of all the toasts.
+
+## Toast components
+
+**Note:** You shouldn't need to directly used the `<Toast />` and `<Toasts />` components to display Toast notifications. This section is provided for completeness.
```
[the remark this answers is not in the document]
maxime-rainville: I think it's convenient to have it there, so people going through the pattern library know how to interact with those Toast component they are looking at.

I'm not oppose to moving that content to the main doc site if you think it makes more sense there.

client/src/components/Toasts/Toasts.scss
maxime-rainville: I multiplied the value by a factor of 2 because there's two steps to the transition:
- fading out by transitioning opacity from `1` to `0`
- sliding up the toast notification below by reducing the max-height of the toast from `250px` to `0`.

I want both of those animation to take `$transition-speed-mid` seconds to complete. There is at least a little bit of logic to my madness.

client/src/boot/registerComponents.js
maxime-rainville: Doesn't look like it makes a difference. Removing.

client/src/components/Tabs/TabItem.js
maxime-rainville: Sacha has provided the thumbs up of approval.

client/src/components/Toasts/Toast.js
maxime-rainville: Change the icon as requested and remove the background. The X look misaligned a little bit as well, so I tweak the line height of the button so the x would be aligned with the middle of the first line

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,160 @@
...
+  height: auto;
+  border: 0;
+  border-left: 3px solid;
+  color: #666;
+  left: 0;
+  background: #fff;
+  opacity: 1;
+  box-shadow: $z-depth-1;
+  transition: box-shadow $transition-speed-fast ease;
+  width: 250px;
```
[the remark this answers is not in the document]
maxime-rainville: I cull the superfluous properties. A lot of those were copied over from the old toast styles.

The max-height is actually use for the slide up transition. This was remakably annoying to get working because the height of the toast is variable based on what content is in it.

client/src/containers/ToastsContainer/ToastsContainer.js
```diff
@@ -0,0 +1,44 @@
...
+
+// Just here for testing
+const magicToast = (display, idx) => ({
+  text: 'I just spawn more of myself.',
+  type: types[idx % types.length],
+  actions: [
+    { label: 'Same type', onClick: () => display(magicToast(display, idx)) },
+    { label: 'Next type', onClick: () => display(magicToast(display, idx + 1)) }
+  ]
+});
```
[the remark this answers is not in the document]
maxime-rainville: I'm removing the logic.

For future reference, I've put this test code in a gist https://gist.github.com/maxime-rainville/c09c7c47f48b64fbacb890ca9f7257fa

client/src/components/Toasts/Toasts.scss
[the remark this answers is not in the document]
maxime-rainville: I had to bring back the font properties because there's something in admin that's messes around with it. This doesn't happen in the pattern library.

templates/SilverStripe/Admin/Includes/ToastsContainer.ss
```diff
@@ -0,0 +1 @@
+<div class="toasts-container"></div>
```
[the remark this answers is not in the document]
maxime-rainville: Removed this file.

client/src/legacy/ToastsContainer.js
maxime-rainville: Done.

maxime-rainville:
Addressed all the inline comments.

> Found an issue where if a toast is triggered while you are hovering over the toast container, it doesn't pause

I'll have a look, but I think the problem here is that the mouse events only fires when the mouse moves. When an elements appears under the mouse or slides up over the mouse, that probably doesn't fire an event. Short of tracking the position of the cursor and trying to guess if the toast intersects with it, they're might not be an obvious fix to this.

> The close icon doesn't render correctly in IE 11 (as shown below)

I'll have a look tomorrow.

> Toasts are announced by VoiceOver (yay!) as "Dismiss (contents of message)" (hmm) and focus is _not_ applied to them, nor does it seem possible to focus them before they disappear (eek). The Toast pattern [isn't exactly a champion of accessible design](https://adrianroselli.com/2020/01/defining-toast-messages.html) in the first place, so I'm not sure how much you'll be able to improve this situation. In particular, I'm wondering whether we should drop actions as a feature - screenreader users effectively can't use them, so it doesn't seem like a good idea to provide them to developers. Will be chatting to @silverstripe-ux about this tomorrow.

I'll see if I can wrap the main message in something that screen readers will read. We probably don't want to auto-focus on the Toast when it shows. That will probably be even worst accessibility and that will prevent them from auto-dismissing.

You probably shouldn't put a very important action in something that disappears in 3 sec ... so in practice, these actions should always have an alternative way to be completed any way.

If the end goal is to provide a better alternative to the `.form-alert` message, we probably need those actions, because a lot of those messages contains links to other things.

Cheddam:
> I think the problem here is that the mouse events only fires when the mouse moves.

I'm not so sure about that - if I click the 'Next type' action on the demo toast, without moving my mouse, the new toast doesn't pause, but the other demo toast remains paused. That suggests to me that newly spawned toasts simply aren't being paused if they spawn when a pause is already in place.

> If the end goal is to provide a better alternative to the `.form-alert` message, we probably need those actions, because a lot of those messages contains links to other things.

Yeah, I hadn't considered that some existing toasts already contain links too. I still think this is a dangerous pattern, so it might be worth adding a note about accessibility to the new documentation for the feature (primarily to warn developers not to put important operations that can't be performed elsewhere in toasts.)

client/src/containers/ToastsContainer/ToastsContainer.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Wires the Toasts component into the redux store.
+ */
+const ToastsContainer = ({ toasts, actions: { success, display, dismiss, pause, resume } }) => {
+  // Just here for testing
+  useEffect(() => {
+    success('The Toast container has been initialised.');
+    display(magicToast(display, 0));
+  }, []);
```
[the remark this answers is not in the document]
maxime-rainville: Removed

maxime-rainville:
I've fixed the following things:
- IE11 close button alignment
- Toast notification added while the ToastContainer is in a _paused_ state will start off paused
- Extended the toast display time to 6 seconds
- The `warning` and `error` redux action will now create toasts notification without timeout under the assumption that those are more important and shouldn't auto-dismiss.
- Added some usage guideline to the pattern-lib doc

The screen reader I'm using on Linux was systematically reading new toast notifications as they appeared and we're using all the aria attribute as recommended by the official bootstrap doc. I'm not sure what else we can do here.

I had problem accessing the actual actions on the toast. In practice, those actions should always have an alternative means to be performed. I've added a comment to that effect in the usage guideline. 

The accessibility article @Cheddam sent over suggest things like having a "Notification drawer" where you can retrieve previous notifications. That's probably beyond the scope of this card, but we can look at this in a follow up issue if we think it's worthwhile.

client/src/components/Toasts/Toasts.scss
sachajudd: Could there be a comment added to describe why you've done this for future reference?

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,157 @@
...
+    margin: map_get($spacers, xs) map_get($spacers, xs) map_get($spacers, xs) map_get($spacers, sm);
+    display: flex;
+    flex-flow: row-reverse nowrap;
+    align-items: flex-start;
+  }
+
+  &__content {
+    flex: 1 0;
+    margin-right: map_get($spacers, sm);
+    color: $gray-600;
```
sachajudd, line 114:
```suggestion
    color: $gray-700;
```

sachajudd, line 114:
```suggestion
    color: $gray-700;
```

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,157 @@
...
+  &--success,
+  &--good,
+  &--green {
+    border-color: map_get($theme-colors, success);
+  }
+
+  &--notice,
+  &--info,
+  &--blue {
+    border-color: map_get($theme-colors, info);
```
sachajudd, line 73: Will stick with Paul's initial designs for colour for now. Easy to change later. It's not a theme colour but is under $colors so assume it'd be the below? Or even just $cyan?

```suggestion
    border-color: map_get($colors, cyan);
```

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,157 @@
...
+  &__action {
+    flex: 1 0;
+    color: $gray-600;
+    text-decoration: none;
+    margin: 0;
+    border-radius: 0;
+
+    &:hover,
+    &:focus {
+      background: $gray-100;
```
sachajudd, line 148: In our designs I have #F1F3F6 on hover but can't seem to find a variable for this. Okay to leave as $gray-100 for now.

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,157 @@
...
+    opacity: 0;
+    max-height: 0;
+    margin-bottom: 0;
+  }
+}
+
+
+.toast {
+  margin: 0 0 map_get($spacers, sm) 0;
+  border-radius: 3px;
```
sachajudd, line 42: Should be 4px. Probably could be a variable too? 
```

File: client/src/components/Toasts/Toasts.scss [hunk 40]
```diff
@@ -0,0 +1,157 @@
...
+  font-family: inherit;
+  font-size: inherit;
+  word-wrap: break-word;
+  max-height: 250px;
+  border: 0;
+  border-left: 3px solid;
```
sachajudd, line 48: Will need to be 4px if the border-radius is changing to 4px.

```suggestion
  border-left: 4px solid;
```

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,157 @@
...
+    flex: 1 0;
+    color: $gray-600;
+    text-decoration: none;
+    margin: 0;
+    border-radius: 0;
+
+    &:hover,
+    &:focus {
+      background: $gray-100;
+      color: $gray-700;
```
sachajudd, line 149: Colour should stay the same on hover. I thought we could remove it altogether but I see you're overriding the btn-link hover.

```suggestion
      color: $gray-600;
```

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,157 @@
...
+  width: 250px;
+
+  &:not(:last-child) {
+    margin-bottom: map_get($spacers, sm);
+  }
+
+  &:hover {
+    box-shadow: $z-depth-2;
+  }
+  animation: $transition-speed-mid ease-out 0s 1 toastFadeIn;
```
sachajudd, line 62: Is this just hanging out randomly on a line by itself or is it applied to something higher up I can't see? Can you move it up if so?

sachajudd (changes requested):
Some minor CSS recommendations and variable fixes. Otherwise looking awesome Max! 🙂

client/src/components/Toasts/Toasts.scss
Cheddam: Makes sense @maxime-rainville, I agree with Sacha on adding a note about it, does look quite arbitrary at a glance.

client/src/components/Toasts/Toasts.scss
sachajudd: I'll update the designs as this isn't the case there and works different in practice.

client/src/components/Toasts/Toasts.scss
```diff
@@ -0,0 +1,157 @@
...
+    margin-bottom: 0;
+  }
+}
+
+
+.toast {
+  margin: 0 0 map_get($spacers, sm) 0;
+  border-radius: 3px;
+  font-family: inherit;
+  font-size: inherit;
```
Cheddam, line 44: `font-family` really doesn't seem to do anything here. Removing `font-size` reduces the size of the content, but I think that'd make more sense to be directly set in `.toast__content`.

Cheddam (changes requested):
Pretty close, just a few minor things. I'm still getting wonky output from VoiceOver - I'll tinker and see if I can improve it before we merge.

client/src/components/Toasts/Toasts.scss
maxime-rainville: I've looked up the base bootstrap style https://github.com/twbs/bootstrap/blob/2bceb3e79d848a101e7aa5fc4b2020e681b1dfc3/scss/_toasts.scss

I'll override the bootstrap variables instead.

client/src/components/Toasts/Toasts.scss
maxime-rainville: Weird that the linter didn't throw a fit about this one. I've moved it up.

client/src/components/Toasts/Toasts.scss
maxime-rainville: I added a blurb.

maxime-rainville:
Fixed the behat test, but we'll need this PR to be merged first.

Cheddam (approved):
Awesome work, @maxime-rainville!

client/src/boot/registerComponents.js
```diff
@@ -98,5 +99,6 @@ export default () => {
     HtmlEditorField,
     NumberField,
     PopoverOptionSet,
+    ToastsContainer
```
robbieaverill, line 102:
```suggestion
    ToastsContainer,
```

client/src/components/Toasts/ToastActions.js
```diff
@@ -0,0 +1,59 @@
...
+        // Dismiss the toast after it has fired its action
+        onDismiss();
+      }
+    } };
+
+  return <Button color="link" className="toast__action" {...props}>{label}</Button>;
+};
+
+/**
+ * Display a list of toast actions. Up to 2 actions can be displayed.
```
robbieaverill, line 27: Sorry if you've already covered this, but why 2?

client/src/state/toasts/ToastsActions.js
```diff
@@ -0,0 +1,106 @@
...
+  return display({ text, type: 'success' });
+}
+
+/**
+ * Display an error toast
+ * @param {string} text
+ * @returns {reduxThunk}
+ */
+export function error(text) {
+  return display({ text, type: 'error', stay: true });
```
robbieaverill, line 73: Is it worth making the `stay` argument available for devs to pass? I might not want my errors/warnings to stick around.

client/src/state/toasts/ToastsTypes.js
```diff
@@ -0,0 +1,28 @@
...
+ * @property {number} timeout Reference to a timeout that will call dismissCallback.
+ *     Clearing this timeout will pause the toast dismissal.
+ */
+
+/**
+ * Definition of the Toasts state stored in the redux store.
+ * @typedef {Object} ToastState
+ * @property {boolean} paused Whatever the toast container is currently paused.
+ * @property {ToastOption[]} toasts List of current toast.
+ */
```
robbieaverill, line 27: I don't see any code - is this for typescript?

package.json
robbieaverill: This is kind of dangerous. We've had problems before with Reactstrap changes in core. In theory modules should bring their own Reactstrap versions of course. Let's just make sure we give it enough time after merge to test thoroughly before releasing.

robbieaverill (approved):
I am sooooo stoked to see this happening! Awesome @maxime-rainville, thank you! Thorough test coverage which I appreciate too 😄 

I left a couple of super minor comments.

In terms of your polyfill for `jquery.noticeAdd()`, I assume that the arguments we accept there are the same as what the original library provides for? Just sanity checking that really.

sachajudd (approved):
Looks good Max!

client/src/components/Toasts/ToastActions.js
maxime-rainville: That's what the designs call for. I didn't pick Paul's brain on why two. My guess is that it's a size constraint.

client/src/state/toasts/ToastsActions.js
maxime-rainville: `error`, `info`, `warning` and `success` are meant to be shorthand for `display`. That's why I made them as simple as possible, with the thinking that if you want a bit more control you should call display directly.

The rational for having warning/error stick around was that those are more important, so the message should stay there by default. notice/success can be missed or ignored without much downside, so it's OK to let them timeout.

client/src/state/toasts/ToastsTypes.js
maxime-rainville: Na it's not typescript. JSDoc is smart enough to load those type definitions across files, so you get a slightly smarter auto-complete. I would have left them in the main file where they are used, but those definitions are use across many of the toast redux files, so I figured it would be better to define them in one spot.

maxime-rainville:
> In terms of your polyfill for jquery.noticeAdd(), I assume that the arguments we accept there are the same as what the original library provides for? Just sanity checking that really.

@robbieaverill The only difference is that the new implementation will ignore the `stayTime` attribute. In practice we weren't using this anywhere, and it would have made the redux timeout logic meaningfully more complex.

maxime-rainville:
I've just bump the webpack dependency to 1.7.0 and squash everything into one commit. Behat PR has been merged.

Just waiting the build to be green :crossed_fingers:

maxime-rainville:
Looks like merging this PR might have broken the following fields because of the ReactStrap update:
- Readonly HiddenField,
- Readonly CheckboxSetField
- LookupField
- Readonly HTMLEditorField
- Readonly Single Select

maxime-rainville:
The following builds may have been broken because of this PR:
- campaign admin https://travis-ci.org/github/silverstripe/silverstripe-campaign-admin/jobs/730137400
- versioned-admin https://travis-ci.org/github/silverstripe/silverstripe-versioned-admin/jobs/730138441

Those failures are semi-expected because we bump reactstrap from v6 to v8. That's a bullet we were going to have to bite at some stage any way.

PR #1072, opened 2020-07-08
FIX: Decouple ModelAdmin URL slug from class.

sminnee:
The default slug of a sanitised classname will still be in place, but
now the key can be set to a slug and the class provided in a dataClass
property.

This has a couple of benefits:

 - Slugs can be shorter (ModelAdmin URLs are verbose, to say the least)
 - The same class can be added to a ModelAdmin multiple times, for
   example with different filters.

If the dataClass is the same as the slug, backwards compatibility is
preserved. If the slug differs, then code that expected to find the
Slug in $this->modelClass will instead have to check $this->modelTab.

However, as this doesn’t break any upgrades (since the slug will never
differ on those) my view is that it’s minor-upgrade safe.

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -56,12 +56,14 @@ abstract class ModelAdmin extends LeftAndMain
      * Extended notation with options (e.g. custom titles):
      * <code>
      * array(
-     *   'MyObjectClass' => array('title' => "Custom title")
+     *   'MyObjectClass' => ['title' => "Custom title"]
+     *   'urlslug' => ['title' => "Another title", 'dataClass' => MyNamespacedClass::class]
      * )
      * </code>
      *
      * Available options:
      * - 'title': Set custom titles for the tabs or dropdown names
+     * - 'dataClass': The class name being managed. Defaults to the key. Useful for making shorter URLs or placing the same class in multiple tabs
      *
      * @config
      * @var array|string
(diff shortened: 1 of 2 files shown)


code/ModelAdmin.php
```diff
@@ -56,12 +56,14 @@ abstract class ModelAdmin extends LeftAndMain
      * Extended notation with options (e.g. custom titles):
      * <code>
      * array(
-     *   'MyObjectClass' => array('title' => "Custom title")
+     *   'MyObjectClass' => ['title' => "Custom title"]
+     *   'urlslug' => ['title' => "Antoher title", 'dataClass' => MyNamespacedClass::class]
```
robbieaverill, line 60:
```suggestion
     *   'urlslug' => ['title' => "Another title", 'dataClass' => MyNamespacedClass::class]
```

sminnee:
The failure is just PHPCS

code/ModelAdmin.php
```diff
@@ -459,8 +469,10 @@ public function getManagedModels()
         // Normalize models to have their model class in array key
         foreach ($models as $k => $v) {
             if (is_numeric($k)) {
-                $models[$v] = array('title' => singleton($v)->i18n_plural_name());
+                $models[$v] = ['dataClass' => $v, 'title' => singleton($v)->i18n_plural_name()];
                 unset($models[$k]);
+            } else if (is_array($v) && !isset($v['dataClass'])) {
```
ScopeyNZ, line 474:
```suggestion
            } elseif (is_array($v) && !isset($v['dataClass'])) {
```

PR #1078, opened 2020-07-14
non static call to static method

lpostiglione:
As getRequiredPermissions is declared static "static::" should be used instead of "$this->"


michalkleiner:
Although static methods can be used non-statically, if they don't use $this, it's a bit confusing and inconsistent in terms of expectations. So if the method doesn't need anything static inside, could it be changed to an instance method? Is there any reason for keeping it as it is?

emteknetnz:

emteknetnz:

PR #1080, opened 2020-07-16
Follow up: Add BC support for getCMSValidator

chrispenny:
Follow up for: #1013 

Need to add BC support for the old `getCMSValidator()` method in case someone is on an older version of `framework`.


kinglozzer:
Thanks @chrispenny!

PR #1082, opened 2020-07-28
FIX TinyMCE bug adding a link after removing a link

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-cms/issues/2530

Issue is that after removing a link, you can end up with 3x textNodes e.g. removing the link from

`The <a href="http://example.com">quick</a> brown fox`

Becomes

```
The
quick
brown fox
```

This messes up tinymce.js `getRng()` and subsequent `isCollapsed()` functions when you select the word "quick" to insert a link

This PR makes use of the https://developer.mozilla.org/en-US/docs/Web/API/Node/normalize to join the adjacent text nodes so this no longer happens


emteknetnz:
@maxime-rainville I have moved the function as recommended

PR #1083, opened 2020-07-29
FIX UsedOnTable tweaks

emteknetnz:
**Update: Don't merge this. We'll do this work separately in a different PR on an issue in sprint with further enhancements.  Keep this PR open in a draft state for now so that it's easy to refer back to it**

Tweaks for ~https://github.com/silverstripeltd/product-issues/issues/222~ https://github.com/silverstripe/silverstripe-asset-admin/issues/903 so that the used on table resembles what's in the DSM

DSM https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/asset/componentContainers/5f152068f6772b33db2e52b0/tab/design?mode=preview

Modifications:

1) Remove draft/modified badges
2) Combine 2nd and 3rd columns together to match design in DSM
3) Upper case first letter of words
4) Show "Untitled" if the record does not have a title

Elemental block's have a `singlular_name` that is lowercase e.g. 'content block'.  Use `ucwords()` so they match the other `Types` on the used on table

**Before**

image

**After**

image

Note: the `<tr>` hover state that shows a highlighted row is an existing generic table style.  It's honestly kind of weird though I just left it alone

image

The PR previously had it so that you could click the `<tr>` as if you were clicking the nested `<a>` hyperlink, though this had poor accessibility so I simply remove that from the PR


client/src/components/UsedOnTable/UsedOnTable.scss
```diff
@@ -1,16 +1,25 @@
...
+}
+
 .used-on__badge {
   border-color: #cf3f00;
   color: $state-draft;
   background-color: $state-draft-bg;
 }
 
-.used-on__col--index {
+.used-on__index {
```
robbieaverill, line 11: Are you able to make this change without changing the class names? I'd consider this a breaking change, devs may have CSS or JavaScript referencing this class name

code/Forms/UsedOnTable.php
```diff
@@ -85,8 +85,8 @@ public function usage(HTTPRequest $request)
 
             $usageData[] = [
                 'id' => $userId,
-                'title' => $user->getTitle(),
-                'type' => $user->singular_name(),
+                'title' => $user->getTitle() ?: '(Untitled)',
```
robbieaverill, line 88:
```suggestion
                'title' => $user->getTitle() ?: _t(__CLASS__ . '.UNTITLED', '(Untitled)'),
```

client/src/components/UsedOnTable/UsedOnTable.js
```diff
@@ -64,23 +63,24 @@ class UsedOnTable extends PureComponent {
...
-      : null;
-
     const titleLabel = (link)
-      ? <a className="used-on__edit-link" href={link}>{title} {badge}</a>
-      : <span>{title} {badge}</span>;
+      ? <a className="used-on__edit-link" href={link}>{title}</a>
+      : <span>{title}</span>;
+
+    const handleTrClick = (event) => {
+      event.target.querySelector('a').click();
```
ScopeyNZ, line 74:
```suggestion
      window.location = link
```

I assume this is the intention of the TR click. I don't think using DOM query API to manually trigger the click is very obvious, and probably more prone to bugs.

client/src/components/UsedOnTable/UsedOnTable.js
```diff
@@ -64,23 +63,24 @@ class UsedOnTable extends PureComponent {
...
+    };
 
     return (
-      <tr key={id}>
-        <td className="used-on__col--index">{index + 1}</td>
-        <td className="used-on__col--title">{titleLabel}</td>
-        <td className="used-on__col--type">{type}</td>
+      <tr key={id} className="used-on__row" onClick={handleTrClick}>
```
robbieaverill, line 78: I think you also need to give `role=button` style attributes when making things like this clickable (for accessibility)

emteknetnz:
@ScopeyNZ @robbieaverill I've made the requested changes

client/src/components/UsedOnTable/UsedOnTable.js
```diff
@@ -64,23 +63,24 @@ class UsedOnTable extends PureComponent {
...
     } = data;
 
-    const badge = (state)
-      ? <span className={classnames('badge', 'used-on__badge', `status-${state}`)}>{state}</span>
-      : null;
-
     const titleLabel = (link)
-      ? <a className="used-on__edit-link" href={link}>{title} {badge}</a>
-      : <span>{title} {badge}</span>;
+      ? <a className="used-on__edit-link" href={link}>{title}</a>
```
maxime-rainville, line 70: This approach is fine, but we could have made this a simple functional component which would have produce cleaner code. Something to keep in mind next time.

client/src/components/UsedOnTable/UsedOnTable.js
```diff
@@ -64,23 +63,24 @@ class UsedOnTable extends PureComponent {
...
+      : <span>{title}</span>;
+
+    const handleTrClick = () => {
+      window.location = link;
+    };
 
     return (
-      <tr key={id}>
+      <tr key={id} className="used-on__row" onClick={handleTrClick} role="button">
```
maxime-rainville, line 78: This might cause some accessibility problems. I don't think you can tab to a `<tr>` element.

client/src/components/UsedOnTable/UsedOnTable.js
```diff
@@ -64,23 +63,24 @@ class UsedOnTable extends PureComponent {
...
-      ? <span className={classnames('badge', 'used-on__badge', `status-${state}`)}>{state}</span>
-      : null;
-
     const titleLabel = (link)
-      ? <a className="used-on__edit-link" href={link}>{title} {badge}</a>
-      : <span>{title} {badge}</span>;
+      ? <a className="used-on__edit-link" href={link}>{title}</a>
+      : <span>{title}</span>;
+
+    const handleTrClick = () => {
```
maxime-rainville, line 73: I would rather we found a way to use a proper <a> tag for this.

clarkepaul:
Yeah that's not going to work unfortunately. The title shouldn't be the link or the focusable element, it should be the whole row as per the hover state indicates. For example, Reports in the CMS are tab-able and do not show underline of the headings. 

I know there is no actual Gridfield type component in Storybook so no standards to follow. But effectively it should appear to operate as a Gridfield item.

These Bootstrap styles should be builtin already as an example (one way of doing it) https://getbootstrap.com/docs/4.5/components/list-group/#flush

emteknetnz:
@clarkepaul I'm not clear what your feedback is, is it:
a) Please redo, this time make the table rows accessible, OR
b) Yup, the table rows accessible, lots of rework required, this is current fine as is for now, merge it
?

clarkepaul:
I'm saying it wouldn't meet the definition of done we have if we were to build it as new functionality. It wasn't in the acceptance criteria to make accessible so could be debatable as to whether this should be included. I would hope that if it was an easy-ish thing to do you would just include some improvement there to reduce double handling, but if you think this should be handled separately I understand it'll blow the issue out. I think it's up to you really to make a call, justify what approach fits best, raise it with team cc @brynwhyman . 

If it's a no go then we'd raise another issue.

emteknetnz:
@clarkepaul new issue makes sense here, seems like there's quite a bit to consider.

The design tweaks done so far are actually scope creep from the original issue which was only for not showing particular things such as FileLink's and ElementalAreas on the used-on-table.  The only reason the design tweaks are here are because you linked to the DSM and I thought it looked pretty easy so why not? =)

Re the `<tr>` click, in this piece of work we could either:
a) keep as is for now, refactor in the next issue
b) remove it, rely on the `<a>`

Let me know how you'd like to proceed

client/src/components/UsedOnTable/UsedOnTable.js
robbieaverill: You can if you tell the browser it's a button. It may need a tabindex prop too though. Ideally we avoid doing this if possible, but GridField based UIs are pretty table dominated at the moment.

robbieaverill: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#Basic_button_example

maxime-rainville:
I'm not sure this is that much more work. You would just need to swap a few `<table />`, `<tr />`, `<td />` elements other elements and use a few judicious `display: table|table-row|table-cell` css rules. Or you could probably achieve a similar results with flex boxes.

Is there something I'm missing?

emteknetnz:
I think it's a substantial amount of work.

The core issue to me is that accessibility is currently in a pretty bad state, for instance I press the tab key in firefox and I don't go the things such at 'Details'|'Permissions'|'Used On' tabs, nor the `<a>` links in the used on table.

The design improvements here are no less accessible then things were previously

I'll reiterate that the design improvements here ARE NOT in the linked issue and are just scope creep that I just did them as a bonus.  At this point I'm inclined to just remove them since they're effectively blocking this pull-request from being merged

brynwhyman:
The design tweaks are scope creep to what we had previously planned to do here. I think they look much better but it's clear it means more work to get it to a good state. 

I'd suggest removing changes related to "Whole row clickable instead of just the hyperlink" and any other changes that result in regressed usability. For any improvements that aren't built, let's raise a new issue and treat it separately. cc @silverstripeux

emteknetnz:
@brynwhyman it's not regressed usability.  They table was already non-accessible.  I'd suggest keeping things as they are.

code/Forms/UsedOnTable.php
```diff
@@ -85,8 +85,8 @@ public function usage(HTTPRequest $request)
 
             $usageData[] = [
                 'id' => $userId,
-                'title' => $user->getTitle(),
-                'type' => $user->singular_name(),
+                'title' => $user->getTitle() ?: _t(__CLASS__ . '.UNTITLED', '(Untitled)'),
```
emteknetnz, line 88: This should be `Untitled` rather than `(Untitled)` to match what was agreed [here](https://github.com/dnadesign/silverstripe-elemental/issues/574#issuecomment-675830833)

emteknetnz:
@clarkepaul I've removed the `<tr>` click since that's what was causing the discussion around accessibility

I've updated the description of this PR re-list what the modifications this PR does (main two are remove draft badges and combine 2nd and 3rd columns so that it closer matches DSM)

Are you happy with the state this PR is in now?

clarkepaul:
Yup sounds right (without looking at the code). Can discuss further tomorrow.

bergice:
Should we consider doing something about this: https://github.com/dnadesign/silverstripe-elemental/pull/817#issuecomment-671243867 ?

emteknetnz:
Closed in favor of https://github.com/silverstripe/silverstripe-admin/pull/1097

PR #1086, opened 2020-07-30
FIX Unload previous preview when new URL is not available

Cheddam:
Issue https://github.com/silverstripe/silverstripe-cms/issues/1945#issuecomment-869322412

Some pagetypes (and other previewable objects) may not have any available preview states. A primary example of this is RedirectorPage. When switching the editor to one of these pagetypes, the prior preview was retained, though covered up by a broken overlay. This change, along with the removal of the broken overlay, means that a simple 'No preview available' message will be successfully displayed in these situations.

This accompanies a [template fix in the CMS module](https://github.com/silverstripe/silverstripe-cms/pull/2572).


emteknetnz:
Seems like these PR's should be released as a patch rather than a minor?

Cheddam:
I raised them as minor changes for a little extra safety, but they're probably fine to go into a patch. I'll rebase and take a look at the Behat failures.

emteknetnz:
@Cheddam This PR hasn't had any activity for a while :-)  Needs a rebase and travis is red.  Are you going to do further work on this or should we now close it?

bergice:
Better than before, but it may need some more polishing:

* When changing view mode, a scroll bar sometimes shows up at the bottom of the screen:
image
    * This also causes some weird behavior where opening the toggle button will hide the scroll bar immediately after clicking down which then causes the user to click up on one of the options in the toggle which causes it to close.
* I would consider removing the toggle and preview width buttons or marking them disabled with a tooltip. Maybe even just not show any preview at all. (This is a nice to have though)

Cheddam:
As noted in the other PR, I can reproduce the weird overflow bug without these PRs present, so I think that should be raised and addressed separately.

Cheddam:
Ran Behat locally, can verify that this failure only occurs when the [related CMS PR](https://github.com/silverstripe/silverstripe-cms/pull/2572) is not present.

emteknetnz (approved):

emteknetnz:
Have rebased and retarted the PRs to 1.8/4.8. Tested in firefox and chrome, works great.  Will merge when PRs when Travis goes green.

emteknetnz (changes requested):
Behat tests need to be fixed

I noted that CMS will required an additional composer constraint here https://github.com/silverstripe/silverstripe-cms/pull/2572#pullrequestreview-689063382 - not sure if there needs to be on this side as well (likely not)

emteknetnz:
Behat test failures are because the corresponding cms test has not been merged https://github.com/silverstripe/silverstripe-cms/pull/2572

We should merge that first, then rerun the tests on this one

maxime-rainville (approved):
Tested locally ... works great.

Issue #1018, opened 2020-03-03
Toast notifications position to move below the top navigation toolbar


clarkepaul:
This is a simple suggestion to move the toast notifications just under the top navigation toolbar. This new position would still be familiar enough to users, it would block the top right tabs etc.

I'm suggesting we apply CSS top: 68px to the  toast wrapper. Note the screenshot is missing the drop-shadow around the toast notification [raised here](https://github.com/silverstripe/silverstripe-admin/issues/1017).
<img width="199" alt="Pasted_Image_4_03_20__11_06_AM" src="https://user-images.githubusercontent.com/555033/75825469-4e073300-5e0a-11ea-8849-b10a12972e51.png">

Problem: It is often that users will perform a page action and want to navigate away to one of the other main tab areas (e.g. page settings), and therefore need to wait until the notification disappears.

There are other alternative positions which might also work but I feel the top right corner is still a fairly good spot. Centre top would be hard to tell what it would be covering, bottom left would cover preview actions, bottom left it would be over menu actions.

michalkleiner:
What about adding a close button (small x in the corner) to the notification to be able to dismiss it immediately by the user? Or just by clicking anywhere on it?

Our experience, related and unrelated to this topic at the same time, is that users often miss the notification completely on wider screens in the edit mode when it's just too far out of the main area of focus. Perhaps a thing to factor in as well.

clarkepaul:
They do have a close icon in the corner but its probably a matter of it being small and slightly hard to get to when you are already trying to get to a tab. When hovered they remain in view and don't disappear so that people can read them properly.

As for them not being noticed, I think most positions have the same issue so it might be more about making the look of them stand out more? Positioned top centre might make them more noticeable but that might also annoy people and not sure if the stacking effect in the centre works as well when multiple notifications come in.

emteknetnz:
Fixed in https://github.com/silverstripe/silverstripe-admin/pull/1094

michalkleiner:
Sorry, yes, I missed the close button in the small screenshot. Also, now with the dropped shadow they look better and a bit more noticeable.

I'm no UX expert but I feel like the position of the notification shouldn't care about what it overlays if it's related to the whole CMS window. Moving it down below the tabs nav makes it seem nested inside the content tab area, not 'top-level', if you know what I mean.

Over a range of clients and projects we haven't yet had feedback similar to the stated problem 
[...]

PR #1094, opened 2020-08-14
FIX Move toast notification so that tabs behind it are clickable

emteknetnz:
Fix for https://github.com/silverstripe/silverstripe-admin/issues/1018

Before
image

After
image


emteknetnz:
@clarkepaul based on your [comment](https://github.com/silverstripe/silverstripe-admin/issues/1018#issuecomment-674576359) in the parent issue - is this still something you want to be merged

clarkepaul:
Yes please. People have requested it, that mention I made of moving tabs isn't going to happen anytime soon.

bergice (commented):
From a code point of view this looks OK.

However, I'm inclined to agree with @michalkleiner that it makes assumptions that the toast will always be displayed on a page with the navbar present:

> I'm no UX expert but I feel like the position of the notification shouldn't care about what it overlays if it's related to the whole CMS window. Moving it down below the tabs nav makes it seem nested inside the content tab area, not 'top-level', if you know what I mean.

Will let @clarkepaul make the final call. Would be good to see the related toast issue and who initially reported this:

> This was raised by developers through another toast issue, will link if I can find it but was quite a few mentions about it. I can only hope its for genuine reasons as It's unlikely for a user to complain about this type thing through our CMS feedback form. Personally I would like to move the top right tabs, because I know when testing with new users they don't look far right for things like settings.

clarkepaul:
I've had a decent dig around in older issues but couldn't find mention of who and how often this was requested. It was never the main focus those issues it was mentioned in from memory, but more of a followup on conversations had... stuff about moving the position to top middle of CMS, or bottom right. I'm personally all good to close this issue without the PR, if it comes up again we can dig this one up again. I agree with whats been said above.

robbieaverill:
It's reasonably likely that this was raised by one of the Silverstripe team members, since the toast notifications being above the CMS tabs meant that Behat tests would often fail when trying to click on CMS tabs. I agree with @clarkepaul that if this is the only driver it's probably not worth changing.

PR #1097, opened 2020-08-25
NEW Show nested DataObjects on a single row in the file used on table

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1098

Investigation issue https://github.com/silverstripeltd/product-issues/issues/249

DON'T MERGE YET

build with `yarn dev` for now because jest tests haven't yet been updated

image


code/Forms/UsedOnTable.php
```diff
@@ -70,30 +74,119 @@ public function getSchemaDataDefaults()
...
+                'type' => ucwords($dataObject->singular_name()),
+                //'state' => $state,
                 'link' => $link,
+                'class' => get_class($dataObject),
+                'parentID' => $parentID,
+                'parentClass' => $parentClass,
+                'ancestorPageID' => $ancestorPageID,
+                'isPage' => $isPage,
             ];
+            $this->extend('updateUsageArray', $arr, $dataObject);
```
emteknetnz, line 133: This whole approach of looking `->Parent()` may be a little pointless in that there's not many (any?) non-page DataObjects that follow this pattern (at least in kitchen-sink).  It may be more sense to not look for `->Parent()` by default and instead rely on this new extension hook.

michalkleiner: What about `File` and `Folder` or `TaxonomyTerm`? Those come to mind as using the `Hierarchy` extension which provides has_one `Parent`.

sachajudd (changes requested):
Can you please remove the two font-weight's of 500 from your PR? These should just be default which is 400 and will come through automatically. I'll update the designs to showcase this more clearly. Could you please ensure the title is 13px and the description is 11px also.

emteknetnz:
@sachajudd sorry I should have updated description.  This PR won't get merged as is basically just used as reference for future work, possibly some of it will be copy-pasted.  We ended up spinning up a new issue for used on table feature work.
https://github.com/silverstripe/silverstripe-admin/issues/1098

I'll close this PR.

emteknetnz:
Code form this PR has been somewhat moved to https://github.com/silverstripe/silverstripe-admin/pull/1121

PR #1101, opened 2020-08-26
FIX: Fix ModelAdmin slug flex when getManagedModels() is overridden.

sminnee:
The slug flexibility provided in #1072 introduced a bug, which this fixes:

The logic added to getManagedModels() won’t apply if the method is overridden and custom values returned

Note that the bug introduced in 1072 breaks ArchiveAdmin in versioned-admin.

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -154,7 +154,7 @@ abstract class ModelAdmin extends LeftAndMain
             $this->modelTab = key($models);
         }
 
-        $this->modelClass = $models[$this->modelTab]['dataClass'];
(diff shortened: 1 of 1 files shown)


emteknetnz:
Bugfix is targeting `1` instead of `1.6` because the regression was cause by a PR that targeted `1`

@sminnee `Note that this breaks ArchiveAdmin in versioned-admin.` Are you saying that THIS pr causes the breakage? Or the PREVIOUS pr caused the breakage?

sminnee:
1072 broke archiveadmin, this fixes it.

emteknetnz:
OK cool, will look to test/merge in the next week or so

Issue #886, opened 2019-05-27
Gridfield Better Buttons don't handle sorting and pagination


maxime-rainville:
Our better button implementation in core seems to have some issues.
- [ ] When using the next button, you can't seem to go more than 2 entries past the current page
- [ ] When sorting the gridfield, the next/previous button don't respect your sort order.

https://youtu.be/Mg8sptRZKc8

## Parent issue:
- https://github.com/silverstripeltd/product-issues/issues/254

## Pull request
* [x] https://github.com/silverstripe/silverstripe-admin/pull/1103
* [ ] https://github.com/silverstripe/silverstripe-framework/pull/9827
* [ ] https://github.com/silverstripe/silverstripe-admin/pull/1238

roopamjain01:
I am also getting the issue with previous and next button, please find the issue reproduction detail below.

My Silverstripe version is 4.4 and grid field next and previous button functionality working randomly.

I have grid with 25 records. I have two pages in my grid, means item per page is 15 in the grid.

When I click on first record on the first page in the grid, I can press next button till the 16th record and then next button is disabled on the 17th record.

But when I start from second page in the gird from 18th number record then next record should be the 19th number record but currently after 18th record I am getting 2nd number record as a next record and after that I can see next button till the 16th record and on the 17th record again next button is disabled.

In my case I expect next button should work for all the 25 records in the grid and next button should be disabled on the 25th record. But next button disabled after 17th records according to the current code in GridFieldDetailForm_ItemRequest.php at line 540($limit = $itemsPerPage + 2).

In second senario, when I click on the 15th number of record on the page 1, I can see next button only for 15th and 16th record and then again next button is disabled on the 17th record, but I think I should be able to see all the record from 15th to 25th (all 10 recods) and next button should be enabled.

Same issue with previous button as well, when I click on the 25th number record on page 2, I can see previous button till 16th record and then previous button is disabled on 15th record on page 1. But I expect previous button should work till first record in the grid.

Please let me know If you need any other information.

bergice:
The pagination will either be incorrect or missing in `\SilverStripe\Forms\GridField\GridFieldDetailForm_ItemRequest::getAdjacentRecordID`, meaning that the resulting list will be filtered to the wrong section which is why the previous/next buttons will be disabled.

Ideally we should not depend on the pagination offset and items per page as it's possible to open a record by URL without passing the query string to the detail form.

So we'd need to find the index of the current record and then offset from there somehow.

[...]

PR #1103, opened 2020-08-28
BUG Do not override grid state with search parameters

maxime-rainville:
There's a bit of logic hat overrides the girdstate when you click the detail button with your search parameter, which might not reflect your current gridstate.

This video illustrate the original problem and the end results when the patch is applied.

https://youtu.be/Jw-zdpkXe-Q

# Parent issue
* https://github.com/silverstripe/silverstripe-admin/issues/886


sminnee:
I think that we should get better about embedding grid state changes into the URL using the history js api

Benefits

* back button from a detail form goes back to the correct state
* copy/paste grid field URL goes to the state
* saving a multi-page editable gridfield returns to the right page

The main worry I would have is multiple grid fields on a single form, to check that the grid state is linked to the right grid.

I raise this here because I'm not sure if such work would conflict with this PR?

maxime-rainville:
An alternative approach could be to make sure that any grid state change are immediately reflected in the URL. Even if we go with this kind of approach, I don't think we should directly read the GET parameters from the URL. We should have some sort of abstraction that can tell us what the current grid state is.

When you first access the a given URL, that abstraction would be hydrated with the grid state provided in the URL. Then has the grid sate changes, the abstraction would update the URL using the History JS API.

sminnee:
> An alternative approach could be to make sure that any grid state change are immediately reflected in the URL

Sorry that's what I meant by "embedding grid state changes into the URL using the history js api"

I believe that with that in place, the server-side render of GridField should be able to do what's needed without special post-render treatment being needed in the JavaScript. At the very least, I would start by setting that theory using some handcrafted URLs.

I agree that the actual manipulation of the gridstate on the client side should be captured in a shared helper of some kind.

bergice (commented):
I couldn't get this working.

When going back to the previous URL from the detail view, nothing is filtered and the query params are still empty.
When clicking the back button, the query params are present, but it's not filtering anything.

maxime-rainville:
> When going back to the previous URL from the detail view, nothing is filtered and the query params are still empty.
When clicking the back button, the query params are present, but it's not filtering anything.

That's not exactly what this thing was trying to fix (although yes, we should fix that). 

The bug I was trying to fix is this.
- Access a ModalAdmin
- Sort by one column.
- Click on one entry. You'll be taken to the EditDetail form. Notice that when using "better-button" to navigate the result, the sorting order is honoured.
- Return to the list view.
- Sort by a different column.
- Click on one entry. Try navigating results using "better-button".

Expected results: The next result returns matches the second sorting order.
Actual results: The original sorting order selected is still used.

PR #1114, opened 2020-09-21
BUG Temporary fix to use the most up-tot-date version of chromedriver

maxime-rainville:
ChromeDriver 85.0.4183.83 fixed a security issue however it introduce a bug that causes chromedriver to crash when interacting when some user dialogue. This is preventing us from a running an admin test.

ChromeDriver 85.0.4183.87 fixes the issue, but it hasn't hit the Ubuntu packages repo yet. This is a temporary fix to bypass  this issue.

# Related work
* blocks https://github.com/silverstripe/silverstripe-admin/pull/1068


dnsl48 (approved):
Beautiful!

PR #1115, opened 2020-09-21
Try using chromium.chromedriver

maxime-rainville:
Experimental PR to see if chromium.chromedriver makesa a difference when running PRs.


maxime-rainville:
Doesn't work.

PR #1120, opened 2020-09-24
Disable model admin import from csv by default

UndefinedOffset:
Import from csv is a very powerful feature in the core of the cms however it seems to be one that is rarely used by our clients, and when they do they seem to get themselves into more trouble than it's worth, so we end up disabling it in 99% of our model admins. I wonder if it makes sense to switch the default value of this to false given that it seems to be the exception to the rule that this is really needed and can practically be used given the broad uses of model admin.

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -106,7 +106,7 @@ abstract class ModelAdmin extends LeftAndMain
      * This variable can be a boolean or an array.
      * If array, you can list className you want the form to appear on. i.e. array('myClassOne','myClassTwo')
      */
-    public $showImportForm = true;
+    public $showImportForm = false;
 
     /**
      * Change this variable if you don't want the gridfield search to appear.
--- a/tests/php/ModelAdminTest/OverridenModelAdmin.php
+++ b/tests/php/ModelAdminTest/OverridenModelAdmin.php
@@ -42,6 +42,8 @@ class OverridenModelAdmin extends ModelAdmin implements TestOnly
         'updateGridFieldConfig' => 0,
     ];
 
+    public $showImportForm = true;
+
 
     protected function getGridField(): GridField
     {


robbieaverill (approved):
Sounds like a decent justification to me.

cc @silverstripe/core-team

robbieaverill:
We would definitely need a line in the changelog explaining the change though

sminnee:
Yeah I think we'd need to spell this out in the upgrade guide for 4.7, so people don't pull it in without realising, in the case where they required the feature.

UndefinedOffset:
I agree, it would need to be one of those things that is spelled out in the readme for sure. Pretty sure the import csv button has been on by default since the 2.x days.

chillu:
I think the CSV import is a pretty powerful way to work around ModelAdmin batch edit limitations, but I also acknowledge the potential foot gun. I think a better answer would be to have a CSV import preview which clearly states what's about to happen (overwrite X rows, delete Y rows, etc). I would prefer if we improved the user docs on this feature, and continue to empower users to do things without asking devs for every little thing (in their mind). You could also add this as a warning for developers early on in the "how to create modeladmin" docs.

What are your specific concerns on where users misunderstand or misuse the functionality?

chillu:
My worry with disabling that is that developers will read this changelog, not realise that users have a specific workflow that involves CSV imports, and just don't bother with re-enabling it.

UndefinedOffset:
I agree that a docs update is needed as well regardless how this goes whether it's merged or not. I think it needs to be spelt out very clearly to users in the user docs what this can do and to developers. I also agree that regardless whether this is merged or not a preview of what is going to happen with an additional approval to continue is important. I think that it also needs to be much clearer in general for users what the schema is for the import so invalid files are less likely to happen or will not cause destruction of data.

In the case of what brought this around we had a client use this feature and replace this entire products database (luckily not an e-commerce system just a gallery) with a handful of completely blank products. We can't confirm for sure that the import was used or what they uploaded since our logging isn't complete enough but we were able to replicate the aftermath using it. It would seem that the admin imported an invalid csv which the import interpreted as blank products.

This is a very powerful tool and I agree one that definitely has its place but I personally think given the broad uses of model admin. That it perhaps is one feature that should be opt-in rather than opt-out to prevent cases like above where a default feature could cause big issues if it's not used properly. I think if it's explained clear enough in the change logs and the docs as a whole turning it off by default may not have such a negative impact.

chillu:
> We can't confirm for sure that the import was used or what they uploaded since our logging isn't complete enough but we were able to replicate the aftermath using it

Sounds like we need to syslog whenever a CSV import is performed, incl. the author and file name, and possibly even the effect it had (updated X records, deleted Y records).

> It would seem that the admin imported an invalid csv which the import interpreted as blank products.

I'd rather address this bug than disable the feature. That's how I'd expect CSV import to work:

 * Where a unique identifier (`ID` by default) column is found, update existing records
 * Where columns are missing on updates, *don't* delete existing values
 * Where columns are empty on updates, do delete existing values
 * Where a unique identifier column isn't found, create new records. We could potentially add a checkbox to confirm this behaviour, since I'd imagine its often triggered accidentally

> That it perhaps is one feature that should be opt-in rather than opt-out to prevent cases like above

Sorry, but I haven't heard a specific bug report in this comment. I'd rather start by making the feature safer to use than to disable it. I would be less staunch on this if we had good batch editing abilities in ModelAdmin (incl. find and replace), but we don't - so spreadsheet editing in an essential advanced user tool. And that shouldn't require users begging a developer to enable a feature they probably don't even know exists if it's disabled by default.

bergice:
I don't think disabling the CSV importer is a great outcome here, so I'll close this PR for now, but I'm happy with follow-up issues / PR's.

Here are some other ideas:

- (quick win) Add CSV import / export permissions (only enabled by default for Administrators).
- Add review capabilities to advanced workflow so certain users would need to approve the import first.
- Add a preview table screen before applying a CSV import so you can see how it will affect the data before causing potentially catastrophic alterations (this would be nice as it is normally scary hitting that import button as you don't always know how the importer will alter your data).

PR #1124, opened 2020-09-25
Removing BatchActionParameters as it's not used anymore

bergice:
Resolves https://github.com/silverstripe/silverstripe-cms/issues/1350


maxime-rainville (changes requested):
Can't remove a public method in a minor release. Add a `@deprecated` notice instead or retarget the PR to version 2.

PR #1125, opened 2020-09-28
FIX Address regression caused by reactstrap upgrade

maxime-rainville:
This PR aims to fix a few regression introduce by upgrading to reactstrap 8. It's not meant to address all the regressions ... just the one preventing merger of https://github.com/silverstripe/silverstripe-asset-admin/pull/1111

This build demonstrate that asset-admin is run against this branch https://travis-ci.org/github/silverstripe/silverstripe-asset-admin/jobs/731138827

There's a dedicated issue for getting a more comprehensive fix for all reacstrap regression https://github.com/silverstripe/silverstripe-admin/issues/1126

# Parent issue
* resolves https://github.com/silverstripe/silverstripe-admin/issues/504


client/src/components/SingleSelectField/SingleSelectField.js
```diff
@@ -98,14 +51,33 @@ class SingleSelectField extends Component {
...
+      options.unshift({
+        value: '',
+        title: this.props.data.emptyString,
+        disabled: false,
+      });
     }
 
-    return field;
+    return (
+      <Input type="select" {...this.getInputProps()}>
```
Cheddam, line 68: @maxime-rainville Is there built-in support for the `readOnly` prop in this Reactstrap component? Or does this change remove the readOnly mode?

maxime-rainville: The `else` statement in `getInputProps` above disables the select field when the field is readonly.

I was getting weird errors with the current split logic, so I figured simplifying the code was a nice side effect of this fix.

Cheddam: Oh gotcha, that was tricky to spot!

Cheddam (changes requested):
Couple of changes to make 👍

client/src/components/SingleSelectField/SingleSelectField.js
```diff
@@ -11,57 +11,6 @@ class SingleSelectField extends Component {
...
-   */
-  getReadonlyField() {
-    let label = this.props.source
-      && this.props.source.find((item) => item.value === this.props.value);
-
-    label = typeof label === 'string'
-      ? label
-      : (this.props.value || '');
-
-    return <Input plaintext {...this.getInputProps()}>{label}</Input>;
```
emteknetnz, line 27: This seems very unnecessary

I've fixed it in a much simpler way to match the existing functionality here https://github.com/silverstripe/silverstripe-admin/pull/1128/files#diff-a5a378b351641dae0c882471702eedbfR27

emteknetnz (changes requested):
I don't understand why this PR is changing functionality when it's supposed to fix regressions.  The functionality changes will need to be reverted.

I've linked a PR in a comment that shows a much simpler way to fix the regression issues with hidden fields

I do very much like the addition of the unit tests to test that the components render without errors, so it would be get those merged in.

emteknetnz:
I'm closing this PR because it has been superseded by https://github.com/silverstripe/silverstripe-admin/pull/1128.  It includes relevant changes that will allow the merging of the toast notification PR once the PR has been merged.

@maxime-rainville I've put you down as a Co-Author in the commit message on the new PR

Issue #1098, opened 2020-08-25
Used on table - display nested DataObjects


emteknetnz:
### Overview
Currently the File Used on table presents duplicate information. If a file appears on a block, the used on table will have a record for the block, the page (and previously the elemental area). This duplicate information results in a rather confusing UI, where some rows contain links and others don't. These duplicates need to be reduced to one record.

To do this, we want to ensure there's enough context still given to what the object is and what the 'parent' object is. E.g for a file block, it would be useful to still see what 'parent' page it relates to.

### ACs
- For each use of a file only one record will show on the Used on table
- For each use of a file, show the object type it is being used on (e.g block, blog post, etc.) and a meaningful 'parent' object (e.g block page, blog, etc.)
- Ensure that files used multiple times within an object are still counted separately
- Each table row should include a link, directing the user to view the file in use (note: this may be further enhanced in: https://github.com/silverstripe/silverstripe-admin/issues/1100)
- There is a clear path to understand how links on the UsedOnTable to UserForm FileSubmissions can be added to the table (to then be tracked in https://github.com/silverstripe/silverstripe-userforms/issues/963)
- There is a clear path (including docs) for Developers to understand how to add their own data types to the table, if not added automatically
- This does not degrade performance from existing functionality where there are at least 20 relationships present in the table **Steve: OR we implement lazy loading on the used on table data**

### Designs
[Edit file / Used on tab](https://projects.invisionapp.com/dsm/silver-stripe/silverstripe-cms/nav/build/asset/componentContainers/5f152068f6772b33db2e52b0/tab/design?mode=preview)

### Early investigation notes
Estimated points: 8

Parent investigation issue https://github.com/silverstripeltd/product-issues/issues/249

Needs testing against kitchen sink to see how well it works in practice, more unit tests, possibly more extension PR's created for various data object types

#### PRs created as part of investigation (don't merge, though reuse bits)
- https://github.com/silverstripe/silverstripe-admin/pull/1097 (closed)

### Grooming notes
- It's likely that we'll run into performance issues due to the number of queries that could be required for each row
 - One idea to mitigate the performance issues could be to not load the table until the tab is selected to view (maybe this already happens?) **Steve: this sometimes happens, sometimes not, needs further investigation**
- There was a note about using pagination to support performance issues too

### PRs
- [x] https://github.com/silverstripe/silverstripe-framework/pull/9735
- [x] https://github.com/silverstripe/silverstripe-admin/pull/1139
[...]

PR #1129, opened 2020-10-02
API Introduce hooks to help components figure out what tab they are in.

maxime-rainville:
This PR refactors our Tab implementation and adds a few hooks to make it easy to figure out what tab a component is on.

The new implementations uses react context to figure out what the parent tab is.

# Related issues/PRs
* resolves https://github.com/silverstripe/silverstripe-admin/issues/1098
* blocks https://github.com/silverstripe/silverstripe-admin/pull/1121


client/src/components/Tabs/Tabs.js
```diff
@@ -1,139 +1,31 @@
...
-    return (
-      <div {...containerProps}>
-        <div className="wrapper">
-          {nav}
-          <TabContent activeTab={activeTab || this.getDefaultActiveKey()}>
+import TabNav from './TabNav';
+import getDefaultActiveKey from './getDefaultActiveKey';
+import { TabContext } from 'hooks/useTabContext';
+
+function Tabs({
```
maxime-rainville, line 11: I refactored this entire thing. In hindsight, I probably didn't need to do that, but it makes it generally better so it's worth keeping.

client/src/components/Tabs/Tabs.js
```diff
@@ -1,139 +1,31 @@
...
+}) {
+  const containerProps = { className: classnames([className, extraClass]), id };
+  const currentTab = activeTab || getDefaultActiveKey(defaultActiveKey, children);
+
+  return (
+    <div {...containerProps}>
+      <div className="wrapper">
+        {!hideNav && <TabNav currentTab={currentTab} onToggle={activateTab}>{children}</TabNav>}
+        <TabContent activeTab={currentTab}>
+          <TabContext.Provider value={{ activeTab: currentTab }}>
```
maxime-rainville, line 22: This context bit is the only real change here.

client/src/components/Tabs/getDefaultActiveKey.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Determines a default tab to be opened and validates the given default tab.
+ * Replaces the given default tab if it is invalid with a valid tab.
+ *
+ * @param defaultActiveKey
+ * @param children
+ * @returns {string}
+ */
+export default function getDefaultActiveKey(defaultActiveKey, children) {
```
maxime-rainville, line 11: Still need to write unit test for this guy.

client/src/components/UsedOnTable/UsedOnTable.js
```diff
@@ -4,6 +4,13 @@ import classnames from 'classnames';
 import i18n from 'i18n';
 import provideUsedOnData from './provideUsedOnData';
 
+import { useTabFirstShow } from 'hooks/useTabContext';
```
maxime-rainville, line 7: All the changes in the file are here for demo purposed. We'll need to revert them once we are done.

client/src/hooks/useTabContext.js
```diff
@@ -0,0 +1,57 @@
...
+    return;
+  }
+
+  const { isOnActiveTab } = tabContext;
+
+
+  const [shownOnce, setShownOnce] = useState(false);
+  useEffect(
+    () => {
+      console.log('trigerring useTabFirstShow');
```
maxime-rainville, line 46:
```suggestion
```

client/src/hooks/useTabContext.js
```diff
@@ -0,0 +1,57 @@
...
+      if (isOnActiveTab) {
+        console.log('trigerring useTabFirstShow callback');
```
maxime-rainville, line 48:
```suggestion
```

maxime-rainville:
@silverstripeux Remember how I said that we needed to remove the fade-in transition on the tabs because of the reactsrap upgrade. This PR might allow us to bring the transition back ... if you want it.

client/src/components/Tabs/Tabs.js
emteknetnz: Should `<TabContext.Provider>` wrap both `<TabNav>` + `<TabContent>`?  I'm very far from an expert, I'm just looking at https://reactjs.org/docs/context.html#updating-context-from-a-nested-component and it seems like `<TabContext.Provider>` should wrap everything and there needs to be a `<TabContext.Consumer>` that at least in the example in the react docs, seems like it broadcasts up.  So the `<TabNav>` (or `<Tab>`?) should be wrapped in  `<TabContext.Consumer>` like the `ThemeTogglerButton` in the docs? - So clicking on a `<Tab>` would update the state object on this `<Tabs>` component?

emteknetnz (commented):
Looking good, using Context seems like the correct approach

Seems like there's a bit more to do here, so I'd recommend marking this PR draft cos it's a bit too soon to start reviewing syntax etc

Have you thought about how this solution would allow us to "only load tab content when the tab is clicked"?

client/src/components/Tabs/Tabs.js
maxime-rainville: No. `<TabNav>` will be told directly what tab is currently selected. Context is useful for components that don't have a direct relation to their parent tab.

sachajudd:
@maxime-rainville not too fussed with the transition as this is the only place that actually used the fade in/out tabs. All the other tabs are not built this way so the pattern would actually be inconsistent across the CMS. I'd like all tabs to use them but for now no harm is caused about leaving this out for now.

maxime-rainville:
@sachajudd It was already showing on all the react tabs before I disabled it when we upgraded to reactstrap 8. It's trivial to bring back, so I'll just do it.

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,86 @@ clickable tab.
...
+    return 
+      <p>
+        The {activeTab} tab is currently active.
+        I'm on the {currentTab} tab which is {isOnActiveTab ? 'visible' : 'invisible'}.
+      </p>;
+  }
+}
+
+// `injectTabContext` is a higher-order-component (HOC). 
+// It can be applied to class and functional components.
```
emteknetnz, line 98: Seems like we'd only want to use it on class components though.  Kind of confusing saying we could use it on functional components, meanwhile we have 2 other functional specifically for functional components

client/src/components/Tabs/TabNav.js
```diff
@@ -0,0 +1,31 @@
...
+    );
+
+  return tabs ?
+    <Nav tabs role="tablist">{tabs}</Nav> :
+    null;
+}
+
+TabNav.propTypes = {
+  currentTab: PropTypes.string,
+  onToggle: PropTypes.func
```
emteknetnz, line 28:
```suggestion
  onToggle: PropTypes.func
  name: PropTypes.name
```
Since using props.name in this file

Should probably make everything .isRequired too

client/src/components/Tabs/Tab.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import { NavItem, NavLink } from 'reactstrap';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+
+/**
+ * Render an individual link for the tabset.
+ */
+function Tab({ title, disabled, active, tabClassName, onToggle }) {
```
emteknetnz, line 9: docblock parameter types

client/src/components/Tabs/TabItem.js
```diff
@@ -1,12 +1,30 @@
...
-const TabItem = ({ name, className, extraClass, disabled, children }) => (
-  <TabPane tabId={name} className={`${className} ${extraClass}`} disabled={disabled}>
-    {children}
-  </TabPane>
-);
+/**
+ * Wraps the content of a tab.
+ * @returns {JSX.Element}
+ */
+const TabItem = ({ name, className, extraClass, disabled, children }) => {
```
emteknetnz, line 10: docblock parameter types

client/src/components/Tabs/TabNav.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import { Nav } from 'reactstrap';
+import PropTypes from 'prop-types';
+import Tab from './Tab';
+
+/**
+ * Builds the tabset navigation links, will hide the links if there is only one child
+ */
+function TabNav({ currentTab, children, onToggle }) {
```
emteknetnz, line 9: docblock parameter types

client/src/components/Tabs/Tabs.js
```diff
@@ -1,139 +1,40 @@
...
 import * as Actions from 'state/tabs/TabsActions';
+import TabNav from './TabNav';
+import getDefaultActiveKey from './getDefaultActiveKey';
+import useTabContext, { TabContext } from 'hooks/useTabContext';
+
+/**
+ * Displays a tab set that cam be used to break up a form into smaller chunks.
+ * @returns {JSX.Element}
+ */
+function Tabs({
```
emteknetnz, line 15: docblock parameter types

emteknetnz (changes requested):

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,86 @@ clickable tab.
  * `tabClassName` (string): Class to use for the tab.
 
  _NOTE:_ For other properties, please refer to the [reactstrap Tabs](https://reactstrap.github.io/components/tabs/) documentation.
+
+# Interacting with Tabs from a child component
+
+Components inside a tab can retrieve information about their parent using these method:
+- `useTabContext` which is a hook returning the tab context for functional components
```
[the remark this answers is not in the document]
maxime-rainville: `useTabContext` is a custom react hook. The react convention is to call hooks `use` something (e.g. `useState`, `useEffect`, `useContext`).

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,86 @@ clickable tab.
...
+- `injectTabContext` which injects the tab context as props in your class component
+- `useTabFirstShow` which is a hook that allows you to execute a callback when a tab is first shown.
```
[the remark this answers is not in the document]
maxime-rainville: Same deal ... it's a hook.

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,86 @@ clickable tab.
...
+export function MyComponent() {
+  // useTabContext only works inside functinal components
+  const tabContext = useTabContext();
+  if (!tabContext) {
+    return <p>I'm not in a tab.</p>;
+  }
+
+  const {activeTab, currentTab, isOnActiveTab} = tabContext;
+  
+  return
```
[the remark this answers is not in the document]
maxime-rainville: Brackets are not necessary here. This should work as is.

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,86 @@ clickable tab.
  * `tabClassName` (string): Class to use for the tab.
 
  _NOTE:_ For other properties, please refer to the [reactstrap Tabs](https://reactstrap.github.io/components/tabs/) documentation.
+
+# Interacting with Tabs from a child component
+
+Components inside a tab can retrieve information about their parent using these method:
```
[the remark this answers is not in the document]
maxime-rainville: `injectTabContext` is an HOC. It could be used with a functional component as well. Not sure there's much point dividing up a list of three items.

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,86 @@ clickable tab.
...
+```
+
+## Example of `injectTabContext`
+
+```jsx
+import React, {Component} from 'react';
+import {injectTabContext} from 'hooks/useTabContext';
+
+class MyComponent extends Component {
+  render({tabContext}) {
```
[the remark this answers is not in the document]
maxime-rainville: `tabContext` would be a prop here. So it would come in as a property that need destructuring.

client/src/components/Tabs/README.md
maxime-rainville: Some people prefer to use component composition. To be honest, the only reason I created `injectTabContext` was because `provideUsedOnData` is a class component. I would have preferred to just use hooks.

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,86 @@ clickable tab.
...
+```jsx
+import React from 'react';
+import {useTabFirstShow} from 'hooks/useTabContext';
+
+export function MyComponent({loadDataAction}) {
+  // `useTabFirstShow` is a Silverstripe specific hook. It will triger the provided function once 
+  // the component is visible. The tabContext is passed back to the provided method.
+  // `useTabFirstShow` can be usefull to delay an expensive request until it's actually needed.
+  // Because `useTabFirstShow` is a hook, it can only be called from a functional component.
+  // Use `injectTabContext` and lifecyle methods to achieve the same result in a class component.
```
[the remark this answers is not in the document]
maxime-rainville: The doc assumes some degree of familiarity react. Don't really want to be giving a react crash course here. We could link off to the main react doc.

client/src/components/Tabs/TabNav.js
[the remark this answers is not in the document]
maxime-rainville: It's a functional component. `this` won't exist here.

client/src/components/Tabs/TabNav.js
maxime-rainville: `name` is only use while looping over the children. It's not a prop of TabNav.

client/src/components/Tabs/Tabs.js
```diff
@@ -1,139 +1,40 @@
...
+  hideNav, children, activeTab, className, extraClass, id, activateTab, defaultActiveKey
+}) {
+  const containerProps = { className: classnames([className, extraClass]), id };
+  const currentTab = activeTab || getDefaultActiveKey(defaultActiveKey, children);
+  const tabContext = useTabContext();
+  const nextTabContext = {
```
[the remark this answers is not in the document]
maxime-rainville: It's common when working with react life cycle methods to refer to the updated props passed to a component as `nextProps` and to the existing props as simply `props`.

client/src/components/Tabs/getDefaultActiveKey.js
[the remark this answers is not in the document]
maxime-rainville: That's what it was called when it was part of Tabs. But since we are blowing the entire thing up any way, makes sense to rethink the name.

client/src/components/Tabs/README.md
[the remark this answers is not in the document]
emteknetnz: I think there is, basically

If you have a functional component then do this:
...

If you have a class component then do this:
...

Doesn't seem like much extra work to make the instructions better?

client/src/hooks/useTabContext.js
```diff
@@ -0,0 +1,49 @@
...
+  const readyToShow = !tabContext || tabContext.isOnActiveTab;
+  const [shownOnce, setShownOnce] = useState(false);
+  useEffect(
+    () => {
+      if (readyToShow) {
+        setShownOnce(true);
+        callback(tabContext);
+      }
+    },
+    [shownOnce || readyToShow]
```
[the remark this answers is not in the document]
maxime-rainville: That's how `useEffect()` works and is meant to be used. It expects a callback and a list of parameters. On every render, react will evaluate if the list of parameters has changed and fire the callback if it has.

The change you suggest would probably break the implementation.

client/src/components/Tabs/README.md
emteknetnz: How about

`// It is used to add context to class components.  It can also be used on functional components.`

client/src/components/Tabs/README.md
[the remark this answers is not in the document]
maxime-rainville: It's not clear to me this makes the doc better. You could argue that `injectTabContext` and `useTabContext` should be together  because they essentially do the same thing and `useTabFirstShow` should be on its own.

At the of of the day, these 3 methods do related things. I think it makes sense to discuss them as a whole.

client/src/components/Tabs/README.md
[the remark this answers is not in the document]
emteknetnz: Tried locally, PHPstorm complains
image

No complaint with brackets

image

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,86 @@ clickable tab.
...
+    return <p>I'm not in a tab.</p>;
+  }
+
+  const {activeTab, currentTab, isOnActiveTab} = tabContext;
+  
+  return 
+    <p>
+        The {activeTab} tab is currently active.
+        I'm on the {currentTab} tab which is {isOnActiveTab ? 'visible' : 'invisible'}.
+    </p>;
```
emteknetnz, line 71:
```suggestion
        I&apos;m on the {currentTab} tab which is {isOnActiveTab ? 'visible' : 'invisible'}.
    </p>; 
```

client/src/components/Tabs/getDefaultActiveKey.js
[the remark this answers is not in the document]
emteknetnz: There's also mentions on it in README.md that'll refer to 'activeKey' need to be quickly updated too

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,88 @@ clickable tab.
...
+```
+
+## Example of `injectTabContext`
+
+```jsx
+import React, {Component} from 'react';
+import {injectTabContext} from 'hooks/useTabContext';
+
+class MyComponent extends Component {
+  render({tabContext}) {
```
emteknetnz, line 83: I'm not sure the `render({tabContext}) {` is correct

It threw errors on my local

However this worked as expected

```
render() {
  this.props.tabContext;
```

maxime-rainville: You're right ... I've been doing too much functional components.

client/src/components/Tabs/getDefaultActiveKey.js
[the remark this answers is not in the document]
maxime-rainville: In second thought, what we are returning here is the unique name of a tab. It is functionally a key.

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,90 @@ clickable tab.
...
+
+## Example of `injectTabContext`
+
+```jsx
+import React, { Component } from 'react';
+import { injectTabContext } from 'hooks/useTabContext';
+
+class MyComponent extends Component {
+  render() {  
+    const {tabContext} = this.props;
```
emteknetnz, line 84:
```suggestion
    const { tabContext } = this.props;
```

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,90 @@ clickable tab.
...
+export default injectTabContext(MyComponent);
+```
+
+## Example of `useTabFirstShow`
+
+```jsx
+import React from 'react';
+import { useTabFirstShow } from 'hooks/useTabContext';
+
+export function MyComponent({loadDataCallback}) {
```
emteknetnz, line 112:
```suggestion
export function MyComponent({ loadDataCallback }) {
```

client/src/components/Tabs/README.md
```diff
@@ -33,3 +33,90 @@ clickable tab.
...
+ 
+
+## Example of `useTabContext`
+
+```jsx
+import React from 'react';
+import useTabContext from 'hooks/useTabContext';
+
+export function MyComponent() {
+  // useTabContext only works inside functinal components
```
emteknetnz, line 59:
```suggestion
  // useTabContext only works inside functional components
```

emteknetnz (dismissed):
Minor updates to docs

emteknetnz:
## Testing a class component

Instructions and functionality for `injectTabContext` are good - I've updated the used on table PR to show [how to integrate](https://github.com/silverstripe/silverstripe-admin/pull/1121/files#diff-3514d28e97c2c38ae74fd1f1e89707d9R30)

## Testing a functional component

Instructions and functionality for `useTabContext` are good, I replaced the used on table with them and it worked as expected

Instructions and functionality for `useTabFirstShow` are also good - I used a nested component for this.  The console.log() was only called once even after I selected the tab multiple times

image

I replaced the contents of `UsedOnTable.js` to test the functional components with the following:

```
import provideUsedOnData from './provideUsedOnData';
import React from 'react';
import useTabContext from 'hooks/useTabContext';
import { useTabFirstShow } from 'hooks/useTabContext';

function consoleLog() {
  console.log('my console log');
}

function MyInnerComponent({ loadDataCallback }) {
  // this should only get called once not matter how many times we click on the tab
  useTabFirstShow(loadDataCallback)
  return <div>My inner component</div>;
}

function UsedOnTable() {
  const tabContext = useTabContext();
  if (!tabContext) {
    return <p>I'm not in a tab.</p>;
  }
  const { activeTab, currentTab, isOnActiveTab } = tabContext;

  return (
    <div>
      <p>
        The {activeTab} tab is currently active.
        I'm on the {currentTab} tab which is {isOnActiveTab ? 'visible' : 'invisible'}.
      </p>
      <MyInnerComponent loadDataCallback={consoleLog} />
    </div>
  );
}

export { UsedOnTable as Component };

export default provideUsedOnData(UsedOnTable);

```

client/src/components/Tabs/TabNav.js
maxime-rainville: Done

emteknetnz:
@maxime-rainville Just needs a rebase to sort merge conflicts.  Happy to MOG after that.

PR #1136, opened 2020-10-13
FIX Add onunmatch() method to avoid _super() call failure

Cheddam:
Entwine's `_super()` method attempts to call the current method on every 'parent' matching block, and fails if a block doesn't contain that method. This particular omission was breaking a `_super()` call from `LeftAndMain.js` (in the [`.loading` handler block](https://github.com/silverstripe/silverstripe-admin/blob/1.6/client/src/legacy/LeftAndMain.js#L1067))

If anyone has further insight into Entwine's nuances and would suggest a different path to resolution, I'd welcome it, as this seems a bit brittle (and I'm surprised we don't hit it more often.)


emteknetnz:
This looks wrong :-/

Can we update entwine instead? We would need to do some sort of "if exists" [here](https://github.com/silverstripe/silverstripe-admin/blob/1/thirdparty/jquery-entwine/src/jquery.entwine.js#L348 ) I think?

dhensby:
> Can we update entwine instead?

Not really - it's not maintained and was pretty much written for SS. We'll have to put up with it's quirks until we can retire it...

Cheddam:
I've briefly tested an adjustment to the [`one()` method](https://github.com/silverstripe/silverstripe-admin/blob/1/thirdparty/jquery-entwine/src/jquery.entwine.js#L206) that skips calling the function if it doesn't exist, and it seems to work fine.

We'll likely need to do some [remedial work](https://github.com/silverstripe/silverstripe-framework/issues/3384) on Entwine before it disappears anyway, so we can't be too afraid to tweak it. I'll get a PR up with the adjustment (applied directly to the copy in admin for now, as I'm not sure we have a method of pulling in upstream changes from the original repo at the moment) and get some broader testing done so we can make sure it doesn't cause any other regressions.

Since that fix will supersede this one, I'm closing this PR for now.

PR #1141, opened 2020-10-19
FIX Patch Entwine to ignore missing methods in one()

### LINKED ISSUE #1138: Missing onunmatch() method triggers exceptions in CMS Edit Forms
**Affected versions:** >=1.6.2, 1.7.x-dev

[A fix for tab routing](https://github.com/silverstripe/silverstripe-admin/pull/1030) introduced a regression in CMS Edit Form contexts, where many actions will trigger an exception due to a missing `unonmatch()` method in [TabSet.js](https://github.com/silverstripe/silverstripe-admin/blob/1.6/client/src/legacy/TabSet.js#L11). This can lead to UI failures, like getting stuck in a loading state after reorganising a GridField with sortable rows.

An initial PR was raised [here](https://github.com/silverstripe/silverstripe-admin/pull/1136) to fix the specific case, but a better long-term fix would be to stop Entwine from trying to run non-existent methods. We've avoided making any adjustments to Entwine for at least 3 years, but the change required would be a tiny conditional statement in the [`one()` method](https://github.com/silverstripe/silverstripe-admin/blob/1/thirdparty/jquery-entwine/src/jquery.entwine.js#L206).

#### PRs

- [x] #1141

Cheddam:
`one()` wasn't checking for the presence of the given method on the instance before trying to call it. This fix should be transparent, and resolves an issue with the changes in #1030.

Fixes #1138.


emteknetnz:
Have run this through silverstripe-installer travis which does cross module behat and it's green

https://travis-ci.org/github/creative-commoners/silverstripe-installer/builds/737204335

Issue #1065, opened 2020-07-01
Date field is cut off in search filter panel


clarkepaul:
**Background**
This issue covers things which were not completed as part of https://github.com/silverstripe/silverstripe-admin/pull/1035.
Date fields have a natural min-width due to the length of the date, and 2-3 field controls depending on the browser the field is being viewed with. We have forced the width to be smaller than the field should be displayed at in the filter panel so that the fields can remain side by side, and reduce the impact. This has caused a separate issue of the dates being cut off out of view.

Date fields are cut off
![image](https://user-images.githubusercontent.com/555033/86297164-b3e7b800-bc4e-11ea-97c4-c4490f7cb0f7.png)

**Recommendation**
Increase the width of the filter panel for all views and stack the fields for mobile viewports.

- increase the width of the panel for desktop users to 375px (bigger viewport than 768px), and for smaller viewports (less than 768px) go 100% width.
- for viewports less than 768px remove the CSS which positions the panel 53px from the right.
- and lastly for viewports less than 375px stack the fields rather than having them inline (change flex-direction: row to column)

PR #1142, opened 2020-10-20
FIX Adjust search form widths to prevent date cutoff

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/1065

**1050px width**

image

**700px width**

image

**350px width**

image


michalkleiner:
Should the 700px width have the fields span the whole width (50/50 with the gutter), similar to the other fields? Or is there a specific reason why they are narrower?

I guess it could work like so: min-width to prevent the value truncation and allow for the date picker, span the full width of the container and stack if the minimal-width fields can't fit.

bergice (approved):
Looks good.

Only suggestion would be to change `max-width: calc(50% - #{$spacer / 4});` when the screen size is very small and make the fields full width as they get cut off:
image

However, they would do that previously anyways, and no one is ever going to use such a small screen size (200~px).

PR #1143, opened 2020-10-21
Fix toggle rendering

bergice:
Chrome has a bug when rendering small circles (`border-radius: 50%`). This change fixes that for the toggle button on the bottom left of the screen.

Before:
image

After:
image


maxime-rainville (changes requested):
Didn't manage to replicate this issue in CMS `4.6.x-dev` with Chrome `87.0.4280.66`. Can you confirm this is still an issue?

bergice:
> Didn't manage to replicate this issue in CMS `4.6.x-dev` with Chrome `87.0.4280.66`. Can you confirm this is still an issue?

Just tested again and I'm pretty sure this is only an issue when zooming the page: 
image

maxime-rainville:
I see it when I zoom-out to 90%. When I zoom-in it looks OK.

The fix does slightly alter the look of the radio button. IF @silverstripeux is happy with it, we can merge.

# Current
image

# With PR
image

sachajudd:
I don't see how changing the width and height would fix the problem. At 90% it's still going to break with this change, and actually look worse IMO.
image

michalkleiner:
It's also weirdly a button with a radio-like indicator — should it rather be a true checkbox since it only has two states toggling a single setting?

maxime-rainville:
@bergice Do you want to have another crack at this or should we close this PR?

PR #1144, opened 2020-10-22
FIX: TreeDropdownField wouldn't render children when using TreeBaseID (fixes #954)

### LINKED ISSUE #954: TreeDropdownField: Fails to load items when TreeBaseID is set
Version: 1.4.2

When using `TreeDropdownField` with `setTreeBaseID` child items fail to load e.g.

```php
{

    $myBaseID = 3;

    $fields->addFieldToTab(
        'Root.Main',
        TreeDropdownField::create(
            'NestedObjectID',
            'Nested Object',
            NestedObject::class
        )
        ->setTreeBaseID($myBaseID)
    );

    return $fields;
}
```

__Steps to reproduce__

1. Expand the "Nested Object" dropdown
2. Show children of an item by clicking the right arrow button

__Expected__
The item expands, showing child items.

__Actual__
An XHR request is send, the response comes back, but the items are not loaded and the field becomes un-selectable.

I've put together this example including Vagrantfile and sspak.
https://github.com/flashbackzoo/ss4-treedropdownfield-baseid-bug

# Pull request
- [x] https://github.com/silverstripe/silverstripe-framework/pull/9743/
- [ ] https://github.com/silverstripe/silverstripe-admin/pull/1144

kinglozzer:
#954 

Fix depends on https://github.com/silverstripe/silverstripe-framework/pull/9743, however the form field continues to function as it does currently with that patch so no need to adjust composer constraints.


maxime-rainville (approved):
Build got confused because the PR was open when we were still using Travis.ORG.

I pushed a fake commit to force the build to run on Travis.COM

Tested locally and it works great.

Just need to remove my fake commit ... Merge on green.

maxime-rainville:
Build is still red, but this is pre-existing. Since we're probably never going to do another 1.6 release anyway, I'm disinclined to spend hours down a behat rabbit hole, so I'll just merge this as-is.

PR #1145, opened 2020-10-28
BUG Hide TabNav when only one Tab is present

maxime-rainville:
When doing work on the tabs, we introduced a bug that shows tab nav for single tab.

The Image Placement form on the Media modal is a good example.


michalkleiner:
We are undoing this on every project to always show the tabs nav even for a single tab. We use nested tabsets to organise fields in the CMS and it may mean that the user loses track of which tab/nested set of fields they're looking at without the tabs nav.

If this could be conditional or configurable, somehow, it would be great.

Cheddam (approved):

Cheddam:
@michalkleiner Out of curiosity, what method are you using to 'undo' this at the moment? It does make sense to me to have this be configurable, but it's a bit out of scope for this fix. Bear in mind too that this is a change to the React component, which is only used in the Files area I believe (the Pages area is still driven by `.ss` templates.)

michalkleiner:
@Cheddam using a hacky way through entwine, targeting tabsets with `->addExtraClass("ss-tabset-always-visible")`

```
/**
 * File: LeftAndMain.EditForm.TabSetAlwaysVisible.js
 */
(function ($) {

    $.entwine('ss', function ($) {

        /**
         * Re-show tabs when only one is available and has been hidden by the standard CMS functionality
         */
        $('.cms-edit-form .ss-tabset-always-visible.ss-tabset-tabshidden').entwine({
            onmatch: function (e) {
                this._super(e);

                this.find("> ul:first").show();
            },
            onunmatch: function(e) {
                this._super(e);
            }
        });
    });

}(jQuery));
```

Cheddam:
@michalkleiner Gotcha. If you'd like to raise a PR to make this configurable in core we'd be happy to take a look! My initial thinking is that it'd be best to introduce a PHP API (e.g. `$tabset->setAlwaysVisible(true)`) and then use that in the template logic and expose it as schema for the React component to use (via `TabSet::getSchemaStateDefaults()`.)

@maxime-rainville This just needs a rebase and then we can proceed with merging 👍

PR #1088, opened 2020-08-07
GraphQL 4 Compatibility 

unclecheese:

PR #1148, opened 2020-11-02
NEW: GraphQL 4 Compatibility

unclecheese:
Moved from: https://github.com/silverstripe/silverstripe-admin/pull/1088


_config/graphql-schema-default.yml
```diff
@@ -0,0 +1,30 @@
+---
+Name: 'graphql-schema-default'
```
chillu, line 2: Why would `silverstripe/admin` need to influence the default schema? You might not even run admin to use GraphQL, right? Shouldn't that either be in the GraphQL module, or in code advising how to set up your own schema?

code/GraphQL/_legacy/EqComparator.php
```diff
@@ -0,0 +1,35 @@
...
+use SilverStripe\GraphQL\TypeCreator;
+
+if (!class_exists(TypeCreator::class)) {
+    return;
+}
+
+/**
+ * Class EqComparator
+ * @package SilverStripe\Admin\GraphQL
+ * This class shims the schema with GraphQL 4 compatable filtering
```
chillu, line 16:
```suggestion
 * This class shims the schema with GraphQL 4 compatable filtering
 *
 * @internal Use GraphQL v4 instead
```

code/GraphQL/_legacy/FieldFormatter.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+
+namespace SilverStripe\Admin\GraphQL;
+
+use SilverStripe\Core\Convert;
+
+class FieldFormatter
```
chillu, line 8:
```suggestion
/**
 * @internal Use GraphQL v4 instead
 */
class FieldFormatter
```

code/GraphQL/_legacy/IDFilterType.php
```diff
@@ -0,0 +1,35 @@
...
+use SilverStripe\GraphQL\TypeCreator;
+
+if (!class_exists(TypeCreator::class)) {
+    return;
+}
+
+/**
+ * Class IDFilterType
+ * @package SilverStripe\Admin\GraphQL
+ * This class shims the schema with GraphQL 4 compatible filtering, e.g. readOne
```
chillu, line 16:
```suggestion
 * This class shims the schema with GraphQL 4 compatible filtering, e.g. readOne
 *
 * @internal Use GraphQL v4 instead
```

composer.json
```diff
@@ -25,7 +25,7 @@
         "php": "^7.1 || ^8"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7",
+        "sminnee/phpunit": "^5.7",
```
chillu, line 28: Was this specifically required for this PR, or just something that's not merged upstream yet as parallel work?

chillu (changes requested):

_config/graphql-schema-default.yml
unclecheese: It's always been this way: https://github.com/silverstripe/silverstripe-admin/blob/1/_config/injector.yml

It's a lot of boilerplate to light up a new schema, so the core recipe ships with a prebaked, unrouted schema. "Activating" the default schema is just routing it.

Interesting that it's tied to admin and not graphql, though. I'm not sure what went into that decision, but this is just legacy.

code/GraphQL/_legacy/FieldFormatter.php
unclecheese: OK, it seemed a bit overkill to me to mark the classes themselves as internal. What we really care about is _are you using them in the schema_, which is why the setters are internal (the only way to use them).

composer.json
unclecheese: Will try removing it, but I was getting a fatal without it.. some kind of incompatibility in Sebastian\Comparator with PHPUnit. I noticed a lot of other core modules are using Sam's fork.

_config/graphql-schema-default.yml
chillu: The file you linked to (`injector.yml`) only configures the `admin` schema, not the `default` schema. I can't find any relevant [mention](https://github.com/silverstripe/silverstripe-admin/search?l=YAML&q=default) of "default" in the admin module config before this PR.

chillu (changes requested):

_config/graphql-schema-default.yml
unclecheese: Ah yup, you're right. This has always been in graphql. Needs to be moved.

composer.json
chillu: Yeah no worries, `recipe-cms` uses it as well.

chillu (approved):

PR #1157, opened 2020-11-16
GraphQL BACKWARD_COMPAT flag fix

chillu:
It assumed the defaults would pull in silverstripe/graphql:3,
but due to the loosened constraints ("3 || 4") it pulls in silverstripe/graphql:4.


chillu:
This build should be fixed once https://github.com/silverstripe/silverstripe-versioned/pull/312/ is merged

PR #1160, opened 2020-12-02
MNT Travis shared config, use sminnee/phpunit

emteknetnz:
sass-lint changes are so that sass-lint doesn't run on vendor/silverstirpe/asset-admin =)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,105 +1,45 @@
-language: php
+version: ~> 1.0
 
-dist: xenial
-
-services:
-  - mysql
-  - postgresql
-  - xvfb
-
-cache:
-  directories:
-    - $HOME/.composer/cache/files
-
-addons:
-  apt:
-    packages:
-      - tidy
-      - chromium-chromedriver
-      - chromium-browser
+import:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -1,105 +1,42 @@
...
-    - COMPOSER_ROOT_VERSION=1.x-dev
-    - SS_BASE_URL="http://localhost:8080/"
-    - SS_ENVIRONMENT_TYPE="dev"
-    - RECIPE_VERSION=4.x-dev
+    - COMPOSER_ROOT_VERSION="1.7.x-dev"
+    - REQUIRE_RECIPE="4.7.x-dev"
 
-matrix:
-  fast_finish: true
+jobs:
```
Cheddam, line 11: Could you add a small explanation about the need for a manual job matrix? I'm assuming this has to do with the use of `silverstripe/installer`, but would be useful for developers looking to adjust in the future.

Cheddam (commented):
Pretty much good to go, just one suggestion.

.travis.yml
emteknetnz: Have added explanation

Issue #1163, opened 2020-12-16
ToastsContainer may call undefined  (1.7)


lekoala:
**Affects version:** `admin 1.7`, `CMS 4.7`

**Steps to recreate:**
1. Error is thrown when loading a modeladmin record directly (I got a "validation error" sometimes for some unknown reason so far)
2. Because of the validation error, it tries to display a toast message

> Uncaught` TypeError: Cannot read property 'dispatch' of null

Triggered from 

> noticeAdd: function(e) {
>       (0,
>       d.default.reducer.store.dispatch)((0,
>       f.display)(e))
>   }

I guess the reducer.store may not be initialized yet due to load order/sequence

Everything is fine (except the validation error but that's on my end i guess) if I load the record from a list because the store is already initialized

**Reverting back to 4.6 :** the error is gone.

lekoala:
After looking a bit into it, it seems that the changes about validation in 1.7 have some side effects

https://github.com/silverstripe/silverstripe-admin/blob/97d561481b6905585c4417212111f5567e5b751d/client/src/legacy/LeftAndMain.EditForm.js#L165

Somehow it's trying to show a validation error when there is none

a validationerror is added to my Form_ItemEditForm (in js, not visible in source code) and that triggers the message

It's added by

https://github.com/silverstripe/silverstripe-admin/blob/97d561481b6905585c4417212111f5567e5b751d/client/src/legacy/LeftAndMain.EditForm.js#L201

Commenting that line, and the problem is gone.

If I look at my tabs, they are all valid, but I have sub tabs on my project, and some of them show a little info icon although they are just fine.

I'm pretty sure it's because of this

https://github.com/silverstripe/silverstripe-admin/blob/97d561481b6905585c4417212111f5567e5b751d/client/src/legacy/LeftAndMain.EditForm.js#L171

It's flagging any tab with alert as being invalid, but actually, I have alert info boxes displayed and they crash the whole thing. It should just crash with alerts from the validation system itself, not all regular alerts

PR #1164, opened 2020-12-16
Fix overzealous alerts

lekoala:
Regular alert should not be reported as invalid tabs

For the whole story, see #1163


kinglozzer:
Thanks for investigating this @lekoala - I also use alerts in the same way for displaying help/info to users and had encountered the same issue!

Do we need to check for other “invalid” class names too? `.error`, `.bad` and `.validation` are ones that I’ve previously seen used though I’m not sure if the form validation logic still sometimes adds these.

lekoala:
I assumed the same as the code above. Alert error makes sense since it's bootstrap classes. I wouldn't check for alert bad or anything like that... Maybe message bad is still used somewhere but i believe it's probably better to push a single convention for that system in order to keep it predictable.

kinglozzer:
@lekoala could you run this through the build process?

lekoala:
> @lekoala could you run this through the build process?

Honestly I haven't setup the build process and I'm not sure how long it would take to properly do it. I just made the pull request to have some progress on that issue. It's probably easier for someone that is actively working on the project to run the build process.

kinglozzer (approved):

kinglozzer:
Ping @Cheddam @emteknetnz, any objections to this?

tldr; using `LiteralField::create('Notice', '<div class="alert info">Info box for CMS user</div>')` results in the tab being highlighted as containing errors after saving, due to the presence of the `.alert` class. This PR changes the code responsible to look for `.alert.error` instead.

PR #1165, opened 2020-12-16
Defensive change

lekoala:
I'm not sure why, but on page load, the reducer.store may not be available. This change ensures the page loads regardless of the reducer. i know it's probably not a good idea to log stuff to the console, but i have no idea how to fix the missing reducer

See #1163

### DIFF
--- a/client/src/legacy/ToastsContainer.js
+++ b/client/src/legacy/ToastsContainer.js
@@ -28,6 +28,11 @@ jQuery.entwine('ss', ($) => {
 ((jquery) => {
   jquery.extend({
     noticeAdd(options) {
+      // If reducer is not available, log to the console
+      if(!Injector.reducer) {
+        console.log(options.type + ': ' + options.text);
+        return;
+      }
       // Manually dispatch a redux display event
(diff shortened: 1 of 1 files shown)


emteknetnz (changes requested):
This isn't really a solution?

If you think that it's simply a timing issue, would it make more sense to wrap this in something like window.setInterval() that periodically checks if Injector.reducer.store exists before running code that references it?

lekoala:
> This isn't really a solution?
> 
> If you think that it's simply a timing issue, would it make more sense to wrap this in something like window.setInterval() that periodically checks if Injector.reducer.store exists before running code that references it?

I agree, it's really ugly :-) setInterval seems even worse in my opinion. The real fix would be to understand why the reducer.store is not available, it probably means that the load order must be wrong somewhere.
Another option is to wrap things to store things in a temporary array if the reducer store is not available and when it is, process the array.

GuySartorelli:
I'm going to close this PR - I don't think we would ever want to just print out to the console like this. If this is still an issue I'm in favour of fixing whatever the actual issue is, but this PR doesn't do that.

Issue #853, opened 2019-03-22
TreeDropdownField hover state text colour broken


michalkleiner:
Assuming text colour on hover over the item's parent should be white for the chevron and the text itself, the text colour doesn't change.

![hover colour issue](https://user-images.githubusercontent.com/233342/54799905-14561e80-4cc4-11e9-9073-94712c9d0780.png)

### ACs
- Renders as expected in the CMS _and_ the pattern library
- The additional points laid out in https://github.com/silverstripe/silverstripe-admin/issues/853#issuecomment-623765986 are addressed

michalkleiner:
Is there a standardised way how SS tests colour contrast? If I do a fix for this, might be good to test it as well.

clarkepaul:
Some designs here for the original development, as for contrast we should have done checks in design when these were created although variables in the CMS might not be exactly matching anymore. I would think using "near to" variables would be acceptable. https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5acd939108f7520011a65e5c

Seems the main fix for this is just white text.

sachajudd:
Note: The field border colour also does not match the other fields, should be #ced5e1.

emteknetnz:
Fixed in https://github.com/silverstripe/silverstripe-admin/pull/1166

PR #1166, opened 2020-12-21
FIX Treedropdown css

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/853


client/src/components/TreeDropdownField/TreeDropdownField.scss
```diff
@@ -96,14 +101,22 @@
   }
 }
 
-.is-focused:not(.is-open)>.Select-control,
-.is-open>.Select-control,
-.Select-menu-outer {
+.Select-arrow-zone {
+  vertical-align: initial;
```
bergice, line 105: This seems to actually just move the arrow up by about 1 pixel instead of vertically centering it.

bergice (changes requested):
We should probably also fix the text colour while we're at it.

client/src/components/TreeDropdownField/TreeDropdownField.scss
emteknetnz: I'm not sure about this? See screenshot below

emteknetnz:
Re the arrow alignment, testing in chrome:

Before:
image

After:
image

client/src/components/TreeDropdownField/TreeDropdownField.scss
bergice: image

This is what it looks like for me.
Left is the current look.
Right is with vertical-align: initial; applied.

Looks to me like we don't need to do anything here.

client/src/components/TreeDropdownField/TreeDropdownField.scss
```diff
@@ -96,14 +101,22 @@
...
+
+.Select-control {
+  border-color: $gray-200;
   box-shadow: none;
+}
+
+.Select.is-focused:not(.is-open) > .Select-control,
+.is-open > .Select-control {
   border-color: $brand-primary;
 }
```
bergice, line 116: image

Need to move box shadow to `.Select.is-focused:not(.is-open) > .Select-control`.

bergice (changes requested):

client/src/components/TreeDropdownField/TreeDropdownField.scss
emteknetnz: Have added `box-shadow: none` so that the box shadows doesn't show

The box shadows comes from react-select - the DSM linked in the parent issue has no box-shadows so this seems correct https://projects.invisionapp.com/dsm/silver-stripe/silver-stripe/asset/components/5acd939108f7520011a65e5c

client/src/components/TreeDropdownField/TreeDropdownField.scss
emteknetnz: Have removed the vertical align

PR #1169, opened 2021-01-06
Fix field type not displaying on userform

bergice:
Resolves https://github.com/silverstripe/silverstripe-userforms/issues/874


maxime-rainville (commented):
Couldn't we fix this in userform specifically? Seems like this could have impacts in other part of the CMS.

bergice:
> Couldn't we fix this in userform specifically? Seems like this could have impacts in other part of the CMS.

We could, but we'd need to add bootstrap to userforms.

maxime-rainville:
You can probably get away with adding a more specific CSS selector to [userforms-cms.scss](https://github.com/silverstripe/silverstripe-userforms/blob/5/client/src/styles/userforms-cms.scss) to override the admin rule.

emteknetnz:
As Max has suggested, recommend that you open a PR on userforms and update https://github.com/silverstripe/silverstripe-userforms/blob/5/client/src/styles/userforms-cms.scss

PR #1170, opened 2021-01-26
Try cleaning up old life cycle methods + look at search component issue

maxime-rainville:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/1185

This PR removes references to deprecated React Lifecycle methods which should clean up our JS console and make it easier to debug actual problems.

It also addresses a semi-related issue in the search component where the Redux-form state would not be cleared after hitting the X button:
- Access a Search component.
- Pick some options in the "advanced filter" form.
- Run your search
- Clear your search by clicking the X button
- Look at the "Advanced Filter" form again

4.7 outcome: The form values still show in the form
Outcome in this patch: The form values have been cleared


client/src/components/Search/Search.js
```diff
@@ -226,6 +218,7 @@ class Search extends Component {
     }
 
     const formSchemaUrl = (props && props.formSchemaUrl) || this.props.formSchemaUrl;
+    console.dir(formSchemaUrl);
```
michalkleiner, line 221: Remove?

client/src/legacy/GridField.js
```diff
@@ -619,6 +620,9 @@ $.entwine('ss', function($) {
         });
       }
 
+      console.clear();
```
michalkleiner, line 623: Debugging leftover?

client/src/components/Search/Search.js
michalkleiner: Ah, sorry, didn't notice it was a Draft PR.

maxime-rainville:
Note for peer reviewers: I honnestly don't understand what some of those bits of logic are meant to do. Some of those usage looked somewhat dubious or like they weren't actually doing anything. I mostly tested by running the front end code and seeing if anything broke.

emteknetnz (changes requested):
There's far too much going on in this pull-request, this one will need to be split into separate PR's.  I've counted the following things going on here that are separate things:

1. Fixing a redux bug on the Search component
2. Globally updating deprecated react lifecyle methods to newer methods
3. Refactoring the SearchForm component from a class to function component
4. Refactoring the logic of the UsedOnTable fetching data, also removing a UsedOnTable unit test

maxime-rainville:
I'm going to push back on the idea that needs to be broken up into several PRs.
- The "redux bug on the Search component" is a side effect of using a dodgy life cycle method that was specifically deprecated because it leads to unpredictable results. It makes sense to refactor this bit of code as part of a move remove usage of the deprecated life cycle method.
- "Refactoring the SearchForm" is a small incremental improvement I did to make it easier to debug the parent Search component.
- The "UsedOnTable fetching data" tweak was necessary to remove the deprecated `componentWillReceiveProps` method. 
- The removed test was being skipped and was calling some of those deprecated methods. It makes sense to remove it .

All these bits of code are meant to do the same thing they use to do. The peer reviewer needs to assess that they indeed are doing the same thing and have some good understanding of how those life cycle method works. Splitting the PR in four is not going to change that and  it would add administrative overhead.

emteknetnz:
It's obviously not impossible, it's just unnecessarily hard and awkward to work with and it introduces risk which doesn't need to be there. 

I get they're all trying to achieve the same outcome which is getting rid of old life cycle methods so I can see how it would make sense at one level to do them at the same time.

However we're refactoring separate components in kind of different ways, so standard practice would be to do them in separate PR's?  I think them having the same underlying motivation for changing them (old lifecycle methods) is somewhat irrelevant here.

maxime-rainville:
I'm not sure I see the problem here. The change in each files are quite compartmentalised and do not impact the other files. You could review each file individually without regards to the other files.

However, you do need bits of specialised knowledge about react lifecycle methods to review each file. It makes sense to spend some time upfront acquiring that knowledge and review all the changes in one go.

If you were upgrading a project to be compatible with PHP8, wouldn't you remove all references to old APIs and submit one comprehensive PR?

emteknetnz:
I'm the one being weird so I'll do the extra work :-)

You can open these pull-requests so they go under your name:

https://github.com/silverstripe/silverstripe-admin/compare/1...creative-commoners:pulls/1/life-cycle-search?expand=1
https://github.com/silverstripe/silverstripe-admin/compare/1...creative-commoners:pulls/1/life-cycle-used-on-table?expand=1
https://github.com/silverstripe/silverstripe-admin/compare/1...creative-commoners:pulls/1/life-cycle-methods?expand=1

emteknetnz:
OK, I'm the one being weird :-)

I'll review as is

client/src/components/UsedOnTable/tests/UsedOnTable-test.js
```diff
@@ -49,20 +49,6 @@ describe('UsedOnTable', () => {
         expect(mockActions.loadUsedOn).toBeCalledWith('abc', 'get', 'http://www.bob.co.nz');
         expect(mockComponent).toBeCalled();
       });
-
-      it.skip('should try to load new data if identifier changes', () => {
```
emteknetnz, line 53: We shouldn't delete this unit test, it should be adapted to the new lifecycle method

client/src/legacy/GridField.js
```diff
@@ -603,6 +603,7 @@ $.entwine('ss', function($) {
     },
 
     close() {
+
```
emteknetnz, line 606: Should delete this extra line

client/src/components/Search/SearchForm.js
```diff
@@ -9,61 +9,55 @@ const searchLabel = i18n._t('Admin.SEARCH', 'Search');
...
-    const { visible, expanded, onSearch, onClear,
-      formSchemaUrl, id, identifier, clearable } = this.props;
+  };
+}
 
-    const loadForm = (visible || expanded);
+/**
+ * Display an advanced filter form.
+ */
+function SearchForm({
```
emteknetnz, line 28: Would prefer not have a refactor from class to function component at the same time as all the other stuff, it's a good thing to do just distracting.  No point changing it back though.

emteknetnz (changes requested):
Looks good, main this is the unit test being deleted

Also needs a rebase

client/src/components/Search/SearchForm.js
maxime-rainville: From memory, I ended up refactoring this because I was trying to debug something in `<Search />`. In the end, it ended up being completely unrelated. Still the move to a functional component is an improvement here.

client/src/components/UsedOnTable/tests/UsedOnTable-test.js
maxime-rainville: This test is currently being skipped altogether. I could try writing an equivalent test, but in practice, I'll mostly just be doubling up the test for `useTabContext`.

client/src/legacy/GridField.js
maxime-rainville: Fixed

client/src/components/UsedOnTable/tests/UsedOnTable-test.js
emteknetnz: OK, I didn't notice it was being skipped.  Happy to keep it deleted.

PR #1172, opened 2021-01-29
BUG Update getManagedModelTabs so it can still work with the old output format of getManagedModels

maxime-rainville:
https://github.com/silverstripe/silverstripe-admin/pull/1072 updated the output format of `getManagedModels`. If your `ModelAdmin` has overridden `getManagedModels`, it might still be returning the old format which will cause `getManagedModelTabs` to fall flat on its face. e.g.: [RegistryAdmin](https://github.com/silverstripe/silverstripe-registry/blob/2/src/RegistryAdmin.php#L30)

I'm calling this a semver breakage.

# Related
- https://github.com/silverstripe/silverstripe-framework/pull/9837

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -416,7 +416,9 @@ abstract class ModelAdmin extends LeftAndMain
             $forms->push(new ArrayData(array(
                 'Title' => $options['title'],
                 'Tab' => $tab,
-                'ClassName' => $options['dataClass'],
+                // `getManagedModels` did not always return a `dataClass` attribute
+                // Legacy behaviour is for `ClassName` to map to `$tab`
+                'ClassName' => isset($options['dataClass']) ? $options['dataClass'] : $tab,
                 'Link' => $this->Link($this->sanitiseClassName($tab)),
                 'LinkOrCurrent' => ($tab == $this->modelTab) ? 'current' : 'link'
             )));
--- a/tests/php/ModelAdminTest.php
+++ b/tests/php/ModelAdminTest.php
@@ -2,6 +2,8 @@
 
 namespace SilverStripe\Admin\Tests;
 
+use SilverStripe\Admin\Tests\ModelAdminTest\Contact;
+use SilverStripe\Admin\Tests\ModelAdminTest\Player;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\Session;
 use SilverStripe\Forms\GridField\GridField;
@@ -10,6 +12,7 @@ use SilverStripe\Forms\GridField\GridFieldImportButton;
 use SilverStripe\Forms\GridField\GridFieldPrintButton;
 use SilverStripe\Security\Permission;
 use SilverStripe\Dev\FunctionalTest;
+use SilverStripe\View\ArrayData;
 
 class ModelAdminTest extends FunctionalTest
 {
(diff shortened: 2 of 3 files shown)


tests/php/ModelAdminTest/ContactAdmin.php
```diff
@@ -21,4 +21,9 @@ public function Link($action = null)
         }
         return Controller::join_links('ContactAdmin', $action, '/');
     }
+
+    public function getManagedModelTabs()
+    {
+        return parent::getManagedModelTabs();
+    }
```
maxime-rainville, line 28:
```suggestion
```

This bit is not needed.

maxime-rainville:
@sminnee Can you double check that I'm still honouring the original of your  https://github.com/silverstripe/silverstripe-admin/pull/1072 PR?

maxime-rainville:
I feel the [ModelAdmin doc](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/modeladmin/#modeladmin) should have been updated to mention the new format for `private static $managed_models`. I'll do a doc PR to go along with this.

maxime-rainville:
Here's the matching DOC PR https://github.com/silverstripe/silverstripe-framework/pull/9837

tests/php/ModelAdminTest.php
```diff
@@ -126,4 +129,84 @@ public function testGetGridField()
...
+
+        $models = $admin->getManagedModels();
+
+        $this->assertEquals(
+            [
+                'dataClass' => Contact::class,
+                'title' => 'Contacts'
+            ],
+            $models[Contact::class],
+            'Managed Model that only provide a string get nomalised'
```
emteknetnz, line 145:
```suggestion
            'Managed Model that are class name only get nomalised'
```

tests/php/ModelAdminTest.php
```diff
@@ -126,4 +129,84 @@ public function testGetGridField()
...
+            'Managed Model with an explicit dataClass can have a custom URL key'
+        );
+
+        $this->assertEquals(
+            [
+                'dataClass' => Player::class,
+                'title' => 'Rugby Players'
+             ],
+            $models[Player::class],
+            'Managed Model default to using the array key when a dataClass is not provided'
```
emteknetnz, line 163:
```suggestion
            'Managed Model without a dataClass provided default to using the class name for dataClass
```

tests/php/ModelAdminTest.php
```diff
@@ -126,4 +129,84 @@ public function testGetGridField()
...
+        );
+    }
+
+    public function testGetManagedModelTabs()
+    {
+        $mock = $this->getMockBuilder(ModelAdminTest\MultiModelAdmin::class)
```
emteknetnz, line 169: This seems wrong to me that we're using a mock for this given there's a test class explicitly for this which we can easily instantiate.  It seems like the only reason that it's used is to be able to access the protected method `getManagedModelTabs()`.

I particularly don't like that we're coercing the return value of `getManagedModels()`

Why not just 
```
$admin = new ModelAdminTest\MultiModelAdmin();
$tabs = $admin->getManagedModelTabs()->toNestedArray();
```

?

tests/php/ModelAdminTest/MultiModelAdmin.php
```diff
@@ -0,0 +1,36 @@
...
+
+    public function Link($action = null)
+    {
+        if (!$action) {
+            $action = $this->sanitiseClassName($this->modelClass);
+        }
+        return Controller::join_links('ContactAdmin', $action, '/');
+    }
+
+    public function getManagedModelTabs()
```
emteknetnz, line 32:
```suggestion
    // The purpose of this method is to increase the visibility of ModelAdmin::getManagedModelTabs()
    // from protected to public
    public function getManagedModelTabs()
```

code/ModelAdmin.php
```diff
@@ -416,7 +416,7 @@ protected function getManagedModelTabs()
             $forms->push(new ArrayData(array(
                 'Title' => $options['title'],
                 'Tab' => $tab,
-                'ClassName' => $options['dataClass'],
+                'ClassName' => isset($options['dataClass']) ? $options['dataClass'] : $tab,
```
emteknetnz, line 419: I agree with this solution

I do think there needs to at least be an inline comment explaining that this very unintuitive code (ClassName = $tab) is done for backwards compatibility, otherwise future developer will look at this with great confusion

emteknetnz (changes requested):
Agree with solution, some fairly small changes

tests/php/ModelAdminTest.php
maxime-rainville: `getManagedModelTabs` calls `getManagedModels` internally. The original PR that introduced the bug updated `getManagedModels` so it always uses the `'Player' => ['dataClass' => Player::class, 'title' => 'Ice Hockey Players']` format.

I need to make sure `getManagedModelTabs` doesn't fall over when `getManagedModels` returns the legacy format. The only way I could do that is:
- mock the function like I'm doing right here
- create yet another ModelAdmin test class and override `getManagedModels`

I think the mock function has the advantage of showing exactly the data input we are testing next to the assertion.

code/ModelAdmin.php
maxime-rainville, line 419:
```suggestion
                // `getManagedModels` did not always return a `dataClass` attribute
                // Legacy behaviour is for `ClassName` to map to `$tab`
                'ClassName' => isset($options['dataClass']) ? $options['dataClass'] : $tab,
```

maxime-rainville: Added a comment

tests/php/ModelAdminTest.php
emteknetnz: OK, in that case I'd recommend adding an inline comment explain the mock is explicitly coercing the return value of `getManagedModels()` to match the legacy format

May as well also remove `->expects($this->once())` since the purpose of the mock is not for assertion, it's specially for data generation

emteknetnz (changes requested):
Travis is red

Please add an inline comment explaining the use of mock, and may as well remove `->expects($this->once())`

tests/php/ModelAdminTest.php
maxime-rainville: Added a comment. The `once` call is already there.

emteknetnz (approved):

emteknetnz:
Have merged up

Issue #1156, opened 2020-11-13
Popover (for Elemental) too narrow (1.7 regression) 


purplespider:
The popover used for displaying available Elemental blocks appears to have become too narrow in 1.7 causing block names to become truncated.

**1.6**
<img width="497" alt="Screenshot 2020-11-13 at 15 21 49@2x" src="https://user-images.githubusercontent.com/329880/99089687-8a0fac00-25c5-11eb-8a86-fd7d1cb17160.png">

**1.7**
<img width="433" alt="Screenshot 2020-11-13 at 15 22 35@2x" src="https://user-images.githubusercontent.com/329880/99089698-8c720600-25c5-11eb-882f-653d3c62fbf9.png">

Also, note the more obvious cut off top corners.

As per https://github.com/dnadesign/silverstripe-elemental/issues/848 I believe this is related to work done for the Reactstrap 8 upgrade. But unsure if resolving it is just a cause of increasing the new min-width: https://github.com/silverstripe/silverstripe-admin/blob/5d179b734ed1f97a0f3c853e29539e971c6988f5/client/src/components/PopoverOptionSet/PopoverOptionSet.scss#L15

## Acceptance Criteria
- This has been tested across a range of device sizes

kinglozzer:
Looks to me like the `max-width` we set [here](https://github.com/silverstripe/silverstripe-admin/blob/5d179b734ed1f97a0f3c853e29539e971c6988f5/client/src/components/PopoverOptionSet/PopoverOptionSet.scss#L5) and [here](https://github.com/silverstripe/silverstripe-admin/blob/5d179b734ed1f97a0f3c853e29539e971c6988f5/client/src/components/PopoverOptionSet/PopoverOptionSet.scss#L11) need to also be copied into the nested `.popover` selector. Currently the `max-width` is less than the `min-width`!

maxime-rainville:
Fixed by https://github.com/silverstripe/silverstripe-admin/pull/1174

maxime-rainville:
Looks like the fix has broken the popover even worst https://github.com/silverstripe/silverstripe-admin/pull/1174#issuecomment-779564409

Marking as critical because it looks really bad now

thats4shaw:
@maxime-rainville I'm using Chrome Version 88.0.4324.146.

It seems from my initial investigation that it gets worse the more items you have.

maxime-rainville:
[...]

PR #1174, opened 2021-02-03
FIX Set min-width of popover

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/1156


kinglozzer:
This does make the dropdown slightly wider, but I’m not sure it’s the best fix as it’s still much narrower than it was before the bug was introduced.

`.popover` still has a `max-width: 276px;` rule (from Reactstrap I think) so we’re in the slightly muddy waters of setting a `min-width` and `max-width` that conflict. Would we be better served by just updating the `max-width` instead?

```scss
  > .popover {
    min-width: 280px;
    max-width: $column-width * 2;
  }
```

maxime-rainville (changes requested):
What @kinglozzer said.

Looking at 1.6, these are the `max-width` at different breakpoints:
- 280px by default
- 450px with for displays 544px or wider

emteknetnz:
I've updated the CSS

maxime-rainville (approved):
It doesn't quite look the same way it did in 1.6, but it meets my smelly developer standards.

I tested in Firefox, Chrome and IE11. I did notice that the border radius appears more pronounce in 1.7 than in 1.6.

If @clarkepaul approves, I'm fine with merging.

This video shows what it looks like: https://youtu.be/19Zvsanr7Fw
- Right tab is plain 1.6
- Left tab is 1.7 with the PR

clarkepaul:
I'm fine with that cheers, happy for a merge from my side. FYI in the original designs we didn't specify widths but we designed it at 320px wide to accommodate mobile devices. I'm happy that it's ended up being slightly wider for desktop as people seem to favour long block names.
[Design reference](https://projects.invisionapp.com/share/N5MVKR392DS#/screens/318343221_Add_Block)

maxime-rainville (approved):
Everyone is happy

thats4shaw:
The addition of a max width of 100% seems to cause something like this:
Screen Shot 2021-02-16 at 4 48 17 PM

```
dnadesign/silverstripe-elemental 4.5.0
silverstripe/admin 1.7.2 
```

maxime-rainville:
@thats4shaw I've reopen the parent issue. Can you let us know what browser you are using?
https://github.com/silverstripe/silverstripe-admin/issues/1156

Issue #1122, opened 2020-09-24
Fix icon regression + behat test for tab validation icons


emteknetnz:
We intentionally merged https://github.com/silverstripe/silverstripe-admin/pull/1111 without adding any new behat tests because we need a [custom page type](https://github.com/silverstripe/silverstripe-admin/pull/1111#issuecomment-694613669) though we cannot use `implements TestOnly` because that only works for phpunit tests.

The current mechanism for adding custom dataobjects for behat is not a paritcularly good one, it is include [frameworktest](https://github.com/silverstripe/silverstripe-frameworktest) - so perhaps we need to include this as a dev dependency in admin - although that's not preferred.

**Update:** The aim is to put forward a suggestion to add the test fixture to a new module, see: https://github.com/silverstripe/silverstripe-framework/issues/9712

## ACs
- Add the custom dataobject referenced below to the new module (https://github.com/silverstripe/silverstripe-framework/issues/9712
)
- Finalise the behat test (referenced in this issue)
- Ensure that the new module is referenced in the travis file for silverstripe-admin

## Notes
* There is a [behat test written](https://github.com/silverstripe/silverstripe-admin/pull/1111#issuecomment-694643727) (though will need additional work) that can be used

emteknetnz:
Update: the "new module for behat text fixtures" [had cold water poured on it](https://github.com/silverstripe/silverstripe-framework/issues/9712#issuecomment-700335955), and there's no clear way forward

For this reason I'm just going to keep the status quo and add the new fixtures to frameworktest and release a new version of that

Cheddam:
Fixed in #1179.

PR #1179, opened 2021-02-09
FIX Reinstate tab validation icons

emteknetnz:
Tab validation icons were accidentally removed in https://github.com/silverstripe/silverstripe-admin/pull/1164

This reinstates them, as well as adding behat tests which is for https://github.com/silverstripe/silverstripe-admin/issues/1122


tests/behat/src/AdminContext.php
```diff
@@ -0,0 +1,78 @@
...
+            assertNotNull($element, $message);
+            assertTrue(strpos($element->getText(), $text) !== false, $message);
+        }
+    }
+
+    /**
+     * @When /^I click the "([^"]+)" element$/
+     * @param $selector
+     */
+    public function iClickTheElement($selector)
```
Cheddam, line 71: I'm surprised this doesn't already exist? Surely other behat tests have clicked non-button elements before?

tests/behat/src/AdminContext.php
```diff
@@ -0,0 +1,78 @@
+<?php
+
+namespace SilverStripe\Admin\Tests\Behat\Context;
+
+use Behat\Behat\Context\Context;
+use Behat\Mink\Element\DocumentElement;
+use SilverStripe\BehatExtension\Context\MainContextAwareTrait;
+
+/**
+ * Context used to create fixtures in the SilverStripe ORM.
```
Cheddam, line 10: Is this accurate? 🤔

tests/behat/src/AdminContext.php
emteknetnz: It does exist in 2 or 3 other contexts, however it's not at a "high level" e.g. in behat-extension.  It's less than ideal though pragmatically it really doesn't matter

tests/behat/src/AdminContext.php
emteknetnz: no :p

emteknetnz: Have removed

Cheddam (approved):
MOG

PR #1180, opened 2021-02-14
BUG Allow external link anchor field to work even if the user provides a hash mark

maxime-rainville:
This PR ensures that's at most one `#` before the anchor in a WYSIWYG link.

# Parent issue
- Fixes https://github.com/silverstripe/silverstripe-cms/issues/2634


maxime-rainville:
@clarkepaul has mentioned on the parent issue that he's happy with the fix UX wise.

client/src/legacy/TinyMCE_sslink.js
```diff
@@ -131,14 +131,24 @@ jQuery.entwine('ss', ($) => {
       return Promise.resolve();
     },
 
-    buildAttributes(data) {
-      const anchor = data.Anchor && data.Anchor.length ? `#${data.Anchor}` : '';
-      const href = `${data.Link}${anchor}`;
+    /**
+     *
+     * @param {string|undefined} Anchor
```
emteknetnz, line 136: Looks like this docblock for deconstructed object may not be correct? https://jsdoc.app/tags-param.html#parameters-with-properties

client/src/legacy/TinyMCE_sslink.js
```diff
@@ -131,14 +131,24 @@ jQuery.entwine('ss', ($) => {
...
+     * @param {string} Link
+     * @param {boolean} TargetBlank
+     * @param {string} Description
+     * @returns {{href: string, title: string, target: string}}
+     */
+    buildAttributes({ Anchor, Link, TargetBlank, Description }) {
```
emteknetnz, line 142: Minor thing, though I'd make the argument order { Link, Description, Anchor, TargetBlank } so that it matches the order of the form elements on the frontend

client/src/legacy/TinyMCE_sslink.js
```diff
@@ -131,14 +131,24 @@ jQuery.entwine('ss', ($) => {
...
+      let anchor = Anchor && Anchor.length ? `#${Anchor}` : '';
+      // Make sure, there's no more than one # character at the start of the anchor
+      anchor = anchor.replace(/^\#+/, '#');
```
emteknetnz, line 145:
```suggestion
      anchor = anchor.replace(/^#+/, '#');
```
Don't need to escape the hash character

emteknetnz (changes requested):
Good solution to work with existing data, tested that it works as expected

Some minor stylistic feedback

client/src/legacy/TinyMCE_sslink.js
maxime-rainville: That's what PHPStorm auto-suggested. My guess is that's what it understands. I'm more concerned with keeping PHPStorm happy than JSDoc.

client/src/legacy/TinyMCE_sslink.js
maxime-rainville: Done

emteknetnz:
Merged up

maxime-rainville:
@emteknetnz You should formally approve PRs before you merge them ... otherwise an auditor might look at the PR and think we merged things without reviewing them first.

PR #1181, opened 2021-02-17
FIX Correct sizing of PopoverOptionSet, restore dismissal on blur

### LINKED ISSUE #1156: Popover (for Elemental) too narrow (1.7 regression)

Cheddam:
Fixes #1156.
Fixes dnadesign/silverstripe-elemental#876.


kinglozzer:
This results in a min-width that’s larger than max-width again. To be honest, I’m not sure why we’re messing around with min/max widths at different breakpoints here - it looks to me like a much simpler problem than the existing CSS is making it.

`.popover` has a default `max-width: 276px` we need to override, but as far as I can see that’s pretty-much all we need to do:

```scss
$button-height: 40px;

.popover-option-set {
  display: flex;
  flex-direction: column;
  border-radius: $border-radius; // Change to equal 4px

  > .popover {
    width: calc(100% - 10px); // So the popover doesn't touch the edges of the viewport
    max-width: 450px;
  }
}
```

What do you think?

emteknetnz (approved):
I'd agree that it's backwards that we're essentially using `min-width` as `width`

On the other hand it seems like this PR works and it fixes a regression that was released on a non-stable tag, so, I'm keen to get this fixed.  The fix basically just reverts things to how the behaved previously, so this is fine IMO.

I've tested this in a few  browsers including IE11 and it looks fine

I'll approve though not merge @Cheddam if you're feeling inspired enough to change the usage of min-width/max-width to something a little more normal then I'm happy to test again, otherwise I'm happy for you to click the merge button yourself with the code as is

emteknetnz:
@Cheddam somethings gone weird with it, it's lost the rounded corners at the top that was previously fixed in https://github.com/silverstripe/silverstripe-admin/pull/1176

image

Cheddam:
@kinglozzer Thanks for weighing in!

I've adjusted the PR to reduce the number of properties used, in particular to remove conflicting min/max rules. However, I opted to keep the breakpoint in place, as removing it allowed the container to expand in width before the two-column mode kicked in, making that in-between stage look too wide.

I briefly removed the `overflow: hidden` property, not understanding its purpose, but have reinstated it as this was causing the border-radius issue @emteknetnz noted. The alternative solution to this issue would be to apply border-radius properties to the search / button container elements, but this feels more complex and brittle than simply retaining the overflow-based solution.

emteknetnz (approved):
Have tested latest update, including in IE11, works good

kinglozzer (approved):
Also tested, can confirm it fixes both bugs 👍

kinglozzer:
Thanks @Cheddam!

PR #1182, opened 2021-02-17
BUGFIX: Replace GraphQL3 fragment urls, rely on legacy document state

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/360


client/src/boot/apollo/getGraphqlFragments.js
```diff
@@ -20,8 +20,23 @@ const handleError = response => {
...
+      `${baseUrl}graphql/types`,
+    ]
+    : [
+      `${baseUrl}admin.types.graphql`,
+    ];
+
+  if (!preferStatic) {
+    urls.reverse();
+  }
+  const [primaryURL, fallbackURL] = urls;
```
emteknetnz, line 39: Won't fallback URL be undefined for graphql4?  Seems a bit weird that we'll `response = await fetchSchema(fallbackURL);` below

emteknetnz (approved):
Seems fine, one comment, though happy for you to self-merge if you think things are fine as they are

PR #1187, opened 2021-03-04
POC Add a trait to bootstrap ViewableData into react component

maxime-rainville:
This Proof of Concept showcase how we could make it easier to for developers to create PHP Objects that maps to react component.

To illustrate this we've created a SiteName ViewableData class to replace the current Site name in the top left corner of the CMS.

### DIFF
--- a/client/src/boot/registerComponents.js
+++ b/client/src/boot/registerComponents.js
@@ -1,6 +1,7 @@
 import { Field as ReduxFormField } from 'redux-form';
 import Injector from 'lib/Injector';
 import ActionMenu from 'components/ActionMenu/ActionMenu';
+import CmsSiteName from 'components/CmsSiteName/CmsSiteName';
 import Badge from 'components/Badge/Badge';
 import Button from 'components/Button/Button';
 import BackButton from 'components/Button/BackButton';
@@ -100,5 +101,6 @@ export default () => {
     NumberField,
     PopoverOptionSet,
     ToastsContainer,
+    CmsSiteName
   });
 };
--- a/client/src/bundles/bundle.js
+++ b/client/src/bundles/bundle.js
@@ -100,6 +100,7 @@ require('../legacy/AddToCampaignForm');
 require('../legacy/SecurityAdmin');
 require('../legacy/ModelAdmin');
 require('../legacy/ToastsContainer');
+require('../legacy/BootstrapComponent');
 
 // Legacy form fields
 // Fields used by core legacy UIs, or available to users
--- a/client/src/components/CmsSiteName/CmsSiteName.js
+++ b/client/src/components/CmsSiteName/CmsSiteName.js
@@ -1,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const CmsSiteName = ({ title, baseHref }) => (
+  <div className="cms-sitename">
+    <a href="#" className="cms-sitename__link font-icon-silverstripe font-icon-large" target="_blank">
+    </a>
+    <a className="cms-sitename__title" href={baseHref} target="_blank">{title}</a>
+  </div>
+);
+
+CmsSiteName.propTypes = {
+  title: PropTypes.string,
+  baseHref: PropTypes.string,
+};
+
+CmsSiteName.defaultProps = {
+};
+
+export default CmsSiteName;
--- a/client/src/legacy/BootstrapComponent.js
+++ b/client/src/legacy/BootstrapComponent.js
@@ -1,0 +1,48 @@
+/* global ss */
+import jQuery from 'jquery';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { loadComponent } from 'lib/Injector';
+
+jQuery.entwine('ss', ($) => {
+  $('.js-injector-boot .bootstrap-component').entwine({
+
+    Component: null,
+
+    onmatch() {
+      const cmsContent = this.closest('.cms-content').attr('id');
+      const context = (cmsContent)
+        ? { context: cmsContent }
+        : {};
+
+      const componentName = this.data('component');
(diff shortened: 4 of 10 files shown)


client/src/boot/registerComponents.js
```diff
@@ -100,5 +101,6 @@ export default () => {
     NumberField,
     PopoverOptionSet,
     ToastsContainer,
+    CmsSiteName
```
maxime-rainville, line 104: This bits register our demo component in injector. Any new custom component would have to go through this step.

I nice thing here is that a developer could choose to override our component with their own custom implementation.

client/src/bundles/bundle.js
```diff
@@ -100,6 +100,7 @@ require('../legacy/AddToCampaignForm');
 require('../legacy/SecurityAdmin');
 require('../legacy/ModelAdmin');
 require('../legacy/ToastsContainer');
+require('../legacy/BootstrapComponent');
```
maxime-rainville, line 103: We're registering tho logic to bootstrap the react components. This would only happen once in core.

client/src/components/CmsSiteName/CmsSiteName.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+const CmsSiteName = ({ title, baseHref }) => (
```
maxime-rainville, line 5: This replicates the existing SS templates but in JSX

client/src/legacy/BootstrapComponent.js
```diff
@@ -0,0 +1,48 @@
+/* global ss */
+import jQuery from 'jquery';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { loadComponent } from 'lib/Injector';
+
+jQuery.entwine('ss', ($) => {
```
maxime-rainville, line 7: This bit uses entwine to wire our react component. All Component based off the BootstrapComponent trait would use this generic bootstrapping logic.

code/React/BootstrapComponent.php
```diff
@@ -0,0 +1,152 @@
+<?php
+
+namespace SilverStripe\Admin\React;
+
+use SilverStripe\Core\Convert;
+
+trait BootstrapComponent
```
maxime-rainville, line 7: This trait contains most of the logic that needs to be added to a ViewableData object to be rendered as a React component.

maxime-rainville:
We would probably need a slightly more specialised version of BootstrapComponent for React Form Fields, but this shows the gist of what it would look like.

client/src/legacy/BootstrapComponent.js
blueo: looks good - quite similar to our implementation here: https://github.com/silverstripeltd/tourism1/blob/master/app/javascript/src/entwine/reactFieldBootstrapper.js we've got some extra handling for having an input field that the react component updates

code/React/BootstrapComponent.php
```diff
@@ -0,0 +1,152 @@
...
+        'AttributesHTML' => 'HTMLFragment'
+    ];
+
+    private $attributes = [];
+
+    protected $extraClasses = [];
+
+    public function forTemplate()
+    {
+        $return = $this->renderWith($this->getTemplates());
```
blueo, line 20: might be nice to throw somewhere if we aren't on viewable data, just in case someone expects that to work

code/React/SiteName.php
```diff
@@ -0,0 +1,27 @@
...
+
+use SilverStripe\Admin\LeftAndMain;
+use SilverStripe\Control\Director;
+use SilverStripe\SiteConfig\SiteConfig;
+use SilverStripe\View\ViewableData;
+
+
+class SiteName extends ViewableData
+{
+    use BootstrapComponent;
```
blueo, line 12: looks nice and clean :)

code/React/BootstrapComponent.php
unclecheese: The trait is coupled to `ViewableData`, which to me implies that this should be a subclass, not a trait. What are the tradeoffs for asking devs to subclass `ReactComponent` or something?

unclecheese (commented):
This is really cool. Reduces a lot of the toil around React and will definitely invite more uptake. I just have an architectural issue with the trait that could use some discussion, I think.

blueo:
> This is really cool. Reduces a lot of the toil around React and will definitely invite more uptake. I just have an architectural issue with the trait that could use some discussion, I think.

I'm of two minds about the trait. In support of a trait - i can be applied to existing fields which is great for extending functionality - eg the treedropdown field or multivalue.

unclecheese:
Yeah, fair point. `FormField` extends `RequestHandler` 🙄  ... maybe more composition isn't such a bad idea.

maxime-rainville:
Yeah ... I think the main value of a trait is to be able to retroactively add it to existing classes. My thinking is to create a ReactComponent interface that dev could choose to apply to classes that should be rendered with React. The trait would mostly be a sensible implementation of that interface. 

It's also worth mentioning that technically the only explicit coupling to ViewableData is the call to `renderWith()`. All the Attributes and ExtraClass logic snippets was actually copied over from `Form` and gets duplicated across a number of other classes. I'm thinking it might actually make sense to retroactively extract those bits of code to traits and remove them form `Form` and the other classes in which they appear.

Another side question could be do we want to use the trait approach for form fields as well? 

I was thinking most of those form field already have a react implementations ... we could actually put that "render in react" logic directly in `FormField` and enable it with a config flag. Suddenly, most of our CMS UI could be rendered in react by the flick of a switch. I'm not actually sure if this a good idea or if it would provide any value. 

Even if we decided not to flick that switch on all the time for all the fields, any React based form field would just need to extend `FormField` and have `private static $render_in_react = true` statement.

PR #1195, opened 2021-04-27
Fix asset-admin Editor ParentID/Location dropdown no longer showing any other options

bergice:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/375

Fix asset-admin Editor ParentID/Location dropdown no longer showing any other options.

https://github.com/silverstripe/silverstripe-admin/commit/095b7737a8cfa3ff9fc250264282dad514fd7475#diff-962b5b46986393fbd25451f1639d6017065eff84dd0e9ddf7e3673022b172c1bR700-R727 is causing the dropdown to no longer re-render when opened. This means we now need to include all the options when the dropdown is not opened.


emteknetnz:
There's an issue with the same folder showing twice

image

I created a folder 'My-folder'

I uploaded a file

I refreshed

I clicked on the drop-down to move the file

Tested in firefox

Could you also add a unit test that ensures the number of options in the dropdown is what we expected to see

PR #1196, opened 2021-04-29
Fix TreeDropDownField no longer displaying dropdown options

bergice:
Resolves https://github.com/silverstripe/silverstripe-graphql/issues/375

Revert change in 095b7737a8cfa3ff9fc250264282dad514fd7475 that broke TreeDropDownField dropdown options.

This method should no longer be necessary anyways once we upgrade to react-select >= 2.0


emteknetnz (changes requested):
I'm not sure that going back to using a deprecated lifecycle method is the correct way to fix this.  That method it currently needs the UNSAFE_ prefix to work in react 17+ https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

bergice:
> I'm not sure that going back to using a deprecated lifecycle method is the correct way to fix this. That method it currently needs the UNSAFE_ prefix to work in react 17+ https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

Perhaps upgrading react-select could be done at the same time we upgrade to react 17 in that case.

maxime-rainville (approved):
As much as it pains me to say this, I'm going to agree that reverting this specific change is the best way to address this issue.

Had a quick look to see if we could upgrade to react-select 2. This is definitely not trivial and would require us to potentially update the look and feel of the component.

I'll raise a separate issue to refactor this component ... if there isn't one already.

maxime-rainville:
There's already an issue to upgrade us to react-select v2 https://github.com/silverstripe/silverstripe-framework/issues/9188

I think this issue was raise specifically about that `// Hack! Temporary fix until we can do a proper upgrade of react-select to >= 2.0.` comment.

emteknetnz (approved):
Behat failures are [unrelated](https://github.com/silverstripe/silverstripe-admin/issues/1198)

PR #1208, opened 2021-05-13
FIX ModelAdmin importer use with custom routes

chillu:
Regression from when we introduced $modelTab.
This is backwards compatible since $modelTab falls back to $modelClass.

Also improved $model_importers to fall back to the default per model,
rather than requiring definition of *all* models on a non-null value.


chillu:
Superseded by https://github.com/silverstripe/silverstripe-admin/pull/1216

PR #1209, opened 2021-05-13
FIX ModelAdmin::$model_importers alias support

chillu:
Hasn't been considered when introducing aliases rather than class names
in $managed_models.


chillu:
Replaced by https://github.com/silverstripe/silverstripe-admin/pull/1216

Issue #1213, opened 2021-05-17
Broken builds - Add graphql3 behat tests


emteknetnz:
GraphQL 3 builds are not actually running on admin - the existing behat tests are only running graphql 4

There's also an existing graphql 4 build failure, though this issue won't fix that https://travis-ci.com/github/silverstripe/silverstripe-admin/jobs/505604067

### PRs
- [x] https://github.com/silverstripe/silverstripe-admin/pull/1214

PR #1214, opened 2021-05-17
MNT Ensure graphql 3 versions of behat tests run

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1213

Admin is a bit of strange one that runs both 'admin' and 'cms' behat suites.  It's debatable whether things should be this way, though for now I think we should keep it as is.

This PR will ensure that both behat suites have both graphql 3 and graphql 4 jobs run

The new graphql 3 behat tests are passing, though the graphql 4 build has an existing failure

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,17 +19,33 @@ jobs:
         - REQUIRE_INSTALLER="$REQUIRE_RECIPE"
         - NPM_TEST=1
         - RELEASE=1
-    - php: 7.2
+    - php: 7.3
       env:
         - DB=MYSQL
         - REQUIRE_INSTALLER="$REQUIRE_RECIPE"
         - BEHAT_TEST=1
+        - REQUIRE_GRAPHQL="^3@dev"
         - BEHAT_SUITE="admin"
     - php: 7.3
       env:
         - DB=MYSQL
         - REQUIRE_INSTALLER="$REQUIRE_RECIPE"
         - BEHAT_TEST=1
+        - REQUIRE_GRAPHQL="^3@dev"
+        - BEHAT_SUITE="cms"
+    - php: 7.3
+      env:
+        - DB=MYSQL
+        - REQUIRE_INSTALLER="$REQUIRE_RECIPE"
+        - BEHAT_TEST=1
+        - REQUIRE_GRAPHQL="^4@dev"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -19,17 +19,33 @@ jobs:
...
+        - REQUIRE_INSTALLER="$REQUIRE_RECIPE"
+        - BEHAT_TEST=1
+        - REQUIRE_GRAPHQL="^3@dev"
+        - BEHAT_SUITE="cms"
+    - php: 7.3
       env:
         - DB=MYSQL
         - REQUIRE_INSTALLER="$REQUIRE_RECIPE"
         - BEHAT_TEST=1
+        - REQUIRE_GRAPHQL="^3@dev || 4.x-dev"
```
maxime-rainville, line 41: I don't understand the rational for having a `3 || 4` constraint here. In practice, the 4 version will always be installed no matter what. Having a `3 || 4` constraint just adds useless ambiguity about what is being tested for each matrix entry.

maxime-rainville (changes requested):

.travis.yml
emteknetnz: I'm not sure either at this point.  There would have been a rationale back in the day, presumably there was some cases where graphql 4 wasn't installable and we still wanted the job to still run so it would fallback to graphql 3

At this point though, yeah it doesn't really seem to make any sense

That said, there are several other modules that have `"^3@dev || 4.x-dev"` and I'd rather not have 2 different ways of saying 'run graphql 4'

emteknetnz: I've updated it to `^4@dev` - my understanding is that's pretty much the loosest constraint non-tag constraint there is so it should work with whatever recipes / modules define e.g. `^4.1 / 4.x-dev / 4.5.x-dev`

maxime-rainville (approved):
The travis conf looks OK.

The build are still failing however .. it's complaining about 
```
In Schema.php line 1055:
Config contains invalid keys: modelConfig  
```

If that's an expected failure and it's track somewhere else, feel free to merge.

emteknetnz:
@maxime-rainville yup graphql4 is existing failure https://github.com/silverstripe/silverstripe-admin/issues/1215

PR #1218, opened 2021-05-25
NEW Add dedicated controller to displays errors in CMS

maxime-rainville:
This is an experimental project. The idea is that you could display more useful errors when something goes wrong in the CMS


GuySartorelli:
This isn't needed after https://github.com/silverstripe/silverstripe-admin/pull/1317

Issue #1215, opened 2021-05-17
Broken builds - GraphQL 4 - modelConfig


emteknetnz:
https://travis-ci.com/github/silverstripe/silverstripe-admin/jobs/505604067 (admin suite)
https://travis-ci.com/github/silverstripe/silverstripe-admin/jobs/505604068 (cms suite, same issue)

We've recently started building the graphql 4 schema at the start of a behat test - https://github.com/silverstripe/silverstripe-graphql/pull/380

This seems to have exposed an issue within admin

```
In Schema.php line 1055:
  Config contains invalid keys: modelConfig  
```

### PRs
- [x] https://github.com/silverstripe/silverstripe-admin/pull/1224
- [x] https://github.com/silverstripe/silverstripe-admin/pull/1225

emteknetnz:
Looks like it's frameworktest - using 0.4.0 instead of the more recent 0.4.2 -- needs this commit https://github.com/silverstripe/silverstripe-frameworktest/commit/3c83a619888a95edc93a0f6a095f028fd87cb1c3

maxime-rainville:
All the PRs have been merged ... I've merged up 1.8 into 1. Just waiting to confirm that's green.

maxime-rainville:
Travis is happy

PR #1220, opened 2021-05-27
FIX managed_models slugs can't have hyphens

### LINKED ISSUE #1219: Hyphens in `managed_models` keys cause errors
According [to the docs](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/modeladmin/#defining-the-modeladmin-models) we should be able to have a `$managed_models` key like so:

```php
        'product-category' => [
            'dataClass' => Category::class,
            'title' => 'Product categories'
        ]
```

However, adding one always results in the error: `ModelAdmin::init(): Invalid Model class`

I believe this is due to the `unsanitiseClassName()` call but I'm setting out to prove it shortly.

# Pull request
- [ ] #1218

andrewandante:
Fixes #1219


andrewandante:
Remaining test failures are from upstream 👍

dhensby (approved):

emteknetnz (approved):
Failing travis tests are existing https://github.com/silverstripe/silverstripe-admin/issues/1215

tests/php/ModelAdminTest.php
```diff
@@ -37,6 +38,17 @@ public function testModelAdminOpens()
         $this->assertEquals(200, $this->get('ContactAdmin')->getStatusCode());
     }
 
+    public function testMultiModelAdminOpensEachTab()
```
maxime-rainville, line 41: What's this thing meant to do?

It's [failing our kitchen sink build](https://travis-ci.com/github/silverstripe/cwp-recipe-kitchen-sink/jobs/510738944) and it's not actually making any assertions.

andrewandante: Previously, this would throw an exception in the `doInit()` call by hitting: https://github.com/silverstripe/silverstripe-admin/blob/1/code/ModelAdmin.php#L162

This "asserts" that each tab will open without crashing in the MultiModelAdmin.

Not sure why that would fail in the kitchen sink but pass here - my guess is it would be fixed by setting the session in the same way as in `testExportFieldsDefaultIsSummaryFields()`?

maxime-rainville: Tried defining a session like the other test, but got a different error.

```
Already directed to http://localhost/Security/login?BackURL=%2F; now trying to direct to Security/login?BackURL=%2F
 /home/maxime/Projects/cwp2x/vendor/silverstripe/framework/src/Control/Controller.php:640
 /home/maxime/Projects/cwp2x/vendor/silverstripe/admin/code/LeftAndMain.php:827
 /home/maxime/Projects/cwp2x/vendor/silverstripe/framework/src/Security/Security.php:436
 /home/maxime/Projects/cwp2x/vendor/silverstripe/subsites/src/Extensions/LeftAndMainSubsites.php:372
 /home/maxime/Projects/cwp2x/vendor/silverstripe/framework/src/Core/Extensible.php:468
 /home/maxime/Projects/cwp2x/vendor/silverstripe/framework/src/Control/Controller.php:123
 /home/maxime/Projects/cwp2x/vendor/silverstripe/admin/tests/php/ModelAdminTest.php:49
 /home/maxime/Projects/cwp2x/vendor/sminnee/phpunit/phpunit:52
```

All good ... I'll try to have a go at fixing it.

andrewandante: Thanks mate, let me know how it goes

PR #1221, opened 2021-06-01
MNT Put a placeholder page to point users to the new pattern lib

maxime-rainville:
The URL of the pattern lib will change. This just put a boring bootstrap placeholder page to point users to the new location.

I just copied a [generic template](https://github.com/lewagon/bootstrap-boilerplate) and added a bit of JS to tweak the link to make sure you land on the right page.

You can preview it here: https://creative-commoners.github.io/silverstripe-admin

@silverstripeux Not sure how much value you put in this transition page. If you feel like it's worth styling it a bit let me know.

# Parent issue
- https://github.com/silverstripeltd/product-issues/issues/421


emteknetnz (approved):
Looks great, I'll merge as is.  If we want further UX tweaks later we can open a second PR.

PR #1222, opened 2021-06-01
Use new public/_graphql/ subfolder location

chillu:
Relies on https://github.com/silverstripe/silverstripe-graphql/pull/385


emteknetnz:
This has caused a regression on graphql 3?

We'll be tagging 4.10 which will still have graphql 3 as the default version, so this change should be going out in 4.11

image

unclecheese:
Did this happen in a test suite, or in a dev environment? The `isLegacy` check should prevent that URL from ever being seen. https://github.com/silverstripe/silverstripe-admin/pull/1222/files#diff-91a46241bc2775e1723704f6746eeb70a244c3f445cbeb4e9c8f4e33c9d463f3R25

Check the body tag and see if that attribute is there?

emteknetnz:
Dev environment

`document.body.getAttribute('data-graphql-legacy')` returns null

unclecheese:
I'm not able to replicate this one with a vanilla 4.x-dev install with graphql 3.x-dev. I'm getting a `data-graphql-legacy` attribute on the body tag. Can you check CMSDialog.ss in admin, and / or `GraphQLLegacyExtension` in admin to make sure they're functioning correctly?

emteknetnz:
Now I'm getting console warnings for `http://bugfix2.test/admin.types.graphql` ? 

Both of these files are in admin - I wonder if they're sporadically loading in a different order?

**graphql-legacy.yml:**
```
SilverStripe\Admin\LeftAndMain:
  graphql_legacy: true
```

**graphql-schema-admin.yml:**
```
SilverStripe\Admin\LeftAndMain:
  graphql_legacy: false
```

unclecheese:
New PR to get rid of the 404. https://github.com/silverstripe/silverstripe-admin/pull/1265

PR #1224, opened 2021-06-03
MNT Remove pattern-lib, bump frameworktest, resolve merge-conflict

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1215

PR #1225, opened 2021-06-03
MNT Restore NPM test

maxime-rainville:
https://github.com/silverstripe/silverstripe-admin/pull/1224/ accidentally removed the NPM test

# Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1215

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,5 +41,3 @@ jobs:
         - BEHAT_TEST=1
         - REQUIRE_GRAPHQL="^4@dev"
         - BEHAT_SUITE="cms"
-    - env:
-        - NPM_TEST=1
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 version: ~> 1.0
 
 import:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -41,3 +41,5 @@ jobs:
         - BEHAT_TEST=1
         - REQUIRE_GRAPHQL="^4@dev"
         - BEHAT_SUITE="cms"
+    - env:
+        - NPM_TEST=1
```
emteknetnz, line 45: instead of this, could you update the provision to standard-jobs-fixed-npm.yml?  That'll add in an npm job using the shared config

emteknetnz (changes requested):

.travis.yml
emteknetnz, line 45: now that the provision has been updated we should be able to remove these lines

maxime-rainville: :facepalm:

maxime-rainville, line 45:
```suggestion
```

PR #1226, opened 2021-06-04
BUG Fix broken ModelAdmin test

maxime-rainville:
Resolves https://github.com/silverstripe/silverstripe-admin/issues/1219

The `testMultiModelAdminOpensEachTab` test fails when subsite is installed which fails the sink. Not 100% sure why.

This PR makes the test more resilient by explicitly logging in a user and setting a session.

### DIFF
--- a/tests/php/ModelAdminTest.php
+++ b/tests/php/ModelAdminTest.php
@@ -38,14 +38,33 @@ class ModelAdminTest extends FunctionalTest
         $this->assertEquals(200, $this->get('ContactAdmin')->getStatusCode());
     }
 
+
+
+
     public function testMultiModelAdminOpensEachTab()
     {
         $admin = new MultiModelAdmin();
+        $this->logInWithPermission();
         foreach ($admin->getManagedModelTabs()->toNestedArray() as $tab) {
             $request = new HTTPRequest('GET', $tab['Link']);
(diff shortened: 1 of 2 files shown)


tests/php/ModelAdminTest.php
```diff
@@ -38,14 +38,33 @@ public function testModelAdminOpens()
...
     {
         $admin = new MultiModelAdmin();
+        $this->logInWithPermission();
         foreach ($admin->getManagedModelTabs()->toNestedArray() as $tab) {
             $request = new HTTPRequest('GET', $tab['Link']);
             $request->setRouteParams(['ModelClass' => substr($tab['Link'], strlen('admin/multi/'))]);
+            $request->setSession(new Session([]));
             $admin->setRequest($request);
             $admin->doInit();
+            $this->assertEquals(
```
maxime-rainville, line 54: When writing tests, it's not enough that a bit of code doesn't throw an exception. It could be doing something completely unexpected without throwing an exception.

It's also difficult afterwards to understand what the test is trying to accomplish.

andrewandante: Yep fair enough, threw it together too quickly. Odd that this particular test requires the admin login and some others don't - nice work debugging.

PR #1235, opened 2021-07-01
Update CMSDialog.ss

lekoala:
Otherwise you might get admin.types.graphql 404 (Not Found) errors when using the template with graphql v3


emteknetnz (approved):
Looks good - matches what happens to the `<body>` tag in LeftAndMain.ss https://github.com/silverstripe/silverstripe-admin/blob/1/templates/SilverStripe/Admin/LeftAndMain.ss#L10

PR #1237, opened 2021-07-08
ENH Avoid "new" keyword to instantiate injectable CMSMenuItem

GuySartorelli:
There's no point CMSMenuItem being injectable if it isn't actually injected.


dhensby (approved):

dhensby:
thanks

PR #1241, opened 2021-07-29
FIX TreeDropdownField error when title not present or null

wilr:


michalkleiner:
How can it happen there's no title or it's not at least an empty string? Is this possibly covering an issue somewhere else?

wilr:
@michalkleiner Here is the response from the GraphQL query. This is on `has_one` to `File` which is not linked to anything already.

Screen Shot 2021-07-29 at 2 42 38 PM

emteknetnz (changes requested):
Seems fine, would prefer it on a single line.

I think it's fair to assume that title will always be a null able string and not a bool, so 

`const title = option.title ? option.title.toLocaleLowerCase() : '';` should also be fine is the line ends up being too long

Could you run `yarn build` and push updated JS build files

PR #1242, opened 2021-07-29
MINOR: Add parseShortcodes: false

unclecheese:
Depends on https://github.com/silverstripe/silverstripe-graphql/pull/398 (kind of.. won't break anything if you merge it before)


GuySartorelli:
@unclecheese Do you still have context about this in your brain? If so, can you please help me understand what this does and why we want it?

GuySartorelli:
Due to the lack of response and the fact that there's no context for this, I'm going to close it.

PR #1246, opened 2021-08-26
ENH Pass form validation result to client

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/764

Allows the ValidationResult to be passed to the client as part of a pjax response

Wrapping the JSON in a `<script type="application/json">` is something of a workaround due to the expectation that pjax returns HTML.  If it's not wrapped in a `<script type="application/json">` tag it will fail here https://github.com/silverstripe/silverstripe-admin/blob/1/client/src/legacy/LeftAndMain.js#L775


code/LeftAndMain.php
```diff
@@ -1476,6 +1488,20 @@ public function getEditForm($id = null, $fields = null)
         return $form;
     }
 
+    /**
+     * Convert and arry of data to JSON and wrap it in an HTML tag as pjax is used and jQuery will parse this
```
michalkleiner, line 1492:
```suggestion
     * Convert an array of data to JSON and wrap it in an HTML tag as pjax is used and jQuery will parse this
```

michalkleiner:
That's an interesting mechanism with the header, didn't know there was something like that. Always learning :)

Looks ok to me, let's wait for the tests.

brynwhyman:
Just came across this PR which I thought was worth noting here incase yo uhadn't seen it: https://github.com/silverstripe/silverstripe-elemental/pull/921

code/LeftAndMain.php
ChloeSartorelli: Done

emteknetnz (commented):
I'd tick "Approve" if I wasn't the original author on this PR

michalkleiner (approved):

michalkleiner:
Approved and merged, but now I wonder whether it should've targeted 2 as an enhancement? But seeing the target branch was updated last week by @GuySartorelli I'm not sure. LMK.

GuySartorelli:
On its own, yes this is just an enhancement and belongs in 2 - but it was done specifically to work with https://github.com/silverstripe/silverstripe-elemental/pull/930 in order to solve https://github.com/silverstripe/silverstripe-elemental/issues/764 - so it's part of a wider work to fix a bug that's pretty egregious, and I think in this case warrants being released in CMS 4 as a patch.

PR #1252, opened 2021-09-30
ENH: Allow use of the preview panel in any LeftAndMain subclass.

GuySartorelli:
For some reason, the preview panel template which is referred to in `LeftAndMain::PreviewPanel()` only exists in the cms module for the `CMSMain` controller.

All of the other code for using a CMS preview is available for any arbitrary `LeftAndMain` subclass and is available in this module. There should be no dependency from this module on the cms module to allow previews, and previews should work for all `LeftAndMain` subclasses in this module (e.g. `ModelAdmin`).

Required for #1250

Edit: Depends on silverstripe/silverstripe-framework#10112

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -39,6 +39,7 @@ en:
     MenuToggleAuto: Auto
     MenuToggleStickyNav: 'Sticky nav'
     NavigateUp: 'Navigate up a folder'
+    NO_PREVIEW: 'No preview available'
     PERMAGAIN: 'You have been logged out of the CMS.  If you would like to log in again, enter a username and password below.'
     PERMALREADY: 'I''m sorry, but you can''t access that part of the CMS.  If you want to log in as someone else, do so below.'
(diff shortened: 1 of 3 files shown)


GuySartorelli:
I'm considering putting the `ViewModeSelector` (currently would be added through an extension to `GridFieldDefailForm_ItemRequest`'s `ItemEditForm` only via an extension in #1250) into the `LeftAndMain_EditForm.ss` template (mirroring its placement in `CMSMain_EditForm.ss`) as a part of this pull request, and taking it out of the extension from #1250.

This would require being able to check if the form has the extra class `cms-previewable` (as we don't want the selector there for non-previewable DataObjects) - which requires adding a method like `hasExtraClass` to the `Form` model in framework. Frankly this check should be made in the `CMSMain_EditForm.ss` template as well.

If nobody points out any issues with that in the next week or so I'll go ahead and raise the new PR in framework, and update both this and #1250 appropriately.

michalkleiner (approved):

emteknetnz (approved):

PR #1255, opened 2021-10-05
[CVE-2021-36150] Escape html tag characters in link text

maxime-rainville:
Backporting fix for CVE-2021-36150


maxime-rainville:
Commit approved through security process

PR #1262, opened 2021-11-02
MNT Remove js prefix from require statements

emteknetnz:
eslint-config PR
- [ ] https://github.com/silverstripe/eslint-config/pull/8

Related:
- [ ] https://github.com/silverstripe/silverstripe-asset-admin/pull/1238
- [ ] https://github.com/silverstripe/silverstripe-campaign-admin/pull/212
- [ ] https://github.com/silverstripe/silverstripe-ckan-registry/pull/256
- [ ] https://github.com/silverstripe/silverstripe-cms/pull/2700
- [ ] https://github.com/silverstripe/silverstripe-tagfield/pull/196


client/src/bundles/vendor.js
```diff
@@ -11,7 +11,7 @@ require('json-js');
 
 // jQuery plugins require that the jQuery object is exposed as a global
 // webpack.ProvidePlugin is used to ensure that jQuery and $ are provided to all includes
-require('script-loader!../../../thirdparty/jquery/jquery.js');
+require('script-loader!../../../thirdparty/jquery/jquerys');
```
maxime-rainville, line 14: I think there's a typo in the statement.

```suggestion
require('script-loader!../../../thirdparty/jquery/jquery');
```

maxime-rainville (changes requested):

client/src/bundles/vendor.js
emteknetnz: Ha fixed

client/src/lib/tests/ShortcodeSerialiser-test.js
```diff
@@ -1,5 +1,7 @@
 /* global jest, jasmine, describe, beforeEach, it, pit, expect, process */
-
+/* eslint-disable
+ import/extensions
```
maxime-rainville, line 3: It's not clear to me why this test can't use a normal `import` statement or why we need a `dontMock` statement here.

client/src/lib/tests/router-test.js
```diff
@@ -1,5 +1,7 @@
 /* global jest, jasmine, describe, beforeEach, it, expect, process */
-
+/* eslint-disable
+ import/extensions
+*/
 jest.dontMock('../Router.js');
```
maxime-rainville, line 5: Maybe try a normal `import` statement and/or remove the `dontMock` statement

maxime-rainville (changes requested):
When we are disabling a single thing, I think it would be better to use `// eslint-disable-next-line no-alert`.

Some of those `require` calls look weird. Before we go ahead and add linting exceptions, I would just like us to confirm that the weirdness is actually necessary.

client/src/boot/BootRoutes.js
```diff
@@ -1,5 +1,7 @@
 /* global window, document */
-
+/* eslint-disable
+ no-alert
```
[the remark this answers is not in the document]
emteknetnz: Updated

client/src/legacy/DateField.js
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable
+ import/extensions
+*/
```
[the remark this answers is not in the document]
emteknetnz: Doesn't seem like it's required, removed it, ran `yarn build` and I can still filter SiteTree's using the DateField

client/src/lib/tests/ShortcodeSerialiser-test.js
emteknetnz: update,d works as an import statement

client/src/lib/tests/router-test.js
emteknetnz: update,d works as an import statement

emteknetnz:
@maxime-rainville have made changed

> // eslint-disable-next-line no-alert.

I've retained the file-wide disables consistency with the other rules we ignore file wide e.g. `import/no-unresolved`

maxime-rainville (approved):
Maxime is happy

PR #1264, opened 2021-11-08
ENH Open GridField rows in a new tab with Shift/Ctrl/Meta key

gurucomkz:
There's often a need to keep the GridField context of the current page and open a record from it in a separate tab. Currently it can be achieved with 3 steps:
1) LMC on the row's Actions Menu
2) RMC on "Edit" option
3) pick "Open in new tab" option

Here I suggest letting users use Shift/Ctrl/Meta keys with the left mouse click to open GridField records in a new tab, just like everywhere else.


GuySartorelli:
Hi,
Thanks for this, that's a great idea. Sorry it's taken so long to get a review on this.

I've retargetted to the `2` branch and rebased.

Can you please make the following changes?
1. Open in new tab if the user does a middle mouse click
2. Remove meta key from this - I have tried clicking on a bunch of links with the meta key held down and it did not open them in new tabs.
3. Either remove shift from this, or make shift open in a new _window_ which seems to be the correct behaviour based on my testing.
4. Abstract that functionality into an entwine function, so we can just call the one function from everywhere that needs that behaviour.

kourawealthtech:
Hi @GuySartorelli 
Thanks for the feedback.

1. Sure, great point.
2. Meta is needed for MacOS - there Ctrl is replaced by Cmd(Meta)
3. For some reason I failed to make Shift open a new window (I came to conclusion that browsers keep this feature to themselves), however decided to keep the function for those who still have a habbit of using Shift key (I've seen such people). Point was to make adapt to people's habbits as much as possible. I had a great feedback from users even with such unconventional Shift key behaviour.
4. Sure

gurucomkz:
Hi @GuySartorelli 
Thanks for the feedback.

1. Sure, great point.
2. Meta is needed for MacOS - there Ctrl is replaced by Cmd(Meta)
3. For some reason I failed to make Shift open a new window (I came to conclusion that browsers keep this feature to themselves), however decided to keep the function for those who still have a habit of using Shift key (I've seen such people). Point was to make it adapt to people's habits as much as possible. I had a great feedback from users even with such unconventional Shift key behaviour.
4. Sure

gurucomkz:
@GuySartorelli I have a question about your suggestion to make the function generally available in `entwine`.
All the examples here https://docs.silverstripe.org/en/5/developer_guides/customising_the_admin_interface/javascript_development/ suggest that the functions are attached to a selector. Does "general availability" means that I must use a `"*"` selector?

GuySartorelli:
All of the scenarios where we want this are on gridfield rows - so either the selector for a gridfield row or for the gridfield as a whole are where I'd put it (whichever makes it more convenient to call from where it's needed).

GuySartorelli:
@gurucomkz There hasn't been any action on this PR for a while. Do you still want to work on this?

lozcalver:
@GuySartorelli As we have maintainer access I’ve rebased this, can we merge it as is? It’d be a really nice UX improvement for the CMS.

I can’t really spot what more can be abstracted now:

- `showDetailView` arguably _is_ the abstraction
- `GridField.js` has its own version of `showDetailView` as it _could_ be being used outside of the CMS, so that can’t be combined with the `LeftAndMain.js` version like I initially thought when I looked at this
- `$('.grid-field .action.no-ajax, .grid-field .no-ajax .action:button')` intentionally avoids calling `showDetailView()` so that it doesn’t accidentally end up calling LeftAndMain’s `loadPanel()`

GuySartorelli:
> I can’t really spot what more can be abstracted now

Ideally we wouldn't have the same code in three places - we should just have one function that can be called from three places. I don't mind how that's done, but my thinking is that both `showDetailView` functions and the `onclick` handler which all have this functionality should call a single function.

> GridField.js has its own version of showDetailView as it could be being used outside of the CMS

By "outside the CMS" do you mean in the front-end?
That is not a supported use-case, as indicated by the fact that the javascript for gridfield is all in the `bundle.js` rather than having its own dist file. I would be very surprised if `client/src/legacy/GridField.js` on its own would handle everything a gridfield should be able to do.

If you are concerned about `GridField.js` relying on `LeftAndMain.js`, we could add the new function somewhere else and `import` it perhaps.

All that said - if you see some clear roadblocks to abstracting that logic out, I would prefer to merge this rather than close it.

lozcalver:
> Ideally we wouldn't have the same code in three places - we should just have one function that can be called from three places.

Agreed, but I can’t see how it can be helped as they all behave slightly differently:

1. [This copy](https://github.com/silverstripe/silverstripe-admin/blob/cb593d5794033311e428bd4e5a7418a5850fdca3/client/src/legacy/GridField.js#L620-L637) is explicitly opting-out of AJAX, so we can’t call the abstracted `showDetailView()` like we do elsewhere because that will likely result in a AJAX call via `showPanel()`
2. [This copy](https://github.com/silverstripe/silverstripe-admin/blob/cb593d5794033311e428bd4e5a7418a5850fdca3/client/src/legacy/GridField.js#L123-L130) is the base GridField implementation. I also don’t know if GridField does actually work outside the CMS, but regardless of that we can’t remove or replace it (explained below)
3. [This copy](https://github.com/silverstripe/silverstripe-admin/blob/cb593d5794033311e428bd4e5a7418a5850fdca3/client/src/legacy/LeftAndMain.js#L1256-L1272) is similar to the one above but adds AJAX loading

2 and 3 both need to exist separately or we’d break the [`cms-loading-ignore-url-params` functionality](https://github.com/silverstripe/silverstripe-admin/blob/cb593d5794033311e428bd4e5a7418a5850fdca3/client/src/legacy/LeftAndMain.js#L1256-L1272), as this needs to fall back to the base GridField implementation.

I looked at abstracting out just the “open in new tab” part to a `showDetailViewInNewTab()` function but it doesn’t actually save anything, it just results in _more_ duplication of the `if (e && (e.metaKey || e.ctrlKey || e.shiftKey))` checks instead as these then have to be moved out of `showDetailView()`.

If you think I’m missing something obvious feel free to give it a shot, otherwise I think we should just merge this.

GuySartorelli:
1 and 2 above seem like they're functionally the exact same code, unless I'm missing something? Here's a diff between the relevant parts:

```diff
 if (e && (e.metaKey || e.ctrlKey || e.shiftKey)) {
-  var newtab = window.open(this.actionurl(), '_blank');
+  var newtab = window.open(url, '_blank');
   newtab.focus();
 } else {
-  window.location.href = this.actionurl();
+  window.location.href = url;
 }
```

I was thinking of something like this:

```js
function whateverYouWantToCallThisFunction(event, url, callback) {
    if (event.metaKey || event.ctrlKey || event.shiftKey)) {
        const newtab = window.open(url, '_blank');
        newtab.focus();
    } else {
        callback();
    }
}
```

Then for each of your "this copy" links above:

1. becomes
    ```js
    const url = this.actionurl();
    whateverYouWantToCallThisFunction(e, url, () => window.location.href = url);
    ```
1. becomes
    ```js
    whateverYouWantToCallThisFunction(e, url, () => window.location.href = url);
    ```
1. becomes
    ```js
    whateverYouWantToCallThisFunction(e, url, () => $('.cms-container').loadPanel(url));
    ```

Only the if/else clause is duplicated, so that's all we need to replace. I'm not proposing changing anything that exist before or after the conditional clauses.

lozcalver:
Okay, I’ve pushed up another commit with those changes. Let me know if you have a better name for the function or want me to squash the commits. Thanks for bearing with me @GuySartorelli 😅

client/src/legacy/GridField.js
```diff
@@ -120,8 +120,16 @@ $.entwine('ss', function($) {
         }
       }, ajaxOpts));
     },
-    showDetailView: function(url) {
-      window.location.href = url;
+    showDetailView: function(url, event) {
+      this.maybeOpenUrlInNewTab(event, url, () => window.location.href = url);
+    },
+    maybeOpenUrlInNewTab: function(event, url, openInSameTabCallback) {
```
michalkleiner: I'd just name it `openUrl`, the method name doesn't always need to convey all the logic of the method body in its name.

GuySartorelli (approved):
LGTM and works well.
I'm not too fussed about the name, but I'll hold off merging for a day or so in case either of you feel strongly about it.

lozcalver:
I had to rebase this, so I renamed the function to `openUrl` at the same time

michalkleiner (approved):

GuySartorelli (approved):
Thanks! Works great locally.

PR #1268, opened 2021-11-22
DEP Update JS dependencies

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/505

Needed to make an update to the `DatetimeField.js` after the upgrade to moment-js and change px to rem in _bootstrap-variables.scss to keep boostrap happy (didn't like a calc that had a mix of px + rem)

moment-js added a bunch of locale files

May be some CSS regressions that get picked up during beta, looks like there's one on toasts padding (I'm assuming this came from admin) - however I think this is fine merging as is, raising and issue fixing during the beta

image


maxime-rainville (approved):
LGTM

PR #1272, opened 2021-12-04
FIX LeftAndMain_MenuToggle limit to 4 characters rather then 3

LiamKearn:
FIX LeftAndMain_MenuToggle limit to 4 characters rather then 3 in order to support 4.xx. Fixes display of 4.10 as 4.1


LiamKearn:
@michalkleiner over here! Sorry my bad was a 2 minute `.dev` style PR haven't worked out branch convention yet :)

Issue #1274, opened 2021-12-06
Version number gets truncated on 4.10.0-beta1


maxime-rainville:
With `4.10`, we need to display an extra character in that help menu version number. This is getting truncated right now.

![image](https://user-images.githubusercontent.com/1168676/144919055-58c6d228-8f71-458a-9780-32797ee31f08.png)

# Acceptance criteria
- [ ] Any combination of `4.XX` can be displayed in the menu without being truncated.
- [x] The CWP specific logic is removed/deprecated.

# Stretch Goal 
- Update wording to "Silverstripe CMS"

# Related issue
- https://github.com/silverstripe/silverstripe-admin/issues/1256

# PRs
- [x] https://github.com/silverstripe/silverstripe-admin/pull/1276
- [x] https://github.com/silverstripe/cwp-core/pull/105

michalkleiner:
@maxime-rainville https://github.com/silverstripe/silverstripe-admin/pull/1273
You can review and merge it if you're happy with it.

clarkepaul:
If we can lower the middle S of Silverstripe that would be a good addition, don't worry about adding "CMS" to it as it goes over a double line and doesn't look good. Would need to drop the font-size to 11px and thats way too small.

maxime-rainville:
All PRs have been merged.

PR #1276, opened 2021-12-07
FIX Allow double digit minor version numbers

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1274

Also updates the the `application_name` config from 'SilverStripe' to 'Silverstripe CMS';


code/LeftAndMain.php
```diff
@@ -1755,6 +1755,21 @@ public function CMSVersionNumber()
...
+    /**
+     * Returns the minor version e.g. 4.10
+     * Will handle 4.10.x-dev by removing .x-dev
+     *
+     * @return string
+     */
+    public function CMSMinorVersionNumber(): string
+    {
+        $version = $this->CMSVersionNumber();
+        if (preg_match('#^([1-9])\.([1-9][0-9]?)#', $version, $m)) {
```
michalkleiner, line 1767: This will have the same problem as we have now if we got to three-digit versions — unlikely but not impossible.

I'm also not 100% sure we should display the same thing for stable and dev versions, e.g. 4.10 for both 4.10 and 4.10.x-dev. Could the hover show the full thing?

code/LeftAndMain.php
```diff
@@ -1755,6 +1755,21 @@ public function CMSVersionNumber()
         return $lockModules[$moduleName];
     }
 
+    /**
+     * Returns the minor version e.g. 4.10
+     * Will handle 4.10.x-dev by removing .x-dev
+     *
+     * @return string
+     */
+    public function CMSMinorVersionNumber(): string
```
maxime-rainville, line 1764: My gut feeling would be to just update `CMSVersionNumber` to drop the patch release version instead of creating a new method ... it's clearly only meant to be rendered in the template, and we don't use it anywhere else.

We also don't make any formal promise about what we'll be returning there beyond saying its a string.

code/LeftAndMain.php
maxime-rainville: I'm with Michal on supporting 3-digit minor version. We'll probably never get there, but the regex for supporting a version number like 123.456.789 is simpler to understand any way.

```suggestion
        if (preg_match('#^(\d+)\.(\d+)#', $version, $m)) {
```

There could be some interesting value in adding some bit more detail about the stability of what you have installed ... if only for our own sake. However, I think that goes beyond the scope of this card which is just to make sure we don't confuse people when they upgrade to 4.10. We have a related card to convert the help menu to React and implement some updated "Silverstripe CMS" branding. I think that discussion would make more sense there.

maxime-rainville (changes requested):

code/LeftAndMain.php
michalkleiner: Fair enough re: the stability aspect and more info, happy to leave that for the other card.

Since `#^(\d+)\.(\d+)#` isn't used for validation but rather just extracting the major and minor version, knowing that the version is a valid composer version, this simplified approach is fine. Using the [official semver regex](https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string) would be an overkill here.

emteknetnz:
De-wrap it, remove 'CMS'

tests/php/LeftAndMainTest.php
```diff
@@ -180,4 +181,38 @@ public function testGetHelpLinks()
...
+     */
+    public function provideTestCMSVersionNumber()
+    {
+        return [
+            ['4.9.1', '4.9'],
+            ['4.10.5', '4.10'],
+            ['4.9.x-dev', '4.9'],
+            ['4.10.x-dev', '4.10'],
+            ['myfork', 'myfork'],
+        ];
```
michalkleiner, line 216: Can we test the 3-digit version as well? Thanks

```suggestion
        return [
            ['4.9.1', '4.9'],
            ['4.10.5', '4.10'],
            ['4.236.7, '4.236'],
            ['4.9.x-dev', '4.9'],
            ['4.10.x-dev', '4.10'],
            ['myfork', 'myfork'],
        ];
```

yarn.lock
```diff
@@ -3611,7 +3611,7 @@ debug@^3.1.0, debug@^3.2.7:
   dependencies:
     ms "^2.1.1"
 
-debuglog@*, debuglog@^1.0.1:
+debuglog@^1.0.1:
```
michalkleiner, line 3614: The change doesn't require yarn.lock file change, might be a bi-product of the CSS build? Normally I'd reset this.

emteknetnz: May as commit it otherwise it'll just show up in future PRs

tests/php/LeftAndMainTest.php
emteknetnz: Updated

code/LeftAndMain.php
```diff
@@ -1740,19 +1740,22 @@ public function CMSVersion()
 
     /**
      * Return the version number of the CMS, ie. '4.2.1'
```
michalkleiner, line 1742:
```suggestion
     * Return the version number of the CMS in the 'major.minor' format, e.g. '4.2'
```

michalkleiner (approved):
Last minor method comment tweak and all good from me. Thanks!

code/LeftAndMain.php
emteknetnz: Updated

michalkleiner: Missed the patch deletion from the original version example @emteknetnz

client/src/components/Menu/Menu.scss
```diff
@@ -771,6 +771,10 @@
       .cms-help__badge {
         display: block;
       }
+
+      .cms-help__logo {
+        min-width: 26px;
```
maxime-rainville, line 776: What is the purpose of this line? Nothing changes when I disable it.

maxime-rainville (changes requested):

maxime-rainville:
@clarkepaul This is what the update help menu looks like.
image

client/src/components/Menu/Menu.scss
emteknetnz: I had issues when cwp-core was installed with it squishing the logo

That line was previously in the custom.css file in cwp-core, that now that's been deprecated seems like it makes more sense here.

Since it possibly has no effect, it has no real downside to it being here

clarkepaul:
Can you just make it a little more vertically centred please? try moving the word Silverstripe down 2-3px and leave the version badge (if the badge also goes down lift it up 2px). Basically needs to all look aligned.

emteknetnz:
I'm not seeing that locally, either with cwp-core installed or not (though that should no longer make a different now that we're removed the cwp-core template + css)

image

clarkepaul:
@emteknetnz 's version looks better, is that because the 2 of you use different browsers? If so I guess it was always like that.

client/src/components/Menu/Menu.scss
maxime-rainville: You had extra CSS rule in CWP-core which have now been removed. Keeping dead-weight code around for nothing just increases our technical debt.

Take the rule out.

maxime-rainville (changes requested):

emteknetnz:
@maxime-rainville updated.  Have re-added the yarn.lock file which got removed here https://github.com/silverstripe/silverstripe-admin/commit/7dbe0547c84fe82e2f6be2d3a047929254e64057

michalkleiner:
@emteknetnz it looks like the last force-push removed the change and only added the yarn.lock, can you please check?

emteknetnz:
Opps :-)

Note vendor.js folder has updated even though there's no JS changes here

The yarn.lock file was missing form `1.10` branch.  I've copied in the yarn.lock file from the `1` and run `yarn build`.  I've also rebased this branch on top of `1.10`

emteknetnz:
@maxime-rainville the min-width rule is required.  I think it may be to do with recent update to dependencies in yarn.lock

image

In Firefox 94 it get rendered as 'flex-shrink' ... I'm not entirely sure how though

This is WITHOUT the rule

image

Note: this is WITHOUT the cwp-core module installed, that has no affect on this.

client/src/components/Menu/Menu.scss
```diff
@@ -771,6 +771,10 @@
       .cms-help__badge {
         display: block;
       }
+
+      .cms-help__logo {
+        flex-shrink: 0;
```
maxime-rainville, line 776: Put this on line 693 with the rest of logo rules.

maxime-rainville (changes requested):

client/src/components/Menu/Menu.scss
emteknetnz: updated

code/LeftAndMain.php
```diff
@@ -1739,20 +1739,23 @@ public function CMSVersion()
     }
 
     /**
-     * Return the version number of the CMS, ie. '4.2.1'
+     * Return the version number of the CMS in the 'major.minor' format, e.g. '4.2.1'
```
michalkleiner, line 1742:
```suggestion
     * Return the version number of the CMS in the 'major.minor' format, e.g. '4.2'
```

emteknetnz: Updated

maxime-rainville (approved):
Looks good.

Issue #1279, opened 2022-01-10
Inconsistent state between current URL and actual content if XHR request fails


patricknelson:
## TODO (emteknetnz):
- [x] Cross repo run of behat using vendor-code-patcher e.g. asset-admin etc
  [pass] https://github.com/emteknetnz/silverstripe-installer/actions/runs/1715647443
- [ ] ~~Write a behat test - replication steps and new required step-definition in a PR comment~~
  - this does not seem possible - cookie is HttpOnly meaning you cannot delete it with JS.  Nor can log out user serverside since php has context of behat, not the silverstripe site i.e. Security::getCurrentUser() is null

## Original issue

It appears that the CMS will enter into an inconsistent mixed state if the current XHR/AJAX server request happens to return a erroneous HTTP code (i.e. the browser will point to the **new URL** for a page that failed to load while actually showing the **old content**). 

A common use case for this would be if the user's session has timed out (or is no longer valid) and the user attempts to navigate to a new page. This presents the login dialog. However, if the user logs in, the new URL is still presented while the old panel is still visible (and the CMS is now in a mixed/inconsistent state).  See below.

**Steps to reproduce:**

1. Vanilla SilverStripe installation (`framework` and `cms` 4.9.1 and `admin` 1.9.0 at time of writing)
2. Navigate to "Home"
3. Clear cookies (including your session cookie), effectively logging you out in the same way as a typical timed out session
4. Click "About Us"
5. Log back in

![silverstripe URLs and session reauthentication modal](https://user-images.githubusercontent.com/4269377/148714536-ca670e38-6168-457a-8a20-2ab7afe815c6.gif)

# Acceptance criteria
- [ ] URL state is constant after an error.
- [ ] URL state is constant after a relogin.

## PR
- https://github.com/silverstripe/silverstripe-admin/pull/1280

patricknelson:
@maxime-rainville I'm about to submit a PR with a proposed fix for this. Should it be branched off of (and against) `1` or `1.10` since it's more of a bug fix? 🤔

emteknetnz:
@patricknelson easiest to branch off 1.10 because it makes retargetting it to 1 later on easier.  Which branch it should target depends on the extent of PHP API changes i.e. adding a new public or protected method in PHP means it should go into 1, though if there's only a new private method then it can go into 1.10 and be released much sooner.

patricknelson:
[...]

PR #1280, opened 2022-01-10
FIX Inconsistent state between current URL and actual content if XHR request fails

patricknelson:
Proposed fix for #1279. Basically, abstract the ability to reverse to previously known good working state so that it can _**also**_ be called in case the current `$.ajax` request fails. That way we don't end up in a mixed/inconsistent state between the URL and the content currently visible on page.

~~**IMPORTANT:** This is currently a WIP since I'm having issues on my Win10 machine getting the `npm run build` process to work due to issues finding `eslint`. I'm containerizing the production build right now so it'll work on my machine so I can also compile the build/dist files (I just have to wait for `yarn` to finish downloading the entire Internet).~~


patricknelson:
Ok, containerized and was able to do a production build (thanks to `docker-compose.yml` in #1281).

patricknelson:
Pushed a fix (can squash/rebase later) since I noticed some builds failing. Also saw that unfortunately `window.ss.router.back()` is using `setTimeout()` (old version probably but same code: https://github.com/visionmedia/page.js/blob/master/page.js#L632), so there's probably a bug in the pausing of state since it technically is unpausing state BEFORE the actual .back() call actually runs.

Noticed this myself when on the page history tab. If you:

1. Reload (start fresh) on page history tab
2. Delete your session cookie
3. Click settings (triggering the server error HTTP response on next XHR)

It will call `.back()` instead of `window.ss.router.show()` which is a synchronous function. Unfortunately, `.back()` isn't synchronous (and doesn't use promises!) and just uses `setTimeout()` resulting in it re-triggering `handleStateChange` since it's no longer actually paused. 🤦‍♂️

patricknelson:
Quick note: I had already branched this off of `1.10` but accidentally made the PR against `1` 😄

client/src/legacy/LeftAndMain.js
```diff
@@ -643,6 +640,33 @@ $.entwine('ss', function($) {
       return promise;
     },
 
+    /**
+     * Reverts the window.history state back to the last known good state. See this.handleStateChange().
+     */
+    reverseStateChange: function() {
```
emteknetnz, line 646: do we also need to call `this.setStateChangeCount(this.getStateChangeCount() + 1);`, or `this.setStateChangeCount(this.getStateChangeCount() - 1);` ?

client/src/legacy/LeftAndMain.js
```diff
@@ -628,7 +615,17 @@ $.entwine('ss', function($) {
         headers: headers,
         url: historyState.path || document.URL
       })
+      .fail((xhr, status, error) => {
+        // Ignoring aborts: The server request failed, so prevent a mixed UI state by rolling back to previously
+        // succesfully loaded URL (consistent with currently loaded content).
+        if (xhr.readyState !== 0) {
```
emteknetnz, line 621: why do we need to test for `xhr.readyState !== 0`?

emteknetnz (changes requested):
Seems good, a couple of questions

client/src/legacy/LeftAndMain.js
patricknelson: I noticed that when switching between page settings tabs, the XHR was requested multiple times but the history state was updated only once. The first request was nearly immediately aborted (resulting in `.readyState` of 0).

patricknelson: p.s. While I happened to stumble upon this buggy side effect (due to inconsistent/repeated XHR's) I noticed that pushing this line also fixed some [failing tests](https://app.travis-ci.com/github/silverstripe/silverstripe-admin/jobs/554830938#L1751) as well. So, I guess it didn't just affect the page settings/history tabs.

client/src/legacy/LeftAndMain.js
patricknelson: I only abstracted [this block of code](https://github.com/silverstripe/silverstripe-admin/pull/1280/files#diff-3f04cea4046f53f0d7ea15fb030d44ab684baf2290fa998b5e2462feb8fe5fd8L575-L587) to a separate method (and ideally it'd be private/protected) so that it could be called multiple times within the legacy `handleStateChange` method. So, no I don't think so. 

Maybe I should consider declaring it as a function/closure inside of the `handleStateChange` method?

patricknelson: Oh rats, sorry I misread your comment. Yeah that's probably a good idea.

patricknelson: Sorry for the delay on this @emteknetnz.

Went ahead and pushed an extra line to also decrement that `StateChangeCount` property, see e129d254e728f6a7fcf3dca75b315286ab88debf. Good idea to keep the API consistent (albeit I don't see it referenced anywhere _within_ the module). Best not to break anything referencing it for a bug fix.

emteknetnz:
@patricknelson Is there anyone this would be testable via behat?  Behat's what we use for end-to-end testing and it basically just gets chrome to click buttons and enter text.  I wouldn't expect you to write the behat test, I can handle that myself, I'm just wondering what steps we would need to include here

patricknelson:
I'm not familiar with behat directly @emteknetnz but there are some great steps to reproduce the bug, see https://github.com/silverstripe/silverstripe-admin/pull/1280#issuecomment-1009541773. Basically:

1. Ensure logged in.
1. Check current URL (e.g. `/admin/pages/edit/show/1`).
1. Delete session cookies (replicates effect of session timing out)
1. Click site tree element to navigate to another page (e.g. located at `/admin/pages/edit/show/2`)
1.  Login dialog should appear. Current URL should still be `/admin/pages/edit/show/1`.

Before this change, current URL would be the new one `/admin/pages/edit/show/2`. Does that help? Not sure if in behat you can alter cookies/session or if it runs JS (maybe via ChromeDriver?)

emteknetnz:
I could probably define a custom behat step that runs some JS to delete all the site cookies.  

I'll put a todo to create this myself in a seperate PR.  I'll also get something spun up to do a cross repo run of behat using this PR on asset-admin, campagin admin, etc, simply because changing js navigation code is somewhat risky

Leave this with me, once I'm happy I'll probably merge this as is

patricknelson:
Ok thanks @emteknetnz! I’ll keep an eye on this ticket and the separate PR you create since I’m interested to see what that’d look like.

emteknetnz (commented):
I think I'm generally happy with this.  I've run all the core behat suites on a custom CI through using this PR and it seems fine https://github.com/emteknetnz/silverstripe-installer/actions/runs/1715647443

I was unable to create a behat test specifically for this PR however due to the inability to delete HttpOnly cookies via JS or logout the current member via PHP because we're in a behat context controlling the browser rather than inside silverstripe i.e. Security::getCurrentMember() is null

I'll manually test this PR hopefully tommorow and then merge

patricknelson:
Thank you. I understand how time consuming it is doing this so I really appreciate the effort digging into this.

emteknetnz (dismissed):
After following replication steps using fork (not sure if it was precisely or not) I got this error in firefox

image

patricknelson:
Ugh, sorry that's a goof up on my part @emteknetnz. I accidentally used a regular function closure instead of an arrow function and so `this` was bound to the wrong object in the `setTimeout()` (which I'll add, by the way, certainly **is** hacky and sadly is needed due to it being used in the `page.js` library we depend on here 😕)

GuySartorelli:
Tested locally and it works great! @patricknelson Can you please rebase and then run `yarn build` again to resolve the merge conflict? Then I'll be happy to approve and merge.

GuySartorelli:
1.10 please. We'll release this as a patch release since it's a fix with no API changes.

> Do you know why that file is committed? I'm sure there's a logical explanation for why, it's just never something I do (precisely for this reason)

We don't currently have an automated process to build the file after the PR has been merged, so this ensures that when we do releases the js and css have already been built. We're looking at adding quite a few github actions to automate things and this is one of the things on the cards - but for now it's done the old school way

patricknelson:
I'm actually working on this right now, just had to start fresh with a new node 10 container, so I have to wait for `yarn` to finish downloading half the Internet before I can run the build, commit and push. 😅 Wanna get this out the door before it goes stale again!

patricknelson:
Ok @GuySartorelli please poke at it now.

GuySartorelli:
I'll just wait for travis to do its checks first ;p

GuySartorelli:
The travis failures are unrelated to this PR:
- Graphql v4 wants framework 4.11, so it naturally can't be installed for 4.10
- There is some php 7.3 syntax error in a file in the frameworktest module, which is obviously unrelated.

GuySartorelli (approved):

GuySartorelli:
@patricknelson Tagged as [1.10.3](https://github.com/silverstripe/silverstripe-admin/releases/tag/1.10.3)

PR #1281, opened 2022-01-10
NEW Add docker-compose.yml for better cross environment compatibility.

patricknelson:
Just recommendation to facilitate easier development for those of us not running *nix or MacOS. I found it very easy to perform `npm run dev` on my host machine, but `npm run build` failed repeatedly for unknown reasons. Containerizing it and simply running in the `node:10` container seemed to work like a charm. 

Enabled my build for #1280

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,22 @@
+##
+## To get started
+## 1. Start the container by running:  docker-compose up -d
+## 2. Drop into a shell via:           docker-compose exec node ash
+## 3. Initialize "node_modules" via:   yarn
+##
+## Further development and builds can be performed via "npm run ..." (see scripts in package.json)
+##
+
+services:
+  node:
+    # Node version per "engines" in package.json
+    image: node:10-alpine
+    working_dir: /build
+
+    # Keeps container running so you can login directly (good for debugging or manual task running).
+    entrypoint: tail -f /dev/null
+    environment:
+      NODE_ENV: development
+
+    volumes:
+      - .:/build


michalkleiner:
As much as I appreciate this addition making the module development easier, I'm leaning towards not having it within the codebase. I'd possibly put it into docs, but that might be just me.

patricknelson:
I'd say we should either do this  or at least find ways to make local development more Windows-friendly for developers.

There were some issues getting things like `NODE_ENV=development ...` commands embedded in `package.json` to work but that was easily mitigated by installing https://github.com/laggingreflex/win-node-env. But unfortunately `eslint` wouldn't work at all, regardless of what I tried, even though I installed everything precisely the same way as in the container. This was _in addition_ to ensuring I had exactly the right node version for everything to work (easy to workaround though these days with tools like `nvm`).

patricknelson:
p.s. I don't even know where in the docs this could possibly go. For me picking this up and finding the bug in `silverstripe-admin` module, I checked the `README.md` which just links to the documentation website, which wasn't helpful for me getting started contributing to this particular module. I'm aware of the general contributions documentation but not specific to how to work with this module's webpack workflow.

emteknetnz:
Sorry, I don't want this in the admin module

I don't think any Silverstripe repo, including docs, is the correct place to have any docker/vagrant/other type of environment sample files.  Local environments is not something we're interested in spending time on maintaining.  There are a multitude of sample environment files available on the internet for people to copy and modify

patricknelson:
That’s ok @emteknetnz, I get it.

This was just a quick fix. If I do happen to find the root cause of why this wasn’t working on Windows host machines (and it wasn’t just an silly problem on my end), I’ll see if I can maybe submit a PR on that in particular.  Is that fine? 

Thanks for the guidance.

emteknetnz:
@patricknelson npm on Windows (even WSL) is notorious having a large number of little things that just don't work right.  I personally don't place any value on view getting JS to build natively on win10, and even if we did get it to work, there's no guarantee that it will works in a years time.  So I don't think there's any value in an effort to try and get it to run natively on Win10.  I think you simply need to reply on containers here.

PR #1283, opened 2022-01-18
FIX Apply patch and run yarn build

emteknetnz:
https://github.com/silverstripe/silverstripe-admin/pull/1280

I seemed to get a slightly different bundle.js when running `yarn build` locally.  

However the travis git diff on the rebuild bundle.js passed.

I'd prefer to use the original PR though.  Will probably close this


emteknetnz:
Issue was a nbsp being converted to a regular space

PR #1285, opened 2022-01-30
FIX Right-align preview on the first record

emteknetnz:
This [PR](https://github.com/silverstripe/silverstripe-admin/pull/1252) introduced a small bug where the first DataObject being added had a left-aligned navigator instead of right aligned

image

Subsequently added DataObjects have it correctly right aligned

image

This PR ensures the first DataObject is correctly right aligned

image


michalkleiner (approved):
This probably indicates that the system behind where/how the components are added/displayed could use a restructure/rethink, which is out of scope here.

Merge on green.

PR #1286, opened 2022-02-07
ENH: Prefer dependency injection for GridFieldComponents.

GuySartorelli:
`GridFieldComponent`s packaged with silverstripe/framework will be injectable as of 4.11.0 (see silverstripe/silverstripe-framework#10204)
Explicitly invoking the injector here instead of using `create()` allows backwards compatibility with framework < 4.11.0 while ensuring dependency injection is still used from 4.11.0 onwards.

If core team would prefer, I can instead use the `::create()` static method which was added via the `Injectable` trait - but that will also require updating the constraint for the framework dependency to `^4.11`.

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -6,6 +6,7 @@ use SilverStripe\Control\Controller;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Convert;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\BulkLoader;
 use SilverStripe\Dev\Deprecation;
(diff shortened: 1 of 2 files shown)


GuySartorelli:
Once this has been approved I'll go through other repositories and check for `GridFieldComponent`s that are being instantiated with the `new` keyword but I am waiting to see if this is the preferred way of injecting them before I do so.

michalkleiner (approved):
Both approaches work for me, I'd prefer the `::create` one but I'm not sure what's the appetite on forcing `^4.11` onto folks e.g. if someone uses admin in a different version to get newer stuff but wants to stay on older cms/framework. That would force them to alias those dependencies as well.
Will leave for @maxime-rainville or @emteknetnz to make the final call.

dhensby (approved):

PR #1288, opened 2022-02-14
ENH Allow setting options for extra css/js in LeftAndMain.

GuySartorelli:
I've also made the js/css tests a little more robust to make sure the scripts and styles are actually in script/style tags, not just arbitrarily printed in the body.


michalkleiner (approved):
Nice enhancement, thanks Guy!

Merge on green.

GuySartorelli:
Might take a couple of tries to get the regex for the tests just right. I forgot to include the cache querystring.

GuySartorelli:
@michalkleiner Got it on the second try. All tests are now green. Thanks for approving this one so fast!

PR #1291, opened 2022-02-22
BUG Add missing dependency to build pattern library

maxime-rainville:
The pattern library is throwing some errors for `isPlainObject.isPlainObject is not a function`. Adding the `is-plain-object` library to admin's dev dependencies fixes it.

image

# Parent issue
- https://github.com/silverstripe/silverstripe-pattern-lib/issues/2


emteknetnz (approved):
MOG

PR #1294, opened 2022-03-07
ENH PHP 8.1 compatibility

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10250

### DIFF
--- a/code/CMSBatchActionHandler.php
+++ b/code/CMSBatchActionHandler.php
@@ -280,11 +280,11 @@ class CMSBatchActionHandler extends RequestHandler
      */
     protected function cleanIDs($csvIDs)
     {
-        $ids = preg_split('/ *, */', trim($csvIDs));
+        $ids = preg_split('/ *, */', trim($csvIDs ?? ''));
         foreach ($ids as $k => $id) {
             $ids[$k] = (int)$id;
         }
-        return array_filter($ids);
+        return array_filter($ids ?? []);
     }
 
     /**
@@ -297,7 +297,7 @@ class CMSBatchActionHandler extends RequestHandler
     {
         $actions = static::registeredActions();
         $recordClass = $this->recordClass;
-        $actions = array_filter($actions, function ($action) use ($recordClass) {
+        $actions = array_filter($actions ?? [], function ($action) use ($recordClass) {
             return $action['recordClass'] === $recordClass;
         });
         return $actions;
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -341,7 +341,7 @@ class LeftAndMain extends Controller implements PermissionProvider
             // Trim leading/trailing slash to make it easier to concatenate URL
             // and use in routing definitions.
             'name' => $name,
-            'url' => trim($this->Link(), '/'),
+            'url' => trim($this->Link() ?? '', '/'),
             'form' => [
                 'EditorExternalLink' => [
                     'schemaUrl' => $this->Link('methodSchema/Modals/EditorExternalLink'),
@@ -794,7 +794,7 @@ class LeftAndMain extends Controller implements PermissionProvider
             $response->addHeader('X-Controller', static::class);
         }
         if (!$response->getHeader('X-Title')) {
-            $response->addHeader('X-Title', urlencode($title));
+            $response->addHeader('X-Title', urlencode($title ?? ''));
         }
 
         // Prevent clickjacking, see https://developer.mozilla.org/en-US/docs/HTTP/X-Frame-Options
@@ -926,7 +926,7 @@ class LeftAndMain extends Controller implements PermissionProvider
         // Get default class title
         $title = static::config()->get('menu_title');
         if (!$title) {
-            $title = preg_replace('/Admin$/', '', $class);
+            $title = preg_replace('/Admin$/', '', $class ?? '');
         }
 
         // Check localisation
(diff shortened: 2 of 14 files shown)


code/LeftAndMain.php
```diff
@@ -1765,7 +1765,7 @@ public function CMSVersionNumber()
             return '';
         }
         $version = $lockModules[$moduleName];
-        if (preg_match('#^([0-9]+)\.([0-9]+)#', $version, $m)) {
+        if (preg_match('#^([0-9]+)\.([0-9]+)#', $version ?? '', $m)) {
```
GuySartorelli, line 1768: By this point if `$lockModules[$moduleName]` was not null (i.e. `!isset()`) we've already returned out of the method. No need for null coalescing here.

code/MemberImportForm.php
```diff
@@ -37,7 +37,7 @@ public function __construct($controller, $name, $fields = null, $actions = null,
             $importer = new MemberCsvBulkLoader();
             $importSpec = $importer->getImportSpec();
 
-            $columns = implode(', ', array_keys($importSpec['fields']));
+            $columns = implode(', ', array_keys($importSpec['fields'] ?? []));
```
GuySartorelli, line 40: This is ultimately the same as the `GroupCsvBulkLoader` import spec as mentioned above.
No need for null coalescing here.

GuySartorelli (changes requested):
Found a lot of situations where null coalescing just isn't necessary - and I stopped because there were just too many of them. We've discussed this a bit in person but for the sake of having this recorded somewhere can you please respond with your thinking on this? It's definitely overkill the way this is being done so far and it'll be good to have it recorded that this is a conscious decision.

emteknetnz:
> It's definitely overkill the way this is being done so far and it'll be good to have it recorded that this is a conscious decision.

Yes agree it's overkill, though I'm not sure how else we can do this.

The code writing simply looks to see if a variable is being passed in as a parameter, rather than a scalar or a nested native function call with a single return type.

Ideally the code writer would have more smarts about it where it could judge if modification being made to a variable could lead it to being in a nullable state.

However even if we did have a method that could trace back changes made to variables, we're still stuck with the fact that the vast majority of Silverstripe methods have dynamic return types, and the docblock return types are very often wrong, so you cannot rely on them.  Since variables are largely a result of calls to Silverstripe methods, we won't very far with an automated cleanup.

At a future date, once Silverstripe is fully strongly typed, I would like to do an automated cleanup run that does trace what type variables, and then based on that remove the redundant null coalescing statements

code/CMSBatchActionHandler.php
```diff
@@ -280,11 +280,11 @@ protected function buildAction($class)
      */
     protected function cleanIDs($csvIDs)
     {
-        $ids = preg_split('/ *, */', trim($csvIDs));
+        $ids = preg_split('/ *, */', trim($csvIDs ?? ''));
         foreach ($ids as $k => $id) {
             $ids[$k] = (int)$id;
         }
-        return array_filter($ids);
+        return array_filter($ids ?? []);
```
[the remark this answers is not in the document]
emteknetnz: Correct fix here would be some defensive programming like adding a new line about this such as `$ids = is_array($ids) ? $ids : [];`

However since this hasn't come up in any unit test failures, I think we should just leave it as is.  (There will be thousands of other examples similar to this)

GuySartorelli:
In general I agree - doing this in an automated fashion we're either going to go overboard and miss a few cases, or be light-handed and miss _heaps_ of cases. But on the otherhand doing this all by hand (like I was for the review) is a long laborious process with little payoff. Ultimately having some null coalescing operators where they aren't strictly needed isn't going to hurt anybody, and like you say it'll be much easier to go back and clean that up in a future major release when things are strongly typed.

I'll give Max and other core committers a chance to review this discussion before moving on but I think this is ultimately aan acceptable approach.

maxime-rainville:
I think Steve and I had a similar discussion in a previous Zoom call. Generally, my mind is in the same place yours are. There's not really a way to support PHP8.1 without warning and without having a big automating process that will type cast everything.

To a degree, this will bite us in the ass later once we strongly type everything ... and will potentially neglect some of the benefit of strongly typing because it will obfuscate logic errors. 

In the end, we have to begrudgingly accept the automated approach and as a short term necessary evil.

GuySartorelli (approved):
As per the above conversation, the purposes of this PR is to avoid passing null (specifically null) from built-in functions that as of php 8.1 will throw deprecation warnings if null is passed in. The approach is necessarily heavy-handed, and while there are many situations where it isn't needed, it would be prohibitively laborious to find all of those situations by hand.

With that context, this looks good to me.

code/ModelAdmin.php
```diff
@@ -620,7 +620,7 @@ public function import($data, $form, $request)
 
         // File wasn't properly uploaded, show a reminder to the user
         if (empty($_FILES['_CsvFile']['tmp_name']) ||
-            file_get_contents($_FILES['_CsvFile']['tmp_name']) == ''
+            file_get_contents($_FILES['_CsvFile']['tmp_name'] ?? '') == ''
```
GuySartorelli, line 623: An error is emitted whether this is an empty string or null.
> Uncaught ValueError: Path cannot be empty

Is a warning emitted _as well_ as the error if this is an empty string?

GuySartorelli (commented):
Just a question here before I merge.

code/ModelAdmin.php
emteknetnz: both, php 8.1
```
php -a
php > file_get_contents(null);
PHP Deprecated:  file_get_contents(): Passing null to parameter #1 ($filename) of type string is deprecated in php shell code on line 1

Deprecated: file_get_contents(): Passing null to parameter #1 ($filename) of type string is deprecated in php shell code on line 1
PHP Warning:  Uncaught ValueError: Path cannot be empty in php shell code:1
Stack trace:
#0 php shell code(1): file_get_contents()
#1 {main}
  thrown in php shell code on line 1

Warning: Uncaught ValueError: Path cannot be empty in php shell code:1
Stack trace:
#0 php shell code(1): file_get_contents()
#1 {main}
  thrown in php shell code on line 1

```

PR #1295, opened 2022-03-10
DEP Update eslint-config

emteknetnz:
Issue https://github.com/silverstripe/eslint-config/issues/9


maxime-rainville (approved):
We'll fix the broken build separately.

PR #1299, opened 2022-04-06
FIX Start react event delegation before page router.

GuySartorelli:
React doesn't start its event delegation until it has rendered a component which has an event handler attached to it.
The page router has a global click handler associated with it, which will take all internal link click events which point at an admin section and direct the user to that section via pjax unless the event has had `preventDefault()` called on it.
Without this PR, calling `preventDefault()` on a link's click event has no effect, because the page router's event handler was registered first, so it was processed before the event ever got passed to react's handlers.

## Parent issue
silverstripe/silverstripe-asset-admin#1007


GuySartorelli:
Behat test failures are pre-existing and are unrelated to this change (e.g. see https://github.com/silverstripe/silverstripe-admin/runs/5828586404)

emteknetnz (approved):
Ran it through a cross module behat test - all green

https://github.com/emteknetnz/silverstripe-installer/actions/runs/2145505469

Issue #1304, opened 2022-05-04
DataObjects that return an empty PreviewURL result in 404 in preview panel


GuySartorelli:
It is perfectly valid for a `CMSPreviewable` object to return `null` or an empty string from the `PreviewURL()` method - for example a dataobject which expects to belong to a page may return an empty value if it does not currently have a page relation set.

In this case the preview panel should display the "no preview available" message instead of spitting out a 404 - and ideally would be collapsed when loading this object (which is the same behaviour you would have for objects that are not `CMSPreviewable`.

## Acceptance Criteria
- [ ] If a `DataObject` which implements `CMSPreviewable` has no preview URL in the current context, the preview panel displays a relevant message instead of an error page
- [ ] The preview panel is collapsed by default for the above mentioned situation
- [ ] When navigating to an edit form for a `DataObject` that does have a preview URL, the previously used mode (e.g. split, preview, or edit) is used (i.e. the above collapsing is not stored)

## PRs
- silverstripe/silverstripe-cms#2737
- #1305

PR #1305, opened 2022-05-04
FIX Collapse the preview panel by default if no preview is available.

GuySartorelli:
This makes the editor experience more streamlined for DataObjects that _can_ have a preview (and therefore the preview panel is present) but do not have a preview in the current context (e.g. an elemental block owned by a non-previewable DataObject).

## Dependencies (won't work without these PRs)
- silverstripe/silverstripe-cms#2737

## Parent issue
- #1304


michalkleiner:
Should it be collapsed or disabled? Sorry been a while for me — what happens when you open the panel and there's no preview available? Is there a 'No preview available' state? All good if that's already the case.

GuySartorelli:
> what happens when you open the panel and there's no preview available? Is there a 'No preview available' state?

Yup! Theoretically at least. There's a related PR to this in the description that makes sure null or empty preview links will display the "no preview available" message. Unfortunately the classes that handle that are in CMS... we should move them to admin at some point.

> Should it be collapsed or disabled?

I think it should be collapsed - the "no preview" message will display if users choose to expand it. This reflects the fact that the data object _is_ previewable, there's just no preview for it in the current context.
That said, I'm not a UX expert by any means... if anyone has reasoning why it should be disabled I'm happy to make that change.

GuySartorelli:
Changed base from `1` to `1.11` as this is a fix that should be included in the 1.11.0 release.

GuySartorelli:
I've created an issue (#1306 ) for adding test coverage for previews in admin - there currently aren't any in admin itself, so it will be quite a big task that I won't be able to get done in the current sprint.

sabina-talipova (approved):

Issue #1308, opened 2022-05-18
make lazy-loadable a global feature


lekoala:
Currently, if you have lazy content inside tabs, you need to implement custom js code to handle this
Which is a bit of a shame, because there is already some code related to this for the gridfield

https://github.com/silverstripe/silverstripe-admin/blob/d3fc48cde0aef36518d0b7efd11dd5711bef6be4/client/src/legacy/TabSet.js#L105-L113

Would the ss dev consider making looking for any ".lazy-loadable" node and trigger a lazy-load event on it ? Actually, it could also be applied to the gridfield. The current implementation feels like a missed opportunity to open this feature for other fields.

lekoala:
@GuySartorelli i can certainly give it a try, the basic idea is quite simple and yet has many advantages :)

lekoala:
@GuySartorelli so, i'm working on it and i've discovered that lazy loading is only delayed if there is a parent ss-tabset.. But the cms-tabset doesn't have that class (and that seems to be a deliberate choice).

https://github.com/silverstripe/silverstripe-admin/blob/d3fc48cde0aef36518d0b7efd11dd5711bef6be4/client/src/legacy/GridField.js#L24

That means that for a GridFieldDetailForm, lazy loading of the GridField is always triggered right away. Is that the desired outcome ?

lekoala:
@GuySartorelli so I gave it a try
https://github.com/silverstripe/silverstripe-admin/pull/1309

this is a fairly conservative PR, since i keep both ways of doing things.
fixing the cms-tabset selector seems like a nice bonus, since I don't believe excluding it was intended in this case as it defeats the initial purpose.

GuySartorelli:
> i've discovered that lazy loading is only delayed if there is a parent ss-tabset.. But the cms-tabset doesn't have that class (and that seems to be a deliberate choice). ... fixing the cms-tabset selector seems like a nice bonus, since I don't believe excluding it was intended in this case

According to [the docs](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/cms_architecture/#tabs) the `cms-tabset` class is actually the class that _should explicitly_ work, so I'd say you've found a bona fide defect:
> The base behaviour is applied through adding a class .cms-tabset to a container.

I'll take a look at the PR today and let you know if it needs any changes.

lekoala:
[...]

Issue #1248, opened 2021-09-13
Silverstripe CMS UI Dev doc is a complete nightmare


maxime-rainville:
I just went through our existing docs to try to figure out how a new developer who is looking to do a JS/CSS PR could get started. And it's a complete useless non-sense. There's bits of information spread across a dozen different articles. There's no clear place to start and no natural progression between the articles. Most of the articles discuss very esoteric details. Even the individual articles jump around from topic to topic without a clear purpose.

# What articles do we have right now
This is a list of the articles I could find discussing JS development and a brief summary of what each article discuss.

## [Javascript Development](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/javascript_development/#jquery-jquery-ui-and-jquery-entwine-our-libraries-of-choice)

 - A very brief overview of our build stack without going into React/Redux/Appollo
 - jQuery, jQuery UI, jQuery.entwine, jQuery Plugins, jQuery UI Widgets
 - PJAX

Most of this page can probably be scrapped. It's mostly just giving a very bad and confusing jQuery crash course.

## [Introduction to the "React" layer](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/reactjs_redux_and_graphql/)

- Very brief overview of React
- Very brief overview of Redux
- Very brief overview of GraphQL and Apollo
- The Injector API
  - A very weird aside about middlewares
  -  Registering new services
  - Transforming services using middleware
  - Controlling the order of transformations
  - Injector context
- Customising React components with Injector
  - Using dependencies within your React component
  - Using the injector directly within your component
  - Using Injector to customise forms
  - Using Injector to customise Redux state data
    - Lots of weird side points
  - Using Injector to customise GraphQL queries
    - Lots of very advanced topics

This article is way too long. Some bits could be reuse but need to be rethought to have a more sensible progression. For example the Injector stuff.

The React, Redux, GraphQL/Apollo overviews are better than the jQuery overview, but also kind of pointless.

## [CMS Architecture](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/cms_architecture/)

- How to install and build JS dependencies (with  an outdated mehtod that won't work)
- Pattern library
- How to change your Admin URLs
- PHP only things about
  - Templates and Controllers
  - Layout and Panels
  - Forms
- JavaScript through jQuery.entwine
- JavaScript and CSS dependencies via Requirements and Ajax
- Client-side routing
  - For entwine
[...]

PR #1309, opened 2022-05-23
Implement global lazy loadable support for tabs

lekoala:
For context, see issue
https://github.com/silverstripe/silverstripe-admin/issues/1308


client/src/legacy/GridField.js
```diff
@@ -21,7 +21,7 @@ $.entwine('ss', function($) {
       }
 
       if (this.is('.grid-field--lazy-loadable') && (
-        (this.closest('.ss-tabset').length === 0) || (this.data('gridfield-lazy-load-state') === 'force') )
+        (this.closest'.ss-tabset, .cms-tabset').length === 0) || (this.data('gridfield-lazy-load-state') === 'force') )
```
michalkleiner, line 24:
```suggestion
        (this.closest('.ss-tabset, .cms-tabset').length === 0) || (this.data('gridfield-lazy-load-state') === 'force') )
```

lekoala:
@michalkleiner sorry my bad, this is what happens when you simply edit stuff through github ;-)

client/src/legacy/TabSet.js
```diff
@@ -46,22 +56,35 @@ $.entwine('ss', function($){
...
+     * @param {Object} panel
+     */
+    triggerLazyLoad: function(panel) {
+      panel.find('.grid-field--lazy-loadable, .lazy-loadable').each((i, el) => {
+        const $el = e(el);
+        // Avoid triggering all lazy loadable scripts when using nested tabs
+        if ($el.closest('.ss-tabset, .cms-tabset').is(this)) {
+          // Trigger dedicated entwine function (see GridField.js)
+          if(typeof $el.lazyload === 'function') {
+            $el.lazyload();
```
GuySartorelli, line 71: I'm not sure where that `e` variable comes from but I assume it's meant to be the jQuery var?
We don't tend to prefix any variables in javascript with `$`, so either rename `$el` to something else or just remove that declaration.
```suggestion
        const element = $(el);
        // Avoid triggering all lazy loadable scripts when using nested tabs
        if (element.closest('.ss-tabset, .cms-tabset').is(this)) {
          // Trigger dedicated entwine function (see GridField.js)
          if(typeof element.lazyload === 'function') {
            element.lazyload();
```

client/src/legacy/TabSet.js
```diff
@@ -46,22 +56,35 @@ $.entwine('ss', function($){
...
+          }
+          // Also dispatch a native event for scripts not using entwine
+          el.dispatchEvent(new Event("lazyload"), { once: true });
```
GuySartorelli, line 74:
```suggestion
          el.dispatchEvent(new Event('lazyload'), { once: true });
```

GuySartorelli, line 74:
```suggestion
          el.dispatchEvent(new Event('lazyload'), { once: true });
```

client/src/legacy/TabSet.js
```diff
@@ -46,22 +56,35 @@ $.entwine('ss', function($){
       });
     },
 
+    /**
+     * @func triggerLazyLoad
+     * @desc Find all the lazy loadable gridfield and nodes in the panel and trigger their reload.
+     * @param {Object} panel
+     */
+    triggerLazyLoad: function(panel) {
+      panel.find('.grid-field--lazy-loadable, .lazy-loadable').each((i, el) => {
```
GuySartorelli, line 65: The `.lazy-loadable` class should be documented in the lazy loading part of the [relevant documentation](https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/02_CMS_Architecture.md#tabs)

GuySartorelli, line 65: The `.lazy-loadable` class should be documented in the lazy loading part of the [relevant documentation](https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/02_CMS_Architecture.md#tabs)

client/src/legacy/TabSet.js
```diff
@@ -96,22 +119,7 @@ $.entwine('ss', function($){
         $(this).attr('href', document.location.href.replace(/#.*/, '') + matches[0]);
       });
     },
-
-    /**
-     * @func lazyLoadGridFields
-     * @desc Find all the lazy loadable gridfield in the panel and trigger their reload.
-     * @param {Object} panel
-     */
-    lazyLoadGridFields: function(panel) {
```
GuySartorelli, line 105: Instead of removing this entirely (which could be called by someone's own project code for whatever reason they may have), move this up into the `$('.ss-tabset, .cms-tabset')` scope, and have `triggerLazyLoad` call this function instead of handling gridfields itself.

GuySartorelli, line 105: Instead of removing this entirely (which could be called by someone's own project code for whatever reason they may have), move this up into the `$('.ss-tabset, .cms-tabset')` scope, and have `triggerLazyLoad` call this function instead of handling gridfields itself.

client/src/legacy/TabSet.js
GuySartorelli: Also call `lazyLoadGridFields()` instead of handling gridfields directly here.

GuySartorelli (changes requested):
Looks good in general - the approach is nice and general, and it's resolving some problems with lazy-loadable gridfields that probably have been unnoticed (or just ignored) for a while.

The new `.lazy-loadable` class and the native event should be added to the [relevant documentation](https://github.com/silverstripe/silverstripe-framework/blob/4/docs/en/02_Developer_Guides/15_Customising_the_Admin_Interface/02_CMS_Architecture.md#tabs) and there are a few other fairly small changes here to make.

GuySartorelli:
Please also run `yarn build` and commit the dist files to make the CI happy.

lekoala:
@GuySartorelli hey, sorry, this is probably the worst pull request ever. I did work from the uncompiled source because I have issues running yarn build locally. I'm reviewing things with a fresh mind to make sure it's more like a proper PR :-)

client/src/legacy/TabSet.js
lekoala: yes sorry, that's because I was using the uncompiled source and made a copy paste and forgot to change back the jquery selector... i'm implementing your suggestions

lekoala:
@michalkleiner ah makes sense
it might still be worth an upgrade. trying to run with version 10 using nvm i get this
error @eslint/eslintrc@1.3.0: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0". Got "10.24.1"

supporting newer major versions might be a hassle, but staying on older versions might be just as bad...

i mean, if there is a full guide somewhere on how to use the build tools, I'm all for it, but installing node 10 and running yarn build doesn't work (at least on windows with the way i've installed things)

kinglozzer:
I’m pretty sure our client-side build process only works with Node 10. FWIW I use [nvm](https://github.com/nvm-sh/nvm) to handle switching between Node versions for different projects, we already have an `.nvmrc` file in this repo so running `nvm use` in the directory should be enough

lekoala:
@kinglozzer thanks for your help on this

nvm use gives me "invalid version" as a result (maybe windows specific?). still, i get pick my version and use a valid branch

removing node_modules was a good idea, but it doesn't solve the issue

i get this

```
yarn build
yarn run v1.22.18
$ yarn && yarn lint && yarn test && NODE_ENV=production webpack -p --bail --progress
[1/5] Validating package.json...
[2/5] Resolving packages...
[3/5] Fetching packages...
[4/5] Linking dependencies...
warning " > bootstrap@4.6.1" has unmet peer dependency "jquery@1.9.1 - 3".
warning " > react-router-config@4.4.0-beta.8" has incorrect peer dependency "react-router@4.4.0-beta.1".
warning " > @storybook/react@3.4.12" has unmet peer dependency "babel-core@^6.26.0 || ^7.0.0-0".
warning " > babel-jest@23.6.0" has unmet peer dependency "babel-core@^6.0.0 || ^7.0.0-0".
warning " > storybook-addon-jsx@5.4.0" has unmet peer dependency "babel-core@^6.26.0".
[5/5] Building fresh packages...
$ eslint client/src && sass-lint -v
The system cannot find the path specified.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
```

so it seems it's either eslint or sass-lint that doesn't get installed. `yarn add sass-lint` works fine, but `yarn add eslint` doesn't work and gives: `error @eslint/eslintrc@1.3.0: The engine "node" is incompatible with this module. Expected version "^12.22.0 || ^14.17.0 || >=16.0.0". Got "10.24.1"
error Found incompatible module.`
not even sure why it's not there in the first place when doing a regular `yarn install` (which doesn't trigger any error). is eslint installed globally or something like that ?

I've tried with node v12, it doesn't work due to engine restriction. I really think upgrading to at least node 12 would make sense as eslint doesn't seem to support at all node v10.

GuySartorelli:
Node 10 works fine for me, eslint and all. I don't undertand why nvm heckles you about the version... did you `nvm install 10` before `nvm use`?

> if there is a full guide somewhere on how to use the build tools, I'm all for it

This is something [we're aware it pretty lacking](https://github.com/silverstripe/silverstripe-admin/issues/1248) at the moment. But improving documentation is a lengthy process and nobody has made time to do it yet.

> I really think upgrading to at least node 12 would make sense

I agree in general - but this isn't something we're going to look at doing in the immediate short term.

GuySartorelli:
@lekoala I've built the js and pushed it to your PR branch

client/src/legacy/TabSet.js
```diff
@@ -17,49 +21,15 @@ $.entwine('ss', function($){
...
-    /**
-   * Lightweight wrapper around jQuery UI tabs for generic tab set-up
-   */
-  $('.ss-tabset').entwine({
-    IgnoreTabState: false,
-
+    
     onadd: function() {
       this.on("tabsactivate", (function (event, {newPanel}) {
         this.lazyLoadGridFields(newPanel);
```
GuySartorelli, line 27: Remove this line and call it from inside `triggerLazyLoad()` instead. That way anyway calling `triggerLazyLoad()` from their own code (for whatever reason they may have) will not have to also call `lazyLoadGridFields()` separately.
```suggestion
```

client/src/legacy/TabSet.js
```diff
@@ -17,49 +21,15 @@ $.entwine('ss', function($){
...
+        this.triggerLazyLoad(newPanel);
       }).bind(this));
       this.on("tabscreate", (function(event, {panel}) {
         this.lazyLoadGridFields(panel);
```
GuySartorelli, line 31:
```suggestion
```

client/src/legacy/TabSet.js
```diff
@@ -96,7 +66,22 @@ $.entwine('ss', function($){
...
     },
-
+    
+    /**
+     * @func triggerLazyLoad
+     * @desc Triggers a "lazyload" event on all lazy-loadable nodes
+     * @param {Object} panel
+     */
+    triggerLazyLoad: function(panel) {
+      panel.find('.lazy-loadable').each((i, el) => {
```
GuySartorelli, line 76:
```suggestion
      this.lazyLoadGridFields(newPanel);
      panel.find('.lazy-loadable').each((i, el) => {
```

GuySartorelli (changes requested):
Finally got around to taking a look at this post-refactor. It looks like some of the functionality (`onremove`, `redrawTabs`, `rewriteHashLinks`, etc) may have been intentionally omitted from `.cms-tabset`.
https://github.com/silverstripe/silverstripe-admin/blob/eb192f4f34d5462dd7ce6f27d6763aa80a2afd23/themes/cms-forms/templates/SilverStripe/Forms/CMSTabSet.ss#L1

Can you please undo the refactoring? The only change to `.cms-tabset` should be that it now includes lazy loading.

client/src/legacy/TabSet.js
```diff
@@ -17,6 +17,39 @@ $.entwine('ss', function($){
 
       this._super();
     },
+    
+    onadd: function () {
+        this.on(
+            "tabsactivate",
+            function (event, { newPanel }) {
+                this.triggerLazyLoad(newPanel);
```
GuySartorelli, line 25:
```suggestion
                this.lazyLoadGridFields(newPanel);
                this.triggerLazyLoad(newPanel);
```
And remove the `tabsactivate` event handler in the `.ss-tabset`-exclusive entwine scope

client/src/legacy/TabSet.js
```diff
@@ -17,6 +17,39 @@ $.entwine('ss', function($){
...
+            }.bind(this)
+        );
+        this.on(
+            "tabscreate",
+            function (event, { panel }) {
+                this.triggerLazyLoad(panel);
```
GuySartorelli, line 31:
```suggestion
                this.lazyLoadGridFields(panel);
                this.triggerLazyLoad(panel);
```
And remove the `tabscreate` event handler in the `.ss-tabset`-exclusive entwine scope

client/src/legacy/TabSet.js
```diff
@@ -17,6 +17,39 @@ $.entwine('ss', function($){
 
       this._super();
     },
+    
+    onadd: function () {
+        this.on(
+            "tabsactivate",
```
GuySartorelli, line 23:
```suggestion
            'tabsactivate',
```

client/src/legacy/TabSet.js
```diff
@@ -17,6 +17,39 @@ $.entwine('ss', function($){
...
+            function (event, { newPanel }) {
+                this.triggerLazyLoad(newPanel);
+            }.bind(this)
+        );
+        this.on(
+            "tabscreate",
```
GuySartorelli, line 29:
```suggestion
            'tabscreate',
```

client/src/legacy/TabSet.js
```diff
@@ -17,6 +17,39 @@ $.entwine('ss', function($){
...
+        this._super();
+    },
+    
+    /**
+     * @func triggerLazyLoad
+     * @desc Find all the lazy loadable fields in the panel and trigger their reload.
+     * @param {Object} panel
+     * @param {string} selector
+     */
+    triggerLazyLoad: function (panel, selector = ".lazy-loadable") {
```
GuySartorelli, line 43:
```suggestion
    triggerLazyLoad: function (panel, selector = '.lazy-loadable') {
```

client/src/legacy/TabSet.js
```diff
@@ -17,6 +17,39 @@ $.entwine('ss', function($){
...
+        panel.find(selector).each((idx, el) => {
+            var $el = $(el);
+            var lazyEvent = el.dataset.lazyEvent || "lazyload";
+            if ($el.closest(".ss-tabset, .cms-tabset").is(this)) {
```
GuySartorelli, line 47:
```suggestion
            var lazyEvent = el.dataset.lazyEvent || 'lazyload';
            if ($el.closest('.ss-tabset, .cms-tabset').is(this)) {
```

client/src/legacy/TabSet.js
```diff
@@ -17,6 +17,39 @@ $.entwine('ss', function($){
...
+        );
+        this.on(
+            "tabscreate",
+            function (event, { panel }) {
+                this.triggerLazyLoad(panel);
+            }.bind(this)
+        );
+        this._super();
+    },
+
```
GuySartorelli, line 36: Move `lazyLoadGridFields` into this entwine scope.

GuySartorelli (changes requested):
Sorry it's taken me so long to look at this again.

I still think it makes sense for `lazyLoadGridFields` to be moved into the `$('.ss-tabset, .cms-tabset')` entwine set - I had intended this when I said "The only change to .cms-tabset should be that it now includes lazy loading" but I should have been more explicit about it.
I'm certain that was intended, since `.cms-tabset` is checked inside that function and [the docs](https://docs.silverstripe.org/en/4/developer_guides/customising_the_admin_interface/cms_architecture/#tabs) say that's the class that should explicitly work with lazy loading.

If you can make that change along with the other small changes I've requested, I'll build the bundle for you again, test locally, and we can let CI do its thing - if that passes then I'll be happy with it.

We need docs for this as well as I've mentioned before, but that will be in the framework repository anyway so I'm okay so it has to be a separate PR, so I'm okay with merging this PR without docs for now - it would be super cool if you can create a new PR in framework to add docs, but I'm happy to do it if you don't.

GuySartorelli:
> when I started working on the PR, I thought it would be useful as a global feature

I agree, it sounds like something that would be useful to have in core for sure.

> i didn't imagine it would be difficult to run the build pipe ;-) 

Like I said, happy to take care of that part for you.

> now actually i realize that maybe its also overcomplicating things.
If someone need that kind of feature, they can include an additional LeftAndMain script

I don't think it's overcomplicated at all. Technically just about anything can be added to a project with some extra code. ;P I think this change makes sense in core though.

> that is moving towards react anyway

There hasn't been any real attempt to convert entwine functionality to react for quite a while, and I don't see it being in the works any time soon either. 

> if you find the PR interesting, maybe it's easier that you proceed and modify things yourself

I think it's pretty close to done, there's just the suggested changes I've mentioned above. If you'd rather not do them that's fine, though - I understand this all takes time so no pressure either way. If you want to close the PR, I might pick it up at some point, otherwise if you want to make those changes then as I've said if it passes testing I'll be happy to merge it.

lekoala:
@GuySartorelli mhh don't mind me, i'm sometimes a bit disappointed by how much time it takes to follow up a PR that would probably take no more than 10 minutes for a solo dev :-)
I've merged the changes you requested and moved the function to the entwine set, it all looks good to me

I'd be happy to make a PR to the docs once it's merged

GuySartorelli (approved):

GuySartorelli:
Looks good, works well locally. Thank you for this, it's a great idea.
Feel free to @ me when you create the docs PR and I'll take a look.

PR #1311, opened 2022-06-23
FIX GridField - no spacing below

sabina-talipova:
### DESCRIPTION
Space above text field was added by using margin-top. 

### Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/9154


GuySartorelli (commented):
Is this problem specifically for when a `TextField` is below a `GridField`? Or are their other fields which don't have enough space from below a `GridField`?

Does adding a margin to the top of `TextField` cause there to now be _too much_ space above it for some other scenario?

I feel like adding something _below `GridField`_ might be a more suitable solution.

sabina-talipova:
There are some inconsistencies how spaces around fields are created.
TextField expects that field above should have a space(margin or padding) below. 
GridField has different view depends on does it have items or not, does it have a description or not. So and gap between GridField every time different. 

 It would be nice to make all fields are using the identical approach in CMS 5

GuySartorelli:
This still doesn't look quite right to me in some situations... in particular when there's a description on the `GridField` there's actually less space between the gridfield (and its description) and the field below than when there is no description. This is especially noticable if there's no paginator.
Additionally, if there's no paginator but there is a message on the `GridField`, the message is touching the bottom of the `GridField`.

I expect there are a lot of other varied interactions depending on what components are included on the gridfield, and that's without even starting to test different fields being below the gridfield...

I think at a minimum, there should be a horizontal bar after the `GridField` similar to the one after `FieldGroup`, which will much more clearly mark where the `GridField` ends. But regardless, I think we probably need to spend some time looking at all of the spacing above and below all fields, and around the gridfield with various combinations of components, to make sure everything is nice and uniform... and I think that needs to be done as a larger piece of work than this PR.

sabina-talipova:
I close this PR for now.
I did some testing and found that many of the fields have paddings for some of the components of this composite field, which in turn creates extra space between the fields and the spaces between the different fields look negligent. Also, some fields do not have a line before or after the field, which in turn also looks negligent.
Most likely we should investigate and test how we add spaces between all fields with all additional attributes (description, additional fields, etc.) in a complex in different sections of the CMS. Then, probably, to create an additional wrapper for any field in CMS, which will determine the spaces between the fields following each other.
A qualified opinion from the designers may be required

Issue #1232, opened 2021-06-27
No error handling for non-existient CMS section


maxime-rainville:
If you try to access a CMS section that doesn't exist, you get an unstyled error page. e.g.: Try to access `/admin/boom` in the CMS 

![image](https://user-images.githubusercontent.com/1168676/123539531-59fcce00-d78e-11eb-9f04-f77a501eb8ca.png)

Presuming you are logged in, I would expect to see the left navigation so you can choose a section that does exist and some useful error messages.

## PRs
- https://github.com/silverstripe/silverstripe-errorpage/pull/67
- https://github.com/silverstripe/silverstripe-admin/pull/1317

GuySartorelli:
Closing in favor of https://github.com/silverstripe/silverstripe-cms/issues/2669 as it's effectively a subset of the same problem.

PR #1317, opened 2022-07-04
ENH Display error messages in admin with an admin context

GuySartorelli:
Currently if there is an error in any /admin route, there is no admin-specific way of handling the errors.
This means that if silverstripe/errorpage is installed the user is shown the frontend error page (if there is one) and if it isn't installed or doesn't have a page for the given error, a simple text message is displayed.

This PR renders an error message in the context of the admin, so that users can simply select an appropriate section in the left menu.

The error detail only appears in Dev mode. 

**In this example I tried to navigate to /admin/badonk and an appropriate 404 error is displayed**
image

**In this example I tried to navigate to /admin/pages/save and an appropriate forbidden error is displayed**
image

## Parent issues
- fixes https://github.com/silverstripe/silverstripe-admin/issues/1232
- fixes https://github.com/silverstripe/silverstripe-cms/issues/2669

## Depends on
- https://github.com/silverstripe/silverstripe-errorpage/pull/67

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -176,6 +176,11 @@ class LeftAndMain extends Controller implements PermissionProvider
     protected $pageID = null;
 
     /**
+     * Set by {@link LeftAndMainErrorExtension} if an http error occurs
+     */
+    private string $httpErrorMessage;
+
+    /**
      * Assign themes to use for cms
      *
      * @config
@@ -771,6 +776,32 @@ class LeftAndMain extends Controller implements PermissionProvider
 
         // Set default reading mode to suppress ?stage=Stage querystring params in CMS
         Versioned::set_default_reading_mode(Versioned::get_reading_mode());
+    }
+
+    public function afterHandleRequest()
+    {
+        if ($this->response->isError()) {
+            $this->init();
+            $errorCode = $this->response->getStatusCode();
+            $errorType = $this->response->getStatusDescription();
+            $defaultMessage = _t(
+                self::class . '.ErrorMessage',
+                'Sorry, it seems there was a {errorcode} error.',
+                ['errorcode' => $errorCode]
+            );
+            $this->response = HTTPResponse::create($this->render([
+                'Title' => $this->getApplicationName() . ' - ' . $errorType,
+                'Content' => $this->renderWith($this->getTemplatesWithSuffix('_Error'), [
+                    'ErrorCode' => $errorCode,
+                    'ErrorType' => $errorType,
+                    'Message' => DBField::create_field(
+                        'HTMLFragment',
+                        _t(self::class . '.ErrorMessage' . $errorCode, $defaultMessage)
+                    ),
+                ]),
+            ]), $errorCode, $errorType);
+        }
+        parent::afterHandleRequest();
     }
 
     public function handleRequest(HTTPRequest $request)
@@ -1997,5 +2028,24 @@ class LeftAndMain extends Controller implements PermissionProvider
     public function getVersionProvider()
     {
         return $this->versionProvider;
+    }
+
+    /**
+     * Get the HTTP error message if one has occurred during HandleRequest.
+     */
+    public function getHttpErrorMessage(): string
+    {
+        return $this->httpErrorMessage;
+    }
+
+    /**
+     * Set the HTTP error message when one occurs during HandleRequest.
+     * Called by {@link LeftAndMainErrorExtension}
+     * @internal
+     */
+    public function setHttpErrorMessage(string $message): self
+    {
+        $this->httpErrorMessage = $message;
+        return $this;
     }
 }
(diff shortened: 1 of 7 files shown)


code/LeftAndMain.php
```diff
@@ -773,6 +778,32 @@ protected function init()
...
+            $errorCode = $this->response->getStatusCode();
+            $errorType = $this->response->getStatusDescription();
+            $defaultMessage = _t(
+                self::class . '.ErrorMessage',
+                'Sorry, it seems there was a {errorcode} error.',
+                ['errorcode' => $errorCode]
+            );
+            $this->response = HTTPResponse::create($this->render([
+                'Title' => $this->getApplicationName() . ' - ' . $errorType,
+                'Content' => $this->renderWith($this->getTemplatesWithSuffix('_NotFound'), [
```
emteknetnz, line 794: Should rename the new template to `_Error` or something generic as probably not all errors are 'not-found'

templates/SilverStripe/Admin/Includes/LeftAndMain_NotFound.ss
```diff
@@ -0,0 +1,24 @@
...
+		<div class="cms-content-header-info vertical-align-items flexbox-area-grow">
+			<div class="breadcrumbs-wrapper">
+				<span class="cms-panel-link crumb last">
+					$ErrorType
+				</span>
+			</div>
+		</div>
+    </div>
+
+	$Tools
```
emteknetnz, line 13: Indentation

Also this template has a mix of 2 spaces and 4 spaces (possibly there's tabs in here?)

lang/en.yml
```diff
@@ -33,6 +33,12 @@ en:
     CollapsePanel: 'Collapse panel'
     DELETED: Deleted.
     DropdownBatchActionsDefault: 'Choose an action...'
+    ErrorDetail: 'Error detail: {detail}'
+    ErrorMessage: '<p>Sorry, it seems there was a {errorcode} error.</p>'
```
emteknetnz, line 37: We should not have `<p>` tags in the translations - they should be directly in the template

lang/en.yml
```diff
@@ -33,6 +33,12 @@ en:
...
+    ErrorMessage403: '<p>Sorry, it seems the action you were trying to perform is forbidden.</p>'
+    ErrorMessage404: '<p>Sorry, it seems you were trying to access a page that doesn''t exist.</p>'
```
emteknetnz, line 39: It's not really a page that does not exist - it's a route that does not exist.  Maybe a "section" is better language?

e.g. `/admin/security/asdfasdfsadf` is will say this page does not exist, though you wouldn't call `/admin/security` a page

emteknetnz (changes requested):
Much nicer having error messages in the CMS then going to the front-end

A few things

code/LeftAndMain.php
GuySartorelli: Ahh yup my bad. I was originally only treating 404 but forgot to change this when I generalised it. Will fix.

lang/en.yml
GuySartorelli: This is done _all over_ the place already.

lang/en.yml
GuySartorelli: True. I think "section" has the same problem as it may be for example a specific ID in an edit form that has no associated data object.
How about something like "You were trying to access a section or object that doesn't exist"? Or simply "The URL you were trying to access resulted in a 'Not Found' error"?

templates/SilverStripe/Admin/Includes/LeftAndMain_NotFound.ss
GuySartorelli: There were tabs. I've converted those to 4 spaces.

lang/en.yml
emteknetnz: That's not a great reason :-)

All we need to do is change the template from `$Message` to `<p>$Message</p>`, unless I've missed something

emteknetnz (changes requested):

lang/en.yml
GuySartorelli: It was my argument for it not being worth the time it takes to change rather than for it being a good idea in the first place lol. Also not worth another round of ping pong though so I've moved the tag to the template.

emteknetnz (approved):

emteknetnz:
@GuySartorelli looks like this has caused regressions - https://github.com/silverstripe/silverstripe-cms/runs/7571102762?check_suite_focus=true

`1) SilverStripe\CMS\Tests\Controllers\CMSMainTest::testCreationOfTopLevelPage
Error: Typed property SilverStripe\Admin\LeftAndMain::$httpErrorMessage must not be accessed before initialization`

GuySartorelli:
#1331 fixes that.

PR #1321, opened 2022-07-06
Enhance batch action parameters

mooror:
# Description
This is the second of two PRs (the first was made in the cms module) for implementing the enhancement described in the ["Batch actions form - Custom parameter fields enhancement"](https://github.com/silverstripe/silverstripe-cms/issues/2744) issue

# Changes
+ Added the parameter fields created by CMSMain's BatchActionParameters() method to the BatchActionsForm
+ Added javascript for displaying/hiding parameter fields when the batch action dropdown is changed
+ Added CSS styles to make the parameter fields render properly under the batch action dropdown


mooror:
One thing I noted was that each of the commits was labeled with a different author. I was having trouble running builds on windows, so I used another workstation which was configured with a different author so :man_shrugging: hopefully that won't be an issue

GuySartorelli:
> each of the commits was labeled with a different author. I was having trouble running builds on windows, so I used another workstation which was configured with a different author so man_shrugging hopefully that won't be an issue

It's not a problem. The commits will be squashed anyway before merging. If you want to control the name used in the changelog, you might want to squash the commits yourself so that you can control who the commit is authored by.

client/src/legacy/LeftAndMain.BatchActions.js
```diff
@@ -374,11 +374,23 @@ $.entwine('ss.tree', function($){
     onchange: function(e) {
       var form = $(e.target.form),
         btn = form.find(':submit'),
-        selected = $(e.target).val();
+        selected = $(e.target).val(),
+        actionUrlParts = selected.split("/"),
```
GuySartorelli, line 378:
```suggestion
        actionUrlParts = selected.split('/'),
```

client/src/legacy/LeftAndMain.BatchActions.js
```diff
@@ -374,11 +374,23 @@ $.entwine('ss.tree', function($){
...
+        actionName = actionUrlParts[actionUrlParts.length - 1];
 
       // Refresh selected / enabled nodes
       $('#Form_BatchActionsForm').refreshSelected();
 
+      // Hide / display action parameter fields
+      var parameterFields = $("#BatchActionParameters_" + actionName);
```
GuySartorelli, line 385:
```suggestion
      var parameterFields = $('#BatchActionParameters_' + actionName);
```

client/src/legacy/LeftAndMain.BatchActions.js
```diff
@@ -374,11 +374,23 @@ $.entwine('ss.tree', function($){
...
+      if (parameterFields.length) {
+        parameterFields.siblings().hide();
+        parameterFields.show();
+        $("#BatchActionParameters").slideDown();
+      } else {
+        $("#BatchActionParameters").slideUp();
```
GuySartorelli, line 391:
```suggestion
        $('#BatchActionParameters').slideDown();
      } else {
        $('#BatchActionParameters').slideUp();
```

client/src/legacy/LeftAndMain.BatchActions.js
```diff
@@ -374,11 +374,23 @@ $.entwine('ss.tree', function($){
...
+      }
+
       // TODO Should work by triggering change() along, but doesn't - entwine event bubbling?
       this.trigger("chosen:updated");
```
GuySartorelli, line 395:
```suggestion
      this.trigger('chosen:updated');
```
I know you didn't touch this line, but as with the other PR we may as well tidy some of the surrounding code a little while we're at it.

client/src/styles/legacy/_style.scss
```diff
@@ -1320,6 +1320,25 @@ html, body {
   }
 }
 
+// Tree batch action parameter fields
+#BatchActionParameters {
```
GuySartorelli, line 1324: Can we target this without using an ID selector?

code/LeftAndMain.php
```diff
@@ -1584,6 +1585,7 @@ public function BatchActionsForm()
             $actionsMap[$action->Link] = $action->Title;
         }
 
+        $cmsMain = singleton(CMSMain::class);
```
GuySartorelli, line 1588: CMSMain may not exist (it's perfectly valid to use `silverstripe/admin` in a project without `silverstripe/cms`).
Use the `updateBatchActionsForm()` extension hook in the cms module instead.

GuySartorelli (changes requested):
A few changes mostly from a code quality POV

code/LeftAndMain.php
mooror: Yeah I thought that as well, but then I noticed that the LeftAndMain class already imports the `SilverStripeNavigator` class from the CMS module on [line 9](https://github.com/silverstripe/silverstripe-admin/blob/1/code/LeftAndMain.php#L9)
```
use SilverStripe\CMS\Controllers\SilverStripeNavigator;
```

But if you'd rather use an extension, I can totally do that. In fact that is how I'm implementing it in my own installations. Oh but what should I call the `LeftAndMainExtension` subclass and file? or is there an existing extension I should put it in?

GuySartorelli: Yes please. The fact that it's using `SilverStripeNavigator` is already something we're going to have to unpick - it shouldn't do that.

> what should I call the LeftAndMainExtension subclass and file?

Anything you think suitably describes the intention of the class. Maybe `LeftAndMainBatchActionsExtension` or something along those lines.
It also doesn't have to subclass `LeftAndMainExtension` specifically - there are no methods in that class that are useful here. May as well just subclass `Extension`.

mooror: > Yes please. The fact that it's using SilverStripeNavigator is already something we're going to have to unpick - it shouldn't do that.

Oh yeah then we definitely don't want to add more dependence on the CMS module.

> Maybe LeftAndMainBatchActionsExtension or something along those lines.

Totally makes sense. I'll go ahead and create that as a Extension subclass and add it to my other PR. Or should it be a separate PR? Theoretically it shouldn't effect the impact of the first PR (in CMS) if we add it to that, as even though the LiteralField will be added to the BatchActionsForm, it won't be displayed due to the `display:none;`

GuySartorelli: > add it to my other PR. Or should it be a separate PR?

Add it to the PR you already created - it's part of the same work of making batch actions parameters usable.

client/src/styles/legacy/_style.scss
mooror: Yeah I could, but to keep the css from impacting other areas, I think I would need to add at least one class to the `#BatchActionParameters` LiteralField.

For that I could eight create a class called `cms-batch-action-parameters` (to match with the existing `cms-batch-actions` naming) and keep the current nesting of the SCSS rules. Or, if we wanted, I could call the class `action-parameters`, and we could nest all of the rules under `cms-batch-actions` (so the resulting generated rule would be `.cms-batch-actions .action-parameters .form__field-description`). Which would keep everything together in the SCSS file, but also add an extra level of specificity (which is unneeded)

What do you think?

GuySartorelli: > I think I would need to add a class to the #BatchActionParameters LiteralField (in the CMS PR)

That sounds like a good solution to me. Looking at the other styles in that file, it seems that nesting `.action-parameters` inside `.cms-batch-actions` would be most consistent (assuming the markdown is in fact nested that way).

code/LeftAndMain.php
mooror: Should the `LeftAndMainBatchActionsExtension` be placed in the base of the `code/` folder in the CMS module, or in a sub-folder (like `code/Extensions` or something)?

GuySartorelli: Maybe put it next to `LeftAndMainPageIconsExtension` (in `code/Controllers/`)

mooror:
Just as a heads up, commit [68abd1a](https://github.com/silverstripe/silverstripe-admin/pull/1321/commits/68abd1a01b2a6651846aade095ed829a1c114e1b) adds extra functionality for resetting parameter fields before they are displayed. This fixes an issue I found that was causing field values to be kept even after the "Go" button was pressed and the action successfully completed (which is not what we want).

I placed the reset functionality in the dropdown onchange event (instead of the "go" button click event) so that it would also clear values when the user changed actions (via the action dropdown). My thinking is that batch actions make large changes across many records, and we want to reduce the chances for user error.

GuySartorelli (approved):
Works well locally. Without the accompanying PR in cms this causes no problems. With the accompanying PR this works great.
Thank you for this contribution!

Issue #1314, opened 2022-06-30
GridField. Create Unique URLs for Gridfield state.


sabina-talipova:
### Description
At the moment, there are few issues with navigation in the GridField.

- There is a problem with the "Back" button on the GridField child page in the header. When working with children of a filtered, sorted GridField, or when you are on a specific page, the back control button in the child's header navigates to the first page of an unfiltered, unsorted GridField.

- There is also a problem that it is not possible to create a link to the results of a filtered and sorted GridField list, as well as use the "Back" / "Forward" button in the browser.

### SPIKE 
- https://app.zenhub.com/workspaces/cms-squad-5d12de441659a92b9fa61677/issues/silverstripeltd/product-issues/254

### AC's
- [x] Keep all state in URL via an hash reference that maps to a state store either in the session or local storage.
- [x] URL keeps user's search / sort / pagination parameters.
- [x] Search parameters are stored either an hash mapping to a server side grid state or in session storage.
- [x] Optimise URL to keep long search parameters as well as work with many tabs with different GridFields. 
- [x] When you perform an action in the gridfield, it updates the URL.
- [x] The browser history is hijacked so that the back/next button can be use to navigate your gridfield state.
- [x] Directly accessing a gridfield URL with a state hash restore hydrates that gridfield with that state for the user with a given session.

### Notes
- Need to validate if it's OK to change the URL on every state change and if changing the URL should affect the back button behaviour.
  - e.g.: If you change the sort order 15 times, should it create 15 entries in your history anh require you to click the back button 15 times to get back to your original state.

### PR's
- silverstripe/silverstripe-framework#10331
- https://github.com/silverstripe/silverstripe-admin/pull/1326

GuySartorelli:
PRs merged, no outstanding tasks. Closing.

PR #1322, opened 2022-07-07
ENH Keep State and show Search field if any keywords are presented. 

sabina-talipova:
## Description
- [x] Add a check for the presence of a filter in the state that is stored in the URL. If exist, keep Search field open.
- [x] Keep state in URL when new filter was applied or GridField was sorted

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1314


client/src/legacy/GridField.js
```diff
@@ -180,6 +194,19 @@ $.entwine('ss', function($) {
...
+    },
+
+    keepStateInHistory: function() {
+      const newURLSceme = $(this).find('.gridfield-actionmenu__container').data('schema');
+      if (newURLSceme && newURLSceme.length > 0) {
+        newURLSceme.filter( e => {
+          if (e.type === 'link') {
+            const searchParam = e.url.includes('?') ? '?' + e.url.split('?')[1] : '';
+            const historyState = $.extend({}, {path: window.location.pathname + searchParam}, this.getState());
+            history.replaceState(historyState, '', window.location.pathname + searchParam);
```
GuySartorelli, line 206: Make sure we're not just throwing away other GET vars that were already present

GuySartorelli (changes requested):
Haven't done a full review, this is just to check the approach. Looks good overall, just one small change.

client/src/legacy/GridField.js
sabina-talipova: Thank you. I'll check parameters that we pass in `GridField.php` class. Since, here I just grab all search params from scheme, that we also pass to Edit button and as URL for click. So, would be better to be sure that we have a correct value in the scheme.
What do you think?

GuySartorelli: This is the URL for the gridfield we're currently looking at - which needs to include any GET parameters that are _already_ in the URL.
So for example if I'm on some page `/admin/pages/edit/show/2?some-var=some-value` and I click paginate on a gridfield, the `some-var=some-value` shouldn't be removed. These may be here for any reason beyond gridfield, and removing them may cause problems in some project-specific code that relies on them for whatever reason.

The edit and view buttons don't need these, because they're relevant only while you're on the page that holds the gridfield, _not_ for the item edit form.

sabina-talipova:
CLOSE

PR #1326, opened 2022-07-19
ENH Keep State and show Search field if any keywords are presented

sabina-talipova:
## Description
- [x] Add a check for the presence of a filter in the state that is stored in the URL. If exist, keep Search field open.
- [x] Keep state in URL when new filter was applied or GridField was sorted

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1314


sabina-talipova:
Open instead https://github.com/silverstripe/silverstripe-admin/pull/1322

client/src/legacy/GridField.js
```diff
@@ -180,6 +194,19 @@ $.entwine('ss', function($) {
...
+    },
+
+    keepStateInHistory: function() {
+      const newURLSceme = $(this).find('.gridfield-actionmenu__container').data('schema');
+      if (newURLSceme && newURLSceme.length > 0) {
+        newURLSceme.filter( e => {
+          if (e.type === 'link') {
+            const searchParam = e.url.includes('?') ? '?' + e.url.split('?')[1] : '';
+            const historyState = $.extend({}, {path: window.location.pathname + searchParam}, this.getState());
+            history.replaceState(historyState, '', window.location.pathname + searchParam);
```
GuySartorelli, line 206: This might be a good way to preserve existing URL parameters while adding the ones from the element:
```js
const params = [window.location.search, (new URL(e.url)).searchParams.toString()].join('&');
```

client/src/legacy/GridField.js
```diff
@@ -180,6 +194,21 @@ $.entwine('ss', function($) {
       if (hasLegacyFilterHeader) {
         this.find('.sortable-header th:last').html(content);
       }
+    },
+
+    keepStateInHistory: function() {
+      const newURLSceme = $(this).find('.gridfield-actionmenu__container').data('schema');
```
GuySartorelli, line 200:
```suggestion
      const newURLSchema = $(this).find('.gridfield-actionmenu__container').data('schema');
```

client/src/legacy/GridField.js
sabina-talipova: DONE

client/src/legacy/GridField.js
```diff
@@ -180,6 +194,21 @@ $.entwine('ss', function($) {
...
+      if (newURLSceme && newURLSceme.length > 0) {
+        newURLSceme.filter( e => {
+          if (e.type === 'link') {
+            const url = [window.location.origin, e.url].join('/');
+            const params = (new URL(url)).searchParams.toString();
```
GuySartorelli, line 205: This is still not retaining any GET params in the URL before sorting.
So if you start off at `http://my-local.dev/admin/gridfield/?some-var=some-val` the `some-var=some-val` won't be kept.
We need to make sure we keep the `window.location.search` params like so:
```suggestion
            const url = [window.location.origin, e.url].join('/');
            const params = [window.location.search, url.searchParams.toString()].join('&').substring(1);;
```
Note that if `window.location.search` has a value, it will include a `?` at the start, but if it _doesn't_ have a value, the result of the `join` will have a `&` at the start. So either way we need to trim off the first character of `params` which is what the `substring` is there for.

GuySartorelli (changes requested):
This PR should target the `1` branch - it doesn't make any sense to release this separately from https://github.com/silverstripe/silverstripe-framework/pull/10331

The UX advice we received indicated the following behaviour for the history state:
- If you navigate to a `ModelAdmin` or page that has a gridfield and then immediately click 'back' in your browser, you should go back to the previous page you were on.
- If you navigate to a `ModelAdmin` or page that has a gridfield and then change the state of that gridfield (either one time or many times e.g. sort, filter, paginate a couple of times), when you click 'back' in your browser you should go back to the _initial default state_ of the gridfield. Pressing 'back' again will then take you back to the page you were previously on.

The first of those is met, but the second is not.

client/src/legacy/GridField.js
sabina-talipova: DONE

client/src/legacy/GridField.js
```diff
@@ -19,9 +19,13 @@ $.entwine('ss', function($) {
...
         this.injectSearchButton(false);
       }
+      
+      if (this.hasFilters()) {
+        this.injectSearchButton(true);
+      };
 
       if (this.is('.grid-field--lazy-loadable') && (
-        (this.closest('.ss-tabset').length === 0) || (this.data('gridfield-lazy-load-state') === 'force') )
+        (this.closest('.ss-tabset, .cms-tabset').length === 0) || (this.data('gridfield-lazy-load-state') === 'force') )
```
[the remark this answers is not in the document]
sabina-talipova: Sorry, probably it appeared after rebase. I can remove it, but it exist in branch 1 already, just in case.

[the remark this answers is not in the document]
GuySartorelli: Yeah in the `1` branch it's there to support a change in lazy-loading behaviour. When you retarget the `1` branch it will be there but won't count as a change so should be fine once you swap to targetting `1`.

code/SecurityAdmin.php
```diff
@@ -274,7 +275,11 @@ public function GroupImportForm()
      */
     public function Backlink()
     {
-        return false;
+        $params = $this->getRequest()->getVars();
+        return Controller::join_links(
+            $this->Link(),
+            '?' . http_build_query($params ?? [])
+        );
```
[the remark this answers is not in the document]
sabina-talipova: It doesn't work if we are on the second or next levels. It works for Breadcrumb link for Company, but doesn't work for Breadcrumb link for Employees in particular Company.

client/src/legacy/GridField.js
sabina-talipova: I add params after request to URL (history.state) only to provide ability to share this result (link) and navigate by using Browser's buttons.

sabina-talipova: We cannot keep custom params in URL from POST request.

code/SecurityAdmin.php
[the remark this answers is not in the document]
GuySartorelli: It should work if you apply the recommended change in framework: https://github.com/silverstripe/silverstripe-framework/pull/10331#pullrequestreview-1050165675

client/src/legacy/GridField.js
GuySartorelli: > to provide ability to share this result (link) and navigate by using Browser's buttons.

Yes, that's exactly my point. When navigating back/forward using the browser's back/forward buttons, we should keep any custom GET variables in the query string that have been added through project code, as it may be important for custom functionality.

> We cannot keep custom params in URL from POST request.

That's absolutely fine. The recommended change _won't_ keep custom POST params in the URL, nor would we ever want to.

tests/behat/features/gridfield-search.feature
```diff
@@ -4,21 +4,53 @@ Feature: Search in GridField
...
+    Given the "Company" "Walmart" with "Category"="Retail"
+    And the "Employee" "Alen" with "Company"="1"
+    And the "Employee" "Bill" with "Company"="1"
+    And the "Employee" "Ford" with "Company"="1"
+    And the "Company" "ExxonMobil" with "Category"="Oil"
+    And the "Company" "Test" with "Category"="Other" 
     And I am logged in with "ADMIN" permissions
     And I go to "/admin/test"
-    
+ @run
```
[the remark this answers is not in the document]
sabina-talipova: DONE

tests/behat/features/gridfield-search.feature
```diff
@@ -4,21 +4,53 @@ Feature: Search in GridField
...
     And I press the "Enter" key in the "SearchBox__Name" field
     Then I should see "Walmart" in the "#Form_EditForm" element
-    But I should not see "ExxonMobil" in the "#Form_EditForm" element
-    And I should not see "Test" in the "#Form_EditForm" element
+    But I should not see "ExxonMobil" in the ".col-Name" element
+    And I should not see "Test" in the ".col-Name" element
     And I click "Walmart" in the "#Form_EditForm" element
     Then I should see "Walmart"
     And I should see "Walmart" in the ".breadcrumbs-wrapper" element
+@run2
```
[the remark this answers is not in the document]
sabina-talipova: DONE

client/src/legacy/GridField.js
```diff
@@ -180,6 +194,36 @@ $.entwine('ss', function($) {
       if (hasLegacyFilterHeader) {
         this.find('.sortable-header th:last').html(content);
       }
+    },
+
+    keepStateInHistory: function() {
+      const newURLScheme = $(this).find('.gridfield-actionmenu__container').data('schema');
```
GuySartorelli, line 200:
```suggestion
      const newURLSchema = $(this).find('.gridfield-actionmenu__container').data('schema');
```
[URL Scheme](https://en.wikipedia.org/wiki/List_of_URI_schemes) means something specific - we're dealing with schema here.

tests/behat/features/gridfield-search.feature
```diff
@@ -4,21 +4,53 @@ Feature: Search in GridField
...
     When I press the "Open search and filter" button
     And I press the "Advanced" button
     Then I should see a "#Form_CompaniesSearchForm_Search_Name.no-change-track" element
     And I fill in "SearchBox__Name" with "Walmart"
     And I press the "Enter" key in the "SearchBox__Name" field
     Then I should see "Walmart" in the "#Form_EditForm" element
-    But I should not see "ExxonMobil" in the "#Form_EditForm" element
-    And I should not see "Test" in the "#Form_EditForm" element
+    But I should not see "ExxonMobil" in the ".col-Name" element
+    And I should not see "Test" in the ".col-Name" element
```
GuySartorelli, line 24: There are multiple `.col-Name` elements and this probably only checks the first one. Check for it anywhere on the page instead.
"Test" is a hard one since it exists in the side admin area and breadcrumbs - maybe pick something more unique? I've suggested "Test Company" below (you'll need to change that elsewhere in this feature as well)
```suggestion
    But I should not see "ExxonMobil"
    And I should not see "Test Company"
```

GuySartorelli: Please also do the same for the other `.col-Name` references below.

client/src/legacy/GridField.js
```diff
@@ -180,6 +194,36 @@ $.entwine('ss', function($) {
...
+      const params = [window.location.search, (new URL(link)).searchParams.toString()].join('&').substring(1);
+      const newrequest = [];
+      const reqString = [];
+      params.split('&').forEach(param => {
+        const newVal = param.split('=');
+        newrequest[newVal[0]] = newVal[1] ? newVal[1] : '';
+      });
+      Object.keys(newrequest).forEach(param => {
+        reqString.push([param, newrequest[param]].join('='));
+      });
```
GuySartorelli, line 224: Can you please explain what this is doing? We should already have a valid query string before you do any of this.

GuySartorelli (changes requested):
Please also add behat tests for navigating back/forward in the browser.
This can be done with `When I move backward one page` and `When I move forward one page`

Also, have you addressed my earlier comment about the UX advice not being implemented correctly?

client/src/legacy/GridField.js
sabina-talipova: This just clean up request string from identical keys and leaves only new one. Otherwise, we just concatenate new state for some key to existing GET request that already has this key. 
Like this : ?**test**=3&**gridState-SilverStripe-FrameworkTest-Model-Company-0**=%7B"GridFieldSortableHeader"%3A%7B"SortColumn"%3A"Name"%7D%7D&**gridState-SilverStripe-FrameworkTest-Model-Company-0**=%7B"GridFieldSortableHeader"%3A%7B"SortColumn"%3A"Name"%2C"SortDirection"%3A"desc"%7D%7D

tests/behat/features/gridfield-search.feature
sabina-talipova: I run behat tests in my local and it didn't work. Can do tests again.

client/src/legacy/GridField.js
sabina-talipova: `newSchema` ?

client/src/legacy/GridField.js
GuySartorelli: Probably just add a comment above this that explains you're removing duplicate entries. It's a bit hard to understand that at a glance.

sabina-talipova:
> 

Could I clarify behavior for "Forward" button in browser? What is your expectations?

GuySartorelli:
> Could I clarify behavior for "Forward" button in browser? What is your expectations?

If you click the "back" button in the browser, and then click the "forward" button in the browser, you should be looking at the exact same page with the same state as you had before you clicked "back".

sabina-talipova:
> If you click the "back" button in the browser, and then click the "forward" button in the browser, you should be looking at the exact same page with the same state as you had before you clicked "back".Z

Sorry, Guy, but I completely disagree with this approach.
We don't have any navigation tools for GridField list of items. 
Just imagine, user sorted and filtered list and now he is on some page of results. Then he accidentally or somehow clicks "Back" button and lost his results. Then he decided to use "Forward" button to come back to his previous state, but he stay on the same page.
If you test Google, how it works. It keeps each step in the history to provide users with ability to navigate back and forward.
Probably, better solution to provide "Clear" button for sorting and filtering.

client/src/legacy/GridField.js
sabina-talipova: DONE

GuySartorelli:
Just putting this here for the sake of anyone coming across this PR in the future:
We had a follow-up UX session and the following advice came out of it (note that "page" here should not be interpreted to mean pagination):

- Clicking back in the browser should take you to the _previous page_ you were on
- Clicking forward in the browser should take you forward one page

i.e:
1. I am at /admin/pages
2. I navigated to some model admin which has a gridfield
3. I paginate through, sort, or filter the gridfield
4. I click back - this takes me back to /admin/pages
5. I click forward - this takes me to the model admin gridfield with whatever sort/pagination/filtering I had applied before step 4.

tests/behat/features/gridfield-search.feature
sabina-talipova: I returned `And I should not see "Test" in the ".col-Name" element `, since `And I should not see "Test Company"` fails in GitHub actions.

client/src/legacy/GridField.js
```diff
@@ -180,6 +199,44 @@ $.entwine('ss', function($) {
...
+          if (e.type === 'link') {
+            const reqString = this.buildURLString(e.url)
+            const searchParam =  reqString ? '?' + reqString.join('&') : '';
+            const historyState = $.extend({}, { page: window.location.pathname + searchParam }, this.getState());
+            history.replaceState(historyState, '', window.location.pathname + searchParam);
+          }
+        })
+      }
+    },
+    /**
```
GuySartorelli, line 217:
```suggestion

    /**
```

GuySartorelli (changes requested):
One _very_ minor change to make.

client/src/legacy/GridField.js
```diff
@@ -180,6 +199,45 @@ $.entwine('ss', function($) {
...
+
+    keepStateInHistory: function() {
+      const newSchema = $(this).find('.gridfield-actionmenu__container').data('schema');
+      if (newSchema && newSchema.length > 0) {
+        newSchema.filter( e => {
+          if (e.type === 'link') {
+            const reqString = this.buildURLString(e.url)
+            const searchParam =  reqString ? '?' + reqString.join('&') : '';
+            const historyState = $.extend({}, { page: window.location.pathname + searchParam }, this.getState());
+            history.replaceState(historyState, '', window.location.pathname + searchParam);
```
GuySartorelli, line 212:
```suggestion
            window.ss.router.replace(window.location.pathname + searchParam, undefined, undefined, false);
```
Turns out we have to use the special custom page router that Silverstripe admin uses... see https://github.com/visionmedia/page.js/blob/master/page.js#L678 for what this is actually doing.

client/src/legacy/GridField.js
```diff
@@ -29,6 +33,11 @@ $.entwine('ss', function($) {
       }
 
       this.data('gridfield-lazy-load-state', 'ready');
+
+      $(window).on('popstate', function (e) {
+        if (e.originalEvent.state === null) { return; }
+        location.reload();
+      });
```
GuySartorelli, line 40: This is what's causing the behat failures.
```suggestion
```

GuySartorelli (changes requested):

GuySartorelli (approved):

Issue #1332, opened 2022-07-29
Regressions caused by new error handling in LeftAndMain


GuySartorelli:
`LeftAndMain` handles errors differently after https://github.com/silverstripe/silverstripe-admin/pull/1317 was merged.
This appeared to work fine but is causing regressions as detailed below.
Some regressions were resolved by https://github.com/silverstripe/silverstripe-admin/pull/1331 but the below are still happening

https://github.com/silverstripe/silverstripe-mfa/runs/7573617039?check_suite_focus=true
Multiple failures wherein the new error output doesn't contain some strings that the tests expect

https://github.com/silverstripe/silverstripe-security-extensions/runs/7573614718?check_suite_focus=true
> SilverStripe\SecurityExtensions\Tests\Control\SudoModeControllerTest::testActivateFailsWithGetRequest
TypeError: SilverStripe\Admin\LeftAndMain::setHttpErrorMessage(): Argument `#1` ($message) must be of type string, null given, called in /home/runner/work/silverstripe-security-extensions/silverstripe-security-extensions/vendor/silverstripe/admin/code/AdminErrorExtension.php on line 21

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1334

PR #1334, opened 2022-07-31
FIX Don't overwrite JSON errors with templated error page

GuySartorelli:
The new error handling logic was overwriting JSON error messages, which are most likely responses for AJAX requests. A more complete fix would be to find out why they weren't being detected as AJAX requests and resolve that - but that could lead to even more regressions, and would be time consuming besides.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1332

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -780,7 +780,10 @@ class LeftAndMain extends Controller implements PermissionProvider
 
     public function afterHandleRequest()
     {
-        if ($this->response->isError() && !$this->request->isAjax()) {
+        if ($this->response->isError()
+            && !$this->request->isAjax()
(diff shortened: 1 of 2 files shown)


emteknetnz (approved):
This seems appropriate as we use the "hardcoded" string `'application/json'` elsewhere in LeftAndMain.js to denote XHR requests

`HTTPRequest::isAjax()` looks like a sub-optimal function as it requires setting querystring or header values client side, instead of a smarter automatic detection, so likely it's giving the wrong result in many instances

**HTTPRequest.php**
```php
    public function isAjax()
    {
        return (
            $this->requestVar('ajax') ||
            $this->getHeader('x-requested-with') === "XMLHttpRequest"
        );
    }
```

Issue #1347, opened 2022-08-25
Tidy up array syntax


GuySartorelli:
For some reason there are a _lot_ of instances of `array()` instead of `[]` in this module.

## Acceptance criteria
- [x] All instances of `array()` are replaced with `[]`
- [x] phpcs config is updated to check for this.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1348

maxime-rainville:
Done.

PR #1348, opened 2022-08-25
ENH Use short array syntax

GuySartorelli:
This has been done on various other modules already - looks like this one was missed at the time.
The phpcs rule was taken from cms - IMO it makes a lot of sense to add this to linting just to keep things tidy. It's part of PSR12 so we'll likely globally adopt it anyway.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1347


maxime-rainville (approved):
LGTM

PR #1352, opened 2022-08-28
DEP Update thirdparty jQuery libaries

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1351

Issue #1351, opened 2022-08-28
Update thirdparty jQuery libraries


emteknetnz:
There are unmanaged jquery libraries in the thirdparty directories such as jquery-entwine.

## ACs
- [x] jquery libraries in the thirdparty directories need to be updated to work with jQuery 3.6.0+
- [x] Behat on silverstripe/installer is green

## Stretch goal
- [ ] Migrate from thirdparty folder to be regular npm dependencies that get built with `yarn build`

## Notes
- This need to be targetted to the `4` branch and will be released in `4.12`

### Previous POC PRs - for reference only, start a new PR for the actual work
- https://github.com/silverstripe/silverstripe-admin/pull/1352 (can probably ignore this)
- https://github.com/silverstripe/silverstripe-admin/pull/1353 (a slightly different approach)
- https://github.com/silverstripe/silverstripe-admin/pull/1357 (most recent attempt)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1375
- https://github.com/silverstripe/silverstripe-framework/pull/10524
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1298
- https://github.com/silverstripe/silverstripe-cms/pull/2792
- https://github.com/silverstripe/developer-docs/pull/108

GuySartorelli:
I have spun off some of the jQuery library upgrades as separate cards since they can (and in some cases must) be done separately:
- https://github.com/silverstripe/silverstripe-admin/issues/1376
- https://github.com/silverstripe/silverstripe-admin/issues/1377

GuySartorelli:
Note: While working on this I have also run into existing console errors that were caused by the following issues
- https://github.com/silverstripe/silverstripe-elemental/issues/950
- https://github.com/silverstripe/silverstripe-admin/issues/1385

emteknetnz:
Linked PRs have been merged

PR #1353, opened 2022-08-29
ENH Upgrade jquery to 1.12

GuySartorelli:
A slightly different approach from https://github.com/silverstripe/silverstripe-admin/pull/1352

- This approach has updated jquery UI from the thirdparty dir, but has completely removed jquery itself from there in favour of installing it via yarn.
- There are various small changes in src javascript code to resolve bugs - they are not all the best resolutions, this was just a POC to see how far I could get.
- I tried updating entwine from https://github.com/hafriedlander/jquery.entwine but I didn't see any new issues arise or existing issues resolved from doing so - so I reverted back to what was already in admin.
- I manually removed some code from entwine which was causing problems - that code is in both the original and updated versions so updating wouldn't resolve that part.

There are comments in files changed about why some changes were made.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1351


client/src/legacy/LeftAndMain.ActionTabSet.js
```diff
@@ -124,7 +124,8 @@ $.entwine('ss', function($) {
           containerSouth = activePanel.parents('.toolbar--south');
           if (containerSouth){
             // If container is the southern panel, make tab appear from the top of the container
-            padding = activeTab.offset().top - containerSouth.offset().top;
+            const offset = activeTab.offset();
+            padding = offset ? (offset.top - containerSouth.offset().top) : 0;
```
ChloeSartorelli, line 128: New tabs are now an empty object - there is no offset() function on them - so this just checks if there was an offset or not to avoid "that is undefined" error.

client/src/legacy/LeftAndMain.js
```diff
@@ -221,17 +221,18 @@ $.entwine('ss', function($) {
...
+      //   $('.ss-loading-screen').append(
+      //     '<p class="ss-loading-incompat-warning"><span class="notice">' +
+      //     'Your browser is not compatible with the CMS interface. Please use Internet Explorer 8+, Google Chrome or Mozilla Firefox.' +
+      //     '</span></p>'
+      //   ).css('z-index', $('.ss-loading-screen').css('z-index')+1);
+      //   $('.loading-animation').remove();
+
+      //   this._super();
+      //   return;
+      // }
```
ChloeSartorelli, line 235: This will need to be replaced with some more appropriate third-party solution - for now I removed it because jquery doesn't have a browser property anymore.

thirdparty/jquery-entwine/dist/jquery.entwine-dist.js
```diff
@@ -748,35 +748,6 @@ Sizzle is good for finding elements for a selector, but not so good for telling
...
-					this.removeEventListener(original, $.event.special[fix].handler, true);
-				},
-				handler: function(e){
-					arguments[0] = $.event.fix(e);
-					arguments[0].type = fix;
-					return $.event.handle.apply(this, arguments);
-				}
-			};
-		});
-	}
```
GuySartorelli, line 779: `$.support` is deprecated and `$.browser` doesn't exist - and anyway we aren't interested in supporting such old browsers so the easy fix to the errors this caused was to remove it.

ChloeSartorelli, line 779: `$.support` is deprecated and `$.browser` doesn't exist - and anyway we aren't interested in supporting such old browsers so the easy fix to the errors this caused was to remove it.

thirdparty/jquery-entwine/dist/jquery.entwine-dist.js
```diff
@@ -1465,23 +1436,6 @@ catch (e) {
...
-		if (el.attachEvent) {
-			var isSupported = (eventName in el);
-			if (!isSupported) {
-				el.setAttribute(eventName, "return;");
-				isSupported = typeof el[eventName] === "function";
-			}
-
-			$.support.changeBubbles = isSupported;
-		}
-	}
```
ChloeSartorelli, line 1484: `$.support` is deprecated - and anyway we aren't interested in supporting such old browsers so the easy fix to the errors this caused was to remove it.

GuySartorelli:
Closing in favor of https://github.com/silverstripe/silverstripe-admin/pull/1375

Issue #1350, opened 2022-08-28
SecurityAdmin should extend ModelAdmin


maxime-rainville:
SecurityAdmin while being functionally equivalent in practically every way to a standard ModelAdmin doesn't extend ModelAdmin.

This adds extra technical debt because any fix or new feature we apply to the regular ModelAdmin needs to be replicated in SecurityAdmin.

In CMS5, we should aim to refactor SecurityAdmin to extend ModelAdmin.

## Acceptance criteria
- [x] SecurityAdmin is refactored to extend ModelAdmin and remove redundant logic in CMS5
- [x] It is still possible to import Group and Members
- [x] Review where the warning came from. Remove the "Caution: Removing members from this list will remove them from all groups and the database" warning or talk to Design team to find a better solution.

## Related
- https://github.com/silverstripe/silverstripe-admin/pull/963
- https://github.com/silverstripe/silverstripe-admin/pull/1359
- https://github.com/silverstripe/developer-docs/pull/80
- https://github.com/silverstripe/silverstripe-admin/issues/1302

emteknetnz:
I don't know if we'll even the background to the warning "Caution: Removing members from this list will remove them from all groups and the database" - it was added 11 years ago in Silverstripe 2.4 - https://github.com/silverstripe/silverstripe-cms/commit/033d27ac3f8d4ccae2578774d46fdf445299698f

I think it should just be removed and we do not need any design input here. It seems completely obvious that, yes, when you click the delete button on a record from the thing that manages records, that the record is then deleted :-)

GuySartorelli:
All ACs met and PRs merged.

PR #1359, opened 2022-09-06
API Convert SecurityAdmin to be a ModelAdmin

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1350

Continued from where Max got to with https://github.com/silverstripe/silverstripe-admin/pull/963 - I've included him as a  co-author in the commit message


code/MemberImportController.php
```diff
@@ -0,0 +1,59 @@
+<?php
+
+namespace SilverStripe\Admin;
+
+use SilverStripe\Control\Controller;
+use SilverStripe\Security\Permission;
+use SilverStripe\View\Requirements;
+
+class MemberImportController extends Controller
```
emteknetnz: Moved logic off SecurityAdmin to a new controller because ModelAdmin isn't allowed to have random controller endpoints on it, as every subpath on a ModelAdmin is assumed to be a managed_model

code/SecurityAdmin.php
```diff
@@ -2,316 +2,66 @@
...
-    {
-        if (!Permission::check('ADMIN')) {
-            return null;
+        // Limit import to admin since the import logic can affect assigned permissions
+        if (!Permission::check('ADMIN') || str_contains($name, 'role')) {
+            $config->removeComponentsByType(GridFieldImportButton::class);
         }
-
-        /** @var Group $group */
-        $group = $this->currentPage();
```
emteknetnz, line 238: I didn't copy this logic over MemberImportController because it always returned null, so it did nothing

emteknetnz:
Unit tests failing because symfony 6 PRs need to be merged. Re-run tests after that

code/SecurityAdmin.php
```diff
@@ -2,316 +2,66 @@
...
+        Member::class,
+        Group::class,
+        PermissionRole::class
+    ];
 
     private static $url_segment = 'security';
 
-    private static $url_rule = '/$Action/$ID/$OtherID';
-
     private static $menu_title = 'Security';
```
ChloeSartorelli, line 57:
```suggestion
    private static $menu_title = 'Security';

    private static $menu_priority = 0;
```
The menu priority is overridden in `ModelAdmin` to -0.5 which lumps all model admins together below the built-ins. We should keep it at 0 so it stays where it has always been.

code/SecurityAdmin.php
```diff
@@ -2,316 +2,66 @@
...
 /**
  * Security section of the CMS
  */
-class SecurityAdmin extends LeftAndMain implements PermissionProvider
+class SecurityAdmin extends ModelAdmin implements PermissionProvider
 {
+    private static $managed_models = [
+        Member::class,
+        Group::class,
+        PermissionRole::class
```
ChloeSartorelli:
```suggestion
        'users' => [
            'title' => 'Users',
            'dataClass' => Member::class
        ],
        'groups' => [
            'title' => 'Groups',
            'dataClass' => Group::class
        ],
        'roles' => [
            'title' => 'Roles',
            'dataClass' => PermissionRole::class
        ],
```
This gives nicer URLs and retains the old title for the Users tab.

code/SecurityAdmin.php
```diff
@@ -2,316 +2,66 @@
...
+        if (!Permission::check('ADMIN') || str_contains($name, 'role')) {
+            $config->removeComponentsByType(GridFieldImportButton::class);
         }
-
-        /** @var Group $group */
-        $group = $this->currentPage();
-        $form = new MemberImportForm($this, __FUNCTION__);
-        $form->setGroup($group);
-
+        $this->extend('updateEditForm', $form);
```
ChloeSartorelli: We can't have this here - that hook is already called when we use `parent::getEditForm()` so this would result in the same hook being called twice for the same form, which would give unexpected results (e.g. fields being added twice)

code/SecurityAdmin.php
```diff
@@ -2,316 +2,66 @@
...
 namespace SilverStripe\Admin;
 
-use SilverStripe\Control\HTTPRequest;
-use SilverStripe\Control\HTTPResponse;
-use SilverStripe\Core\Convert;
+use Exception;
 use SilverStripe\Core\Injector\Injector;
-use SilverStripe\Forms\FieldList;
-use SilverStripe\Forms\Form;
 use SilverStripe\Forms\GridField\GridField;
```
ChloeSartorelli:
```suggestion
```
Not needed anymore - see below

code/SecurityAdmin.php
```diff
@@ -2,316 +2,66 @@
...
-     *
-     * @param HTTPRequest $request
-     * @return HTTPResponse
-     */
-    public function roles($request)
-    {
-        return $this->index($request);
-    }
-
     public function getEditForm($id = null, $fields = null)
```
ChloeSartorelli: The only changes we're making here are for the gridfield config - we should use `getGridFieldConfig()` for this instead. e.g:
```php
    protected function getGridFieldConfig(): GridFieldConfig
    {
        $config = parent::getGridFieldConfig();
        // Limit import to admin since the import logic can affect assigned permissions
        if (!Permission::check('ADMIN')) {
            $config->removeComponentsByType(GridFieldImportButton::class);
            return $config;
        }

        $memberImportController = Injector::inst()->get(MemberImportController::class);
        /** @var GridFieldImportButton $importButton */
        $importButton = $config->getComponentByType(GridFieldImportButton::class);
        switch ($this->modelClass) {
            case Member::class:
                $importButton
                ->setImportIframe($memberImportController->Link('memberimport'))
                ->setModalTitle(_t(__CLASS__ . '.IMPORTUSERS', 'Import users'));
                break;
            case Group::class:
                $importButton
                ->setImportIframe($memberImportController->Link('groupimport'))
                ->setModalTitle(_t(__CLASS__ . '.IMPORTGROUPS', 'Import groups'));
                break;
            case PermissionRole::class: {
                $config->removeComponentsByType(GridFieldImportButton::class);
            }
        }
        return $config;
    }
```

code/SecurityAdmin.php
ChloeSartorelli: Note that this obviously makes the titles not localisable - we'll need to also override either `getManagedModels()` or `getManagedModelTabs()` - I've included an example of overriding `getManagedModels()` for convenience but either method will do just fine
```php
    public function getManagedModels()
    {
        $models = parent::getManagedModels();
        // Ensure tab titles can be localised
        foreach ($models as $key => $spec) {
            switch ($spec['dataClass']) {
                case Member::class:
                    $spec['title'] = _t(__CLASS__ . '.Users', 'Users');
                    break;
                case Group::class:
                case PermissionRole::class:
                    $spec['title'] = singleton($spec['dataClass'])->i18n_plural_name();
            }
        }
        return $models;
    }
```

code/MemberImportController.php
```diff
@@ -0,0 +1,59 @@
+<?php
```
ChloeSartorelli: This class, as well as the `MemberImportForm` and `GroupImportForm` classes are obsolete.
Remove these classes and their corresponding javascript/styles. Add the following to `SecurityAdmin` instead:
```php
    private static $model_importers = [
       'users' => MemberCsvBulkLoader::class,
       Member::class => MemberCsvBulkLoader::class,
       'groups' => GroupCsvBulkLoader::class,
       Group::class => GroupCsvBulkLoader::class,
    ];
```
Note: We have to add both the model tab reference _and_ the class name as keys for the importers because `ModelAdmin` currently checks for `$importers[$modelClass]` in some places and `$importers[$this->modelTab]` in others. This is a bug that we should fix in CMS 5 but up to you if it's done separately or with this PR. If you want this to be handled separately let me know and I'll create a new card for it.

Then in `getGridFieldConfig()` remove the `setImportIframe()` calls and reference to `MemberImportController`.

Finally, if you want to include the advanced usage details from the old `MemberImportForm` and `GroupImportForm` classes (e.g "Groups can be assigned by the <em>Groups</em> column. Groups are identified by their <em>Code</em> property multiple groups can be separated by comma. Existing group memberships are not cleared." etc) you can override `ImportForm()` and add them to the form that `ModelAdmin` already creates. You'll also need to add `ImportForm` as an allowed action if you do that since allowed actions are only inherited for methods that aren't overridden:
```php
    private static $allowed_actions = [
        'ImportForm',
    ];
```

GuySartorelli (changes requested):
It feels a bit suspect that there's no javascript or css being removed - is there really no double-up of javascript or css for the old SecurityAdmin that is now handled by existing ModelAdmin javascript? E.g. inside `client/src/legacy/SecurityAdmin.js` or `client/src/styles/legacy/_SecurityAdmin.scss`?

code/MemberImportController.php
emteknetnz: > If you want this to be handled separately let me know and I'll create a new card for it.

New card would be great thanks.  I'll add an inline comment in here for now explaining why there's duplicates

code/SecurityAdmin.php
emteknetnz: Removed

code/SecurityAdmin.php
```diff
@@ -3,314 +3,145 @@
...
+                . '<ul>'
+                . '<li>Allowed columns: <em>{columns}</em></li>'
+                . '<li>Existing users are matched by their unique <em>Code</em> property, and updated with any new values from '
+                . 'the imported file.</li>'
+                . '<li>Groups can be assigned by the <em>Groups</em> column. Groups are identified by their <em>Code</em> property, '
+                . 'multiple groups can be separated by comma. Existing group memberships are not cleared.</li>'
+                . '</ul>'
+                . '</div>',
+                ['columns' => $columns]
+            )
```
ChloeSartorelli: The allowed columns is already in the `ModelAdmin` import form - it feels like overkill to be replacing that with a less-standard way of displaying that same information.

From what I can tell, the only thing this _adds_ on top of what `ModelAdmin` has in the form already is the second two bullet points there which only apply to `Member` - and the first point I think isn't correct so we should probably remove that. `Member` has no `Code` property.
There's also some information missing from the `GroupImportForm` which applies only to groups.

I think it would be cleaner to add that information like this:
```php
public function ImportForm()
{
    $form = parent::ImportForm();
    if (!$form) {
        return $form;
    }
    $form->Fields()->removeByName('EmptyBeforeImport');
    $extraInfo = match ($this->modelClass) {
        Member::class => '<p>Groups can be assigned by the <em>Groups</em> column. Groups are identified by their '
            . '<em>Code</em> property, multiple groups can be separated by comma. Existing group memberships are '
            . 'not cleared.</p>',
        Group::class => '<ul><li>Existing groups are matched by their unique <em>Code</em> value, and updated with any new values from the '
        . 'imported file</li>'
        . '<li>Group hierarchies can be created by using a <em>ParentCode</em> column.</li>'
        . '<li>Permission codes can be assigned by the <em>PermissionCode</em> column. Existing permission codes are not '
        . 'cleared.</li></ul>'
    };
    if ($extraInfo) {
        $form->Fields()->insertBefore('_CsvFile', LiteralField::create('ExtraInfo', $extraInfo));
    }
    return $form;
}
```

This way `ModelAdmin` is still in charge of saying what columns are valid, and it does so in a standard format that people will be used to from everywhere else they can import CSV content. The columns mentioned are the same as what you're adding here.
It also means we can get rid of that extra javascript which is essentially just a duplication of the javascript `ModelAdmin` already has but with different selectors.


An alternative, if you'd specifically like this information to be inside the "Show specification" expandable area, is to add a `$ExtraDetails` section into the `ModelAdmin_ImportSpec.ss` template and introduce a new `protected function getExtraImportDetails()` method, which `ModelAdmin` can call to add any extra details into the template. If there are no extra details then it'll just render as it always has, but if there _are_ extra details, it can add a new section inside that expandable area to display it. That feels a lot cleaner and it also has the benefit of providing that functionality to custom `ModelAdmin` in people's projects.

client/src/legacy/MemberImportForm.js
```diff
@@ -7,7 +7,7 @@ $.entwine('ss', function($){
   /**
    * Class: .import-form .advanced
    */
-  $('.import-form .advanced').entwine({
+  $('#Form_ImportForm .advanced').entwine({
```
ChloeSartorelli: We shouldn't need this javascript at all - we should rely on what `ModelAdmin` already provides for showing hidden content instead of duplicating that functionality.

GuySartorelli (changes requested):

code/SecurityAdmin.php
```diff
@@ -3,314 +3,145 @@
...
-        Requirements::clear();
-        Requirements::javascript('silverstripe/admin: client/dist/js/vendor.js');
-        Requirements::javascript('silverstripe/admin: client/dist/js/MemberImportForm.js');
-        Requirements::css('silverstripe/admin: client/dist/styles/bundle.css');
-
-        return $this->renderWith('BlankPage', [
-            'Content' => ' ',
-            'Form' => $this->GroupImportForm()->forTemplate()
+        Requirements::javascript('silverstripe/admin:client/dist/js/MemberImportForm.js');
+        $form->Fields()->removeByName('EmptyBeforeImport');
```
ChloeSartorelli, line 91: Why are we removing that field?
If it's just because it wasn't there in `SecurityAdmin` originally I think we should keep it here.
We should only remove it if there's a very clear reason to not provide this functionality. Perhaps we could remove it for non-admins only if we are worried about implications of people clicking it when they shouldn't

code/SecurityAdmin.php
ChloeSartorelli: I've also just read up on `match` expressions - we need to add a `default` here in case someone adds import functionality to permission codes or adds their own models, or it will raise a `E_FATAL_ERROR`

code/MemberImportController.php
ChloeSartorelli: https://github.com/silverstripe/silverstripe-admin/issues/1364

code/SecurityAdmin.php
emteknetnz: > Perhaps we could remove it for non-admins only

Non-admins aren't even given the option to bulk load from CSV

It feels extremely risky to include this checkbox because if it's accidentally checked then you'll completely destroy your content auditing, as well as the ability for anyone to login.

I don't think there's a valid use case for removing ALL members.  I can't recall any feedback on this part of the CMS saying "please add in the checkbox that lets me delete all members on csv import"

ChloeSartorelli: Fair enough, I'm okay with that.

client/src/legacy/MemberImportForm.js
emteknetnz: Removed

code/SecurityAdmin.php
emteknetnz: Updated

GuySartorelli (approved):
Works great locally and the code makes sense, happy to merge.

Issue #1354, opened 2022-08-30
ModelAdmin is a little difficult to work with


GuySartorelli:
ModelAdmin is a little clunky to work with at the moment - there are a few methods that don't require any instance data but aren't static, and any time you want to get a link for a managed model tab you need to re-implement the logic in `sanitiseClassName()` - and basically build the entire link from scratch every time with all of its component parts.

## Acceptance Criteria
- [x] Methods that don't need to be instance methods should be static.
- [x] There should be an easy way to get a link for a managed model without having to reimplement `sanitiseClassName()`.
- [x] There is a straightforward way to get a canonical edit link for objects managed by ModelAdmin
- [x] The changes are documented appropriately
- [x] Documentation that currently includes re-implementing `str_replace('\\', '-', $class)` is updated

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1360
- https://github.com/silverstripe/silverstripe-frameworktest/pull/123
- https://github.com/silverstripe/developer-docs/pull/83
- https://github.com/silverstripe/silverstripe-admin/pull/1355
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/254
- https://github.com/silverstripe/developer-docs/pull/86

## Related PRs
These are an extended approach for resolving the "There is a straightforward way to get a canonical edit link for objects managed by ModelAdmin" acceptance criterion
- https://github.com/silverstripe/silverstripe-admin/pull/1361
- https://github.com/silverstripe/silverstripe-cms/pull/2778
- https://github.com/silverstripe/developer-docs/pull/85

PR #1360, opened 2022-09-12
NEW New API for managing models in ModelAdmin

GuySartorelli:
Makes it easier to manage models in a ModelAdmin, e.g. for implementing `CMSEditLink()` from the `CMSPreviewable` interface and checking if models are actually managed in the admin.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1354

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -2,9 +2,11 @@
 
 namespace SilverStripe\Admin;
 
+use InvalidArgumentException;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\BulkLoader;
@@ -155,11 +157,11 @@ abstract class ModelAdmin extends LeftAndMain
         }
 
         // security check for valid models
-        if (!array_key_exists($this->modelTab, $models ?? [])) {
+        if (!$this->isManagedModel($this->modelTab)) {
             // if it fails to match the string exactly, try reverse-engineering a classname
             $this->modelTab = $this->unsanitiseClassName($this->modelTab);
 
-            if (!array_key_exists($this->modelTab, $models ?? [])) {
+            if (!$this->isManagedModel($this->modelTab)) {
                 throw new \RuntimeException(sprintf('ModelAdmin::init(): Invalid Model class %s', $this->modelTab));
             }
         }
@@ -184,6 +186,65 @@ abstract class ModelAdmin extends LeftAndMain
     }
 
     /**
+     * Get the link for the tab of a DataObject subclass managed by this ModelAdmin.
+     *
+     * Note that if the modelClass is managed on multiple tabs and you want a link for a
+     * specific tab you should use {@link getLinkForModelTab} instead.
+     *
+     * @throws InvalidArgumentException if $modelClass is not managed by this ModelAdmin.
+     */
+    public function getLinkForModelClass(string $modelClass): string
+    {
+        if (!$this->isManagedModel($modelClass)) {
+            throw new InvalidArgumentException("$modelClass isn't managed by this ModelAdmin.");
+        }
+        return $this->getLinkForModelTab($this->getModelTabForModelClass($modelClass));
+    }
+
+    /**
+     * Get the link for a specific tab on this ModelAdmin.
+     *
+     * If you do not have multiple tabs for a given class you should use {@link getLinkForModelClass}
+     * instead.
+     *
+     * @param string $modelTab This is either the custom url segment for the tab or, if that
+     * was not not defined, it is the fully qualified class name of the managed model.
+     * In the latter case, you should prefer to use {@link getLinkForModelClass}.
+     *
+     * @throws InvalidArgumentException if $modelTab is not a tab on this ModelAdmin.
+     */
+    public function getLinkForModelTab(string $modelTab): string
+    {
+        // Don't use isManagedModel here because a subclass of a managed model may not have its own tab.
+        if (!array_key_exists($modelTab, $this->getManagedModels())) {
+            throw new InvalidArgumentException("$modelTab isn't a tab on this ModelAdmin.");
+        }
+        return $this->Link($this->sanitiseClassName($modelTab));
+    }
+
+    /**
+     * Get the link for editing an object inside this ModelAdmin.
+     *
+     * @throws InvalidArgumentException if $obj is not managed by this ModelAdmin.
+     */
+    public function getEditLinkForManagedDataObject(DataObject $obj): string
+    {
+        $modelTab = $this->getModelTabForModelClass($obj->ClassName);
+        if ($modelTab === null) {
+            throw new InvalidArgumentException("$obj->ClassName isn't managed by this ModelAdmin");
+        }
+        $link = static::join_links(
+            $this->getLinkForModelClass($obj->ClassName),
+            'EditForm/field/',
+            $this->sanitiseClassName($modelTab),
+            'item',
+            $obj->ID
+        );
+        $this->extend('updateEditLinkForDataObject', $link, $obj);
+        return $link;
+    }
+
+    /**
      * Produces an edit form that includes a default {@link \SilverStripe\Forms\GridField\GridField} for the currently
      * active {@link \SilverStripe\ORM\DataObject}. The GridField will show data from the currently active `modelClass`
      * only (see {@link self::init()}).
@@ -202,7 +263,7 @@ abstract class ModelAdmin extends LeftAndMain
         )->setHTMLID('Form_EditForm');
         $form->addExtraClass('cms-edit-form cms-panel-padded center flexbox-area-grow');
         $form->setTemplate($this->getTemplatesWithSuffix('_EditForm'));
-        $editFormAction = Controller::join_links($this->Link($this->sanitiseClassName($this->modelTab)), 'EditForm');
+        $editFormAction = Controller::join_links($this->getLinkForModelTab($this->modelTab), 'EditForm');
         $form->setFormAction($editFormAction);
         $form->setAttribute('data-pjax-fragment', 'CurrentForm');
 
@@ -374,7 +435,6 @@ abstract class ModelAdmin extends LeftAndMain
         return $form;
     }
 
-
     /**
      * You can override how ModelAdmin returns DataObjects by either overloading this method, or defining an extension
      * to ModelAdmin that implements the `updateList` method (and takes a {@link \SilverStripe\ORM\DataList} as the
@@ -389,6 +449,8 @@ abstract class ModelAdmin extends LeftAndMain
      *     return $list->filter('Archived', false);
      * }
      * </code>
+     *
+     * Note: If you override this method you may also need to override getEditLinkForManagedDataObject()
      *
      * @return \SilverStripe\ORM\DataList
      */
@@ -427,7 +489,7 @@ abstract class ModelAdmin extends LeftAndMain
                 // `getManagedModels` did not always return a `dataClass` attribute
                 // Legacy behaviour is for `ClassName` to map to `$tab`
                 'ClassName' => isset($options['dataClass']) ? $options['dataClass'] : $tab,
-                'Link' => $this->Link($this->sanitiseClassName($tab)),
+                'Link' => $this->getLinkForModelTab($tab),
                 'LinkOrCurrent' => ($tab == $this->modelTab) ? 'current' : 'link'
             ]));
         }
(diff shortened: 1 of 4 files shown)


code/ModelAdmin.php
```diff
@@ -183,6 +185,34 @@ public function Link($action = null)
         return parent::Link($action);
     }
 
+    /**
+     * Get the link for a modelTab on this ModelAdmin.
+     * @throws InvalidArgumentException if $modelTab is not managed by this ModelAdmin class.
+     */
+    public function LinkForModelTab(string $modelTab): string
```
michalkleiner: any specific reason why not
```suggestion
    public function getLinkForModelTab(string $modelTab): string
```
?

code/ModelAdmin.php
```diff
@@ -487,6 +518,37 @@ public function getManagedModels()
...
+     */
+    public function getTabForModelClass(string $modelClass): ?string
+    {
+        $managed = $this->getManagedModels();
+        if (is_subclass_of($modelClass, DataObject::class)) {
+            // Check the superclasses as well as the specifically passed-in class
+            $classes = array_reverse(ClassInfo::ancestry($modelClass));
+        } else {
+            // Most likely it was actually a tab name that was passed in, not a class name
+            $classes = [$modelClass];
```
michalkleiner: Not sure we should assume that something not subclassing DataObject should be a candidate here.

ChloeSartorelli: There's a comment there that explains why this is being done - the tab _name_ can be explicitly defined.

The scenarios are:
1. A managed model _class_ is passed in. This method returns the correct tab name.
2. A managed model _tab_ is passed in. This method returns the tab name.
3. Literally anything other than a managed model or tab is passed in. This method returns `null`.

I think it is an early holdover from when this was called from the new `LinkForModelTab` which was just `LinkForModel` and was going to accept a tab or a class into _that_ method. I can remove this part if you want - though in that case I'll probably just remove the check altogether and assume the class _is_ a subclass of `DataObject` in all cases and let it run the checks.

code/ModelAdmin.php
ChloeSartorelli: I was just keeping it similar to `Link()` - I'm happy to throw a get in front of it if you'd prefer.

ChloeSartorelli: Done

code/ModelAdmin.php
emteknetnz: > A managed model tab is passed in. This method returns the tab name.

What does it mean to pass in a `model tab` here?  As in the 'name' of the model tab? 

I'm not sure we should even have logic to support this here given the method name is `getTabForModelClass`

ChloeSartorelli: If the reason to not support that is the name of the method, I'm happy to rename it to `getTabForModel` so it's not explicitly named 'class'. If the reason to not support it anything else then I'm happy to remove that part of the function of this method.

> What does it mean to pass in a `model tab` here? As in the 'name' of the model tab?

Yes. It's the `$this->modelTab`.

tests/php/ModelAdminTest.php
```diff
@@ -270,4 +269,82 @@ public function testGetManagedModelTabs()
...
+            'admin/multi/Player',
+            $admin->getLinkForModelTab('Player')
+        );
+        $this->assertEquals(
+            'admin/multi/' . $this->sanitiseClassName(Player::class),
+            $admin->getLinkForModelTab(Player::class)
+        );
+        $this->assertEquals(
+            'admin/multi/cricket-players',
+            $admin->getLinkForModelTab('cricket-players')
```
emteknetnz, line 315: I'm finding this super confusing. I'm not sure what a "ModelTab" is

Based on these unit tests it can be a Model class name, or 'cricket-players' which I'm assuming is a url-segment.

I wouldn't call either of those a "tab"

Seems like the method name should be something more like `getLinkForClassNameOrModelUrlSegmentWhenUsedInModelAdmin()` (though obviously not this)

emteknetnz (dismissed):
I'm just confused :-)

tests/php/ModelAdminTest.php
emteknetnz: Oh never mind

`$this->modelTab = $this->getRequest()->param('ModelClass');`

:hmm:

It's been a long day :-)

code/ModelAdmin.php
ChloeSartorelli: Note that this is used for `isManagedModel()` which _does_ need to check tabs as well as classes, so if I remove that dual functionality form here I'll need to explicitly add it to that method.

michalkleiner: My first instinct was that the majority of the logic should've been in the `isManagedModel` anyway and then only have the `getTabForModel` to get the name, though I did understand that the way how the config for the tabs is constructed to allow for the title better supports the non-intuitive way. Perhaps we could have a private method normalising the config in a way we could use in both places?

A sidenote to Steve's comment — I would also ditch support for directly passing in a tab name as I'd say that's a non-standard way of defining which model classes the admin manages.

emteknetnz: Also minor thing, should change `getTabForModelClass` to `getModelTabForModelClass` to make things a little clearer

code/ModelAdmin.php
```diff
@@ -183,6 +185,34 @@ public function Link($action = null)
...
+        if (!array_key_exists($modelTab, $this->getManagedModels())) {
+            throw new InvalidArgumentException("$modelTab doesn't exist on this ModelAdmin.");
+        }
+        return parent::Link($this->sanitiseClassName($modelTab));
+    }
+
+    /**
+     * Get the link for editing an object inside this ModelAdmin.
+     */
+    public function getEditLinkForObject(DataObject $obj): string
```
emteknetnz: Should change this to `getEditLinkForDataObject(DataObject $obj)`

emteknetnz (changes requested):
I've just a good look through ModelAdmin to try to straighten up my confusion with the the concepts of ModelClass vs ModelTab.  My understanding is that ModelClass is just the regular MyDataObject::class that everyone is used to, while ModelTab is a ModelAdmin only concept where it can be:

` 'urlslug' => ['title' => "Another title", 'dataClass' => MyDataObject::class]`

OR just

`MyDataObject::class`

Hence why when dealing with "ModelTab" we need account for either passing in `MyDataObject::class` AND `'urlslug'`

If I'm getting confused it's likely that other people will be confused too, so I think we need to be careful with how we name things here

In terms of adding new public API, I think we should only be adding things that deal with ModelClass as that's vastly more intuitive.  For instance I don't think that we should be adding `getLinkForModelTab()`

Other classes such as DataQuery and GridField use "ModelClass"

Because of this I think we should change:

`getLinkForModelTab(string $modelTab)` to `getLinkForModelClass(string $modelClass)`

And try and remove `getTabForModelClass`, or at least make it a private method.

code/ModelAdmin.php
ChloeSartorelli: I don't think that's a necessary change given it has the typehint to clearly indicate what should be passed in - but ultimately it makes little difference so I'll make this change.

GuySartorelli:
I think `getLinkForModelTab()` is important to keep, because it is possible to manage the same class in multiple tabs, and it may be that on one tab you manage one sub-group of that class based on some filter, and in another tab you manage another sub-group based on some other filter - if you want a link to a _specific_ tab then without `getLinkForModelTab()` you have to make that link yourself, which is a situation this PR is trying to avoid.

I definitely agree that _most_ of the API should be based on the model _class_ though because that's _usually_ what people will be dealing with. I'll add an additional `getLinkForModelClass()` _along side_ `getLinkForModelTab()`.
The unit tests will give a clear example of the difference between the two, where the `Player` class has three tabs - two have custom url slugs and one does not. `getLinkForModelClass()` just picks the first one sequentially so if you wanted one of the other two tabs without `getLinkForModelTab()` it's a bit of a pain.

GuySartorelli:
I have:
- renamed `getTabForModelClass()` to `getModelTabForModelClass()` and I have also made it protected
- changed `getModelTabForModelClass()` to explicitly only accept a class name of a `DataObject` subclass
- changed the logic in `isManagedModel()` to explicitly check against model tabs and model classes
- added a new `getLinkForModelClass()` to be used more often than `getLinkForModelTab()` and have documented when to use each one in the PHPDocs - I do think both are needed for reasons I've discussed above
- renamed `getEditLinkForObject()` to `getEditLinkForDataObject()`

The documentation in https://github.com/silverstripe/developer-docs/pull/83 has also been updated to reflect these changes and may make it a little clearer why I think we need `getModelTabForModelClass()` to be at least `protected` (not `private`) and why I think we need to keep `getLinkForModelTab()`

code/ModelAdmin.php
```diff
@@ -183,6 +185,55 @@ public function Link($action = null)
...
+        }
+        return $this->getLinkForModelTab($this->getModelTabForModelClass($modelClass));
+    }
+
+    /**
+     * Get the link for a specific tab on this ModelAdmin.
+     *
+     * If you do not have multiple tabs for a given class you should use {@link getLinkForModelClass}
+     * instead.
+     *
```
emteknetnz, line 209: The whole "modelTab" concept it pretty confusing, I think it's worth clarifying in the method comment that modelTab can either be:
- the url-slug of the modelTab - the reason why this method exists
- the modelClass of the modelTab - and yes this is functionally the same as getLinkForModelClass()

code/ModelAdmin.php
```diff
@@ -183,6 +185,55 @@ public function Link($action = null)
...
+
+    /**
+     * Get the link for editing an object inside this ModelAdmin.
+     */
+    public function getEditLinkForDataObject(DataObject $obj): string
+    {
+        $link = static::join_links(
+            $this->getLinkForModelClass($obj->ClassName),
+            'EditForm/field/',
+            $this->sanitiseClassName($this->getModelTabForModelClass($obj->ClassName)),
```
emteknetnz: `getModelTabForModelClass()` can return null.

I think it's worth doing something like that at the top of the method:

```php
$modelTab = $this->getModelTabForModelClass($obj->ClassName);
if (is_null($modelTabl) {
    throw new InvalidArgumentException(sprintf('There is no ModelTab for ModelClass %s', $obj->ClassName));
}
```

code/ModelAdmin.php
```diff
@@ -183,6 +185,55 @@ public function Link($action = null)
...
+            'item',
+            $obj->ID
+        );
+        $this->extend('updateEditLinkForDataObject', $link, $obj);
```
emteknetnz, line 243: How come this new method gets an extension hook and the others don't?

emteknetnz (changes requested):

code/ModelAdmin.php
ChloeSartorelli: An exception like this is already thrown by `getLinkForModelClass()` which is called from here, but it makes sense to add an explicit exception here too.

code/ModelAdmin.php
ChloeSartorelli: This one is added to support the additional work in https://github.com/silverstripe/silverstripe-admin/pull/1361 for making CMS edit links easier to deal with - it allows developers to modify the edit link returned by a `ModelAdmin` that isn't directly part of their codebase.

I didn't add one to the others because I didn't have a specific use-case for such an extension in mind - and they feel more like the `Link()` method, which itself doesn't have a hook.

If it comes down to either removing this or adding them to the other methods for consistency I'd choose removing this one out of those options.

code/ModelAdmin.php
ChloeSartorelli: Done

code/ModelAdmin.php
ChloeSartorelli: Added messaging to the phpdoc

code/ModelAdmin.php
ChloeSartorelli: Added the exception

code/ModelAdmin.php
emteknetnz: OK cool, leave this one in

code/ModelAdmin.php
```diff
@@ -183,6 +185,65 @@ public function Link($action = null)
...
+        }
+        return $this->Link($this->sanitiseClassName($modelTab));
+    }
+
+    /**
+     * Get the link for editing an object inside this ModelAdmin.
+     *
+     * @throws InvalidArgumentException if $obj is not managed by this ModelAdmin.
+     */
+    public function getEditLinkForDataObject(DataObject $obj): string
```
emteknetnz: Just wondering, would it make sense to change the signature in getEditLinkForDataObject to the following, and get rid of the extension:

```suggestion
    public function getEditLinkForDataObject(DataObject $obj, string $modelTab = ''): string
```

That could be another way to resolve the issue of the multiple tabs managing the same dataobject.  I think it would be more intuitive to put in the url-slug on the model tab than manually updating the edit link in via an extension class

emteknetnz (changes requested):
Looks good, have tested locally

Just wondering if we should get rid of the extension hook and add an optional param instead

code/ModelAdmin.php
ChloeSartorelli: I think that boils down to whether the admin owns the concern or the object owns the concern.
From my point of view it should be the admin owning the concern - everything else about how the admin handles putting or pulls objects into/from different tabs is defined in the admin itself.
If the admin has multiple tabs with different filters, it should be the admin which says "The link for that object is in tab A, _not_ tab B". The object shouldn't care IMO

ChloeSartorelli: Steve and I have talked about this offline - I'm renaming this to `getEditLinkForManagedDataObject()`.
This is a little clearer here but it makes the method added in https://github.com/silverstripe/silverstripe-admin/pull/1361 a lot more comprehensible (I'll rename it to match).

emteknetnz (approved):

PR #1361, opened 2022-09-13
NEW Add an extension to dynamically generate edit URLs

GuySartorelli:
Currently getting an edit link for an object is a bit of a pain - you have to generate a new link based on the predictable gridfield or other admin edit paths. This is partly improved by https://github.com/silverstripe/silverstripe-admin/pull/1360 - but that still doesn't resolve the problem for nested setups (e.g. object in a page's gridfield).

## NOTE
CI won't go green until https://github.com/silverstripe/silverstripe-admin/pull/1360 is merged in as the new tests rely on `ModelAdmin` having a `getEditLinkForManagedDataObject()` method

## Related issue
- https://github.com/silverstripe/silverstripe-admin/issues/1354

### DIFF
--- a/code/CMSEditLinkExtension.php
+++ b/code/CMSEditLinkExtension.php
@@ -1,0 +1,149 @@
+<?php
+
+namespace SilverStripe\Admin;
+
+use LogicException;
+use SilverStripe\CMS\Controllers\CMSMain;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\Director;
+use SilverStripe\Core\Extension;
+use SilverStripe\Forms\FieldList;
+use SilverStripe\Forms\GridField\GridField;
+use SilverStripe\Forms\GridField\GridFieldDetailForm;
+use SilverStripe\ORM\DataObject;
+
+/**
+ * An extension that automatically generates a CMS edit link for DataObjects even if
+ * they are canonically edited in some nested {@link GridField}.
+ * Designed to be used in conjunction with the {@link CMSPreviewable} interface.
+ *
+ * For nested relations (e.g. a DataObject managed in a GridField of another DataObject)
+ * you can apply this extension to both the parent and the child object and the links
+ * will chain down the nested `GridField`s to the root cms_edit_owner.
+ *
+ * You must set a cms_edit_owner config variable which defines the cms edit
+ * owner for this class.
+ * e.g. set this to a {@link LeftAndMain} class:
+ * private static string cms_edit_owner = MyModelAdmin::class;
+ * or to a has_one relation:
+ * private static string cms_edit_owner = 'Parent';
+ *
+ * Note that the cms_edit_owner must implement a getCMSEditLinkForManagedDataObject() method.
+ *
+ * If the cms_edit_owner is a has_one relation, the class on the other end
+ * of the relation must have a CMSEditLink() method.
+ */
+class CMSEditLinkExtension extends Extension
+{
+    private static string $cms_edit_owner = '';
+
+    /**
+     * Get the ModelAdmin, LeftAndMain, or DataObject which owns this object for CMS editing purposes.
+     *
+     * @return LeftAndMain|DataObject|null
+     */
+    public function getCMSEditOwner()
+    {
+        $ownerType = $this->owner->config()->get('cms_edit_owner');
+        if (is_subclass_of($ownerType, LeftAndMain::class)) {
+            return $ownerType::singleton();
+        }
+        return $this->owner->getComponent($ownerType);
+    }
+
+    /**
+     * Get the link for editing an object from the CMS edit form of this object.
+     * @throws LogicException if a link cannot be established
+     * e.g. if the object is not in a has_many relation or not edited inside a GridField.
+     */
+    public function getCMSEditLinkForManagedDataObject(DataObject $obj, string $reciprocalRelation): string
+    {
+        $fields = $this->owner->getCMSFields();
+        $link = $this->getCMSEditLinkForRelation($this->owner->hasMany(false), $obj, $reciprocalRelation, $fields);
+        if (!$link) {
+            throw new LogicException('Could not produce an edit link for the passed object.');
+        }
+        return $link;
+    }
+
+    /**
+     * Get a link to edit this DataObject in the CMS.
+     */
+    public function CMSEditLink(): string
+    {
+        $owner = $this->owner->getCMSEditOwner();
+        if (!$owner || !$owner->exists()) {
+            return '';
+        }
+
+        if (!$owner->hasMethod('getCMSEditLinkForManagedDataObject')) {
+            throw new LogicException('The cms_edit_owner must implement getCMSEditLinkForManagedDataObject()');
+        }
+
+        if ($owner instanceof DataObject) {
+            $relativeLink = $owner->getCMSEditLinkForManagedDataObject($this->owner, $this->owner->config()->get('cms_edit_owner'));
+        } else {
+            $relativeLink = $owner->getCMSEditLinkForManagedDataObject($this->owner);
+        }
+        return Director::absoluteURL($relativeLink);
+    }
+
+    private function getCMSEditLinkForRelation(array $componentConfig, DataObject $obj, string $reciprocalRelation, FieldList $fields): string
+    {
+        $candidate = null;
+        foreach ($componentConfig as $relation => $class) {
+            // Check for dot notation being used to explicitly mark the reciprocal relation.
+            $remoteField = null;
+            if (strpos($class ?? '', '.') !== false) {
+                list($class, $remoteField) = explode('.', $class ?? '');
+            }
+
+            // We're only interested in relations to the $obj class.
+            if (!is_a($obj, $class)) {
+                continue;
+            }
+
+            if ($remoteField) {
+                if ($remoteField === $reciprocalRelation) {
+                    // We've found a direct reciprocal relation, so this is definitely correct.
+                    if ($this->relationIsEditable($relation, $fields)) {
+                        return $this->constructLink($relation, $obj->ID);
+                    }
+                    // If the relation isn't in a gridfield, we have no link for it.
+                    return '';
+                }
+                // We're not interested in unrelated relations.
+                continue;
+            }
+
+            // Check for relations that have gridfields we can build a link from.
+            if ($this->relationIsEditable($relation, $fields)) {
+                $candidate = $relation;
+            }
+        }
+
+        // Only do this if we didn't find a direct reciprocal relation.
+        return $candidate ? $this->constructLink($candidate, $obj->ID) : '';
+    }
+
+    private function relationIsEditable(string $relation, FieldList $fields): bool
+    {
+        $field = $fields->dataFieldByName($relation);
+        return $field
+            && $field instanceof GridField
+            && $field->getConfig()->getComponentByType(GridFieldDetailForm::class);
+    }
+
+    private function constructLink(string $relation, int $id): string
+    {
+        $ownerType = $this->owner->config()->get('cms_edit_owner');
+        $prefix = is_a($ownerType, CMSMain::class, true) ? 'field' : 'ItemEditForm/field';
+        return Controller::join_links(
+            $this->owner->CMSEditLink(),
+            $prefix,
+            $relation,
+            'item',
+            $id
+        );
+    }
+}
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -227,7 +227,7 @@ abstract class ModelAdmin extends LeftAndMain
      *
      * @throws InvalidArgumentException if $obj is not managed by this ModelAdmin.
      */
-    public function getEditLinkForManagedDataObject(DataObject $obj): string
+    public function getCMSEditLinkForManagedDataObject(DataObject $obj): string
     {
         $modelTab = $this->getModelTabForModelClass($obj->ClassName);
         if ($modelTab === null) {
@@ -450,7 +450,7 @@ abstract class ModelAdmin extends LeftAndMain
      * }
      * </code>
      *
-     * Note: If you override this method you may also need to override getEditLinkForManagedDataObject()
+     * Note: If you override this method you may also need to override getCMSEditLinkForManagedDataObject()
      *
      * @return \SilverStripe\ORM\DataList
      */
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -538,11 +538,16 @@ abstract class ModelAdmin extends LeftAndMain
 
         // Normalize models to have their model class in array key
         foreach ($models as $k => $v) {
+            // No custom tab url segment
             if (is_numeric($k)) {
                 $models[$v] = ['dataClass' => $v, 'title' => singleton($v)->i18n_plural_name()];
                 unset($models[$k]);
+            // Custom title but no custom tab url segment
             } elseif (is_array($v) && !isset($v['dataClass'])) {
                 $models[$k]['dataClass'] = $k;
+            // Custom tab url segment but no custom title
+            } elseif (is_a($v, DataObject::class, true)) {
+                $models[$k] = ['dataClass' => $v, 'title' => singleton($v)->i18n_plural_name()];
             }
         }
 
--- a/tests/php/CMSEditLinkExtensionTest.php
+++ b/tests/php/CMSEditLinkExtensionTest.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace SilverStripe\Admin\Tests;
+
+use LogicException;
+use SilverStripe\Admin\Tests\CMSEditLinkExtensionTest\BasicNestedObject;
+use SilverStripe\Admin\Tests\CMSEditLinkExtensionTest\ManagedDataObject;
+use SilverStripe\Admin\Tests\CMSEditLinkExtensionTest\CMSEditModelAdmin;
+use SilverStripe\Admin\Tests\CMSEditLinkExtensionTest\NestedObject;
+use SilverStripe\Admin\Tests\CMSEditLinkExtensionTest\PolymorphicNestedObject;
+use SilverStripe\Dev\SapphireTest;
+
+class CMSEditLinkExtensionTest extends SapphireTest
+{
+    protected static $fixture_file = 'CMSEditLinkExtensionTest.yml';
+
+    protected $usesDatabase = true;
+
+    protected static $extra_dataobjects = [
+        ManagedDataObject::class,
+        BasicNestedObject::class,
+        NestedObject::class,
+        PolymorphicNestedObject::class,
+    ];
+
+    protected static $extra_controllers = [
+        CMSEditModelAdmin::class,
+    ];
+
+    public function testGetCMSEditOwner()
+    {
+        $adminSingleton = CMSEditModelAdmin::singleton();
+        $root = $this->objFromFixture(ManagedDataObject::class, 'root');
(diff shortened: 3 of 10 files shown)


code/HasCanonicalEditLink.php
```diff
@@ -0,0 +1,152 @@
...
+            }
+
+            if ($remoteField) {
+                if ($remoteField === $reciprocalRelation) {
+                    if ($reciprocalRelation === 'reciprocalRelation') {
+                        var_dump([
+                            'HERE',
+                            'relation' => $relation
+                        ]);
+                    }
```
michalkleiner: Should this be removed?

code/HasCanonicalEditLink.php
```diff
@@ -0,0 +1,152 @@
...
+        if (!$link) {
+            throw new LogicException('Could not produce an edit link for the passed object.');
+        }
+        return $link;
+    }
+
+    /**
+     * Get a link to edit this page in the CMS.
+     */
+    public function CMSEditLink(): string
```
michalkleiner: `File` already has `CMSEditLink` method, would this be a problem when applied to e.g. Image or other class extending `File`?

code/HasCanonicalEditLink.php
```diff
@@ -0,0 +1,152 @@
...
+ * private static string canonical_edit_owner = MyModelAdmin::class;
+ * or to a has_one relation:
+ * private static string canonical_edit_owner = 'Parent';
+ *
+ * Note that the canonical edit owner must implement a getEditLinkForObject() method.
+ *
+ * If the canonical edit owner is a has_one relation, the class on the other end
+ * of the relation must have a CMSEditLink() method.
+ */
+trait HasCanonicalEditLink
```
michalkleiner: I was out for a bit and may have missed it, but do we have a naming convention for traits?

I'm a proponent of suffixing the name with the type of class/construct it is, e.g. `SomethingTrait`, `PageController`, `FieldExtension` etc. so it's immediately obvious what it is.

```suggestion
trait CanonicalEditLinkTrait
```

code/HasCanonicalEditLink.php
ChloeSartorelli: Yes, oops. Removed now

michalkleiner:
I remember having issues not being able to get an edit link to objects in past so this would be a welcome addition, however I'm not sure if it should be a trait or somehow integrated more closely with `DataObject` itself or via an automatically applied extension to `DataObject` etc. What is the benefit of using the trait over an extension? Is this opt-in or would it be automatically applied?

code/HasCanonicalEditLink.php
ChloeSartorelli: I can't think of literally any reason someone would do that - but if they did, I'm pretty sure the trait implementation would take precedence.

code/HasCanonicalEditLink.php
ChloeSartorelli: We do have quite a few traits already - some are suffixed with `Trait` but the majority are not.
I think the fact that you declare it _as_ a trait, and you `use` it, make it abundantly clear it's a trait.

`Extension` and `Controller` are a bit different because those aren't built-in language structure things like a trait (we don't say `SiteTreeClass` after all).

GuySartorelli:
> What is the benefit of using the trait over an extension?

Less (if any) overhead is the main benefit - but mostly it just didn't really occur to me to make this as an extension.

> Is this opt-in or would it be automatically applied?

It _has_ to be opt-in. There's no way to know the canonical parent of some arbitrary model. How would we differentiate between the 12 has_one relations and the 3 model admins that all have edit links for some record?
These numbers are arbitrary but that's kinda my point. There's no way for us to determine, at a product-level, what some arbitrary project's model's canonical edit context is.

code/HasCanonicalEditLink.php
emteknetnz: Agree we shouldn't suffix `Trait` on this one (even though I'd actually like it, I also like the `Interface` suffix), because most of the code base doesn't suffix it

I don't like the use of the word `Canonical` at all - it really should just be `CMSEditLink`

I had a look at different traits in the codebase to find a common naming convention, there isn't that much in the way of consistency.

The most common pattern I could find was just calling the trait the feature it implemented e.g.
- trait ImageManiputlation
- trait FlushScheduler
- trait FormMessage

The next most consistent, though less common one was the Configurable / Extensible / Injectable - though those are less "concrete" concepts than just the other traits that just bolt a feature on

This makes a couple of obvious candidates
- trait CMSEditLink
- trait CMSEditLinkable (weird IMO)

I think `CMSEditLink` is the most non-horrible of the two

```php
class MyClass
{
   use CMSEditLink;
```
And I get a `CMSEditLink()` method

ChloeSartorelli: I'll rename the trait to `CMSEditLink`

GuySartorelli:
I've renamed the trait to `CMSEditLink` and renamed the `getEditLinkForObject()` method to `getEditLinkForDataObject()` to match the change in https://github.com/silverstripe/silverstripe-admin/pull/1360

GuySartorelli:
Steve and I have talked about this offline - I'll be converting this into an `Extension` and renaming `getEditLinkForDataObject()` to `getEditLinkForManagedDataObject()` so it's more intuitive.

code/ModelAdmin.php
```diff
@@ -538,9 +538,14 @@ public function getManagedModels()
 
         // Normalize models to have their model class in array key
         foreach ($models as $k => $v) {
+            // No custom tab url segment
             if (is_numeric($k)) {
                 $models[$v] = ['dataClass' => $v, 'title' => singleton($v)->i18n_plural_name()];
                 unset($models[$k]);
+            // Custom tab url segment but no custom title
+            } elseif (is_a($v, DataObject::class, true)) {
+                $models[$k] = ['dataClass' => $v, 'title' => singleton($v)->i18n_plural_name()];
```
ChloeSartorelli: This fix was needed for tests to go green.

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+
+    private function relationIsEditable(string $relation, FieldList $fields): bool
+    {
+        $field = $fields->dataFieldByName($relation);
+        return $field
+            && $field instanceof GridField
+            && $field->getConfig()->getComponentByType(GridFieldDetailForm::class);
+    }
+
+    private function constructLink(string $relation, int $id): string
```
emteknetnz, line 137:
```suggestion
    private function constructCMSEditLink(string $relation, int $id): string
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+
+        if ($owner instanceof DataObject) {
+            $relativeLink = $owner->getEditLinkForManagedDataObject($this->owner, $this->owner->config()->get('canonical_edit_owner'));
+        } else {
+            $relativeLink = $owner->getEditLinkForManagedDataObject($this->owner);
+        }
+        return Director::absoluteURL($relativeLink);
+    }
+
+    private function getLinkForRelation(array $componentConfig, DataObject $obj, string $reciprocalRelation, FieldList $fields): string
```
emteknetnz:
```suggestion
    private function getCMSEditLinkForRelation(array $componentConfig, DataObject $obj, string $reciprocalRelation, FieldList $fields): string
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+        }
+        return $this->owner->getComponent($ownerType);
+    }
+
+    /**
+     * Get the link for editing an object from the CMS edit form of this object.
+     * @throws LogicException if a link cannot be established
+     * e.g. if the object is not in a has_many relation or not edited inside a GridField.
+     */
+    public function getEditLinkForManagedDataObject(DataObject $obj, string $reciprocalRelation): string
```
emteknetnz:
```suggestion
    public function getCMSEditLinkForManagedDataObject(DataObject $obj, string $reciprocalRelation): string
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+ * private static string canonical_edit_owner = 'Parent';
+ *
+ * Note that the canonical edit owner must implement a getEditLinkForManagedDataObject() method.
+ *
+ * If the canonical edit owner is a has_one relation, the class on the other end
+ * of the relation must have a CMSEditLink() method.
+ */
+class CMSEditLinkExtension extends Extension
+{
+    private static string $canonical_edit_owner = '';
```
emteknetnz:
```suggestion
    private static string $cms_edit_owner = '';
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+
+    /**
+     * Get the admin or DataObject which owns this object for CMS editing purposes.
+     *
+     * @return LeftAndMain|DataObject|null
+     */
+    public function getCanonicalEditOwner()
```
emteknetnz:
```suggestion
    public function getCMSEditOwner()
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        $ownerType = $this->owner->config()->get('canonical_edit_owner');
```
emteknetnz:
```suggestion
        $ownerType = $this->owner->config()->get('cms_edit_owner');
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+        if (!$owner || !$owner->exists()) {
+            return '';
+        }
+
+        if (!$owner->hasMethod('getEditLinkForManagedDataObject')) {
+            throw new LogicException('The canonical owner must implement getEditLinkForManagedDataObject()');
+        }
+
+        if ($owner instanceof DataObject) {
+            $relativeLink = $owner->getEditLinkForManagedDataObject($this->owner, $this->owner->config()->get('canonical_edit_owner'));
```
emteknetnz:
```suggestion
            $relativeLink = $owner->getCMSEditLinkForManagedDataObject($this->owner, $this->owner->config()->get('canonical_edit_owner'));
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+        }
+        return $link;
+    }
+
+    /**
+     * Get a link to edit this page in the CMS.
+     */
+    public function CMSEditLink(): string
+    {
+        $owner = $this->owner->getCanonicalEditOwner();
```
emteknetnz:
```suggestion
        $owner = $this->owner->getCMSEditOwner();
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+        if (!$owner || !$owner->exists()) {
+            return '';
+        }
+
+        if (!$owner->hasMethod('getEditLinkForManagedDataObject')) {
```
emteknetnz:
```suggestion
        if (!$owner->hasMethod('getCMSEditLinkForManagedDataObject')) {
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+            throw new LogicException('The canonical owner must implement getEditLinkForManagedDataObject()');
```
emteknetnz:
```suggestion
            throw new LogicException('The canonical owner must implement getCMSEditLinkForManagedDataObject()');
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        if ($owner instanceof DataObject) {
+            $relativeLink = $owner->getEditLinkForManagedDataObject($this->owner, $this->owner->config()->get('canonical_edit_owner'));
+        } else {
+            $relativeLink = $owner->getEditLinkForManagedDataObject($this->owner);
```
emteknetnz:
```suggestion
            $relativeLink = $owner->getCMSEditLinkForManagedDataObject($this->owner);
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        $field = $fields->dataFieldByName($relation);
+        return $field
+            && $field instanceof GridField
+            && $field->getConfig()->getComponentByType(GridFieldDetailForm::class);
+    }
+
+    private function constructLink(string $relation, int $id): string
+    {
+        $ownerType = $this->owner->config()->get('canonical_edit_owner');
```
emteknetnz:
```suggestion
        $ownerType = $this->owner->config()->get('cms_edit_owner');
```

emteknetnz (changes requested):
Mostly naming stuff

Key thing that we missed when we merged https://github.com/silverstripe/silverstripe-admin/pull/1360 was the method name ModelAdmin::getEditLinkForManagedDataObject - it should be getCMSEditLinkForManagedDataObject for consistency with the rest of the code base

A quick string search

vendor/silverstripe/**/*.php

"CMSEditLink(" - 33 results
"EditLink(" - 37 results

Of the "EditLink(" that weren't "CMSEditLink(":
"updateCMSEditLink(" - 1 result
"getFileEditLink(" - 3 results

So you'll also need to rename https://github.com/silverstripe/silverstripe-admin/blob/1/code/ModelAdmin.php#L230 in this PR as well

The other thing is the word "Canonical" is very confusing.  Rename this to "CMSEdit" so that it's standard with everything else

I've left a bunch of recommended edits, though there's a large number of things that will need renaming e.g. the test files, docs, etc

GuySartorelli:
> The other thing is the word "Canonical" is very confusing. Rename this to "CMSEdit" so that it's standard with everything else

"canonical" _is_ correct in this context - e.g. the `<link rel="canonical" href="https://www.example.com/page/" />` meta tag tells search engines what the _canonical_ URL is for a page, in case the page can be accessed via multiple URLs.

But I'll make that change to avoid peer review ping pong and to match the other changes where "canonical" was already changed.

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+        $fields = $this->owner->getCMSFields();
+        $link = $this->getCMSEditLinkForRelation($this->owner->hasMany(false), $obj, $reciprocalRelation, $fields);
+        if (!$link) {
+            throw new LogicException('Could not produce an edit link for the passed object.');
+        }
+        return $link;
+    }
+
+    /**
+     * Get a link to edit this page in the CMS.
```
emteknetnz:
```suggestion
     * Get a link to edit this DataObject in the CMS.
```

ChloeSartorelli: Updated

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+ * You must set a cms_edit_owner config variable which defines the cms edit
+ * owner for this class.
+ * e.g. set this to a {@link LeftAndMain} class:
+ * private static string cms_edit_owner = MyModelAdmin::class;
+ * or to a has_one relation:
+ * private static string cms_edit_owner = 'Parent';
+ *
+ * Note that the cms edit owner must implement a getCMSEditLinkForManagedDataObject() method.
+ *
+ * If the cms edit owner is a has_one relation, the class on the other end
```
emteknetnz:
```suggestion
 * If the cms_edit_owner is a has_one relation, the class on the other end
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+ * of the relation must have a CMSEditLink() method.
+ */
+class CMSEditLinkExtension extends Extension
+{
+    private static string $cms_edit_owner = '';
+
+    /**
+     * Get the admin or DataObject which owns this object for CMS editing purposes.
```
emteknetnz:
```suggestion
     * Get the ModelAdmin, LeftAndMain or DataObject which owns this object for CMS editing purposes.
```

code/CMSEditLinkExtension.php
```diff
@@ -0,0 +1,149 @@
...
+     *
+     * @return LeftAndMain|DataObject|null
```
emteknetnz, line 43:
```suggestion
     * @return LeftAndMain|ModelAdmin|DataObject|null
```

code/CMSEditLinkExtension.php
ChloeSartorelli: ModelAdmin _is_ a LeftAndMain - and they're all admin. I'll make this change but it's super subjective.

code/CMSEditLinkExtension.php
ChloeSartorelli: done

code/CMSEditLinkExtension.php
ChloeSartorelli: ModelAdmin is a LeftAndMain so that's not necessary here. I'll make the change for the text comment but it's not appropriate for a type hint.

code/CMSEditLinkExtension.php
ChloeSartorelli: done

tests/php/CMSEditLinkExtensionTest/BelongsToModelAdmin.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+namespace SilverStripe\Admin\Tests\CMSEditLinkExtensionTest;
+
+use SilverStripe\Admin\CMSEditLinkExtension;
+use SilverStripe\Dev\TestOnly;
+use SilverStripe\ORM\DataObject;
+
+class BelongsToModelAdmin extends DataObject implements TestOnly
```
emteknetnz: Could you please rename this, it's not a "belongs_to" relationship, which is defined as the reciprocal of a has_one relationship when it's a 1-1 relations.

Also, it being suffixed with "ModelAdmin" confused me for a long time because I mistakenly thought it extended `ModelAdmin`

Even just calling this `TestDataObject` would be an improvement

tests/php/CMSEditLinkExtensionTest/BelongsToModelAdmin.php
```diff
@@ -0,0 +1,37 @@
...
+    private static $db = [
+        'Name' => 'Varchar(255)',
+    ];
+
+    private static $has_many = [
+        'ArbitraryRelation' => NestedObject::class,
+        'Nested' => NestedObject::class . '.Parent',
+        'AnotherArbitraryRelation' => NestedObject::class . '.AnotherOfTheSameClass',
+        'BasicNested' => BasicNestedObject::class,
+        'PolyMorphic' => PolymorphicNestedObject::class,
```
emteknetnz:
```suggestion
        'Polymorphic' => PolymorphicNestedObject::class,
```

emteknetnz (changes requested):

tests/php/CMSEditLinkExtensionTest/BelongsToModelAdmin.php
ChloeSartorelli: That wouldn't be an improvement, it's a lot less descriptive. I'll rename it to something that doesn't contain "belongsto" or end with "modeladmin". But given this is a test object I don't see that it matters _all_ that much what it's called.

ChloeSartorelli: Done

tests/php/CMSEditLinkExtensionTest/BelongsToModelAdmin.php
ChloeSartorelli: Done

emteknetnz (approved):

Issue #1362, opened 2022-09-14
DataObjects cannot be made previewable via an extension


GuySartorelli:
Currently for a `DataObject` to be previewable in the CMS it must directly implement the `CMSPreviewable` interface. This ultimately means it's impossible to make a model previewable unless it's directly controlled in your project.

## Acceptance criteria:
- [x] If a `DataObject` has an `Extension` applied which implements `CMSPreviewable`, the model itself doesn't need to implement that interface for it to be previewable.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1363
- https://github.com/silverstripe/silverstripe-cms/pull/2779
- https://github.com/silverstripe/developer-docs/pull/87
- https://github.com/silverstripe/silverstripe-siteconfig/pull/127

GuySartorelli:
Code to test these changes locally (note that the AdminExtension could just be part of the ModelAdmin subclass directly):

## MyModelAdmin.php
```php
<?php

namespace App;

use SilverStripe\Admin\ModelAdmin;

class MyModelAdmin extends ModelAdmin
{
    private static $url_segment = 'test';

    private static $managed_models = [
        MyPreviewable::class,
    ];
}
```

## AdminExtension.php
```php
<?php

namespace App;

use SilverStripe\Core\Extension;
use SilverStripe\View\SSViewer;

class AdminExtension extends Extension
{
    private static $allowed_actions = [
        'cmsPreview',
    ];

    private static $url_handlers = [
        '$ModelClass/cmsPreview/$ID' => 'cmsPreview',
    ];

    public function cmsPreview()
    {
        $id = $this->owner->getURLParams()['ID'];
        $obj = $this->owner->getModelClass()::get_by_id($id);
        if (!$obj || !$obj->exists()) {
            return $this->owner->httpError(404);
        }

        // Include use of a front-end theme temporarily.
        $oldThemes = SSViewer::get_themes();
        SSViewer::set_themes(SSViewer::config()->get('themes'));
        $preview = $obj->forTemplate();

        // Make sure to set back to backend themes.
        SSViewer::set_themes($oldThemes);

        return $preview;
    }
}
```

## MyPreviewable.php
```php
<?php

namespace App;

use SilverStripe\ORM\DataObject;

class MyPreviewable extends DataObject
{
    private static $db = [
        'Name' => 'Varchar',
    ];
}
```

## ModelExtension.php
```php
<?php

namespace App;

[...]

Issue #1306, opened 2022-05-04
Test coverage for preview panel


GuySartorelli:
Currently there is test coverage for `SiteTree` and `BaseElement` objects in preview panels in the cms and elemental modules respectively, but there's no test coverage for previews in general.

The below list is not necessarily an exhaustive list of useful tests that could be implemented.

## Behat tests
- [ ] The preview panel is _not_ present if the object belonging to the current form does not implement `CMSPreviewable`
- [ ] The preview panel is present but collapsed by default if the object belonging to the current form implements `CMSPreviewable` but returns an empty value from `PreviewURL()`
- [ ] The preview panel is present and displays the correct content if the object belonging to the current form implements `CMSPreviewable` and returns a valid relative URL path from `PreviewURL()`
- [ ] Unversioned and Versioned objects are tested, including published and unpublished objects, and swapping preview modes
- [ ] Swapping preview sizes

GuySartorelli:
Note that currently preview relies on some classes that exist in the CMS module - probably those will need to be transplanted.

PR #1363, opened 2022-09-14
ENH Records can be made previewable with an extension

GuySartorelli:
Note: There are currently few if any tests for previewable objects - that will come with https://github.com/silverstripe/silverstripe-admin/issues/1306

Requires https://github.com/silverstripe/silverstripe-cms/pull/2779 to be merged in at the same time

See the parent issue for code to test this locally.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1362

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -1444,7 +1444,7 @@ class LeftAndMain extends Controller implements PermissionProvider
         // Added in-line to the form, but plucked into different view by frontend scripts.
         if ($record instanceof CMSPreviewable || $record->has_extension(CMSPreviewable::class)) {
             /** @skipUpgrade */
-            $navField = new LiteralField('SilverStripeNavigator', $this->getSilverStripeNavigator());
+            $navField = new LiteralField('SilverStripeNavigator', $this->getSilverStripeNavigator($record));
             $navField->setAllowHTML(true);
             $fields->push($navField);
         }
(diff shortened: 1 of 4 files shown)


sabina-talipova:
@GuySartorelli , could you rebase this to branch `2`, please.

GuySartorelli:
Rebased on `2`

code/LeftAndMain.php
```diff
@@ -1676,7 +1676,7 @@ public function printable()
     public function getSilverStripeNavigator()
     {
         $page = $this->currentPage();
```
sabina-talipova:
```suggestion
        $page = $this->currentPage();
        
         if (!$page) {
            return null;
        }
```

sabina-talipova (changes requested):

code/LeftAndMain.php
sabina-talipova: All works good! Just small changes, since it breaks Settings section.

ChloeSartorelli: I've gone one further - `SiteConfig` should be previewable too! See https://github.com/silverstripe/silverstripe-siteconfig/pull/127.

CI should go green without that as well though, as I have incoporated the above suggestion in case other admins don't pass their "current" items in the expected way.

sabina-talipova (approved):

PR #1367, opened 2022-09-19
DEP Upgrade tinymce from 4 to 6

GuySartorelli:
Update to tinymce 6
Also move jquery changetracker which needed some modifications - made sense to move it before making the changes as it was in the wrong place.

## Notes:
- `TinymceInlineToolbar` is not needed anymore - there's built-in functionality for that now.

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10369


_config.php
```diff
@@ -38,7 +37,7 @@
             . "tbody[id|class|style],thead[id|class|style],tfoot[id|class|style],"
             . "#td[id|dir|class|colspan|rowspan|width|height|align|valign|scope|style],"
             . "-th[id|dir|class|colspan|rowspan|width|height|align|valign|scope|style],caption[id|dir|class],"
-            . "-div[id|dir|class|align|style],-span[class|align|style],-pre[class|align],address[class|align],"
+            . "-div[id|dir|class|align|style|data*],-span[class|align|style],-pre[class|align],address[class|align],"
```
ChloeSartorelli: This is needed for embeds. It looks like data attributes were being implicitly allowed in the past, somhow... I need to see whether that was intentional and what we need to do about it (i.e. do we always allow them or just add in the changelog that not all elements allow them by default?)

_config.php
```diff
@@ -15,7 +15,6 @@
 $editorConfig = TinyMCEConfig::get('cms');
 $editorConfig
     ->enablePlugins([
-        'contextmenu' => null,
```
ChloeSartorelli, line 18: This is built-in now

client/src/legacy/jquery.changetracker.js
```diff
@@ -85,8 +96,8 @@ import debounce from 'lodash.debounce';
             case 'default':
               var config = $.extend({ forced_root_block: 'p' }, $field.data('config'));
 
-              var serializer = new tinymce.html.Serializer(config);
-              var parser = new tinymce.html.DomParser(config);
+              var serializer = tinymce.html.Serializer(config);
+              var parser = tinymce.html.DomParser(config);
```
ChloeSartorelli, line 100: There's no constructor for these anymore

_config.php
ChloeSartorelli: That was incorrect - I hadn't noticed the shortcodes for embed and media were being stored in their temporary HTML format.

_config/tinymce.yml
```diff
@@ -2,6 +2,7 @@
 Name: admintinymce
 ---
 SilverStripe\Forms\HTMLEditor\TinyMCEConfig:
-  base_dir: 'silverstripe/admin:thirdparty/tinymce'
+  base_dir: 'silverstripe/admin:client/dist/tinymce'
+  lang_dir: 'silverstripe/admin:thirdparty/tinymce_lang'
```
ChloeSartorelli, line 6: tinymce itself is in `client/dist/` to reflect the fact that it's being put there as part of `yarn build` - i.e. it's not just pasted in there directly, which means we'll get updates through the normal dependency update mechanisms.

The lang dir is in `thidrparty/` to reflect the fact that it's third party code which we've downloaded and dumped in there - it's not grabbed as part of `yarn build` so to updates to this will be manual.

client/src/lib/TinyMCEActionRegistrar.js
```diff
@@ -36,6 +36,17 @@ class TinyMCEActionRegistrar {
...
 
+    /* eslint-disable no-param-reassign */
+    action.type = 'menuitem';
+    // To reduce upgrade pain from tinymce4 to tinymce6 (cms4 to cms5) we've added this
+    // as this would mean any custom plugins using this class to register commands won't
+    // need to update the syntax for doing so
+    if (action.hasOwnProperty('onclick')) {
+      action.onAction = action.onclick;
+      delete action.onclick;
+    }
```
ChloeSartorelli, line 47: I'm happy to remove this if there is any pushback at all - people will probably have to update other parts of their plugins anyway so this doesn't really help _all_ that much.

code/LeftAndMain.php
```diff
@@ -683,6 +683,10 @@ protected function init()
         // file because insufficient information exists when that is being processed
         $htmlEditorConfig = HTMLEditorConfig::get_active();
         $htmlEditorConfig->setOption('language', TinyMCEConfig::get_tinymce_lang());
+        $langUrl = TinyMCEConfig::get_tinymce_lang_url();
+        if ($langUrl) {
+            $htmlEditorConfig->setOption('language_url', $langUrl);
+        }
```
ChloeSartorelli, line 689: Along with the language itself, we need to point to the correct translation file now.

GuySartorelli:
Tests pass locally, but https://github.com/silverstripe/silverstripe-framework/pull/10508 will need to be merged for CI to go green.

sabina-talipova:
Removing `spellchecker` and `contextmenu` plugins impact to the other modules `SilverStripe\SpellCheck\Handling\SpellCheckAdminExtension class`  and `CWP` module (see `cwp\cwp\_config.php`).
There is some issue with `fontawesome` plugin in my local (see `cwp\agency-extensions\_config.php`). I have console error `404. File not found`.

sabina-talipova (approved):
Merge after [silverstripe-framework](https://github.com/silverstripe/silverstripe-framework/pull/10508)

GuySartorelli:
The behat failures were fixed in https://github.com/silverstripe/silverstripe-admin/pull/1370 which was merged after my last rebase. They are unrelated to this PR but I can rebase if you want to see this go green.

PR #1375, opened 2022-10-04
DEP Update jQuery to 3.x

GuySartorelli:
I've added some comments in this PR to help explain why some of the changes have been made.

A full CI run including behat for all modules which have a behat config has been run on kitchen sink. Note that running kitchen sink without this PR also has some failures (because cwp and other modules change defaults which change the test conditions) so compare these two to see if the failures are the same:
- kitchen sink CI with PR: https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/3333342120
- kitchen sink CI without PR: https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/3325921391

I've also run this against installer:
- https://github.com/creative-commoners/silverstripe-installer/actions/runs/3333345210

I have also confirmed that [unclecheese/display-logic](https://github.com/unclecheese/silverstripe-display-logic) still works

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1351


GuySartorelli:
Functionality from `client/src/legacy/LeftAndMain.menu.js` has been removed because it isn't used, and was requiring an entire dependency (for a one-liner) which I would otherwise have had to update.

The example in that file doesn't even work as advertised - it gives the following output:
image
As you can see the styling for this is all over the place. It's a holdover from CMS 3 which [actually broke part-way through CMS 3 itself](https://github.com/silverstripe/silverstripe-framework/pull/4999), and hasn't been fixed or used since.

If anyone does want nested menus, there's [a separate module for that](https://github.com/symbiote/silverstripe-grouped-cms-menu).

Note: That module relies on some of this child menu functionality and related styling, so I won't remove all of it - but I will remove the "flyout" parts, which are not used at all.

client/src/bundles/vendor.js
```diff
@@ -46,16 +46,13 @@ require('expose-loader?moment!moment');
 require('expose-loader?merge!merge');
 
 require('../../../thirdparty/jquery-ondemand/jquery.ondemand.js');
-require('../../../thirdparty/jquery-ui/jquery-ui.js');
-// require('../../../thirdparty/jquery-ui-themes/smoothness/jquery-ui.css');
-require('../../../thirdparty/jquery-entwine/dist/jquery.entwine-dist.js');
+require('../../../thirdparty/jquery-ui/jquery-ui.js'); // there is related css in styles/bundle.scss
+require('../../../thirdparty/jquery-entwine/jquery.entwine.js');
```
ChloeSartorelli, line 50: Moved because we only need this one file - all the rest is unnecessary cruft.

client/src/bundles/vendor.js
```diff
@@ -46,16 +46,13 @@ require('expose-loader?moment!moment');
...
 require('../../../thirdparty/jquery-cookie/jquery.cookie.js');
 require('../../../thirdparty/jquery-query/jquery.query.js');
-require('../../../thirdparty/jquery-form/jquery.form.js');
+require('jquery-form');
 require('jquery-sizes/lib/jquery.sizes.js');
-require('../../../thirdparty/jstree/jquery.jstree.js');
-// require('../../../thirdparty/stree/themes/apple/style.css');
-require('../../../thirdparty/jquery-hoverIntent/jquery.hoverIntent.js');
```
ChloeSartorelli, line 58: This dependency was only used in one place, and the place it was used was for legacy CMS 3 code (refer to https://github.com/silverstripe/silverstripe-admin/pull/1375#issuecomment-1269148790)

client/src/components/Menu/Menu.scss
```diff
@@ -220,7 +220,6 @@
     padding-top: 17px;
   }
 
-  .child-flyout-indicator,
```
ChloeSartorelli, line 223: For all the flyout removals, refer to https://github.com/silverstripe/silverstripe-admin/pull/1375#issuecomment-1269148790

client/src/legacy/LeftAndMain.ActionTabSet.js
```diff
@@ -124,7 +125,8 @@ $.entwine('ss', function($) {
           containerSouth = activePanel.parents('.toolbar--south');
           if (containerSouth){
             // If container is the southern panel, make tab appear from the top of the container
-            padding = activeTab.offset().top - containerSouth.offset().top;
+            const offset = activeTab.offset();
+            padding = offset ? (offset.top - containerSouth.offset().top) : 0;
```
ChloeSartorelli, line 129: This change is required because [offset()](https://api.jquery.com/offset/) returns `undefined` for hidden elements.

> While it is possible to get the coordinates of elements with `visibility:hidden` set, `display:none` is excluded from the rendering tree and thus has a position that is undefined.

client/src/legacy/LeftAndMain.BatchActions.js
```diff
@@ -372,37 +372,42 @@ $.entwine('ss.tree', function($){
...
-        parameterFields.show();
-        $('#BatchActionParameters').slideDown();
-      } else {
-        $('#BatchActionParameters').slideUp();
+      const form = $(e.target.form);
+      const btn = form.find(':submit');
+      const selected = $(e.target).val();
+
+      // When resetting the field after a successful action, there won't be a valid "selected" option.
+      if (selected) {
```
ChloeSartorelli, line 380: See the `onsubmit()` function - the action is intentionally reset to the default option after a successful action has been fired (`self.find(':input[name=Action]').val('').change();`).

This works correctly before this PR - not sure what changed but this started throwing errors on the next line because `selected` was null or undefined. This avoids that issue.

client/src/legacy/LeftAndMain.Menu.js
```diff
@@ -29,50 +29,27 @@ $.entwine('ss', function($){
   $('.cms-panel.cms-menu').entwine({
 
     togglePanel: function(doExpand, silent, doSaveState) {
-      //apply or unapply the flyout formatting, should only apply to cms-menu__list when the current collapsed panal is the cms menu.
+      //apply or unapply the child formatting, should only apply to cms-menu__list when the current collapsed panal is the cms menu.
```
ChloeSartorelli, line 32: Refer to https://github.com/silverstripe/silverstripe-admin/pull/1375#issuecomment-1269148790 for all of the flyout changes in this file.

client/src/legacy/LeftAndMain.js
```diff
@@ -220,26 +220,13 @@ $.entwine('ss', function($) {
      * Constructor: onmatch
      */
     onadd: function() {
-      // Browser detection
-      if($.browser.msie && parseInt($.browser.version, 10) < 8) {
```
ChloeSartorelli, line 224: `$.browser` was [removed in jQuery 1.9](https://api.jquery.com/jquery.browser/) - but this message isn't useful anymore anyway. It's saying to use a REALLY old version if IE, when we don't support IE at all anymore.

Better to just remove the warning altogether.

client/src/legacy/LeftAndMain.js
```diff
@@ -752,13 +739,13 @@ $.entwine('ss', function($) {
      *   (Object) state The original history state which the request was initiated with
      */
     handleAjaxResponse: function(data, status, xhr, state) {
-      var self = this, url, selectedTabs, guessFragment, fragment, $data;
```
ChloeSartorelli, line 755: Those variables aren't ever set or used.

client/src/legacy/jquery.changetracker.js
```diff
@@ -4,7 +4,7 @@ import debounce from 'lodash.debounce';
 
 // Copyright (c) 2009, SilverStripe Ltd.
```
ChloeSartorelli, line 5: Moved this file to be consistent with changes in CMS 5 but also because this isn't actually a third party script - as the copyright and license make clear, this is ours. It doesn't belong in thirdparty especially since we're not maintaining it separately from this module at all.

client/src/legacy/jquery.changetracker.js
```diff
@@ -29,9 +29,20 @@ import debounce from 'lodash.debounce';
...
+ * Setup:
+ * 	jQuery('<my-form>).changetracker();
+ *
+ * Finding out if the form has changed:
+ * 	jQuery('<my-form>).is('.changed');
+ *
+ * Options:
+ * fieldSelector: jQuery selector string for tracked fields (Default: ':input:not(:submit),:select:not(:submit)')
+ * ignoreFieldSelector: jQuery selector string for specifically excluded fields
+ * changedCssClass: CSS class attribute which is appended to all changed fields and the form itself
```
ChloeSartorelli, line 42: This is just copied straight from the readme so we can get rid of the readme file associated with this script.

client/src/legacy/tests/jquery.changetracker-test.js
```diff
@@ -0,0 +1,196 @@
+/* global jest, describe, it, expect */
+
+const jQuery = require('jquery');
```
ChloeSartorelli, line 3: This is just an adaptation of the existing tests so that we can actually test this using jest. Previously we simply weren't testing this.

package.json
```diff
@@ -105,13 +108,15 @@
...
+    "expose-loader": "^0.7.3",
     "fast-glob": "^3.0.1",
     "html-loader": "^0.5.1",
     "is-plain-object": "^5.0.0",
     "jest-cli": "^23.6.0",
     "markdown-loader": "^5.1.0",
+    "sass-loader": "^6.0.5",
     "storybook-addon-jsx": "^5.4.0",
-    "webpack": "^2.6.1"
+    "webpack": "^3.12.0"
```
ChloeSartorelli, line 118: This was necessary to get jquery to install at all. Thankfully the only change required to run this version of webpack was adding `expose-loader` and `sass-loader` as explicit dependencies (previously added via webpack-config) so it's not a big deal and won't affect anything.

thirdparty/jquery-entwine/jquery.entwine.js
```diff
@@ -42,6 +42,9 @@ var Base;
 
 	Base.addMethods = function(props) {
 		for (var name in props) {
+			if (!props.hasOwnProperty(name)) {
```
ChloeSartorelli, line 69: This has been added quite a few times as best practice and to fix a regression where sometimes it tries to include `prototype` in the loop. We only ever want to loop over the properties that are being added by our code, not built-ins.

thirdparty/jquery-entwine/jquery.entwine.js
```diff
@@ -742,80 +744,19 @@ Sizzle is good for finding elements for a selector, but not so good for telling
...
-
-/* src/jquery.focusinout.js */
-
-(function($){	
-	
-	/**
-	 * Add focusin and focusout support to bind and live for browers other than IE. Designed to be usable in a delegated fashion (like $.live)
-	 * Copyright (c) 2007 Jörn Zaefferer
-	 */
-	if ($.support.focusinBubbles === undefined)  {
```
ChloeSartorelli: `$.support` [is deprecated](https://api.jquery.com/jquery.support/), but thankfully since this was just to support IE we can remove it without repercussions.

thirdparty/jquery-entwine/jquery.entwine.js
```diff
@@ -742,80 +744,19 @@ Sizzle is good for finding elements for a selector, but not so good for telling
...
-					arguments[0] = $.event.fix(e);
-					arguments[0].type = fix;
-					return $.event.handle.apply(this, arguments);
-				}
-			};
-		});
-	}
-		
-	(function(){
-		//IE has some trouble with focusout with select and keyboard navigation
```
ChloeSartorelli: More IE-specific stuff we can happily remove.

thirdparty/jquery-entwine/jquery.entwine.js
```diff
@@ -742,80 +744,19 @@ Sizzle is good for finding elements for a selector, but not so good for telling
...
 
 
 /* src/jquery.entwine.js */
 
-try {
-	console.log;
-}
-catch (e) {
-	window.console = undefined;
-}
```
ChloeSartorelli: `console.log` is always available. This was just here to support SUPER old browsers.

thirdparty/jquery-entwine/jquery.entwine.js
```diff
@@ -1008,8 +959,14 @@ catch (e) {
 					if (funcs[i].selector.matches(el)) {
 						var ret, tmp_i = el.i, tmp_f = el.f;
 						el.i = i; el.f = one;
-						try { ret = funcs[i][funcprop].apply(namespace.$(el), args); }
-						finally { el.i = tmp_i; el.f = tmp_f; }
+						try {
+							if (typeof funcs[i][funcprop] == 'function') {
```
ChloeSartorelli, line 1117: Sometimes that wasn't a function which resulted in errors being thrown - so we have to check first.

thirdparty/jquery-entwine/jquery.entwine.js
```diff
@@ -1157,148 +1117,117 @@ catch (e) {
...
 		}
-		patched.patched = true;
-
-		return patched;
-	}
 
-	var version = $.prototype.jquery.split('.');
-	var callbackIdx = (version[0] > 1 || version[1] >= 10 ? 1 : 2);
-
-	// Monkey patch $.fn.domManip to catch all regular jQuery add element calls
```
ChloeSartorelli: `domManip` wasn't even documented - it's an internal function that was removed in 3.x so we can't patch that to see when new nodes are being added. Instead, we now use mutations.

thirdparty/jquery-entwine/jquery.entwine.js
```diff
@@ -1157,148 +1117,117 @@ catch (e) {
...
-	var _domManip = $.prototype.domManip;
-	$.prototype.domManip = function() {
-		if (!arguments[callbackIdx].patched) arguments[callbackIdx] = patchDomManipCallback(arguments[callbackIdx]);
-		return _domManip.apply(this, arguments);
-	}
-
-	// Monkey patch $.fn.html to catch when jQuery sets innerHTML directly
```
ChloeSartorelli: This was always a dangerous way to detect this - but thankfully it isn't necessary because setting `innerHTML` will result in new nodes in the mutation observer.

thirdparty/jquery-entwine/jquery.entwine.js
```diff
@@ -1457,26 +1393,7 @@ catch (e) {
 
 /* src/jquery.entwine.events.js */
 
-(function($) {	
-
-	/** Taken from jQuery 1.5.2 for backwards compatibility */
-	if ($.support.changeBubbles == undefined) {
```
ChloeSartorelli: `$.support` [is deprecated](https://api.jquery.com/jquery.support/) but we don't need this anyway - it's to support old browsers we don't care about.

thirdparty/jquery-ondemand/jquery.ondemand.js
```diff
@@ -6,13 +6,11 @@
  * Adds capabilities for custom X-Include-CSS and X-Include-JS HTTP headers
  * to request loading of externals alongside an ajax response.
  *
- * Requires jQuery 1.5 ($.Deferred support)
- *
- * CAUTION: Relies on customization of the 'beforeSend' callback in jQuery.ajaxSetup()
```
ChloeSartorelli, line 11: This doesn't require jQuery 1.5. It's been running on 1.7 for who knows how long, but I've now upgraded it to work with 3.6 - `$.deferred` is [still very much supported](https://api.jquery.com/jquery.deferred/)

The note about `beforeSend` is unnecessary and frankly misleading here, since that customisation is included in this file anyway.

thirdparty/jquery-query/jquery.query.js
```diff
@@ -1,22 +1,23 @@
 /**
+ * NOTE: This copy has modifications for Silverstripe
  * jQuery.query - Query String Modification and Creation for jQuery
  * Written by Blair Mitchelmore (blair DOT mitchelmore AT gmail DOT com)
  * Licensed under the WTFPL (http://sam.zoy.org/wtfpl/).
  * Date: 2009/8/13
  *
  * @author Blair Mitchelmore
- * @version 2.1.7
+ * @version 2.2.3
```
ChloeSartorelli, line 9: This has basically just been updated with the latest copy I could find

GuySartorelli (commented):
The following deprecated jQuery functions were replaced with their modern counterparts throughout:
- [proxy](https://api.jquery.com/jQuery.proxy/) replaced with the native [Function.prototype.bind](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) function
- [bind](https://api.jquery.com/bind/) replaced with [on](https://api.jquery.com/on/)
- [delegate](https://api.jquery.com/bind/) replaced with [on](https://api.jquery.com/on/)
- [live](https://api.jquery.com/live/) replaced with [on](https://api.jquery.com/on/)
- [unbind](https://api.jquery.com/bind/) replaced with [on](https://api.jquery.com/off/)
- [undelegate](https://api.jquery.com/bind/) replaced with [on](https://api.jquery.com/off/)
- [andSelf](https://api.jquery.com/andSelf/) replaced with [addBack](https://api.jquery.com/addBack/)
- [trim](https://api.jquery.com/jquery.trim/) replaced with the native [String.prototype.trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) function
- [isFunction](https://api.jquery.com/jquery.isFunction/) replaced with comparing the result of `typeof` with 'function'.
- [isArray](https://api.jquery.com/jquery.isFunction/) replaced with the native [Array.isArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray) function
- [ajax.success, ajax.error, and ajax.complete](https://api.jquery.com/jquery.ajax/) replaced with `ajax.done`, `ajax.fail`, and `ajax.always` respectively

Though note that some of these are deprecated, but not yet removed - in those cases I have largely not replaced these in `thirdparty` scripts

client/src/legacy/LeftAndMain.js
```diff
@@ -772,17 +759,15 @@ $.entwine('ss', function($) {
...
       if(xhr.getResponseHeader('Content-Type').match(/^((text)|(application))\/json[ \t]*;?/i)) {
         newFragments = data;
       } else {
 
         // Fall back to replacing the content fragment if HTML is returned
-        fragment = document.createDocumentFragment();
-
-        jQuery.clean( [ data ], document, fragment, [] );
-        $data = $(jQuery.merge( [], fragment.childNodes ));
+        $data = $($.parseHTML(data, document, false));
```
ChloeSartorelli, line 780: `jQuery.clean()` is an undocumented function that was removed in 3.0 - but the functionality is effectively the same as `$.parseHTML()` (the latter of which does the `jQuery.merge()` for us, so we can skip that part).

client/src/legacy/LeftAndMain.js
```diff
@@ -47,7 +47,17 @@ window.ss.debounce = function (func, wait, immediate) {
   };
 };
 
-$(window).bind('resize.leftandmain', function(e) {
+/**
+ * The URL to use for saving and loading tab state
+ */
+window.ss.tabStateUrl = function() {
```
ChloeSartorelli, line 53: This needs to be accessible in `TabSet.js` now

client/src/legacy/LeftAndMain.js
```diff
@@ -928,46 +923,14 @@ $.entwine('ss', function($) {
...
-            self.trigger('tabstaterestored');
-          }
+      const tabsets = this.find('.cms-tabset, .ss-tabset');
+
+      if (tabsets.length) {
+        tabsets.each(function() {
+          const tabset = $(this);
+          const tabsetId = tabset.attr('id');
+          const overrideState = (overrideStates && overrideStates[tabsetId]) ? overrideStates[tabsetId] : null;
+          tabset.restoreState(overrideState);
```
ChloeSartorelli, line 933: The state restore functionality has moved to `TabSet.js` so that each tab can be responsible for restoring its own state.
This is necessary because `onAdd()` is now triggered asyncronously _some time after_ the element is added (see changes in entwine to using mutation observers) which means the tabs may not be initialised yet when this function is called.

client/src/legacy/TabSet.js
```diff
@@ -36,6 +40,46 @@ $.entwine('ss', function($){
         this._super();
     },
 
+    restoreState: function (overrideState) {
```
ChloeSartorelli, line 43: This functionality is mostly copied directly from `LeftAndMain.js` with some minor changes:
- some of the code has been slightly modernised
- if the tabset hasn't been initialised yet, is is marked to defer restoring the state until `redrawTabs()` is called.

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -142,9 +142,11 @@ $.entwine('ss.tree', function($){
               }
             }
           })
-          .bind('move_node.jstree', moveNodeCallback)
+          .on('move_node.jstree', moveNodeCallback)
           // Make some jstree events delegatable
-          .bind('select_node.jstree check_node.jstree uncheck_node.jstree', function(e, data) {
+          .on('select_node.jstree check_node.jstree uncheck_node.jstree', function(e, data) {
+            // Clear out namespace before triggering so entwine handlers are called
+            e.namespace = '';
```
ChloeSartorelli, line 149: This didn't need to be set to blank in the past because of a bug in jQuery where the namespace wasn't actually kept on the event after `triggerHandler()` was called. This bug has been fixed in the current version of jQuery, so we have to manually remove the namespace for it to get to entwine.

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -57,14 +57,16 @@ $.entwine('ss', function($){
...
-     * (Boolean)
+     * sometimes be inadvertently initialised multiple times, but we don't want duplicate messages.
      */
-    ValidationErrorShown: false,
+    getValidationErrorShown: function() {
+      return Boolean(this.data('_validationErrorShown'));
+    },
+    setValidationErrorShown: function(value) {
+      this.data('_validationErrorShown', value);
+    },
```
ChloeSartorelli, line 69: Turning this from an entwine property into regular javascript data means it is tied to the actual DOM element, which means we can rely on it being reset when a new DOM element is added.

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -96,9 +98,6 @@ $.entwine('ss', function($){
         }
       }
 
-      // Reset error display
-      this.setValidationErrorShown(false);
```
ChloeSartorelli, line 100: We can't rely on this being reset BEFORE we fire a validation message, since `onadd` is fired asyncronously to the `redrawTabs()` call which triggers a call to `onafterredrawtabs()` which relies on this value.

emteknetnz (approved):

lekoala:
i'm so glad the team finally decided to upgrade jquery :-) that was a pleasant surprise while reading the changelog

NightJar:
> I have also confirmed that [unclecheese/display-logic](https://github.com/unclecheese/silverstripe-display-logic) still works

@GuySartorelli I'm curious to what version you tested against.
I'm using 2.0.6 (latest 2 line, released Feb '23) and I'm seeing instant errors

`Uncaught TypeError: s.default.entwine is not a function` in `http://localhost/_resources/vendor/unclecheese/display-logic/client/dist/js/bundle.js?m=1676431318`

I'm trying to figure out if there's too much jQuery, an issue with display logic's front end bundle, or it's actually an issue with the change in interface between these two modules.

GuySartorelli:
> I'm curious to what version you tested against.

It was way too long ago for me to remember it, and I didn't note down the version at the time, unfortunately.

NightJar:
> > I'm curious to what version you tested against.
> 
> It was way too long ago for me to remember it, and I didn't note down the version at the time, unfortunately.

That's cool, I thought that might be the case; thanks for responding.
I've been able to identify that the previously deployed display-logic version (2.0.5) still exhibits the issue, meaning it's not a regression (it worked fine in active service with CMS 4.11).

I'm busy trying to assert it's not the fault of my site before I open an issue :)
jQuery is loading, but isn't Entwine isn't binding to it for some reason, although it loads immediately afterward (in normal old school jQuery fashion).

NightJar:
> I'm trying to figure out if there's too much jQuery

✅ 
... It was `debugbar` 🤦 
Site is fine in test/live mode 😄

Issue #1379, opened 2022-10-07
CMS 4.12: Archive admin is broken


GuySartorelli:
Archive admin doesn't load, because it overrides `getManagedModels()` and doesn't include a `dataClass` key.
There are likely other `ModelAdmin` subclasses that do this to, so instead of just patching archive admin, we should make `ModelAdmin` not assume that key will always be there.

## Acceptance criteria
- [x] Archive Admin works again.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1380

## Notes:
It's probably worth spinning off a card for CMS 5 to officially require `'dataClass'` being a key in the array.

michalkleiner:
@GuySartorelli PR approved and merged, did you want to do a patch release?

GuySartorelli:
Awesome, thank you.
A patch release wouldn't be appropriate in this case - the problem was only present in the `4` branch (aka 4.12) which hasn't had a stable release yet.

Closing as this has been merged.

PR #1380, opened 2022-10-07
FIX Don't assume dataclass is in the spec.

GuySartorelli:
Don't assume the dataclass is in the spec of `getManagedModels()` because subclasses may not include it.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1379


maxime-rainville (commented):
Looks like something that ought to be tested?

GuySartorelli:
Added a test

michalkleiner (approved):

PR #1381, opened 2022-10-10
API Strongly-type action method signatures

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10466

### DIFF
--- a/code/CMSBatchAction.php
+++ b/code/CMSBatchAction.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Admin;
 
 use SilverStripe\CMS\Model\SiteTree;
 use SilverStripe\Control\Controller;
+use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injectable;
 use SilverStripe\ORM\SS_List;
@@ -38,7 +39,7 @@ abstract class CMSBatchAction
      * @param SS_List $objs
      * @return string
      */
-    abstract public function run(SS_List $objs);
+    abstract public function run(SS_List $objs): HTTPResponse;
 
     /**
      * Helper method for responding to a back action request
@@ -49,9 +50,8 @@ abstract class CMSBatchAction
      * key => value pairs, the key can be any string: "error" indicates errors, anything
      * else indicates a type of success. The value is an array. We don't care what's in it,
      * we just use count($value) to find the number of items that succeeded or failed
-     * @return string
      */
-    public function response($successMessage, $status)
+    public function response($successMessage, $status): HTTPResponse
     {
         $count = 0;
         $errors = 0;
@@ -67,16 +67,9 @@ abstract class CMSBatchAction
             }
         }
 
-        $response = Controller::curr()->getResponse();
-
-        if ($response) {
-            $response->setStatusCode(
-                200,
-                sprintf($successMessage ?? '', $count, $errors)
-            );
-        }
-
-        return json_encode($status);
+        return HTTPResponse::create()
+            ->setStatusCode(200, sprintf($successMessage ?? '', $count, $errors))
+            ->setBody(json_encode($status));
     }
 
     /**
@@ -88,7 +81,7 @@ abstract class CMSBatchAction
      * @param string $helperMethod The method to call on each of those objects.
      * @param string $successMessage
      * @param array $arguments
-     * @return string JSON encoded map in the following format:
+     * @return HTTPResponse with a body set to JSON encoded map in the following format:
      *  {
      *     'modified': {
      *       3: {'TreeTitle': 'Page3'},
@@ -99,7 +92,7 @@ abstract class CMSBatchAction
      *     }
      *  }
      */
-    public function batchaction(SS_List $objs, $helperMethod, $successMessage, $arguments = [])
+    public function batchaction(SS_List $objs, $helperMethod, $successMessage, $arguments = []): HTTPResponse
     {
         $status = ['modified' => [], 'error' => [], 'deleted' => [], 'success' => []];
 
--- a/code/CMSBatchActionHandler.php
+++ b/code/CMSBatchActionHandler.php
@@ -125,11 +125,8 @@ class CMSBatchActionHandler extends RequestHandler
 
     /**
      * Invoke a batch action
-     *
-     * @param HTTPRequest $request
-     * @return HTTPResponse
      */
-    public function handleBatchAction($request)
+    public function handleBatchAction(HTTPRequest $request): HTTPResponse
     {
         // This method can't be called without ajax.
         if (!$request->isAjax()) {
@@ -160,11 +157,8 @@ class CMSBatchActionHandler extends RequestHandler
 
     /**
      * Respond with the list of applicable pages for a given filter
-     *
-     * @param HTTPRequest $request
-     * @return HTTPResponse
      */
-    public function handleApplicablePages($request)
+    public function handleApplicablePages(HTTPRequest $request): HTTPResponse
     {
         // Find the action handler
         $action = $request->param('BatchAction');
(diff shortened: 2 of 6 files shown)


code/ModelAdmin.php
```diff
@@ -701,13 +701,8 @@ public function ImportForm()
...
      *
      * @todo Figure out ajax submission of files via jQuery.form plugin
-     *
-     * @param array $data
-     * @param Form $form
-     * @param HTTPRequest $request
-     * @return bool|HTTPResponse
      */
-    public function import($data, $form, $request)
+    public function import(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 705: This method returns `false` a few times which obviously won't work with this new method signature.

code/LeftAndMain.php
```diff
@@ -1347,7 +1311,7 @@ public function getNewItem($id, $setID = true)
         return $object;
     }
 
-    public function delete($data, $form)
+    public function delete(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 1314: Returning `HTTPResponse` here requires that the response negotiator will always return `HTTPResponse` - can we guarantee that, even with custom project code?

code/LeftAndMain.php
```diff
@@ -1281,13 +1249,9 @@ protected function getSearchFilter()
...
     /**
-     * Save  handler
-     *
-     * @param array $data
-     * @param Form $form
-     * @return HTTPResponse
+     * Save handler
      */
-    public function save($data, $form)
+    public function save(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 1254: This method sometimes returns `Security::permissionFailure()` which sometimes returns an array.

code/CMSProfileController.php
```diff
@@ -77,7 +78,7 @@ public function canView($member = null)
         return false;
     }
 
-    public function save($data, $form)
+    public function save(array $data, Form $form): HTTPResponse
```
ChloeSartorelli, line 81: In another place in this PR you've changed `return $this->httpError()` to remove the `return` - for consistency the same should probably be done in this method.

code/CMSBatchActionHandler.php
```diff
@@ -125,11 +125,8 @@ public function Link($action = null)
 
     /**
      * Invoke a batch action
-     *
-     * @param HTTPRequest $request
-     * @return HTTPResponse
      */
-    public function handleBatchAction($request)
+    public function handleBatchAction(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 129: Could `$actionHandler->run()` in custom code return something other than `HTTPResponse`? If so, this return type is too restrictive.

code/CMSBatchAction.php
```diff
@@ -67,16 +67,9 @@ public function response($successMessage, $status)
...
-            $response->setStatusCode(
-                200,
-                sprintf($successMessage ?? '', $count, $errors)
-            );
-        }
-
-        return json_encode($status);
+        return HTTPResponse::create()
+            ->setStatusCode(200, sprintf($successMessage ?? '', $count, $errors))
+            ->setBody(json_encode($status));
```
ChloeSartorelli, line 72: This change will mean any modifications made on the Controller's `HTTPResponse` object won't apply to the back action's response - is that an okay behavioural change to make?

GuySartorelli (changes requested):

code/LeftAndMain.php
```diff
@@ -874,17 +852,14 @@ public function redirect($url, $code = 302)
...
             return parent::redirect($url, $code);
         }
     }
 
-    /**
-     * @param HTTPRequest $request
-     * @return HTTPResponse
-     */
-    public function index($request)
+    public function index(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 862: This will mean any custom `ModelAdmin` or other `LeftAndMain` subclasses that want to override `index()` will have to return `HTTPResponse` explicitly, where previously other types were possible. It's unclear if that fails the acceptance criteria since technically that's not a custom action....

GuySartorelli (changes requested):

code/CMSBatchActionHandler.php
emteknetnz: Yes it could. I've reverted the changes to this method

code/CMSProfileController.php
emteknetnz: Updated

code/LeftAndMain.php
emteknetnz: I think this is an OK change to make
a) It's a very obvious error to fix on upgrade to CMS 5 - you'll see it as soon as you go to /admin
b) It's unlikely that ModelAdmin's will even implement `index()` - none of the core classes that `extends ModelAdmin` implement `index()`

code/LeftAndMain.php
emteknetnz: It never returns an array.  The array return is for an anonymous function https://github.com/silverstripe/silverstripe-framework/blob/4/src/Security/Security.php#L324

code/LeftAndMain.php
emteknetnz: We cannot, however I think it's fine to change this here because it seems pretty like a fairly unlikely thing to get swapped out in custom code.  The only response negotiator used in admin is `PjaxResponseNegotiator`

If if it has been swapped out, this error should show up pretty quickly on CMS 5 upgrade during admin smoke testing.  Realistically the only thing other than an HTTPReponse returned here seems like it would be a string, and it's easy enough to just `return HTTPResponse::create()->setBody($mystring);`

code/ModelAdmin.php
emteknetnz: Updated

code/CMSBatchActionHandler.php
emteknetnz: Possibly it could in CMS 4, though we're updating the signature on the abstract method `CMSBatchAction::run()` to include the `HTTPResponse` return type i.e. `abstract public function run(SS_List $objs): HTTPResponse;`

Because we're updating this, any project that has a custom CMSBatchAction (which honestly seems pretty unlikely) will immediately see the the return type on their `run()` method needs to be updated.

code/CMSBatchAction.php
emteknetnz: Yup that's fine, this is just a helper method used by CMSBatchAction::batchaction() which itself is used by the core implementations of batches actions e.g. `CMSBatchAction_Publish::run()`

These helper methods probably shouldn't really have been made public in the first place

code/LeftAndMain.php
ChloeSartorelli: Well colour me embarrassed. Sorry about that.

code/CMSBatchAction.php
ChloeSartorelli: The visibility of the method is irrelevant here imo. You can change a `Controller`'s response from any number of places to change headers, etc.
Though on further reflections I think in this case the old behaviour of fetching the controller and modifying it's response probably was incorrect anyway so no worries.

code/LeftAndMain.php
ChloeSartorelli: It's not that unlikely. I know I've done it before, and given a custom `ModelAdmin` or other `LeftAndMain` subclass is effectively the same as a custom `Controller` (i.e. they both have an `index()` method that should behave the same regarding return types) I don't think we should restrict the return value here.

That said, I'm not going to stubbornly sit on this rock until you revert the change - if you feel strongly about this then you can keep the change, just call it out explicitly in the changelog.

GuySartorelli (changes requested):

code/LeftAndMain.php
emteknetnz: Have retained and updated developer-docs

GuySartorelli (approved):
Don't merge until the other related PRs are ready to avoid breaking CI

PR #1384, opened 2022-10-18
API Update deprecations

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10531

### DIFF
--- a/_legacy/GraphQL/ReadOneLegacyResolver.php
+++ b/_legacy/GraphQL/ReadOneLegacyResolver.php
@@ -18,7 +18,7 @@ if (!class_exists(ReadOne::class)) {
 /**
  * Shim to make readOne work like GraphQL 4
  *
- * @deprecated 1.8.0 Use _graphql directory functionality instead
+ * @deprecated 1.8.0 Use the latest version of graphql instead
  */
 class ReadOneLegacyResolver implements OperationResolver
 {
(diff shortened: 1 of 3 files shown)


GuySartorelli (approved):

GuySartorelli (changes requested):
Versions are incorrect - use `1` major instead.
This will need to be applied to any other deprecations in this module, as well.

_legacy/GraphQL/ReadOneLegacyResolver.php
```diff
@@ -17,7 +18,7 @@
 /**
  * Shim to make readOne work like GraphQL 4
  *
- * @deprecated 4.8..5.0 Use silverstripe/graphql:^4 functionality.
+ * @deprecated 1.8.0 Use _graphql directory functionality instead
```
ChloeSartorelli: Sorry, I should have caught this the first time around. This should say something like "use the latest version of silverstripe/graphql" - the directory isn't the point.

GuySartorelli (changes requested):

_legacy/GraphQL/ReadOneLegacyResolver.php
emteknetnz: Updated

GuySartorelli (approved):
MOG

PR #1386, opened 2022-10-18
FIX Don't try to parse non-existent gridfield state

### LINKED ISSUE #1385: CMS 4.12: console error for any page with an elemental area (due to gridfield js)
When loading a page that has an elemental area, the following console error is thrown:

> Uncaught exception SyntaxError: JSON.parse: unexpected character at line 1 column 1 of the JSON data
    getState GridField.js:146

This is caused by recent changes to gridfield which results in the `getState()` method being called on `onmatch` - which didn't used to happen. Elemental areas are a subclass of `GridField` but don't have normal gridfield state.

This doesn't seem to cause any actual problems but it's a bad look and it could potentially cause regressions in the future.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1386

GuySartorelli:
## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1385


michalkleiner (approved):
LGTM

michalkleiner:
@GuySartorelli one test seems to fail after the merge — https://github.com/silverstripe/silverstripe-admin/actions/runs/3271324582 — do you think the timeout of 3x 200ms is perhaps too short? What else could be the reason for the failure apart from that?

GuySartorelli:
Intermittent failures for behat in ci is extremely common unfortunately - which is what this appears to be since it's only one of the behat jobs failing. 
I've set the job to run again and expect it to pass. 

Neither the underlying problem here nor the fix would actually affect behat runs since they don't seem to affect the actual behaviour. It was just a console warning.

michalkleiner:
Cool, thanks for the clarification 👍

Issue #1312, opened 2022-06-25
Upgrade frontend build stack for CMS5


maxime-rainville:
Our current JS/CSS build stack is horribly outdated. We need to refresh this to a more modern stack.

# Acceptance criteria
- [x] All our JS dependencies are supported
- [x] Default to running [Node v18](https://nodejs.org/en/about/releases/)
- [x] Where an LTS version of a dependency is available, this is preferred to a shorter lived version
- [x] All our module with front-end clients are upgraded to work with the new stack
- [x] Reassess whether there's value swapping to NPM instead of yarn
  - Decision: They're effectively equivalent. Keeping yarn means we don't have to rewire our brains so stick with that for now. In the future we can consider if some faster alternative is worthwhile, since npm and yarn are notoriously slow, but for now that's not a big enough issue to worry about
- [x] Reassess whether there's value in changing the way to build our CSS
  - Decision: Not at this stage.
- [x] This changes is not back ported to CMS4 as part of this card
 

# Note
- If it's possible to back-port this build stack to CMS4, we should consider it. But it's not the primary purpose of this card.
- Because of the weirdness of our current stack, it might be more cost affective to scrap our current build system and swap to something more generic. But we are not sure yet. We are not very attached to the current build system. Who ever picks up this card should make an initial assessment before proceeding either way.
- We are 3 majors behind Webpack.
- We might want to use [the clean option](https://webpack.js.org/guides/output-management/#cleaning-up-the-dist-folder) to ensure our `dist/` directory never has old unused files.

## CI
All PRs are included in this behat-only CI run on installer:
https://github.com/creative-commoners/silverstripe-installer/actions/runs/3709651281

## PRs
- https://github.com/silverstripe/eslint-config/pull/12
- https://github.com/silverstripe/webpack-config/pull/53
- https://github.com/silverstripe/webpack-config/pull/54 **DO THIS ONE FIRST**, I'll need to re-run `yarn install` afterward so we get the new tag.
- https://github.com/silverstripe/silverstripe-admin/pull/1389
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1303
- https://github.com/silverstripe/silverstripe-cms/pull/2795
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/248
- https://github.com/silverstripe/silverstripe-session-manager/pull/123
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/263
[...]

PR #1389, opened 2022-11-03
DEP Upgrade front-end build stack

GuySartorelli:
## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1312


client/src/components/ActionMenu/ActionMenu.js
```diff
@@ -20,9 +20,9 @@ class ActionMenu extends PureComponent {
       toggleCallback(event);
     }
 
-    this.setState({
-      isOpen: !this.state.isOpen
-    });
+    // Force setting state to the end of the execution queue to clear a potential race condition
+    // with entwine click handlers
+    window.setTimeout(() => this.setState({ isOpen: !this.state.isOpen }), 0);
```
ChloeSartorelli, line 25: I've added this in a couple of places defensively - the only place I _know for certain_ it's needed is this one.
The race condition occurs like this:
1. User clicks the button
2. React responds first.
    1. React toggles the menu to close, which _removes_ the button that was clicked
3. Entwine wants to respond to the event, but the button isn't in the DOM so it doesn't.

This solution works because it defers closing the dropdown (and thus removing the button) until the _next event cycle_ (see [the MDN docs for `setTimeout`](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#delay)). This allows the current event cycle to continue (and therefore lets entwine respond to the click event) before closing closing the dropdown.
More about the event loop can be read in [MDN's docs on the subject](https://developer.mozilla.org/en-US/docs/Web/JavaScript/EventLoop).

client/src/components/DatetimeField/tests/DatetimeField-test.js
```diff
@@ -185,7 +185,8 @@ describe('DatetimeField without html5 date time field support', () => {
 
     it('should use iso format of date value in the input field', () => {
       expect(inputField.type).toBe('datetime-local');
-      expect(inputField.value).toBe('2017-01-05T02:23');
+      // Note that the inclusion of seconds or milliseconds is unnecessary but acceptable
+      expect(inputField.value).toMatch(/^2017-01-05T02:23(:22|:22.000)?$/);
```
ChloeSartorelli, line 189: The test was failing because the value included the seconds and milliseconds. The actual functionality was fine, as this is still a valid ISO format.

client/src/components/GridFieldActions/GridFieldActions.js
```diff
@@ -6,21 +6,6 @@ import classnames from 'classnames';
...
-    this.state = {
-      dropdownOpen: false
-    };
-  }
-
-  toggle() {
-    this.setState({
-      dropdownOpen: !this.state.dropdownOpen
-    });
-  }
```
ChloeSartorelli, line 22: The `toggle` method and `dropdownOpen` state were never being referenced anywhere so I've removed them.

client/src/components/Link/Link.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { useInRouterContext, Link as RouterLink } from 'react-router-dom';
+
+/**
+ * This component can be used whenever a component which needs a relative link
+ * might be rendered in both context within a router (e.g. asset admin) and
+ * contexts without a router (e.g. inside an image upload modal).
+ */
+function Link({ children, href, ...props }) {
```
ChloeSartorelli, line 10: This is necessary because the update of react-router resulted in the way routes are added being changed - where the route and url both used to be e.g. `/admin/assets/show/1/edit`, the route is now _relative to the base route_, e.g. `assets/show/1/edit`. This means we have to use a react `Link` component when in a react router context so that the correct full path is on the rendered `<a>` tag's `href` attribute..

client/src/components/ResizeAware/ResizeAware.js
```diff
@@ -8,7 +8,7 @@ import PropTypes from 'prop-types';
  *
  * Adapted from https://github.com/FezVrasta/react-resize-aware created by Federico Zivolo.
  */
-export default class ResizeAware extends Component {
```
ChloeSartorelli, line 11: Move `export` to after the proptypes and defaultprops are declared.

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -20,14 +24,29 @@ const MULTI_EMPTY_VALUE = 'unchanged';
...
+        key: 'react-select',
+        insertionPoint: document.querySelector('title'),
+      }),
+    []
+  );
+
+  return <CacheProvider value={cache}>{children}</CacheProvider>;
+}
+
 class TreeDropdownField extends Component {
```
ChloeSartorelli, line 39: Most of the changes in this class are a result of changes in the API for `react-select`.
See the upgrade guides here: https://react-select.com/upgrade we went from v1 all the way to v5.

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -515,45 +535,86 @@ class TreeDropdownField extends Component {
...
+  renderInput({ children, ...props }) {
+    props.id = this.props.id;
+    return <selectComponents.Input {...props}>{children}</selectComponents.Input>;
+  }
+
+  /**
+   * Render the breadcrumbs.
+   * This sits above the options in nested trees, as a way to navigate back
+   */
+  renderBreadcrumbs(breadcrumbs, { cx, getStyles, getClassNames, ...props }) {
```
ChloeSartorelli, line 549: This used to be in `TreeDropdownFieldMenu` which is no longer needed.

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -641,92 +694,103 @@ class TreeDropdownField extends Component {
...
+
+    const Highlight = ({ children }) => (
+      <span className="treedropdownfield__option-title--highlighted">{children}</span>
+    );
+    return this.props.search.length
+      ? mapHighlight(title ? title : '', this.props.search, Highlight)
+      : title;
+  }
+
+  noOptionsMessage({ inputValue }) {
```
ChloeSartorelli, line 706: This used to be in `TreeDropdownFieldMenu` which is no longer needed.

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -641,92 +694,103 @@ class TreeDropdownField extends Component {
...
+    // The value passed in is an array of all selected option objects
+    // i.e. an id and title key must be present for each selected option
+    const rawValue = Array.isArray(this.props.value) ? this.props.value : [this.props.value];
+    let value = this.props.selectedValues.filter(item => rawValue.includes(item.id));
+
+    if (!value.length && this.props.data.hasEmptyDefault) {
+      value = options[0];
     }
 
-    const resetValue = (this.props.data.hasEmptyDefault && !this.props.visible.length)
```
ChloeSartorelli, line 665: No longer needed

client/src/containers/App/App.js
```diff
@@ -1,13 +1,13 @@
...
-import { renderRoutes } from 'react-router-config';
+import { Outlet } from 'react-router';
 
 /**
  * Empty container for the moment, will eventually contain the CMS menu`
  * and apply to document.body, rather than just one specific DOM element.
  */
 const App = ({ route }) => (
-  <div className="app">{renderRoutes(route.routes())}</div>
+  <div className="app"><Outlet /></div>
```
ChloeSartorelli: Rendering the routes is no longer the responsibility of this component. It's handled by `BootRoutes.js` explicitly.

client/src/legacy/AddToCampaignForm.js
```diff
@@ -69,7 +71,11 @@ jQuery.entwine('ss', ($) => {
       const modalSchemaUrl = `${sectionConfig.form.AddToCampaignForm.schemaUrl}/${id}`;
       const title = i18n._t('Admin.ADD_TO_CAMPAIGN', 'Add to campaign');
 
-      ReactDOM.render(
+      let root = this.getReactRoot();
+      if (!root) {
+        root = createRoot(this[0]);
+      }
+      root.render(
```
ChloeSartorelli, line 78: `ReactDOM.render` was deprecated in react 17 and removed in react 18. It would still _technically_ work but it would result in react behaving like it was still v17. See [this blog post](https://reactjs.org/link/switch-to-createroot) for more details.
See also `_clearModal()` below for the replacement of the old `ReactDOM.unmountComponentAtNode()` method.

client/src/lib/dependency-injection/tests/ReducerContainer-test.js
```diff
@@ -1,6 +1,6 @@
 /* global jest, jasmine, describe, beforeEach, it, pit, expect, process */
 
-jest.mock('../MiddlewareRegistry', () => () => ({ add: jest.fn() }));
+jest.mock('../MiddlewareRegistry', () => jest.fn().mockImplementation(() => ({ add: () => {} })));
```
ChloeSartorelli, line 3: This was necessary to fix an error where the mocked class wasn't being returned correctly. The error was along the lines of `MiddlewareRegistry is not a constructor`

client/src/lib/tests/Backend-test.js
```diff
@@ -156,7 +156,7 @@ describe('Backend', () => {
           'Content-Type': 'application/x-www-form-urlencoded',
         });
 
-      return promise
+      promise
```
ChloeSartorelli, line 159: Jest complained that tests can either return a promise or accept a `done` argument - not both.

client/src/lib/tests/ReactRouteRegister-test.js
```diff
@@ -31,7 +31,7 @@ describe('ReactRouteRegister', () => {
       expect(reactRouteRegister.getRootRoute())
         .toEqual({
           path: '/',
-          routes: jasmine.any(Function),
+          routes: expect.any(Function),
```
ChloeSartorelli, line 34: jasmine isn't even included as a dependency, I have no idea how this test was passing in the past.

client/src/styles/legacy/_actionTabs.scss
```diff
@@ -206,13 +206,12 @@ $border: 1px solid darken(#D9D9D9, 15%);
             border: none;
           }
           &.loading {
-            background: transparent url("../images/network-save.gif") no-repeat ($spacer / 2) center;
+            background: transparent url('../../images/network-save.gif') no-repeat ($spacer / 2) center;
```
ChloeSartorelli, line 209: URLs for assets should be relative to the file which references them. In this case that means looking in `client/src/images/` where before this pointed at `client/src/styles/images/`.
I think the old webpack config probably rewrote the links as relative to the scss entrypoint which is just silly.

code/LeftAndMain.php
```diff
@@ -347,7 +348,8 @@ public function getClientConfig()
             // Trim leading/trailing slash to make it easier to concatenate URL
             // and use in routing definitions.
             'name' => $name,
-            'url' => trim($this->Link() ?? '', '/'),
+            'url' => $url,
+            'reactRoutePath' => preg_replace('/^' . preg_quote(AdminRootController::admin_url(), '/') . '/', '', $url),
```
GuySartorelli, line 352: This is important because react routes now must be relative to the root route - i.e. if the root route is `/admin` then the asset-admin route must now be `assets` where it used to be `/admin/assets`

ChloeSartorelli, line 352: This is important because react routes now must be relative to the root route - i.e. if the root route is `/admin` then the asset-admin route must now be `assets` where it used to be `/admin/assets`

package.json
```diff
@@ -9,10 +9,10 @@
     "webpack": "./node_modules/webpack/webpack.js"
   },
   "engines": {
-    "node": "^10.x"
+    "node": "^18.x"
   },
   "scripts": {
-    "build": "yarn && yarn lint && yarn test && NODE_ENV=production webpack -p --bail --progress",
+    "build": "yarn && yarn test && rm -rf client/dist/* && NODE_ENV=production webpack --mode production --bail --progress",
```
ChloeSartorelli, line 15: Linting has been temporarily removed - see https://github.com/silverstripe/silverstripe-admin/issues/1411

ChloeSartorelli, line 15: `rm -rf client/dist/*` is added so that the entire dist directory is cleared out before building the resources. This means that, for example, if `moment-js` removes some file, we won't be retaining an old copy of it.
I expect this will be one of the more controversial changes and would welcome feedback on it.

ChloeSartorelli, line 15: The `-p` option alias has been removed. `--mode production` must now be used explicitly

package.json
```diff
@@ -38,91 +38,87 @@
...
   "homepage": "https://github.com/silverstripe/silverstripe-admin",
   "dependencies": {
-    "apollo-cache-inmemory": "^1.2.1",
-    "apollo-client": "^2.3.1",
-    "apollo-link": "^1.2.2",
-    "apollo-link-batch-http": "^1.2.1",
-    "apollo-link-error": "^1.0.9",
-    "apollo-link-http": "^1.5.4",
-    "apollo-link-state": "^0.4.1",
-    "babel-polyfill": "^6.26.0",
```
ChloeSartorelli, line 48: `babel-polyfill` is deprecated in favour of `core-js`

package.json
```diff
@@ -38,91 +38,87 @@
...
   "resolutions": {
-    "colors": "1.1.2",
-    "eslint": "^4.6.1"
+    "colors": "1.4.0"
   },
+  "browserslist": [
+    "defaults"
+  ],
   "jest": {
+    "testEnvironment": "jsdom",
```
ChloeSartorelli, line 116: jest has changed the way its tests are run - this is required for most of them to pass.

package.json
```diff
@@ -38,91 +38,87 @@
...
+    "moduleNameMapper": {
+      "^react-dom/client$": "react-dom-16",
+      "^react-dom((/.*)?)$": "react-dom-16$1",
+      "^react((/.*)?)$": "react-16$1"
+    },
```
ChloeSartorelli, line 121: This is a temporary _hacky_ workaround for https://github.com/silverstripe/silverstripe-admin/issues/1419

GuySartorelli:
The following upgrade guides may help explain some of the changes:
- https://react-select.com/upgrade
- https://github.com/remix-run/react-router/discussions/8753
- https://github.com/remix-run/react-router/blob/main/docs/upgrading/v5.md
- https://www.apollographql.com/docs/react/migrating/apollo-client-3-migration/#react-apollo
- https://reactjs.org/blog/2022/03/08/react-18-upgrade-guide.html#updates-to-client-rendering-apis

GuySartorelli:
JS CI build failed due to broken tests and linting errors. Those will be handled separately, see https://github.com/silverstripe/silverstripe-admin/issues/1421 and https://github.com/silverstripe/silverstripe-admin/issues/1411

package.json
emteknetnz: The "controversial" bit here is the use of `rm -rf` which won't work on windows environments using the default "command" terminal.  This isn't controversial in the slightest for me, Silverstripe has very poor native windows default "command" support as far as I know and there's no chance that regular Silverstripe development would work.  I'd expect anyone using a windows machine to be using WSL or docker or similar

emteknetnz (approved):

PR #1393, opened 2022-11-21
API Remove deprecated code

sabina-talipova:
### Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10501


code/LeftAndMain.php
```diff
@@ -1681,7 +1663,6 @@ public function currentPageID()
             return $this->getRequest()->param('ID');
         }
 
-        /** @deprecated */
```
emteknetnz, line 1684: Could you leave this in - we'll need to separately what the `".currentPage"` functionality was - originally it looks like someone intended for it to be removed

code/LeftAndMain.php
```diff
@@ -1698,7 +1679,6 @@ public function setCurrentPageID($id)
     {
         $this->pageID = $id;
         $id = (int)$id;
-        /** @deprecated */
```
emteknetnz: Leave this in

emteknetnz (changes requested):

code/LeftAndMain.php
sabina-talipova: Done

code/LeftAndMain.php
sabina-talipova: Done

emteknetnz (approved):

PR #1395, opened 2022-11-28
ENH Remove unnecessary workaround

GuySartorelli:
This workaround was related to an outdated version of jQuery - the vulnerability it was preventing is no longer present in this version of admin.

I thought about adding a behat test but decided against it, as it would effectively be including a proof of concept exploit in the codebase, which could be used to attack any sites which haven't been updated.
I did however _write_ a test, and tested it on an unpatched 4.11, then patched 4.11, and then on this branch. The test failed on the unpatched 4.11 version and passed in both the patched 4.11 version and this branch.

The main impact of this PR is that if a prototype pollution attack is attempted, instead of the CMS just breaking to protect the user it will continue working as expected - and since the vulnerability simply no longer exists there is no problematic behaviour triggered.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1396


GuySartorelli:
Note: CI is failing due to https://github.com/silverstripe/gha-ci/issues/54

michalkleiner:
Since it's an enhancement and 1.12 is in beta already, should this target `1`?

GuySartorelli:
I'm not sure, to be honest. In a way it's a follow-up to the merge-up of the PR which introduced it to _preserve_ the beta state, which didn't include this workaround.

But in another way it is a change from what's currently on this branch (even though it's changing to back to what was in the tagged beta) which arguably means it should target `1`

I'll retarget if that's the consensus. I'm on the fence, myself.

michalkleiner:
I mean it doesn't fix anything nor it addresses an issue in the beta release, so I would probably vote for `1` (which is why I asked) to avoid setting a precedent where we allow ourselves to enhance beta/RC branches with non-critical stuff.

GuySartorelli:
Fair enough. The workaround this PR reverts hasn't made it to the `1` branch yet (merge ups will happen after stable release to avoid the temporary constraint changes making it to `1`) so I'll just set this PR to draft until the stable release and subsequent merge-ups are complete.

maxime-rainville:
While it's technically an enhancement, the security issue that added this bit of code got release post beta. It's a very low risk change, so I'm happy to sneak it in this time.

emteknetnz (approved):
MOG

PR #1405, opened 2022-12-07
FIX overflow hidden from inactive chosen

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/630


michalkleiner:
Why is the dist JS file being updated, do you know? Should it stay the same when only changing scss source file?

emteknetnz:
User error :-)

GuySartorelli:
@emteknetnz The change that's shown here for `vendor.js` is already present in this branch: https://github.com/silverstripe/silverstripe-admin/blob/1.12-release/client/src/bundles/vendor.js

Have you accidentally squashed the css change on top of an existing commit?

emteknetnz:
Rather than try and rebase this again, I've just redone in https://github.com/silverstripe/silverstripe-admin/pull/1418

Issue #1398, opened 2022-11-28
4.12 regression: Search bar's filter button takes all half the with in some context


maxime-rainville:
We recently introduce an option to spell out the full "Search options". The problem is that when in a constrain space (e.g. like the page search), that takes half of the available width

![image.png](https://images.zenhubusercontent.com/5ac42fb74b5806bc2bcd8e13/b6abd5e5-06b3-4e96-9edf-7f48d008882a)

## Acceptance criteria
- [ ] If the Search bar is less than 640px, the Search options button displays only has a down arrow.

## Note
- Maxime did a magical ResizeAware component that detects how width the search bar is and will allow you to resize elements accordingly.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1407

GuySartorelli:
PR merged

PR #1407, opened 2022-12-08
FIX Search box css for compact size

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1398

This SearchBox component already had the ResizeAware functionality applied to it

576px is used elsewhere in the SearchBox component for the 'compact size' version


client/src/components/Search/SearchBox.js
```diff
@@ -90,6 +91,10 @@ class SearchBox extends Component {
     return this.state.tagWidth + existingPadding;
   }
 
+  calculateInputRightPadding() {
+    return this.state.width < 576 ? 121 : 264;
```
ChloeSartorelli, line 97: Where do `121` and `264` come from? These seem like arbitrary numbers. Maybe add a comment indicating what this is actually doing?

client/src/components/Search/SearchBox.js
```diff
@@ -191,7 +196,10 @@ class SearchBox extends Component {
    */
   renderInput() {
     const { id, searchText, onChange, placeholder, name, borders } = this.props;
-    const style = { paddingLeft: `${this.calculateInputLeftPadding()}px` };
+    const style = {
+      paddingLeft: `${this.calculateInputLeftPadding()}px`,
+      paddingRight: `${this.calculateInputRightPadding()}px`
+    };
```
ChloeSartorelli, line 204: Why both left and right, but before it was only doing left padding?

client/src/components/Search/SearchBox.scss
```diff
@@ -102,6 +102,13 @@ $search-box-height: 53px;
   &.collapsed::before {
     transform: rotate(0deg) translateY(0);
   }
+
+  &.search-box--no-label {
+    font-size: 0;
+    margin-right: 9px;
+    margin-top: 13px;
+    padding-right: 5px;
+  }
```
ChloeSartorelli: Why font-size 0? The more standard approach to something like this would be to put the text into a `<span>` and add the `sr-only` css class to that span when we want to visually hide the text.

client/src/components/Search/SearchBox.js
emteknetnz: Right padding was added to fix the bug :-)

emteknetnz: From memory it was because there's an "x" that shows when you type in the text input in chrome, and without the right padding change it was totally out of alignment after the search button text was removed

client/src/components/Search/SearchBox.js
emteknetnz: Alignment of the 'x' that shows in chrome and matches the location of it in 4.11

image

image

Have added comment

client/src/components/Search/SearchBox.scss
emteknetnz: Updated

GuySartorelli (approved):

PR #1418, opened 2022-12-11
FIX Remove overflow hidden from inactive chosen

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/630

Replaces https://github.com/silverstripe/silverstripe-admin/pull/1405

bundle.js changed happened due to update of node/npm to 10.24.1/6.14.12 to match CI - I needed to update my local (from 10.23.0/6.14.8) otherwise the js test would fail on git diff - update - it still failed :/


GuySartorelli:
Replaced with #1422

Issue #1426, opened 2023-01-03
CMS 4.12: btn-loading icon is broken in admin 1.12 


lekoala:
Affected version : 4.12

since the upgrade, the animated icon is not fully displayed (you see only the first two dots, tested on chrome, windows)

from what i can see, it changed from

```html
<div class="btn__loading-icon">
    <span class="btn__circle btn__circle--1"></span>
    <span class="btn__circle btn__circle--2"></span>
    <span class="btn__circle btn__circle--3"></span>
</div>
```

to 

```html
<div class="btn__loading-icon">
    <span class="btn__circle btn__circle--1">
        <span class="btn__circle btn__circle--2">
            <span class="btn__circle btn__circle--3"></span>
        </span>
    </span>
</div>
```

the spans are nested instead of being siblings

## Pull Requests
- https://github.com/silverstripe/silverstripe-admin/pull/1431

GuySartorelli:
This is added through one of two places:
- `client/src/components/FormAction/FormAction.js`: `getLoadingIcon()`
- `client/src/legacy/LefAndMain.js`: `submitForm()`

In both cases, the markup being added looks like this:
```html
<div className="btn__loading-icon" >
  <span className="btn__circle btn__circle--1" />
  <span className="btn__circle btn__circle--2" />
  <span className="btn__circle btn__circle--3" />
</div>
```

My first guess is that the more recent version of jQuery doesn't expect self-closing span tags maybe? So it's possible changing it to `<span ... ></span>` will resolve this - but I haven't tested that at all (or even verified that I can reproduce this)

**Update:** That solves it. I'll create a PR. I'll only change the entwine part - having these be self-closing is correct for react syntax.

lekoala:
@GuySartorelli yes i can confirm it works. jquery 3.5 introduced some changes for self closing tags so that's probably why.

GuySartorelli:
Merged - fix released as [1.12.2](https://github.com/silverstripe/silverstripe-admin/releases/tag/1.12.2)

Issue #1421, opened 2022-12-13
CMS 5: Fix broken JavaScript tests


GuySartorelli:
In some modules, javascript tests are broken after https://github.com/silverstripe/silverstripe-admin/issues/1312
For those modules, I removed the `yarn test` command from `yarn build` so that the resources can be built.

The tests must be fixed and passing, then `yarn test` must be added back into the `yarn build` command.
Use https://github.com/silverstripe/silverstripe-admin/pull/1389 as an example of some changes you may need and acceptable workarounds (e.g. resorting to react 16 for tests, for now)

## Acceptance Criteria
- [x] `yarn test` runs without failures on all modules
- [x] `yarn test` is added back into the `yarn build` command

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1456
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1330
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/256
- https://github.com/silverstripe/silverstripe-cms/pull/2834
- https://github.com/silverstripe/silverstripe-session-manager/pull/128
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/274

GuySartorelli:
Should the following be added?
- silverstripe/campaign-admin
- silverstripe/errorpage
- silverstripe/reports

emteknetnz:
Linked PRs have been merged

Issue #1411, opened 2022-12-11
CMS 5: JavaScript and CSS linting errors


GuySartorelli:
https://github.com/silverstripe/silverstripe-admin/issues/1312 results in some regressions which we've decided to handle as separate issues. This is one of those.

Part of that work necessarily included updating https://github.com/silverstripe/eslint-config, which has mean our linting rules are updated - but the code hasn't changed.

We need to update the code to clear up all the linting issues.

Note also that `yarn build` no longer runs `yarn lint` temporarily because of this. Once the linting errors are fixed, it should be added back in.

## Acceptance criteria
- [ ] `yarn lint` has no warnings or errors across kitchen sink
- [ ] `yarn lint` is added back in to the `yarn build` command

## Notes
- Within reason, it's OK to suppress some linting rules in the core config.
 
## PRs
- https://github.com/silverstripe/eslint-config/pull/13 
- https://github.com/silverstripe/eslint-config/pull/14
- https://github.com/silverstripe/eslint-config/pull/15
- https://github.com/silverstripe/webpack-config/pull/56

Note there's JS job failures on lots of these which is a result of js tests being broken - these will be fixed in this [issue](https://github.com/silverstripe/silverstripe-admin/issues/1421)

- https://github.com/silverstripe/silverstripe-admin/pull/1435
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1317
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/252
- https://github.com/silverstripe/silverstripe-cms/pull/2814
- https://github.com/silverstripe/silverstripe-frameworktest/pull/134
- https://github.com/silverstripe/silverstripe-login-forms/pull/130
- https://github.com/silverstripe/silverstripe-session-manager/pull/126
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/269

GuySartorelli:
@emteknetnz Are we adding `yarn lint` back into the `yarn build` command now? I think we should.

GuySartorelli:
Released as [1.12.1](https://github.com/silverstripe/silverstripe-admin/releases/tag/1.12.1)

GuySartorelli:
Oops closed the wrong issue

GuySartorelli:
All PRs merged

PR #1431, opened 2023-01-08
FIX Correctly close spans for btn loading icon

GuySartorelli:
The self-closing span tags were being incorrectly interpreted as _nested_ spans, instead of sibling spans. This is most likely a result of the recent update of jQuery.

The correct HTML syntax for these is to _not_ have them self-closing.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1426


GuySartorelli:
Note: The JS job will likely fail due to https://github.com/silverstripe/silverstripe-admin/issues/1411 and https://github.com/silverstripe/silverstripe-admin/issues/1421, which is expected and should not block this PR from being merged.

michalkleiner (approved):
Looks ok, merge when you're happy with the tests, not sure if the failures are related or not. The change is fine on its own.

GuySartorelli:
Rebased and rebuilt js - the actual change is unaltered.

lozcalver (approved):

lozcalver:
> Running git diff
git diff found modified files when it should not have:
M js/bundle.js

Do we block merge for that? If not, feel free to self merge @GuySartorelli

GuySartorelli:
We do, yeah. It means that when CI runs `yarn build` it's not getting the same results I got - in a community PR I'd be concerned they're trying to add something sneaky in there. In this PR I'm not sure what's causing it, but it's something that definitely needs resolving before we can merge.

maxime-rainville:
@kinglozzer An attacker could try to build a bundle that doesn't match the PR and sneak a backdoor in the bundle.

maxime-rainville:
I'm 90% sure the failure is a glitch. I forced another build, which will hopefully fix this.

lozcalver:
@GuySartorelli could you try rebuilding the bundle locally? CI checks are still failing 😞

GuySartorelli:
I have tried that already (I think? I can't see it in the PR history though..... maybe my memory is playing tricks) but will try again. We've had this problem once before with one of Steve's PRs, and when I created an identical PR to try to debug it my PR didn't have this problem.... so if it doesn't work this time I'll try opening a new PR (in draft initially just to see if it works) - and failing that I guess someone can build it locally and do a diff to validate that they get the same results I push up?
We'll see.

GuySartorelli:
There was a change when I `yarn build`ed for some reason... not sure why but hopefully that fixes it.
Hurray for CI catching things yarn itself didn't.

GuySartorelli:
This is finally green and already has two approvals. Self-merging.

GuySartorelli:
@lekoala Released as [1.12.2](https://github.com/silverstripe/silverstripe-admin/releases/tag/1.12.2)

PR #1432, opened 2023-01-09
ENH Resolve various problems in support of fixing broken behat tests

GuySartorelli:
## Note
This PR includes several _separate_ commits. I've kept it in a single PR to avoid merge collisions and because ultimately this PR exists to support resolving broken tests so they're related in that purpose.

Note also that the items exposed in commits 2 and 3 won't be usable in other modules until https://github.com/silverstripe/webpack-config/pull/55 is released, tagged, and published in npm.

## Commit 1
Fixes part of https://github.com/silverstripe/silverstripe-cms/actions/runs/3726652553/jobs/6320241058

After `".Select-multi-value-wrapper" element not found` is resolved by a combination of https://github.com/silverstripe/silverstripe-cms/pull/2811 and https://github.com/silverstripe/silverstripe-framework/pull/10635, we get `The text "About Us" was not found in the text of the element matching css ".treedropdownfield__value-container".`

This PR addresses that by ensuring we _do_ set a value if we have a value and it's available in the options.

## Commit 2
https://github.com/silverstripe/silverstripe-cms/pull/2813 introduces another component (`CreatableSelect`) which uses emotion css, so would have needed its own separate emotion cache provider if we kept it the way it was. Exporting cache provider resolves this problem.

## Commit 3
We previously had an issue with apolloclient not being exposed correctly, but a workaround was found. Since we have to release a new alpha of webpack-config for commit 2 anyway, it makes sense to expose this at the same time.

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/655


client/src/containers/EmotionCssCacheProvider/EmotionCssCacheProvider.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import createCache from '@emotion/cache';
+import { CacheProvider } from '@emotion/react';
+
+/**
+ * This ensures that Emotion's styles are inserted before ours so we can override
+ * the default react-select styling
+ */
+function EmotionCssCacheProvider({ children }) {
```
maxime-rainville, line 9: My instinct would have been to write this as an HOC. But it's also fine the way it is.

maxime-rainville (changes requested):
TreeDropdownField.js is used to render the regular TreeDropdown outside of React views.

This means it can be rendered outside the context of `<App />`.

Your PR seems to be breaking this.

**With the PR**
Screenshot from 2023-01-11 12-15-36

**Without PR**
image

GuySartorelli:
> TreeDropdownField.js is used to render the regular TreeDropdown outside of React views.
> This means it can be rendered outside the context of <App />.
> Your PR seems to be breaking this.

Shoot, good point... that means we're going to have to expose the new cache component (so `silverstripe/cms` at least can use it), which means a new alpha release of `@silverstripe/webpack-config` - but we needed to do that eventually anyway to expose apolloclient so I'll do that at the same time.

GuySartorelli:
@maxime-rainville I've resolved the issue with this one. There's also now a corresponding PR for webpack-config.
Both of these will need to be merged before I can update the cms PR to use the new component.

maxime-rainville (approved):
Tested locally and it works great.

PR #1438, opened 2023-01-12
ENH Allow base URL to not have trailing slash

GuySartorelli:
https://github.com/silverstripe/silverstripe-framework/pull/10538 allows us to have a base URL which doesn't have a trailing slash. This PR updates the admin javascript to stop explicitly expecting one.

## Parent issue
- silverstripe/silverstripe-cms#2780


client/src/lib/urls.js
```diff
@@ -0,0 +1,7 @@
+export const joinLinks = (...links) => {
```
michalkleiner: The name is strange as it's not joining links, but parts of a link/URL. Perhaps `joinUrlParts`? or `joinLinkParts`? Or make it actually a universal helper in the style of `implode` that takes parts and the glue?

Could be something like https://locutus.io/php/strings/implode/ (simplified) or just used as a dependency.

ChloeSartorelli: Yeah I didn't spend much time thinking about the name tbh, it was just meant to mimic the most common use-case for `Controller::join_links()` in php-land. But it doesn't handle all of the cases that method does so you're right that it can't _truly_ join all links as the name might imply.

I'll update the relevant PRs to change this to `joinUrlPaths` as that's really what it's doing - it is only concerned with the _path_ portion of a url.

While I could make it a universal helper, the intention is to explicitly only include _one_ of the character in-between the two parts (to avoid ending up with `some///path` as an extreme example). The `implode` implementation wouldn't do that, and making this more generic I think would necessarily involve adding additional complexity to make that an optional behaviour - and frankly I've spent too long on this already 😅

michalkleiner: Fair enough, changing to `joinUrlParts = (...urlParts) => {` will suffice.

client/src/lib/Router.js
```diff
@@ -27,8 +27,9 @@ function resolveURLToBase(path) {
     return absolutePath;
   }
 
-  // Remove base url from absolute path, save for trailing `/` which Page.js requires
-  return absolutePath.substring(absoluteBase.length - 1);
+  // Remove base url from absolute path, save for leading `/` which Page.js requires
+  const regex = new RegExp(`^${absoluteBase}/?`);
```
emteknetnz: Should regex escape asbsoluteBase e.g. `.` should become `\.`

Though why not just use the substring above? then ltrim any '/' and then prefix '/' back in

client/src/legacy/LeftAndMain.js
```diff
@@ -55,7 +55,7 @@ window.ss.tabStateUrl = function() {
   return window.location.href
     .replace(/\?.*/, '')
     .replace(/#.*/, '')
-    .replace($('base').attr('href'), '');
+    .replace(new RegExp(`^${$('base').attr('href')}/?`), '');
```
emteknetnz: Should regex escape the base e.g. `.` should become `\.`

Or just `ltrim('/')` the base and then prefix it with `'/'` (better imo)

emteknetnz (changes requested):

client/src/lib/urls.js
```diff
@@ -0,0 +1,8 @@
+export const joinUrlPaths = (...links) => {
```
emteknetnz: Unit tests?

emteknetnz (changes requested):

client/src/lib/Router.js
ChloeSartorelli: `rtrim` and `suffix` - it's meant to be on the end. But regardless, there's no simple `rtrim` function in javascript.

And even if there was, this doesn't just "remove and add a slash", it's removing the `absoluteBase` from the `absolutePath`. e.g. if `absoluteBase` is `https://www.example.com/` and `absolutePath` is `https://www.example.com/some-path/`, the result of this operation is `/some-path/`.

You're correct that the variable needs to be escaped, though. I'll use lodash for that.

client/src/legacy/LeftAndMain.js
ChloeSartorelli: There's no `ltrim` or `rtrim` in javascript. I don't see any downsides to using regex for this. I'll escape the string appropriately.

client/src/lib/urls.js
ChloeSartorelli: Done, except I used `paths` instead of `parts` because I think it's more correct (it doesn't distinguish between, say, query strings or schemes etc so you shouldn't use it to join _the different parts of a URL_ together), and because I already made that change before your last comment :p

client/src/lib/Router.js
ChloeSartorelli: Done

client/src/legacy/LeftAndMain.js
ChloeSartorelli: Done

client/src/lib/urls.js
ChloeSartorelli: Added

client/src/legacy/HtmlEditorField.js
```diff
@@ -289,7 +290,7 @@ ss.editorWrappers.tinyMCE = (function() {
       if(cb) href = eval(cb + "(href, node, true);");
 
       // Turn into relative, if set in TinyMCE config
-      if(cu && href.match(new RegExp('^' + tinyMCE.settings['document_base_url'] + '(.*)$'))) {
+      if(cu && href.match(new RegExp('^' + escapeRegExp(tinyMCE.settings['document_base_url']) + '(.*)$'))) {
```
ChloeSartorelli, line 293: This isn't directly related to the _purpose_ of this PR, but I'm adding escapeRegExp in this PR so it makes sense to apply it in the one other place I found that part of a URL is used as a variable in `new RegExp()`.

emteknetnz (changes requested):
New unit test failure

GuySartorelli:
That unit test will fail until the framework PR is merged. Refer to the installer build in the main issue.

emteknetnz (approved):

PR #1429, opened 2023-01-04
Ignore "load" statusText

lekoala:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/1428

PR #1444, opened 2023-01-18
FIX: Don't show success HTTP status messages as popups (fixes #1428)

### LINKED ISSUE #1428: 1.12 model admin show random "load" notification (with CDNs specifically)
Not sure exactly why, but this is happening since I upgraded to 1.12.

If I navigate from a Model Admin screen to a record, I get this

image

It only happens for some type of records, not all I've tried to trim down the fields to see if there is anything specific, but no luck so far.

it only happens the first time I load a record, after that, no more notification (i can go back to the list, load another one, etc...)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1429 (CMS 4)
- https://github.com/silverstripe/silverstripe-admin/pull/1444 (CMS 5)

lozcalver:
Alternative approach to #1429 

## Parent issue
- #1428


client/src/legacy/LeftAndMain.js
```diff
@@ -166,10 +166,15 @@ $.entwine('ss', function($) {
     var origUrl = document.URL,
       url = xhr.getResponseHeader('X-ControllerURL'),
       destUrl = settings.url,
-      msg = xhr.getResponseHeader('X-Status') !== null ? xhr.getResponseHeader('X-Status') : xhr.statusText, // Handle custom status message headers
+      msg = xhr.getResponseHeader('X-Status'),
       msgType = (xhr.status < 200 || xhr.status > 399) ? 'error' : 'success',
       ignoredMessages = ['OK', 'success', 'HTTP/2.0 200'];
 
+    // For statuses that aren't success or redirect, show the status text if a custom message hasn't already been set
+    if (!msg && (xhr.status < 200 || xhr.status > 399)) {
```
michalkleiner:
```suggestion
    if (!msg && msgType === 'error') {
```
or
```suggestion
    if (!msg && msgType !== 'success') {
```

lozcalver:
I’ve rebuilt the JS and squashed

michalkleiner (approved):

GuySartorelli (changes requested):
With this solution, batch actions success toasts are no longer displayed - which probably means we're doing those the wrong way and they probably don't display in http 2.
https://developer.mozilla.org/en-US/docs/Web/API/Response/statusText
> Note that HTTP/2 does not support status messages. 

I think to get this merged we'll need to find everywhere that sets status text with the intention of setting toast messages and use the header instead.
We also can't do this in a patch release - arguably we can't do it in a minor either because it'll break for someone.

GuySartorelli:
Maybe the solution for CMS 4 is to make `ignoredMessages` configurable (somehow), and for CMS 5 we can make the change this PR proposes.

lozcalver:
>With this solution, batch actions success toasts are no longer displayed - which probably means we're doing those the wrong way and they probably don't display in http 2.

I think that’s a separate issue as batch actions have their own separate code for this: https://github.com/silverstripe/silverstripe-admin/blob/7a2de3e8f1f9166c27b0f3abd0a21e8b736f8df5/client/src/legacy/LeftAndMain.BatchActions.js#L285-L304

I did quickly test setting `X-Status` for batch actions, but it doesn’t work because the `statusMessage()` function (called line 303) isn’t globally available, so it just causes an error:

> Uncaught ReferenceError: statusMessage is not defined

We could probably treat these as two separate issues

GuySartorelli:
The fact is that this PR breaks those toast messages - so if we don't change the way those are output, this is definitely a breaking change.

lozcalver:
> The fact is that this PR breaks those toast messages - so if we don't change the way those are output, this is definitely a breaking change.

Ah sorry, I was testing _without_ this PR applied but I was testing on HTTP2 - your assertion above is correct, batch action messages are already broken in HTTP2 environments. So this would potentially be a breaking change in HTTP1 environments

Should I target this PR to `2.0`? We could then re-open https://github.com/silverstripe/silverstripe-admin/pull/1429 as the fix for the `1.x` release line

GuySartorelli:
Yeah I think that's a good idea.
This PR probably shouldn't be merged into `2.0` without those toasts being changed though. Even though it's kind of a separate issue, this PR would still represent a regression without those toasts using the correct API for toast generation.

GuySartorelli:
The window for this being included in Silverstripe CMS 5 as a breaking change has long passed. As discussed above it would be a breaking change for HTTP/1 environments, which is basically anyone running apache without php-fpm.

I'm going to close this for now - @kinglozzer if you still think this is a valuable change, we'll be starting CMS 6 work in the new year, so we'll be making CMS 6 branches for this sort of thing at that time.

PR #1446, opened 2023-01-19
Support buildstack changes for other modules

GuySartorelli:
This is needed to avoid issues, e.g. with multiple copies of `@emotion` being loaded by `react-select/creatable` vs `react-select` vs `react-select/async` in different modules. If they're all exposed from the same source (admin), they all use the same copy of `@emotion` and therefore there are no conflicts or issues with css from different react-select components.

## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/644


GuySartorelli:
Setting this to draft for now - I'll run a final yarn build when I'm sure I won't be making any other changes.
Edit: This can now be reviewed

sabina-talipova (approved):
Look good for me

PR #1450, opened 2023-01-29
FIX Remove editorIdentifier from global scope

sabina-talipova:
### Description
Removed `var editorIdentifier` from global scope in framework PR https://github.com/silverstripe/silverstripe-framework/pull/10665
### Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2816


sabina-talipova:
CLOSE

PR #1451, opened 2023-01-30
NEW migrate functionality from security-extensions module

GuySartorelli:
DO NO SQUASH! The distinct commits add separate features to keep commit history cleaner.

This adds the following features from the [silverstripe/security-extensions](https://github.com/silverstripe/silverstripe-security-extensions) module, which won't have its own CMS 5 compatible version:
- A more generic circular loading spinner for use within react components
- A React HOC which can be applied to form fields to enforce sudo mode for those fields to be changed
- A controller to allow the react HOC to interact with the sudo mode service

The sudo mode service is added in https://github.com/silverstripe/silverstripe-framework/pull/10666 - that PR must be merged before this one will go green.

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/9203


client/src/components/Loading/Loading.js
```diff
@@ -3,7 +3,7 @@ import React, { PureComponent } from 'react';
 import PropTypes from 'prop-types';
 
 /**
- * Renders a SilverStripe loading animation
+ * Renders a Silverstripe loading animation styled after the Silverstripe logo
```
ChloeSartorelli, line 6: Just being a little more explicit here - I had to render it out to find out what it was, especially since it uses the terminology "spinner" in the css classnames.

client/src/components/Loading/CircularLoading.js
```diff
@@ -0,0 +1,28 @@
+import React, { PureComponent } from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+
+class CircularLoading extends PureComponent {
```
ChloeSartorelli, line 5: Renamed from `LoadingIndicator` to make it clear what it is (since we already have a different loading component here).
Removed "Indicator" from the name to be consistently named with the existing "Loading" component.

tests/php/SudoModeControllerTest.php
```diff
@@ -0,0 +1,162 @@
...
+    protected function tearDown(): void
+    {
+        if ($this->securityTokenEnabled) {
+            SecurityToken::enable();
+        }
+
+        parent::tearDown();
+    }
+
+    public function testNotLoggedIn()
```
ChloeSartorelli, line 50: Added this as it wasn't being tested before

tests/php/SudoModeControllerTest.php
```diff
@@ -0,0 +1,162 @@
...
+
+    public function testLoggedInNotExpired()
+    {
+        $response = $this->get(SudoModeController::singleton()->Link('check'));
+        $this->assertSame(200, $response->getStatusCode());
+        $result = json_decode((string) $response->getBody(), true);
+        $this->assertTrue($result['active'], 'Sudo mode should be active immediately after login');
+    }
+
+    public function testSudoModeExpired()
```
ChloeSartorelli, line 75: This is what the `testCheckReturnsFalse()` used to be. The difference is that now we're _explicitly expiring sudo mode_ first. Otherwise it's still active from when we logged in. We're doing that with the rest of the tests below as well. This way we can be sure that the test is _actually_ activating sudo mode (as opposed to `check()` returning true just because it hadn't expired yet)

tests/php/SudoModeControllerTest.php
```diff
@@ -0,0 +1,162 @@
...
+        // Should redirect to login page
+        $this->assertSame(302, $response->getStatusCode());
+        $this->assertStringEndsWith(
+            Controller::normaliseTrailingSlash('/Security/login?BackURL=%2Fadmin%2Fsudomode%2Fcheck'),
+            $response->getHeader('Location'),
+            'Should not be able to check sudo mode when not logged in'
+        );
+    }
+
+    public function testLoggedInNotExpired()
```
ChloeSartorelli, line 67: When we login, sudo mode is automatically enabled. We login in `setUp()`.
This used to be `assertFalse()` in `testCheckReturnsFalse()` because sudo mode used to only be activated if logging in using a form - which was inconsistent behaviour.
The behaviour is now to activate sudo mode when you log in _regardless_ of how you log in - see the framework PR.

GuySartorelli:
Note that tests will fail until https://github.com/silverstripe/silverstripe-framework/pull/10666 is merged

client/src/components/Loading/CircularLoading.js
```diff
@@ -0,0 +1,28 @@
...
+  render() {
+    const { className, size, block } = this.props;
+
+    const classNames = classnames('ss-circular-loading-indicator', className, {
+      'ss-circular-loading-indicator--block': block,
+    });
+
+    return <div style={{ height: size, width: size }} className={classNames} />;
+  }
+}
```
sabina-talipova, line 15:
```suggestion
import React, { memo } from 'react';
import PropTypes from 'prop-types';
import classnames from 'classnames';

const CircularLoading = memo( function CircularLoading({ className, size, block }){
    
    const classNames = classnames(
               'ss-circular-loading-indicator', className, 
               {
                   'ss-circular-loading-indicator--block': block,
               }
      );

    return <div style={{ height: size, width: size }} className={classNames} />;
});
```

sabina-talipova: Could we also start to migrate to pure function React components if we copy logic from another module?
https://beta.reactjs.org/reference/react/PureComponent#migrating-from-a-purecomponent-class-component-to-a-function

sabina-talipova (changes requested):

client/src/components/Loading/CircularLoading.js
ChloeSartorelli: Honestly we've already pushed back the beta from when we originally wanted to release it, I think we should leave refactoring like that for when we have more time

sabina-talipova:
Could you resolve the conflicts, please.

sabina-talipova:
Failed tests require PR https://github.com/silverstripe/silverstripe-framework/pull/10666

GuySartorelli:
Conflicts resolved after rebase

sabina-talipova (approved):
Test passed. Works perfect.

PR #1453, opened 2023-02-02
API Deprecate SecurityAdmin API

emteknetnz:
Issue https://github.com/silverstripeltd/product-issues/issues/627

### DIFF
--- a/code/GroupImportForm.php
+++ b/code/GroupImportForm.php
@@ -11,14 +11,15 @@ use SilverStripe\Forms\Form;
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\Security\Group;
 use SilverStripe\Security\GroupCsvBulkLoader;
+use SilverStripe\Dev\Deprecation;
 
 /**
  * Imports {@link Group} records by CSV upload, as defined in
  * {@link GroupCsvBulkLoader}.
+ * @deprecated 1.13.0 Will be removed without equivalent functionality to replace it
  */
 class GroupImportForm extends Form
 {
-
     /**
      * @var Group Optional group relation
      */
@@ -26,6 +27,9 @@ class GroupImportForm extends Form
 
     public function __construct($controller, $name, $fields = null, $actions = null, $validator = null)
     {
+        Deprecation::withNoReplacement(function () {
+            Deprecation::notice('1.13.0', 'Will be removed without equivalent functionality to replace it', Deprecation::SCOPE_CLASS);
+        });
         if (!$fields) {
             $helpHtml = _t(
                 __CLASS__ . '.Help1',
--- a/code/SecurityAdmin.php
+++ b/code/SecurityAdmin.php
@@ -7,6 +7,7 @@ use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\Form;
 use SilverStripe\Forms\GridField\GridField;
@@ -30,6 +31,8 @@ use SilverStripe\View\Requirements;
 
 /**
  * Security section of the CMS
+ *
+ * Note: In CMS 5 SecurityAdmin will extend ModelAdmin instead of LeftAndMain
  */
 class SecurityAdmin extends LeftAndMain implements PermissionProvider
 {
@@ -43,7 +46,7 @@ class SecurityAdmin extends LeftAndMain implements PermissionProvider
     private static $tree_class = Group::class;
 
     /**
-     * @deprecated 1.13.0 Do not use this options.
+     * @deprecated 1.13.0 Will be removed without equivalent functionality to replace it
      * @config
      */
     private static $subitem_class = Member::class;
@@ -68,9 +71,13 @@ class SecurityAdmin extends LeftAndMain implements PermissionProvider
      *
      * @param HTTPRequest $request
      * @return HTTPResponse
+     * @deprecated 1.13.0 Will be removed without equivalent functionality to replace it
      */
     public function users($request)
     {
+        Deprecation::withNoReplacement(function () {
+            Deprecation::notice('1.13.0', 'Will be removed without equivalent functionality to replace it.');
+        });
         return $this->index($request);
     }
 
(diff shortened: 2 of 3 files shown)


code/GroupImportForm.php
```diff
@@ -11,21 +11,25 @@
...
     /**
      * @var Group Optional group relation
      */
     protected $group;
 
     public function __construct($controller, $name, $fields = null, $actions = null, $validator = null)
     {
+        Deprecation::withNoReplacement(function () {
+            Deprecation::notice('1.13.0', 'Will be removed without equivalent functionality to replace it', Deprecation::SCOPE_CLASS);
+        });
```
ChloeSartorelli, line 32: I don't think that's how `withNoReplacement()` is meant to work.... isn't it intended simply to wrap calls to deprecated code? So you'll wrap `new GroupImportForm()` in a callback passed into `withNoReplacement()` but not the deprecation notice itself.

We have been a _little_ inconsistent with it (I did find a couple of other usages like this one) - but most "without equivalent functionality" deprecation notices are not directly wrapped like this.
We should probably unwrap the ones that _are_ wrapped.

ChloeSartorelli: Applies to others in this PR and probably in other related PRs.

code/SecurityAdmin.php
```diff
@@ -30,20 +31,25 @@
...
  */
 class SecurityAdmin extends LeftAndMain implements PermissionProvider
 {
 
     private static $url_segment = 'security';
 
+    /**
+     * @deprecated 1.13.0 Will be removed without equivalent functionality to replace it
+     */
     private static $url_rule = '/$Action/$ID/$OtherID';
```
ChloeSartorelli:
```suggestion
    private static $url_rule = '/$Action/$ID/$OtherID';
```
`$url_rule` is very much _not_ deprecated - it comes from `LeftAndMain` and can be configured for SecurityAdmin like it can for any LeftAndMain subclass.

code/SecurityAdmin.php
```diff
@@ -246,8 +274,14 @@ public function MemberImportForm()
         return $form;
     }
 
+    /**
+     * @deprecated 1.13.0 Will be removed without equivalent functionality to replace it
```
ChloeSartorelli, line 275: In hindsight these should probably say "without equivalent API" not functionality, because there _is_ equivalent functionality, just not with an API like this.
Too late to change now though.

code/SecurityAdmin.php
```diff
@@ -276,9 +314,13 @@ public function GroupImportForm()
 
     /**
      * Disable GridFieldDetailForm backlinks for this view, as its
+     * @deprecated 1.13.0 Will be removed without equivalent functionality to replace it
      */
     public function Backlink()
```
ChloeSartorelli, line 316: I think we made a mistake by removing this - see `GridFieldDetailForm_ItemRequest::getBackLink()` - we should probably still explicitly exclude security admin from the back link functionality (I'm presuming the reason it was excluded originally is still valid)

GuySartorelli (changes requested):

code/GroupImportForm.php
emteknetnz: When I originally wrote this it was intended to wrap calls, though then we realised later works the same directly wrapping notices.  When deprecating classes we exclusively use the wrap notices method.  Both ways of doing it are legitimate.

All that wrapping calls does is surpress any notices, so, it's the same thing.  Aesthetically it's not as great, though there's zero practical difference for end-users.

I'm not going to wrap calls just for the sake of aesthetics.

code/SecurityAdmin.php
emteknetnz: Reverted

code/GroupImportForm.php
ChloeSartorelli: Is this going to suppress notices if a community module or project uses the import form directly for some reason? If so, we shouldn't wrap the notice call.

code/SecurityAdmin.php
emteknetnz: I don't understand what you're suggesting we do?

It's removed in CMS5 at this point? So you're saying we should add it back it?  Do we know if it's broken? Is there a card to add BackLink back in?

ChloeSartorelli: Nevermind easier to just keep it gone and deprecate.

code/GroupImportForm.php
emteknetnz: Yes.  But there's no replacement.  That's the point of suppressing notice where there's "no replacement", because there's no migration to go to while you're still on CMS 4, so you'll just get spammed with messages that you cannot do anything about.

You can show suppressed messages via `Deprecation::enable(true);` rather than the regular `Deprecation::enable()`

I've updated the PR to developer-docs to note this - https://github.com/silverstripe/developer-docs/pull/150/files#diff-0d0c6e36c4cca126ccacdc42ab8b4765656ee67badcedc061f2e0c21129bfddc

ChloeSartorelli: I would argue that if someone is using something deprecated, the warning _should absolutely always_ be given, and then they can note it down so they know what they need to do when they upgrade to CMS 5. People are only going to enable deprecation notices when they're doing a major upgrade, so it makes sense to tell them _everything_ they're using that's deprecated. They can then make a note of it, wrap it in `withNoReplacement()` and continue.

In either case we need to make sure we're being consistent. Either always wrap it, or never wrap it. I'm in favour of never wrap it but if you want to update this PR to always wrap notices for API that has no replacement and open one for all other modules which have them I'll accept that too.

emteknetnz: It's simply not worth updating it to be consistent at this stage. This is code that's CMS 4 only and won't have much in the way of active development on it.  It makes zero practical difference.

The vast amount `Deprecation::withNoReplacement()` has already been merged and was done in issues such as https://github.com/silverstripe/silverstripe-framework/issues/10542.  I'm not updating that, not at this late stage.

ChloeSartorelli: Most cases, as I mentioned, do not use `withNoReplacement()`. Consistency in this case is to not wrap it in _this_ PR.

emteknetnz: `Deprecation::SCOPE_CLASS` are always in `__construct()`, so we never wrap calls - this is consistent with the rest of the code base

All the other things on SecurityAdmin are controller endpoints, so we'd need to wrap whatever calls controller endpoints in framework, which is just wrong.

I think the way this PR has it is correct

ChloeSartorelli: I'll accept this - we can just make it clear in the documentation how to deal with these.

GuySartorelli (approved):

Issue #1409, opened 2022-12-11
CMS 5: React context no longer has "You have unsaved changes" confirm dialog


GuySartorelli:
https://github.com/silverstripe/silverstripe-admin/issues/1312 results in some regressions which we've decided to handle as separate issues. This is one of those.

React contexts (such as asset-admin and possibly campaign-admin) used to have a "You have unsaved changes" confirm dialog which would be triggered if you tried to navigate away without saving your changes. This was part of react-router, but that functionality is not included with react-router 6.x - so we need to implement it another way.

## Acceptance criteria
- [x] When a user has unsaved changes and tries to navigate in a react context, a confirm dialog is presented asking if they way to navigate away or not
- [x] If the user selects "no" or "cancel" they do not navigate away and can save their changes
- [x] If the user selects "yes" or "ok" they are navigated away and lose their changes

## Notes
- ~~react-router uses [history](https://github.com/remix-run/history) under the hood. We may be able to utilise its [blocking transitions](https://github.com/remix-run/history/blob/dev/docs/blocking-transitions.md) functionality.~~ It looks like despite _saying_ they use this under the hood, they don't _actually_ use this.
- ~~[react-beforeunload](https://www.npmjs.com/package/react-beforeunload) is another possible way to implement this~~ This might be useful _in addition_ to any solution we come up with, but it doesn't actually handle routing _within_ the app, only _out_ of it.
- There is a lot of discussion about this along with other potential solutions in [this issue thread](https://github.com/remix-run/react-router/issues/8139) which is where the above options came from
    - Seems like [they've added support for this](https://github.com/remix-run/react-router/blob/3a44ce30353f89d57386b75e90e05a79b18b2525/decisions/0001-use-blocker.md) in 6.7 which wasn't available when I did the original upgrade. Lets go with that.
    - Docs are still being written so the only thing we have to go on is [this limited example](https://github.com/remix-run/react-router/blob/main/examples/navigation-blocking/src/app.tsx) and [the source code](https://github.com/remix-run/react-router/blob/2b843a6739c6807e683de9097e20909e78f0a1bd/packages/react-router/lib/hooks.tsx#L842)
- Yes we can technically downgrade to the previous version, however while that version is still maintained for now it doesn't get the same attention as v6. It's better for us to keep the upgrade if we can.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1458
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1335
[...]

PR #1458, opened 2023-02-22
FIX Reinstate unsaved changes dialog

GuySartorelli:
This broke when upgrading to react-router 6.4 - they've added a new API for it in 6.7

I've added behat tests for this in asset-admin: https://github.com/silverstripe/silverstripe-asset-admin/pull/1335

## Notes
- There is a pre-existing issue you'll run into while testing this: https://github.com/silverstripe/silverstripe-admin/issues/2116
    - Fixing that is way out of scope for this PR given it's a pre-existing issue from CMS 4.
- There is no custom message for navigating _outside_ of the react context (i.e. refreshing the page, clicking a link in the left navigation panel, etc). That dialog box comes natively with the browser as part of the [`beforeunload` event](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event). Adding a custom message to it is not supported by modern browsers.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1409


GuySartorelli:
UPDATE: The problem mentioned in this comment is resolved. Leaving here for historical context only.

I've got this to a state where it technically meets the acceptance criteria - except after the initial "yes continue I want to navigate away" it _keeps prompting you every time you go somewhere_ until you go back to the original form. It seems the state of the form is stuck as dirty until you _prove_ (by navigating back to it) that you don't have unsaved changes there anymore.

There's a console error which I think is part of the problem when you accept the prompt the _first_ time:
> ``Warning: Cannot update a component (`re`) while rendering a different component (`c`). To locate the bad setState() call inside `c`, follow the stack trace as described in https://reactjs.org/link/setstate-in-render
c@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/bundle.js?m=1677039314:1:219846
div
./client/src/containers/App/App.js/d<@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/bundle.js?m=1677039314:1:219676
s@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/bundle.js?m=1677039314:38:11650
F@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1677035706:1:1672740
N@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1677035706:1:1672136
ue@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1677035706:1:1678678
le@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1677035706:1:1677857
re@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1677035706:1:1675895
./node_modules/react-redux/es/components/Provider.js/t.default@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1677035706:1:1626487
i@http://react-unsaved-changes_09.local/_resources/vendor/silverstripe/admin/client/dist/js/vendor.js?m=1677035706:1:2510593``

I'm pretty sure react isn't able to update the form state at the same time as the blocker state, and for whatever reason the blocker state gets prioritised, so yes the blocker lets you continue, but each time it asks "should I block this new navigation?" the app tells it "yes, the form is still dirty"

client/src/boot/BootRoutes.js
```diff
@@ -6,7 +6,7 @@ import $ from 'jquery';
 import React from 'react';
 import { Provider as ReduxProvider } from 'react-redux';
 import ReactDOM from 'react-dom/client';
-import { BrowserRouter } from 'react-router-dom';
+import { createBrowserRouter, createRoutesFromElements, RouterProvider } from 'react-router-dom';
```
ChloeSartorelli, line 9: We need to use `createBrowserRouter()` and `createRoutesFromElements()` instead of adding the `BrowserRouter` and `Routes` components ourselves because the `RouterProvider` doesn't like us doing it ourself - and we have to use the `RouterProvider` because this instantiates a "data router" which is necessary to use the new `useBlocker` hook.

client/src/boot/BootRoutes.js
```diff
@@ -35,8 +36,7 @@ class BootRoutes {
     const base = Config.get('absoluteBaseUrl');
     pageRouter.setAbsoluteBase(base);
 
-    this.handleBeforeRoute = this.handleBeforeRoute.bind(this);
-    this.handleBeforeUnload = this.handleBeforeUnload.bind(this);
```
ChloeSartorelli, line 39: Neither of these functions is passed into downstream components anymore, so we don't need explicit binds.

client/src/boot/BootRoutes.js
```diff
@@ -109,18 +109,30 @@ class BootRoutes {
...
       <ApolloProvider client={this.client}>
         <ReduxProvider store={this.store}>
-          <BrowserRouter basename={joinUrlPaths(Config.get('baseUrl'), Config.get('adminUrl'))}>
-            <Routes>
-              <Route path={rootRoute.path} element={<rootRoute.component />}>{routes}</Route>
-            </Routes>
-          </BrowserRouter>
+          <RouterProvider router={router} />
         </ReduxProvider>
       </ApolloProvider>
```
ChloeSartorelli, line 137: Ultimately the only differences here are:
- We now have a `RouterProvider` (which is necessary for the new `useBlocker` hook)
- We're passing a component in as a child of the `App` component (`rootRoute.component`)

See https://github.com/silverstripe/silverstripe-admin/pull/1458#discussion_r1115126349 for the reason we had to change the way we're doing the browserrouter etc.

client/src/boot/BootRoutes.js
```diff
@@ -218,22 +225,10 @@ class BootRoutes {
...
-  }
-
-  handleBeforeRoute(content, callback) {
-    if (this.shouldConfirmBeforeUnload()) {
-      return callback(confirm(i18n._t('Admin.CONFIRMUNSAVED', `Are you sure you want to navigate away
-          from this page?\n\nWARNING: Your changes have not been saved.\n\n
-          Press OK to continue, or Cancel to stay on the current page.`)));
-    }
-
-    return callback(true);
```
ChloeSartorelli, line 236: Neither of these is needed anymore

client/src/components/NavigationBlocker/NavigationBlocker.js
```diff
@@ -0,0 +1,28 @@
+import { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { unstable_useBlocker as useBlocker } from 'react-router-dom';
+
+export default function NavigationBlocker({ shouldBlockFn, blockMessage }) {
+  const blocker = useBlocker(shouldBlockFn);
+  useEffect(() => {
```
ChloeSartorelli, line 7: Has to be in an effect to avoid the console error I mentioned in https://github.com/silverstripe/silverstripe-admin/pull/1458#issuecomment-1439421893

client/src/containers/App/App.js
```diff
@@ -6,8 +6,8 @@ import { Outlet } from 'react-router';
  * Empty container for the moment, will eventually contain the CMS menu`
  * and apply to document.body, rather than just one specific DOM element.
  */
-const App = () => (
-  <div className="app"><Outlet /></div>
+const App = ({ children }) => (
+  <div className="app">{children}<Outlet /></div>
```
ChloeSartorelli, line 10: We have to have the blocker as a child of the root component because it has to be a child of the `RouteProvider` and this is the only component we know for sure will be in there.
But because this replaceable via the `reactRouteRegister` we don't even know that the root component will specifically be `App` so we can't just slap it directly here, we have to pass it in as a child prop.

package.json
```diff
@@ -71,8 +71,8 @@
     "react-dom": "^18.2.0",
     "react-load-script": "^0.0.6",
     "react-redux": "^8.0.4",
-    "react-router": "^6.4.2",
-    "react-router-dom": "^6.4.2",
+    "react-router": "^6.7",
+    "react-router-dom": "^6.7",
```
ChloeSartorelli, line 75: The new `useBlocker` hook was only introduced in 6.7 (I'll update the changelog as well to reflect the new dependency constraint)

emteknetnz (approved):
Tested locally, works good, have confirmed it seems to behave the same was it did in 4.12

PR #1460, opened 2023-02-23
ENH Update translations

emteknetnz:
Automated translations update generated using https://github.com/silverstripe/tx-translator


client/lang/src/da.json
```diff
@@ -35,4 +35,4 @@
     "Admin.VERIFY_TO_CONTINUE": "Bekræft for at forsætte",
     "Admin.ENTER_PASSWORD": "Indtast dit kodeord",
     "Admin.VERIFY": "Bekræft"
-}
+}
```
ChloeSartorelli, line 38: It shouldn't be changing whether there's new lines at ends of files I don't think
But that's not a huge deal

client/lang/src/en.json
```diff
@@ -7,7 +7,7 @@
     "Admin.BATCH_UNPUBLISH_PROMPT": "You have {num} page(s) selected.\n\nDo you really want to unpublish",
     "Admin.SELECTONEPAGE": "Please select at least one page",
     "Admin.FormatExample": "Example: {format}",
-    "Admin.NO_SIZE": "N\/A",
+    "Admin.NO_SIZE": "N/A",
```
ChloeSartorelli, line 10: Is this correct or is it a serialisation/deserialisation error?

lang/en.yml
```diff
@@ -100,7 +100,7 @@ en:
   SilverStripe\Admin\Navigator\SilverStripeNavigatorItem:
     WONT_BE_SHOWN: 'Note: this message will not be shown to your visitors'
   SilverStripe\Admin\Navigator\SilverStripeNavigatorItem_Unversioned:
-    PREVIEW': 'Preview'
+    "PREVIEW'": Preview
```
ChloeSartorelli, line 103: This has highlighted an error in the original yml - should just be `PREVIEW: Preview`

client/lang/src/en.json
emteknetnz: I think it's correct and it was wrong before, in the .js file (which is what the .json file is turned into) - it's "N/A" - https://github.com/silverstripe/silverstripe-admin/blob/1/client/lang/en.js#L17

GuySartorelli:
This has a conflict, and doesn't reference an issue... was this an old one that should be closed?

Issue #915, opened 2019-07-29
LeftAndMain.admin_themes should be moved to YAML


kinglozzer:
It’s currently set in PHP:

https://github.com/silverstripe/silverstripe-admin/blob/f4913c5e52d1cf9cd87700bf5f4698eab827a5c1/code/LeftAndMain.php#L184-L187

That makes it impossible to insert a custom admin theme _before_ the default ones, except by using PHP `Config::modify()`.

## PR
- https://github.com/silverstripe/silverstripe-admin/pull/1461

michalkleiner:
Enhancement by @ntd merged. Will be in the next minor release.

PR #1461, opened 2023-02-28
FIX move admin_themes in YAML (#915)

ntd:
Naive fix for #915.

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -182,15 +182,14 @@ class LeftAndMain extends Controller implements PermissionProvider
     private string $httpErrorMessage;
 
     /**
-     * Assign themes to use for cms
+     * Themes to use within the CMS
+     *
+     * Default themes are assigned in _config/themes.yml
      *
      * @config
      * @var array
      */
-    private static $admin_themes = [
-        'silverstripe/admin:cms-forms',
-        SSViewer::DEFAULT_THEME,
-    ];
+    private static $admin_themes = [];
 
     /**
      * Codes which are required from the current user to view this controller.
--- a/_config/themes.yml
+++ b/_config/themes.yml
@@ -1,0 +1,7 @@
+---
+Name: adminthemes
+---
+SilverStripe\Admin\LeftAndMain:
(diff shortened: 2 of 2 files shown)


_config/forms.yml
```diff
@@ -1,6 +1,11 @@
 ---
 Name: adminforms
 ---
+SilverStripe\Admin\LeftAndMain:
+  admin_themes:
+    - 'silverstripe/admin:cms-forms'
+    - '$default'
+
```
michalkleiner: Please move this block into a new file `_config/themes.yml` and give it `Name: adminthemes`, thanks.

michalkleiner (changes requested):

_config/forms.yml
ntd: :heavy_check_mark:

michalkleiner (approved):
Looks ok to me, what do you think @kinglozzer?

michalkleiner:
Any objections here @GuySartorelli?

code/LeftAndMain.php
```diff
@@ -187,10 +187,7 @@ class LeftAndMain extends Controller implements PermissionProvider
      * @config
      * @var array
      */
-    private static $admin_themes = [
-        'silverstripe/admin:cms-forms',
-        SSViewer::DEFAULT_THEME,
-    ];
+    private static $admin_themes;
```
ChloeSartorelli: Add a note in the PHPDoc that this is set in `_config/themes.yml` - else it just looks like we don't have a default value here.

GuySartorelli (changes requested):
I can't think of any way this would cause problems - seems like just a good change to make.

code/LeftAndMain.php
ChloeSartorelli: Probably also assign an empty array by default
```suggestion
    private static $admin_themes = [];
```

michalkleiner: Done @GuySartorelli

GuySartorelli (approved):
Looks good. Merge on green.

PR #1466, opened 2023-03-06
Suppress admin error context for base 'CMS access' permission check (fixes #1443)

### LINKED ISSUE #1443: CMS 4.12 regression: Incorrect template shown when user doesn’t have CMS access
Likely a regression from https://github.com/silverstripe/silverstripe-admin/pull/1317

If a member without any CMS access permissions is logged in and visits /admin, they will see a blank admin area and a generic error message:

Screenshot 2023-01-17 at 10 40 04

Prior to admin 1.12, the user would be shown a frontend template with a more descriptive error message:

Screenshot 2023-01-17 at 10 39 48

## Acceptance criteria
- Users with no CMS access should not see errors in CMS context.
- ~If there's an 403 error page define, users without CMS access who try to access the CMS should see that error page.~ - this was not the old behaviour, and is actually _less_ useful than the old behaviour which gives you a "log in as someone else" button in case you need to log in as a different user.
- If there's no error page, the previous behaviour is restored.

## Note
- `AdminRootController` loops through all the LeftAndMain. At first glance it would make sense to double check if the Member as access to the CMS before looping through any of those section.
- Traditionally, we consider you have CMS Access if you have any `CMS_ADMIN_*` permission.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1466

lozcalver:
Proposed solution for #1443, haven’t run Behat tests locally so this may fail at first

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -182,6 +182,11 @@ class LeftAndMain extends Controller implements PermissionProvider
     private string $httpErrorMessage;
 
     /**
+     * Used to allow errors to be displayed using a front-end template
+     */
+    private bool $suppressAdminErrorContext = false;
+
+    /**
      * Assign themes to use for cms
      *
      * @config
@@ -660,6 +665,7 @@ class LeftAndMain extends Controller implements PermissionProvider
                 ),
             ];
 
+            $this->suppressAdminErrorContext = true;
             Security::permissionFailure($this, $messageSet);
             return;
         }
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Elegant! Works well locally. Can you please retarget to the `1.12` branch? We should release this as a patch.
Edit: I've retargetted it. Looks like there's no clash that would require fixing commits after retargetting.

GuySartorelli (approved):

GuySartorelli:
Released as [1.12.3](https://github.com/silverstripe/silverstripe-admin/releases/tag/1.12.3)

Issue #1416, opened 2022-12-11
Clear up errors and warnings when running yarn commands


GuySartorelli:
Currently when running `yarn dev` and `yarn build`, there are a _ton_ of warnings that pop up. The sheer volume of warnings means we ultimately just ignore them, which means we'll never know if it's indicating an actual problem that needs to be resolved.

Some of these may be related to linting - those will be handled in https://github.com/silverstripe/silverstripe-admin/issues/1411

## Acceptance criteria
- [ ] When running `yarn dev` or `yarn build` in any applicable module in kitchen sink there are no warnings or errors
- [ ] Maybe fail CI if there are warnings?

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1468
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1337
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/258
- https://github.com/silverstripe/silverstripe-ckan-registry/pull/284
- https://github.com/symbiote/silverstripe-advancedworkflow/pull/482
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/277
- https://github.com/silverstripe/silverstripe-externallinks/pull/97

maxime-rainville:
Left some notes on all the PRs.

Can we confirm that the following modules don't need any updates?
- blog
- cms
- elemental
- session-manager
- fluent
- userform
- versioned-admin

sabina-talipova:
Checked another modules:
- blog
- cms
- elemental
- session-manager
- fluent
- userform

They don't have any front-end issues

Add PR for:
- versioned-admin

maxime-rainville:
Great job on fixing all those react life cycle method.

PR #1467, opened 2023-03-07
FIX Changed import for createRoot function

sabina-talipova:
### Description
- A console warning `Warning: You are importing createRoot from "react-dom" which is not supported. You should instead import it from "react-dom/client".` appears on `yarn dev` command. [See source](https://github.com/reactwg/react-18/discussions/125).
Fixed import for `createRoot` function in `BootRoutes.js`

### Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1416


sabina-talipova:
WRONG PR

PR #1473, opened 2023-03-16
FIX Make graphql template abstractions work with gql4

GuySartorelli:
There are a few problems being fixed here:
- update and create templates just simply didn't support GraphQL 4 at all, since they hardcoded `Input` (`input` with a lowercase `i` is standard in v4) and had hardcoded types (`XCreateInputType` and `XUpdateInputType`, where `CreateXInput` and `UpdateXInput` are standard for gql 4).
- readone and delete templates didn't support GraphQL 4 unless you intentionally set your schema to use non-standard uppercase `ID`, where `id` is standard in gql 4
- read template would result in broken queries if you input any parameters, as it would add them to the variables list but then not use the variables.
- Any templates that used `getParams` (used to only be the base query template) would fail if params were passed in, since it was returning `param: $type` where it should have passed in `param: $variablename` where the variable name should be the same as the parameter name anyway.
- The base query template would have failed if there were any params, because it had double brackets around them.

## NOTES:
- Obviously to avoid breaking backwards compatability I've had to include graphql v3 defaults as fallbacks.
After this PR is merged, the issue should _not be closed_ until a follow-up PR is created for CMS 5 which updates those to be graphql v4 defaults as fallbacks.
- This is mostly being done so that documentation in https://github.com/silverstripe/developer-docs/pull/187 actually works. You can use the examples in that PR to help test/review that this code works as expected.

## Issue
- https://github.com/silverstripe/developer-docs/issues/62


GuySartorelli:
Note, while there are some instances of these templates in kitchen sink, the templates I found _weren't actually used anyway_ - this seems to be a feature that was implemented and documented with the intention of being used, but never actually ended up being used in our code which explains why it was allowed to get to this state in the first place.

client/src/lib/dependency-injection/graphql/buildBaseQuery.js
```diff
@@ -12,7 +12,7 @@ const getQueryType = ({ queryType }) => queryType;
 
 const buildReadQuery = (tag = defaultTag) => (
   tag`${getQueryType} ${getOperationName}${getVariables} {
-    ${getQueryName}(${getParams}) {
+    ${getQueryName}${getParams} {
```
ChloeSartorelli, line 15: This would have ended up as `readMyModel((param: $param)) {` which would result in a syntax error.

client/src/lib/dependency-injection/graphql/buildCreateMutation.js
```diff
@@ -1,13 +1,9 @@
...
-    $Input:${getSingularName}CreateInputType!
-  ) {
-    create${getSingularName}(
-      Input: $Input
-    ) {
+  tag`mutation Create${getSingularName}${getMutationVariables} {
+    create${getSingularName}${getMutationParams} {
       ${getFields}
     }
   }
```
ChloeSartorelli, line 9: Turning `$Input:${getSingularName}CreateInputType!` and `Input: $Input` into default fallbacks, which can be overridden by setting explicit params. See the new related test for an example.

client/src/lib/dependency-injection/graphql/buildDeleteMutation.js
```diff
@@ -1,9 +1,9 @@
...
-import { getSingularName } from './helpers';
+import { getSingularName, getIdOnlyVariables, getIdOnlyParams } from './helpers';
 
 const buildDeleteMutation = (tag = defaultTag) => (
-  tag`mutation Delete${getSingularName}($IDs:[ID]!) {
-    delete${getSingularName}(IDs: $IDs)
+  tag`mutation Delete${getSingularName}${getIdOnlyVariables} {
+    delete${getSingularName}${getIdOnlyParams}
   }`
 );
```
ChloeSartorelli, line 8: Turning `$IDs:[ID]!` and `IDs: $IDs` into default fallbacks, which can be overridden by setting explicit params. See the new related test for an example.

client/src/lib/dependency-injection/graphql/buildReadOneQuery.js
```diff
@@ -1,9 +1,9 @@
 import { defaultTag } from './tags';
-import { getSingularName, getFields, getFragments } from './helpers';
+import { getSingularName, getIdOnlyVariables, getFields, getFragments, getIdOnlyParams } from './helpers';
 
 const buildReadOneQuery = (tag = defaultTag) => (
-  tag`query ReadOne${getSingularName}($ID: ID!)  {
-    readOne${getSingularName}(ID: $ID) {
+  tag`query ReadOne${getSingularName}${getIdOnlyVariables}  {
+    readOne${getSingularName}${getIdOnlyParams} {
```
ChloeSartorelli, line 6: Turning `$ID: ID!` and `ID: $ID` into default fallbacks, which can be overridden by setting explicit params. See the new related test for an example.

client/src/lib/dependency-injection/graphql/buildReadQuery.js
```diff
@@ -1,9 +1,9 @@
 import { defaultTag } from './tags';
-import { getPluralName, getVariables, getRootParams, getFields, getFragments } from './helpers';
+import { getPluralName, getVariables, getParams, getFields, getFragments } from './helpers';
 
 const buildReadQuery = (tag = defaultTag) => (
   tag`query Read${getPluralName}${getVariables} {
-    read${getPluralName}${getRootParams} {
+    read${getPluralName}${getParams} {
       ${getFields}
```
ChloeSartorelli, line 7: root params is nonsensical - and would result in parameters being set and used in `getVariables` but then never being used as actual parameters - so the variable values wouldn't be used.

client/src/lib/dependency-injection/graphql/buildUpdateMutation.js
```diff
@@ -1,15 +1,9 @@
...
-    $Input:${getSingularName}UpdateInputType!
-    ${getVariables}
-  ) {
-    update${getSingularName}(
-      Input: $Input
-      ${getParams}
-    ) {
+  tag`mutation Update${getSingularName}${getMutationVariables} {
+    update${getSingularName}${getMutationParams} {
       ${getFields}
```
ChloeSartorelli, line 7: Turning `$Input:${getSingularName}UpdateInputType!` and `Input: $Input` into default fallbacks, which can be overridden by setting explicit params. See the new related test for an example.

client/src/lib/dependency-injection/graphql/helpers.js
```diff
@@ -1,10 +1,19 @@
...
 const paginationFields = {
   limit: 'Int',
   offset: 'Int',
 };
 
+// just outright is the parameters
+const paginationParams = {
+  limit: 'limit',
+  offset: 'offset',
+};
```
ChloeSartorelli, line 15: Needed because `getFields()` wants an object of param name to variable name

client/src/lib/dependency-injection/graphql/helpers.js
```diff
@@ -26,20 +35,14 @@ export const getVariables = ({ params, pagination = true }) => {
...
-  const paramList = Object.entries(items)
-    .map(([paramName, varName]) => (
-      `${paramName}: $${varName}`
+  const paramList = Object.keys(items)
+    .map((paramName) => (
+      `${paramName}: $${paramName}`
     ));
 
   return paramList.length ? `(${paramList.join(', ')})` : '';
 };
```
ChloeSartorelli, line 44: This was resulting in nonsensical queries like
```graphql
query ReadNotes($param: SomeInputType!) {
  readNotes(param: $SomeInputType!) {
    content
    id
  }
}
```
which is obviously incorrect - it should be
```graphql
query ReadNotes($param: SomeInputType!) {
  readNotes(param: $param) {
    content
    id
  }
}
```
which is what it now does.

client/src/lib/dependency-injection/graphql/helpers.js
```diff
@@ -26,20 +35,14 @@ export const getVariables = ({ params, pagination = true }) => {
...
 
-export const getRootParams = ({ args, pagination = true }) => {
-  const fieldParams = args[ROOT_FIELD] || {};
-
-  return getParams({ params: fieldParams, pagination });
-};
```
ChloeSartorelli, line 41: Not used anymore - it's nonsensical

client/src/lib/dependency-injection/graphql/helpers.js
```diff
@@ -55,7 +58,14 @@ export const getFields = ({ args, fields, pagination = true }, stack = [ROOT_FIE
...
 
-    const str = `${field}${getParams({ params: fieldParams, pagination: false })}`;
+    const items = (pagination) ? { ...fieldParams, ...paginationParams } : fieldParams;
+    const paramList = Object.entries(items)
+      .map(([paramName, varName]) => (
+        `${paramName}: $${varName}`
+      ));
+    const paramOutput = paramList.length ? `(${paramList.join(', ')})` : '';
+
+    const str = `${field}${paramOutput}`;
```
ChloeSartorelli, line 68: Using `getParams()` here resulted in `lib/dependency-injection/tests/ApolloGraphqlManager-test.js` failing on its "adds args to fields" test.
Basically, an assumption had been made that `params` in the registered query would be an object of parameter name to variable name - but it should actually be parameter name to _input type_. This is backed up by the fact that `getVariables()` gets the variable _type_ and _name_ both from the params object.

I'm not 100% sure what the point of having variable names here is at all tbh... seems like it should just return the field names. But this makes the test happy without breaking any other tests, so it seems correct.

client/src/lib/dependency-injection/graphql/helpers.js
```diff
@@ -76,3 +86,112 @@ export const getFragments = ({ availableFragments, fragments = [] }) => (
...
+
+function getMatchingParamKey(matchWith, params) {
+  return Object.keys(params).find((paramName) => paramName.toLowerCase() === matchWith);
+}
+
+function getInputVariable({ templateName, singularName, params }) {
+  const inputKey = getMatchingParamKey('input', params);
+  if (inputKey) {
+    return `$${inputKey}: ${params[inputKey]}`;
+  }
```
ChloeSartorelli, line 98: Allows for having graphql compatible variable declarations like `$input: CreateNoteInput!` by declaring the correct param name and type in the params object.

client/src/lib/dependency-injection/graphql/helpers.js
```diff
@@ -76,3 +86,112 @@ export const getFragments = ({ availableFragments, fragments = [] }) => (
...
+
+export function getMutationVariables(obj) {
+  const inputVar = getInputVariable(obj);
+  switch (obj.templateName) {
+    case CREATE:
+      return `(${inputVar})`;
+    case UPDATE:
+      return `(
+        ${inputVar}
+        ${normaliseMutationParamsAndVars(getVariables(obj), inputVar)}
```
ChloeSartorelli, line 122: If setting an input param like `input: 'CreateNoteInput!'` in the params object, we have to make sure to only output this _one_ for both variables and params.

Really I don't think update should allow for anything _other_ than the basic `input` (which would be consistent with what's happening for create) - but that functionality was there already so I can't just remove it.

client/src/lib/dependency-injection/injectGraphql.js
```diff
@@ -31,14 +31,9 @@ const injectGraphql = (key, context) => (DataHandler) => {
...
           error = false;
-        } catch (e) {
+        } finally {
           this.setState({ target, error });
-
-          // re-throw the error, as we do not want to silence it in the console
-          throw e;
         }
-
-        this.setState({ target, error });
```
ChloeSartorelli, line 41: This is unrelated to the rest of the changes and I'll remove it if you want - but it's just a lot cleaner and results in a stack trace from where the exception originated rather than pointing HERE as the origin of the exception.

Basically, it used to be
```js
try {
  doSomething();
} catch (e) {
  finish();
  throw e;
}
finish();
```
and now it's
```js
try {
  doSomething();
} finally {
  finish();
}
```
which does the same thing but results in a more accurate stack trace.

sabina-talipova (commented):
Looks good for me. Nice work!
Could you rebase it to 1.13, please.

client/src/lib/dependency-injection/graphql/helpers.js
```diff
@@ -55,7 +58,14 @@ export const getFields = ({ args, fields, pagination = true }, stack = [ROOT_FIE
     const key = path.join('/');
     const fieldParams = args[key] || {};
 
-    const str = `${field}${getParams({ params: fieldParams, pagination: false })}`;
+    const items = (pagination) ? { ...fieldParams, ...paginationParams } : fieldParams;
+    const paramList = Object.entries(items)
+      .map(([paramName, varName]) => (
+        `${paramName}: $${varName}`
+      ));
+    const paramOutput = paramList.length ? `(${paramList.join(', ')})` : '';
```
michalkleiner, line 66: I'm not sure I'm a big fan of returning this list (and other similar lists) with the brackets included. It would be cleaner if the brackets were explicitly part of the place where the list is used and the list only returned the list, e.g. `readOne${getSingularName}(${getIdOnlyParams})` or as it was `${getQueryName}(${getParams})`.

What's your view on this?

ChloeSartorelli: There are two reasons to include the brackets here:
1. That's how it already works and I want to avoid breaking things while I'm fixing things
2. Including the brackets here means we don't end up with empty brackets in the query if there are no fields/params/variables/whatever to return from these functions.

GuySartorelli:
> Could you rebase it to 1.13, please.

Done, though there were no merge conflicts to resolve.
Wait until @michalkleiner has indicated he's happy with https://github.com/silverstripe/silverstripe-admin/pull/1473#discussion_r1145596322 before merging.

michalkleiner:
> Wait until @michalkleiner has indicated he's happy with [#1473 (comment)](https://github.com/silverstripe/silverstripe-admin/pull/1473#discussion_r1145596322) before merging.

I 👍'd  the comment and resolved the conversation, but could've written up a comment as well I guess. All good with me.

sabina-talipova (approved):

Issue #1452, opened 2023-01-31
CMS 5 - Popper in not a constructor


emteknetnz:
Mousing over the bottom left hand of the admin triggers the following javascript error

![image](https://user-images.githubusercontent.com/4809037/215887122-0d8b5996-d0a8-47d8-9562-da897f633b05.png)

### PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1487

GuySartorelli:
Merged

PR #1487, opened 2023-03-27
FIX Use v1 of popper.js for thirdparty bootstrap

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1452

admin/node_modules has 2x popperjs:
- admin/node_modules/popper.js (v1)
- admin/node_modules/@popperjs (v2)

admin/thirdparty/bootstrap is compatible with v1 popper, but not v2 popper


GuySartorelli:
What was the change to the copy in the third-party dir?

emteknetnz:
the webpack config has
"to: `${PATHS.THIRDPARTY}/popper/popper.min.js`"

It's copying v1 of popperjs to the thridparty dir, previously it was copying v2

GuySartorelli (approved):

GuySartorelli:
Nice and elegant.

PR #1489, opened 2023-03-27
ENH Get storybook to work with CMS 5

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/2842

I made very poor progress on this, I'm going to halt work on this for now

tl;dr of what I learnt:

the 'chalk' module no longer loads via common js 'require' syntax, though that's fine, we don't need it, it's only for colouring the terminal output

the webpack.config.js in ./.storybook is not compatible for an unknown reason

`@storybook` was removed from devDependencies, it will require significant work to get it going again.  The version in CMS 4 is several versions behind


emteknetnz:
Closed in favour of https://github.com/silverstripe/silverstripe-admin/pull/1514

PR #1491, opened 2023-03-30
MNT Revert using map_get() for custom keys

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/34

Reverts some changes from https://github.com/silverstripe/silverstripe-admin/commit/cc63942affd4485cd1850f29981e9970d013b33e#diff-5c0dc5481c744d9451ae984cb32defd734b765840cf78d2e54d0eadacd2c3d7dR224

Fixes https://github.com/silverstripe/silverstripe-login-forms/actions/runs/4559173620/jobs/8042886903

I ran yarn build and there were no changes to the bundle.css which is to be expected as the output is expected to be identical

### DIFF
--- a/client/src/styles/_bootstrap-variables.scss
+++ b/client/src/styles/_bootstrap-variables.scss
@@ -143,6 +143,10 @@ $spacers: (
   3: $spacer,
   4: ($spacer * 1.5),
   5: ($spacer * 3),
+  // Note: Do not use map_get($spacers, xl/lg/sm/xs) within this file only
+  // This will cause a build failure in different modules such as silversripe/login-forms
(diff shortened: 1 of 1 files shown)


GuySartorelli:
@maxime-rainville should probably review this - I don't understand why he requested the original changes so there's clearly some context I'm missing here.

maxime-rainville:
I requested those changes because I thought the "proper way" to define different spacers is to reference `$spacers` map.

Looking at the actual bootstrap source code. I think I was mistaken https://getbootstrap.com/docs/5.0/utilities/spacing/

Happy for this to be reverted.

Issue #1495, opened 2023-04-05
Linked tabs 404 when Silverstripe 5 is installed in a subfolder


UndefinedOffset:
This maybe more a framework issue but noting here because it's being experienced in the admin area. It looks like Silverstripe 5 is being too aggressive at stripping segments from linked tabs such as other model admin tabs or settings & history on pages. For example given Silverstripe 5 is installed in a subfolder of the www root (for example `/ss_clients/silverstripe5`), clicking on a model admin tab (or settings, history, or content on pages) results in a 404.

The url that is requested by the JavaScript is `http://localhost/ss_clients/admin/pages/settings/show/1` rather than the expected `http://localhost/ss_clients/silverstripe5/admin/pages/settings/show/1`. Interestingly the link on the tab is correct, if you right click on the tab and open the link that way (i.e without the JS request) it works fine. The same thing happens in model admin.

composer.json:
```json
{
    "name": "silverstripe/installer",
    "type": "silverstripe-recipe",
    "description": "The SilverStripe Framework Installer",
    "require": {
        "php": "^8.1",
        "silverstripe/recipe-plugin": "^2",
        "silverstripe/vendor-plugin": "~2.0.0@stable",
        "silverstripe/recipe-cms": "~5.0.0@rc",
        "silverstripe-themes/simple": "~3.2.0",
        "silverstripe/login-forms": "~5.0.0@rc"
    },
    "require-dev": {
        "phpunit/phpunit": "^9.5"
    },
    "extra": {
        "project-files-installed": [
            ".htaccess",
            "app/.htaccess",
            "app/_config.php",
            "app/_config/mimevalidator.yml",
            "app/_config/mysite.yml",
            "app/src/Page.php",
            "app/src/PageController.php"
        ],
        "public-files-installed": [
            ".htaccess",
            "index.php",
            "web.config"
        ]
    },
    "config": {
        "process-timeout": 600,
        "allow-plugins": {
            "composer/installers": true,
            "silverstripe/recipe-plugin": true,
            "silverstripe/vendor-plugin": true
        }
    },
    "prefer-stable": true,
    "minimum-stability": "dev"
}
```

Note: hosting platform is Windows 10, with Apache and PHP 8.2.

## Acceptance criteria
- Tab link works when site is installed in a sub folder
- Common CMS functionally is reviewed when installed in a sub folder

## Related
- https://github.com/silverstripe/silverstripe-admin/issues/1494

## PR
[...]

PR #1499, opened 2023-04-12
FIX Ensure page router can handle base URL in subfolder

GuySartorelli:
This was caused by https://github.com/silverstripe/silverstripe-admin/pull/1438 - it turns out the page router _requires_ a trailing slash for the absolute base URL, so we just need to make sure the base URL has one before giving it to the page router.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1495


sabina-talipova (approved):
I've done tests locally and it looks like works fine on projects in subfolders, ex. base name is `http://localhost/ss_project/silverstripe-installer` and `http://localhost/projects/ss_project/silverstripe-installer`.

sabina-talipova:
@GuySartorelli , could you resolve conflict, please.

GuySartorelli:
Conflict resolved

PR #1510, opened 2023-05-11
Update docs and userhelp urls

izzudin96:
Change to Silverstripe 5 urls


GuySartorelli (approved):
Thank you, good catch

GuySartorelli:
While I'm waiting for tests to go green: I notice you swapped the target branch between `2` and `2.0` a couple of times - this is a change that could be released in a patch, so if you want this reflected in projects right away, please retarget to `2.0`.
However, it's also a very minor change, so I have no problems with this waiting for the next minor release (which will happen if it remains targetted at `2`).

izzudin96:
Hi @GuySartorelli, I just swapped it to `2.0`. I wonder if the extra commit would make any different.

Thanks

GuySartorelli:
Thanks.
Yeah those commits will need to be cleared up - the easiest way to do that imo is:

1. Write down the commit hash for your actual change
  `6e1d3f4`
1. Check out the branch you want to target
  `git checkout 2.0`
1. Delete the local copy of your pr branch
  `git branch -D 2`
1. Create a new branch with the same name - now you have a branch with the same name as your PR branch which is based off the branch you want to target - so it doesn't have those extra commits. In your case the branch name is a bit strange because you didn't create a new branch originally but it'll be fine, I promise.
  `git checkout -b 2`
1. Cherry pick your pr commit (this is why we did step 1)
  `git cherry-pick 6e1d3f4`
1. Force push to your remote - this will override the pr.
  `git push --force`

That last step might need some extra bits to match up the local branch with the remote branch - e.g. if your fork is your origin remote:
`git push origin 2 --force`

PR #1511, opened 2023-05-11
FIX Remove redundant cs_CZ.js language file

sabina-talipova:
## Description
Probably, redundant `cs_CZ.js` lang file was created and added as part of [this PR](https://github.com/silverstripe/silverstripe-admin/pull/1505). `cs_CZ.js` file was removed.
## Parent issue
- https://github.com/silverstripe/.github/issues/46


GuySartorelli (changes requested):
This file should _NOT_ be removed - it's not redundant, it's a valid set of translations. Instead, the unit test should be updated.

sabina-talipova:
> Instead, the unit test should be updated.

I don't think so. Test looks like work fine for all other cases. 

Now we have two files for the Czech language `cs_CZ.js` and `cs.js`. According to the logic of the test, we should use only one translation and have only one file for each dialect that we support. For languages in which there are other dialects, for example for Austrian German (de_AT) we use a special file with translation `de_AT.js`, for all other German dialects we use a common file with translation `de.js`.
According to this logic, I can delete the `cs.js` file and replace `'cs_CZ' => 'cs'` with `'cs_CZ' => 'cs_CZ'` in the `$tinymce_lang` TinyMCEConfig config if you think it would be more correct.

I don't know if other varieties of Czech exist, but I don't think we support them anyway.

GuySartorelli:
> According to the logic of the test, we should use only one translation and have only one file for each dialect that we support.

I think the test was written so that if you _try_ to get a specific translation, and it's not available, the more _generic_ translation is used.

For example if I am from Brazil and I want `pt_BR` but that's not available, it should fall back to `pt` if that's available. So I won't get Brazillian Portuguese, but I will get Portuguese.

> have only one file for each dialect that we support.
> I don't know if other varieties of Czech exist, but I don't think we support them anyway.

It's not _us_ supporting the translations. These are provided by TinyMCE. It's not on us to decide what's valid or what isn't. If TinyMCE have provided a translation, we should allow it to be used. The test is just validating that if a specific translation _isn't_ provided by TInyMCE, a more generic one is used if available.


That said - if the files for `cs` and `cs_CZ` have _identical_ translations, then it's a moot point. If you confirm they're identical then I'll merge this as is.

sabina-talipova:
I have downloaded all available versions of TinyMCE Language Packs. `cs.js` and `cs_CZ.js` files exist for versions 4 and 5, but not for 6. It is possible that language files were selectively added in CMS in the past.

Issue #1515, opened 2023-05-25
Bug: TinyMCE editor is unable to add links to strings containing symbols


chrispenny:
## Affected Version

5.0.3

## Description

* Create an `HTMLEditorField` anywhere in the CMS
* Enter an email address as plain text into the `HTMLEditorField`
* Select the email text
* Click the [link] icon in the `HTMLEditorField` to open the dropdown
* Attempt to select any type of link

Result:

```
Uncaught Error: Syntax error, unrecognized expression: chris.penny@silverstripe.com
    at se.error (vendor.js?m=1684280665:1:867765)
    at se.tokenize (vendor.js?m=1684280665:1:875796)
    at se.select (vendor.js?m=1684280665:1:876631)
    at Function.se [as find] (vendor.js?m=1684280665:1:861220)
    at E.fn.init.find (vendor.js?m=1684280665:1:879217)
    at jQuery.each.jQuery.fn.<computed> [as find] (vendor.js?m=1684280665:1:75485)
    at E.fn.init (vendor.js?m=1684280665:1:879706)
    at new e.fn.init (vendor.js?m=1684280665:1:86700)
    at E (vendor.js?m=1684280665:1:854922)
    at RR.<anonymous> (tinymce-cms-aafdf34ef0.js:62:173594)
```

![Screen Shot 2023-05-25 at 12 55 23 PM](https://github.com/silverstripe/silverstripe-framework/assets/505788/429106d2-167f-44d4-a79f-a56f4e11c839)

## Symbols attempted

Ticked to indicate whether they can have links added to them or not

- [ ] !
- [ ] @
- [x] #
- [ ] $
- [ ] %
- [ ] ^
- [ ] &
- [x] *
- [ ] (
- [ ] )

## PR
- https://github.com/silverstripe/silverstripe-admin/pull/1517

emteknetnz:
Linked PR has been merged and released as [2.0.2](https://github.com/silverstripe/silverstripe-admin/releases/tag/2.0.2)

PR #1517, opened 2023-06-02
FIX Syntax error TinyMC when using special characters

sabina-talipova:
## Description
- The problem most likely occurs directly in `tinymce.dom.Selection.getContent()` if the selection contains special characters. Because of this problem, all the "Insert link" Silverstripe plugins did not work.
- Therefore, I removed the `getContent()` method call, as it returned the entire DOM tree as a string, which was subsequently used only to determine whether text was selected and whether the "Link text" input field should be added.
In the case of the `getSel()` method and the subsequent call to the `toString()` method on it, it returns only the text that was selected or an empty string if nothing was selected. This I think is enough to control the display of the "Link text" input field.
- Solution was tested with the following characters:  
`test!`, `test@test`, `test#3`, `test$$4`, `test%`, `test^test`,  `test&test.com`, `test**8`, `test(9)`;
## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1515

### Note
- This problem also exists when selecting text containing special characters and copying it using keyboard shortcuts.
- The likely cause of this problem is discussed [here](https://stackoverflow.com/questions/26440277/custom-tinymce-button-not-working-with-terms-containing-special-characters)


emteknetnz (approved):
Tested locally works good

PR #1523, opened 2023-06-13
FIX Set current locale based on actual defined dictionaries

GuySartorelli:
Previously, the current locale would always be `null` if a 2-character locale was defined in php with `i18n::set_locale()` (e.g. `i18n::set_locale('nl');`), because no dictionaries would have been loaded at the time `detectLocale()` was called.

## Issue
- https://github.com/silverstripe/silverstripe-totp-authenticator/issues/85


client/src/i18n.js
```diff
@@ -28,6 +30,7 @@ class i18n {
    */
   setLocale(locale) {
     this.currentLocale = locale;
+    this.autoDetectLocale = false;
```
ChloeSartorelli, line 33: If someone explicitly sets a locale, we don't want to override their decision whenever a new dictionary is added.

client/src/i18n.js
```diff
@@ -90,6 +93,11 @@ class i18n {
     for (let entity in dict) {
       this.lang[locale][entity] = dict[entity];
     }
+
+    // Re-set current locale in case the new dictionary provides a better match than the old locale.
+    if (this.autoDetectLocale) {
+      this.currentLocale = this.detectLocale();
+    }
```
ChloeSartorelli, line 100: Dictionaries are added by separate `.js` files. This is the most reliable way to ensure that the current locale is detected according to all dictionaries that will be loaded in, allowing for slow network speeds. i.e. at the time each dictionary is loaded in, the most-correct current locale will be set.

We have no way of knowing ahead of time if more dictionaries will be loaded, so we can't reasonably defer until all of them are loaded.

client/src/i18n.js
```diff
@@ -174,10 +182,9 @@ class i18n {
    * Detect document language settings by looking at <meta> tags.
    * If no match is found, returns this.defaultLocale.
    *
-   * @todo get by <html lang=''> - needs modification of SSViewer
```
ChloeSartorelli, line 177: It's already doing this as the first step

client/src/i18n.js
```diff
@@ -207,14 +214,13 @@ class i18n {
 
     // Get locale (e.g. 'en_US') from common name (e.g. 'en')
     // by looking at i18n.lang tables
-    let detectedLocale = null;
     if (rawLocale.length === 2) {
-      for (let compareLocale in i18n.lang) {
+      for (let compareLocale in this.lang) {
```
ChloeSartorelli, line 218: The dictionaries are added to _this_ instance of `i18n`, not as a static property on the class.

emteknetnz (approved):
Test locally, works good

Issue #1512, opened 2023-05-15
TinyMCE not working with elemental if loaded in react context after already being loaded (silverstripe/admin 4.13)


baukezwaan:
Ok, this is kind of an edge case. But really annoing if it is your specific setup, because clients get really weird experiences. 

## Step to reproduce (from editor perspective):
This is when you open the page (not when ajax-navigating, only on first pageload):
![image](https://github.com/silverstripe/silverstripe-elemental/assets/795238/71c710ea-b938-44df-bdae-b45ef7065301)
This is when you open a elemental block:
![image](https://github.com/silverstripe/silverstripe-elemental/assets/795238/eebcb58f-8968-4e99-ac82-ea48b0bd994b)

## Step to reproduce:
1. Install clean silverstripe via composer:

  ```
  "silverstripe/recipe-cms": "4.13.*",
  "dnadesign/silverstripe-elemental": "*"
  ```

2. Setup Page for Elemental use via yml:

```
Page:
  extensions:
    - DNADesign\Elemental\Extensions\ElementalPageExtension
```

3. Setup Page.php for use with HTMLEditor field from Page

```private static $elemental_keep_content_field = true;```

4. Open the CMS and add one or two standard Content-elemental field to the page and save the page.

5. Refresh the page  and open the first Element, and see weird stuff happening. Again, the part 'refresh the page' is important here. The issue does not seem to occur on navigating after pageload / via AJAX.

## Possible solutions / found problems
- The ID of all the tinyMCE fields is the same, when inspectie the source. This explains why the text-part of the editor is made visible in the wrong place
- This seems to be introduced in the 4.13 release of the silverstripe/admin
- Probably somewhere in the update from tinymce from @GuySartorelli (https://github.com/silverstripe/silverstripe-admin/commit/368575b4a994932615512c4537da09e98f1d84cf)

So the quick-fix for production environments:
- Don't upgrade tot 4.13 yet:

```
        "silverstripe/recipe-cms": "4.12.*",
        "dnadesign/silverstripe-elemental": "*"
```

- Update to 4.13, but don't use silverstripe/admin 4.13 and stick to 4.12

```
        "php": "^7.4 || ^8.0",
        "silverstripe/recipe-plugin": "^1.7",
        "silverstripe/recipe-core": "~4.12.0@stable",               <-----------
        "silverstripe/admin": "~1.13.0@stable",
        "silverstripe/asset-admin": "~1.13.0@stable",
[...]

PR #1525, opened 2023-06-14
FIX Don't load a second TinyMCE instance from react

GuySartorelli:
The tinymce script was always loaded from react even if it had already been loaded from entwine or from a separate react instance, which resulted in conflicting tinymce instances.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1512


client/src/components/HtmlEditorField/HtmlEditorField.js
```diff
@@ -66,7 +66,14 @@ class HtmlEditorField extends TextField {
    * render the editor proper
    */
   renderDependencyScript() {
-    return <Script url={this.props.data.editorjs} onLoad={this.handleReady} />;
+    if (!window.tinymce && !window.TinyMCE) {
```
emteknetnz, line 69: This feels like the wrong place for this logic, seems like this method should just do what it says "render dependency script" rather that working out if it should or not.  It all seems kind of messy, particularly calling `handleReady()` in the `setTimeout()` since it's also called by `Script.onLoad`?

The core issue is that sometimes there's a timing issue where the script tag is rendered because the react component is rendered twice before the javascript file has been delivered from the server?

Seems like we should be doing something like this? (untested)

```js
handleReady() {
  if (!window.TinyMCE && window.tinymce) {
    window.TinyMCE = window.tinymce;
  }
  this.setState({ isReady: true, isFetchingScript: false });
}

renderDependencyScript() {
  this.setState({ isFetchingScript: true });
  return <Script url={this.props.data.editorjs} onLoad={this.handleReady} />;
}

render() {
  if (this.state.isReady) {
    return super.render();
  } else if (!this.state.isFetchingScript) {
    this.renderDependencyScript();
  } else {
    // noop
  }
}

```

emteknetnz (changes requested):

client/src/components/HtmlEditorField/HtmlEditorField.js
ChloeSartorelli: The core issue is that sometimes tinymce is already loaded - whether from another react component (basically another copy of this same component e.g. in a separate elemental area) or from entwine or other logic.
In other words the series of events is like this:
1. Either react or entwine or something renders a script tag to load tinymce
2. tinymce loads from that script. The thing that wanted it is happy
3. This react component is rendered (after that all happened successfully)
4. This react component renders a _new_ script tag and loads tinymce again (it doesn't care that it's already loaded)
5. The new script tag finishes loading, handleready() is called, and this component renders
6. conflicts ensue

What you've proposed would a) cause issues because you can't set state during a given render cycle, and b) not work, because we're still in a "not ready" state to begin with, which means we'll always end up running `renderDependencyScript()`, which doesn't check if tinymce already exists or not.

I thought about refactoring `render()` itself to be aware of whether a tinymce instance is already loaded, but there is logic in a few places that executes branches of code when the ready state changes - so we _have_ to change the ready state if we detect that tinymce is already loaded.

Calling `handleReady()` in a timeout avoids issues from setting state during the current render cycle, and is reminiscent of how that same function is called after the script has loaded. Basically, we're saying "either load the script, and then call `handleReady()` asynchronously, or if we don't need to load the script, skip straight to calling `handleReady()` asynchronously.

emteknetnz: OK makes sense

There's some code up the top of the component

```js
    this.state = {
        isReady: !props.data.editorjs
    };
```
where editorjs is the path that later gets mapped to Script.url

Seems like could simply do this? (untested)

```js
    if (!window.TinyMCE && window.tinymce) {
      window.TinyMCE = window.tinymce;
    }
    this.state = {
        isReady: !props.data.editorjs || window.TinyMCE
    };
```
?

emteknetnz (changes requested):

client/src/components/HtmlEditorField/HtmlEditorField.js
ChloeSartorelli: In theory yes, but that means we don't end up executing the code in `componentDidUpdate()` including `registerChangeListener()` which seems important to run and is the sort of thing that would probably cause a new regression we won't notice for 3 months 😅

emteknetnz: Can't you just, refactor it as a functional component instead then :-p

emteknetnz (approved):

PR #1530, opened 2023-06-21
MNT Behat test Gridfield pagination after search

sabina-talipova:
## Description
Added a new test case to test that the Gridfield keeps pagination (when needed) after search.
## Parent issue
- https://github.com/symbiote/silverstripe-queuedjobs/issues/405

### DIFF
--- a/tests/behat/features/gridfield-search.feature
+++ b/tests/behat/features/gridfield-search.feature
@@ -6,13 +6,19 @@ Feature: Search in GridField
   Background:
     Given the "Company" "Walmart" with "Category"="Retail"
     And the "Company" "ExxonMobil" with "Category"="Oil"
-    And the "Company" "Vitol" with "Category"="Other" 
-    And the "group" "EDITOR" has permissions "Access to 'Pages' section" and "Access to 'Test ModelAdmin' section" and "TEST_DATAOBJECT_EDIT"
+    And the "Company" "Chevron" with "Category"="Oil"
+    And the "Company" "ConocoPhillips" with "Category"="Oil"
+    And the "Company" "Saudi Aramco" with "Category"="Oil"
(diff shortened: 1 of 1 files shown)


tests/behat/features/gridfield-search.feature
```diff
@@ -23,3 +29,17 @@ Feature: Search in GridField
...
+    When I go to "/admin/gridfield-test-navigation"
+      And I should see "Walmart" in the "#Form_EditForm" element
+      And I press the "Open search and filter" button
+      And I press the "Advanced" button
+      And I fill in "Search__Category" with "Oil"
+      And I press the "Enter" key in the "Search__Category" field
+    Then I should see "ExxonMobil" in the "#Form_EditForm" element
+      But I should not see "Walmart" in the ".col-Name" element
+      And I should not see "Vitol" in the ".col-Name" element
+      And I should see "1" in the ".pagination-page-number" element
```
emteknetnz:
```suggestion
      And I should see an "#action_pagination_prev[disabled]" element
      And I should see an "#action_pagination_next" element
      And I should not see an "#action_pagination_next[disabled]" element
```
`And I should see "1" in the ".pagination-page-number" element` failed on my local because `1` is the attribute value rather than a text value. Seems like there isn't an existing way to test for attribute values within elements, though the way I've recommended test essentially the same thing

emteknetnz (changes requested):

tests/behat/features/gridfield-search.feature
sabina-talipova: Yeah, sure, because it should test total number of pages. Replaced to `"2"` and added few steps to check that pagination works.
Updated.

emteknetnz (approved):

PR #1531, opened 2023-06-23
ENH hide permission fields we arent using

andrewandante:


GuySartorelli (changes requested):
I'm a little iffy on this change, because entwine functions can be called from custom code. I doubt anyone is using these - but just in case, could you please keep the `showList` and `hideList` functions and just have them call the new show and hide functions?
i.e. something like
```js
showList: function (instant) {
  this.showGroupsList(instant);
  this.showMembersList(instant);
}
```

It probably makes sense for the onmatch (or wherever else) to keep calling the old functions as well, if only for simplicity's sake - but I don't mind whether you do that or not so long as custom code calling show/hide list still does what it should do.

andrewandante:
I've rebased after the latest changes and restored the `showList` and `hideList` methods - though I kept the newer methods inside the `onMatch` and `onChange` because I now want them to do _slightly different_ things i.e. show one + hide one.

The OG methods will now show/hide _both_ the Member field and the Group field.

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -566,22 +566,74 @@ $.entwine('ss', function($){
...
+      } else {
+        this.hideGroupsList();
+      }
+      if (e.target.value === 'OnlyTheseMembers') {
+        this.showMembersList();
+        this.hideGroupsList();
+      } else {
+        this.hideMembersList();
+      }
+    },
```
ChloeSartorelli: This is effectively the exact same code twice - can you please create a new function (maybe "toggleListDisplay" or similar) that takes a value as an argument, and sets the display accordingly?
For `onChange` the value is obviously `e.target.value`. For `onmatch` the value I think is just whichever of the two options `is(':checked')`

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -566,22 +566,74 @@ $.entwine('ss', function($){
...
+      const list = holder.next().filter('.treemultiselect');
+      list.css('overflow', 'hidden');
+      if (instant) {
+        list.hide().css('display', 'none');
+        holder.removeClass('field--merge-below');
+      } else {
+        list.slideUp(() => {
+          holder.removeClass('field--merge-below');
+        });
       }
```
ChloeSartorelli: Given that this is all duplicated from `hideMembersList()`, can you please migrate that shared functionality into a shared function, so we can have DRY code?
Same with the `showGroupsList()` and `showMembersList()` shared functionality.

GuySartorelli (changes requested):
Works well locally - just needs to be a little DRYer

GuySartorelli (approved):
LGTM

Issue #1433, opened 2023-01-10
Revisit JS linting


emteknetnz:
It was decided to suppress new JS linting issues before the CMS 5 beta in [this issue](https://github.com/silverstripe/silverstripe-admin/issues/1411) as it was causing CI failures when PR's were otherwise fine

We should work out what new rules we wish to keep and re-enable them. See the PRs linked from the above issue for the rules which were suppressed.

## Acceptance criteria
- [ ] Each disabled linting rules have been reviewed and a concious decision has been made about whether to keep them or not.
- [ ] JS modules have been updated to pass all linting rules.

## When deciding if a rule should be enabled consider
- Does the rule avoid common pitfalls that are likely to introduce errors in the code?
- Can the rule be fix with automated linting tools?
- Is the rule mostly stylistic and would require a lot of manual work to fix?

#### ESLint config
- https://github.com/silverstripe/eslint-config/pull/17
- https://github.com/silverstripe/eslint-config/pull/18

#### Fixed modules
- https://github.com/silverstripe/silverstripe-admin/pull/1532
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1371
- https://github.com/silverstripe/silverstripe-blog/pull/714
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/271
- https://github.com/silverstripe/silverstripe-cms/pull/2865
- https://github.com/silverstripe/silverstripe-elemental-bannerblock/pull/114
- https://github.com/silverstripe/silverstripe-mfa/pull/501
- https://github.com/silverstripe/silverstripe-realme/pull/114
- https://github.com/silverstripe/silverstripe-session-manager/pull/157
- https://github.com/silverstripe/silverstripe-tagfield/pull/248
- https://github.com/silverstripe/silverstripe-totp-authenticator/pull/125
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/297
- https://github.com/silverstripe/silverstripe-webauthn-authenticator/pull/145
- https://github.com/bringyourownideas/silverstripe-maintenance/pull/198
- https://github.com/silverstripe/cwp-agencyextensions/pull/97
- https://github.com/silverstripe/silverstripe-elemental/pull/1070
- https://github.com/symbiote/silverstripe-advancedworkflow/pull/497

## Update eslint version
- https://github.com/bringyourownideas/silverstripe-maintenance/pull/200
- https://github.com/silverstripe/cwp-agencyextensions/pull/98
[...]

PR #1532, opened 2023-06-27
MNT ESLint Issues following rules from main and import libs

sabina-talipova:
## Description
Turn on some of the ESLint.

## Related PR
- 

## Parent issue
-  https://github.com/silverstripe/silverstripe-admin/issues/1433


client/src/components/ActionMenu/ActionMenu.js
```diff
@@ -22,7 +22,7 @@ class ActionMenu extends PureComponent {
 
     // Force setting state to the end of the execution queue to clear a potential race condition
     // with entwine click handlers
-    window.setTimeout(() => this.setState({ isOpen: !this.state.isOpen }), 0);
+    window.setTimeout(() => this.setState((prevState) => ({ isOpen: !prevState.isOpen })), 0);
```
emteknetnz, line 25: I wouldn't expect to see changes such as these in a linting PR . Is there a reason this needs to be here for linting to pass? Otherwise this should be reverted

client/src/components/PopoverField/PopoverField.js
```diff
@@ -43,7 +43,7 @@ class PopoverField extends Component {
 
     // Force setting state to the end of the execution queue to clear a potential race condition
     // with entwine click handlers
-    window.setTimeout(() => this.setState({ isOpen: !this.state.isOpen }, toggleCallback), 0);
+    window.setTimeout(() => this.setState((prevState) => ({ isOpen: !prevState.isOpen }), toggleCallback), 0);
```
emteknetnz, line 46: I wouldn't expect to see this sort of change in a linting PR. Do we need this change here or can it be reverted?

emteknetnz (changes requested):

client/src/components/PopoverField/PopoverField.js
sabina-talipova: This is the rule [react/no-access-state-in-setstate](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md).
```txt
Usage of this.state inside setState calls might result
in errors when two state calls are called in batch and thus referencing 
old state and not the current state.
``` 
This problem describes [there](https://legacy.reactjs.org/docs/react-component.html#setstate)
```txt
setState() does not always immediately update the component.
It may batch or defer the update until later. This makes reading this.state right after
calling setState() a potential pitfall. Instead, use componentDidUpdate or a setState callback
(setState(updater, callback)), either of which are guaranteed to fire after the update has been applied. 
If you need to set the state based on the previous state, read about the updater argument below.
```

emteknetnz (approved):

emteknetnz (changes requested):
Merge conflict with the bundle - you'll need to rebase and run yarn build

Also realised this PR has the wrong prefix, please use MNT rather than FIX.

emteknetnz (approved):

PR #1540, opened 2023-08-03
MNT Fix behat test

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/84

PR #1541, opened 2023-08-03
FIX Popover toggle

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/84

Revertion of:
- https://github.com/silverstripe/silverstripe-admin/pull/1532/files#diff-9bd1df7cf43a960e5f3725685ce810dca253828c39186382b29e9bf80e57b93eR25
- https://github.com/silverstripe/silverstripe-admin/pull/1532/files#diff-80dd175364b0a71e21d008135c954525081a218f87a03085e938ff9a9c2afa3cR46
- https://github.com/silverstripe/silverstripe-admin/pull/1532/files#diff-b217634e30de77fc1ecff6ccfa50f7feb6789f145bff643efdfc21954f133c47R29

This should fix the broken builds on asset-admin, campaign-admin and sharedraftcontent

I've checked all the other modules that had files updated as part of https://github.com/silverstripe/silverstripe-admin/issues/1433 and while there were others that switched to the `(prevState)` method, none of them were wrapped in `window.setTimeout()` so I left them as they were

Expected to have behat failures which are fixed on https://github.com/silverstripe/silverstripe-admin/pull/1540 **< ensure this is merged-up to `2` first before merging this one**


GuySartorelli:
Looks sensible. As mentioned in https://github.com/silverstripe/silverstripe-admin/pull/1532/files#r1253832517:
> setState() does not always immediately update the component.
It may batch or defer the update until later. This makes reading this.state right after
calling setState() a potential pitfall.

But here we're not accessing state _after_ calling `setState()`, we're accessing it as part of the arguments we're passing _into_ `setState()`. So we don't need the anonymous function here.

client/src/components/ActionMenu/ActionMenu.js
```diff
@@ -22,7 +22,15 @@ class ActionMenu extends PureComponent {
...
     // with entwine click handlers
-    window.setTimeout(() => this.setState((prevState) => ({ isOpen: !prevState.isOpen })), 0);
+    //
+    // eslint disable the next line which references this.state within setState()
+    // https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-access-state-in-setstate.md
+    // Following eslint guidelines this would be implemented as:
+    // window.setTimeout(() => this.setState((prevState) => ({ isOpen: !prevState.isOpen })), 0);
+    // However react ends up confused it's wrapped inside window.setTimeout()
+    // and it may end up setting state twice causing the toggle to open and then immediately close
+    // eslint-disable-next-line
```
ChloeSartorelli:
```suggestion
    // ignore linting rule because following the recommended replacement caused behat failures
    // eslint-disable-next-line no-access-state-in-setstate
```
Explicitly only disable the linting rule we want to disable.

We also don't need to go super in depth here, and in fact doing so just makes this comment distracting rather than helpful IMO.

Please make this change to the other changes as well.

GuySartorelli (changes requested):

client/src/components/ActionMenu/ActionMenu.js
emteknetnz: Updated

GuySartorelli (approved):
Looks okay, merge on green

Issue #1542, opened 2023-08-04
Ship new logo in CMS 5.1


maxime-rainville:
We have a new logo to differentiate Silverstripe the company from Silverstripe the CMS.

## Acceptance criteria
- All instance of the old logo are replaced with the new logo.
- Placement of the new logo is validated with the strategy and UX team

## Notes
- If the loading animation is going to be difficult to replicate with the new logo, just swap it out for a regular spinner. 

## Pull request
- https://github.com/silverstripe/silverstripe-admin/pull/1546/
- https://github.com/silverstripe/silverstripe-installer/pull/358
- https://github.com/silverstripe/silverstripe-login-forms/pull/151
- https://github.com/silverstripe/silverstripe-sharedraftcontent/pull/208
- https://github.com/silverstripe/silverstripe-framework/pull/10922

## Follow up
- https://github.com/silverstripe/.github/issues/102

maxime-rainville:
Got the logos from the design team. They're looking at the loading animation.

[SCMS-logo-white](https://github.com/silverstripe/silverstripe-admin/assets/1168676/d5c0a8d1-69d0-40ce-af32-daa613ac29d4)
[SCMS-glyph-white](https://github.com/silverstripe/silverstripe-admin/assets/1168676/c25c61e9-dabd-478d-8250-f1a95667255e)
[SCMS-glyph-blue](https://github.com/silverstripe/silverstripe-admin/assets/1168676/0b5cbdf9-98a0-43c9-87b3-96cee080d342)
[SCMS-logo](https://github.com/silverstripe/silverstripe-admin/assets/1168676/a6eb4cc4-887e-4661-9d1f-a951cca7d410)

Off the top of my head the places where the logo needs to be updated are:
- The login screen (dark mod and light mod)
- The loading animation
- The help menu
- The pre-JS loading animation (look for `ss-loading-screen` css class)

GuySartorelli:
Here's copies of the logo that I was able to find:

An animated loading spinner:
- https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/images/spinner.gif
- https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/images/spinner%402x.gif
Used in https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/styles/legacy/_style.scss

The `ss-loading-screen` one mentioned by Max:
[...]

PR #1546, opened 2023-08-11
EHN Implement the new logo on the loading animation

maxime-rainville:
## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1542


GuySartorelli:
Is this to also be used in place of the ss-loading-screen animation?
See https://github.com/silverstripe/silverstripe-admin/blob/2/templates/SilverStripe/Admin/Includes/CMSLoadingScreen.ss

Can you please also add a video or gif of it working so we can be sure what you're intending is what we've ended up with?

client/src/components/Loading/Loading.js
```diff
@@ -1,62 +1,87 @@
 /* global window */
-import React, { PureComponent } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 
 /**
  * Renders a Silverstripe loading animation styled after the Silverstripe logo
+ *
+ * `spinner__animation__outline_right` and `spinner__animation__outline_left`
+ * They form the two halves of the logo. They are transaparent atd are use
```
michalkleiner:
```suggestion
 * They form the two halves of the logo. They are transparent and are used
```

client/src/components/Loading/Loading.js
```diff
@@ -1,62 +1,89 @@
...
+function Loading({ containerClass }) {
+  // If we are displaying many loading indicators at the same time, we need to
+  const [id] = useState(uuid());
 
-    return (
-      <div className={containerClass}>
-        <div key="overlay" className="cms-content-loading-overlay ui-widget-overlay-light" />
-        <div key="spinner" className="cms-content-loading-spinner">
-          <div className="spinner">
-            {/* window.location is neccessary for Safari which otherwise prepends the base tag */}
```
maxime-rainville, line 17: I've decided to throw Safari under the bus here. Basically, forcing us to use absolute URL to find those SVGs introduce a bug where sometime if you've updated the URL with react route or added some extra parameter, the SVG path clipping doesn't work.

Using the IDs directly instead of the full URL seems to mitigate this problem.

Basically, I think that having this work consistently in all other browser is better than adding a hack so that Safari users get it looking at it's best some of the time.

Safari user should still get a reasonable animation in most cases. It just won't have the nice curvy bit. Given that those loading animation are only shown for a fleeting second and are pretty small, I doubt anyone will even notice.

client/src/components/Loading/Loading.scss
```diff
@@ -0,0 +1,111 @@
+$dash: 750;
```
maxime-rainville, line 1: I've moved this bit from `_style.scss` so it lives next to the actual component its meant to style.

client/src/font/icons-reference.html
```diff
@@ -1011,6 +1007,14 @@ <h2>CSS mapping</h2>
           <div class="icon font-icon-attention-1"></div>
           <input type="text" readonly="readonly" value="attention-1">
         </li>
+        <li>
+          <div class="icon font-icon-silverstripe"></div>
```
maxime-rainville, line 1011: I've kept the old Siverstripe icon in there, in case we decide we have a use case for showing the company logo.

client/src/images/spinner.gif
maxime-rainville, line 1: This is reference in one SCSS file for some jQuery modal. I don't actually think this is used anywhere. We seem to have a lot of dead code in our CMS UI.

I've updated it anyway, but I think we should have a follow up card to remove dead weight code.

client/src/styles/legacy/_style.scss
```diff
@@ -784,51 +706,19 @@ html, body {
...
-          stroke-dashoffset: 225;
-        }
-        100% {
-          fill-opacity: 0;
-          stroke-dashoffset: 225;
-        }
-      }
-    }
-  }
-
```
maxime-rainville, line 826: This block was only there for the old animation

client/src/components/BrowserWarning/BrowserWarning.scss
maxime-rainville, line 1: This pure dead weight code.

I've removed all the matching images.

client/src/components/Loading/Loading.js
maxime-rainville: Fixed

maxime-rainville:
Here's what this thing looks like. The video looks a bit skipy, but when you look at live it's very smooth.


[Screencast from 17-08-23 19:17:33.webm](https://github.com/silverstripe/silverstripe-admin/assets/1168676/beedb60d-44c4-423d-ae45-bb57f07b2fa7)

I've validated that looks the same in Chrome.

client/src/components/Loading/Loading.js
maxime-rainville: Safari will probably look something like this.

image

emteknetnz (changes requested):
There's stuff changing in this PR for which there's no explanation:
- silverstripe.ttf added /  silverstripe.eot added / silverstripe.woff deleted - I'm not sure why this is happening - based on https://medium.com/@aitareydesign/understanding-of-font-formats-ttf-otf-woff-eot-svg-e55e00a1ef2 seems that .eot is for old versions of IE?
- Deleting BrowserWarning and associated files - this seems like random refactoring which I really don't like in PRs as it clutters their intent - this should have been done in a separate PR / issue

Please explain the first point

Also get this design reviewed. I tested on my local and honestly this looked pretty bad IMO in firefox.  Experience was a flickering icon that had a solid white background (possibly it's supposed to be transparent?).  Particularly bad was switching between model admins where a large icon flickered followed b a smaller icon flickering. 

https://youtu.be/R1nMfWP-ze8 < flickering icon in firefox

I think maybe it would be better to just get rid of the 'loading' icon that pops up in the middle of the CMS altogether?

emteknetnz:
https://youtu.be/R1nMfWP-ze8

I think maybe it would be better to just get rid of the 'loading' icon that pops up in the middle of the CMS altogether?

GuySartorelli:
> I think maybe it would be better to just get rid of the 'loading' icon that pops up in the middle of the CMS altogether?

I'd be against this, as it appears for a scenario where javascript isn't available (see screenshot) and likely in a variety of other similar scenarios that aren't typical, but indicate something's gone wrong. It's not a massively common use case so if people feel strongly about removing it then I won't kick up a fuss, but just an extra scenario to consider.
loading icon if javascript isn't available

It may also be seen for longer periods of time for users with slow internet connections or scenarios where the CMS takes longer-than-usual to load, which is probably the stronger argument for keeping it.

maxime-rainville:
> silverstripe.ttf added / silverstripe.eot added / silverstripe.woff deleted - I'm not sure why this is happening - based on https://medium.com/@aitareydesign/understanding-of-font-formats-ttf-otf-woff-eot-svg-e55e00a1ef2 seems that .eot is for old versions of IE?

Those fonts are managed via [Fontastic](https://fontastic.me/). We get all those fonts provided to us in a ZIP file. Our process when updating the font has always been to take the entire package and copy it to our code base so we have the full reference.

> Deleting BrowserWarning and associated files - this seems like random refactoring which I really don't like in PRs as it clutters their intent - this should have been done in a separate PR / issue

There's a bunch of logos that are only used in the browser warning. The browser warning is not in-line with our minimum browser requirements. In this context, it would be a waste of time to update logos for a warning that is not relevant any more.

> Also get this design reviewed. I tested on my local and honestly this looked pretty bad IMO in firefox.

James will be looking at this once the code review is done.

>  Experience was a flickering icon that had a solid white background (possibly it's supposed to be transparent?).

The only thing that has changed here is the logo. To the extent there's flickering it would have been there before.

Also, keep in mind you are running this thing locally with zero latency. That's not going to be a typical example.

emteknetnz:
OK code is fine, will leave the 'changes requested on' on for now until the design review so it's not accidentally merged before that's been done

client/src/styles/_variables.scss
```diff
@@ -11,7 +11,7 @@ $brand-success: #3fa142;
 $brand-danger: #d40404;
 
 // Offical SilverStripe brand color
-$color-brand: #43c7f4 !default;
+$color-brand: white !default;
```
maxime-rainville, line 14: The title logo is the only place where this logo colour is used.

client/src/styles/legacy/_style.scss
```diff
@@ -791,44 +713,18 @@ html, body {
...
-          stroke-dashoffset: 225;
-        }
-        100% {
-          fill-opacity: 0;
-          stroke-dashoffset: 225;
-        }
-      }
-    }
+  body.loading & {
+    opacity: 1;
```
maxime-rainville: This snippet adds an opacity transition when the `loading` class is removed from the `<body>`.

maxime-rainville (commented):
Here's the feedback from James and my reply.

> Looks great! There’s just a couple of things to tweak if possible…
> “Site title in the top left when the side bar is collapse” - can this be white? We’re no longer using the cyan colour
> > I'll change it to white.
> 
> Everytime I load a page I get the double loading indicator - a flash of the big one and then the small one. Is it supposed to be showing both?
> > That was annoying Steve as well.
> > That's basically how the CMS admin UI is built right now. Addressing this would go way beyond updating the logo.
> > We could use the little logo everywhere or tweak the look of the indicator, but there's not really a short term way we could get rid of the double loading indicator.
> > > display: none on the big one?? lol
> > > > We could do something like transitioning it to opacity 0
> > > > [video](https://github.com/silverstripe/silverstripe-admin/assets/1168676/9d6d12fe-8ed1-49b3-9c17-2652f95a1aa5)
> > > > > That’s definitely better. It was more about when I saw the big one flick on for a micro-second and then seeing the smaller one.

So basically. I change the logo in the top left to white and I added a fadeout to the big logo loading screen.

emteknetnz (changes requested):
While it certainly looks better, the fade that's been added this is a very signficiant UI downgrade so I cannot approve this in its current state

The issue is that the fade adds a significant lag to the responsiveness to the cms because you need to wait for the fade in animation, which applies to the ENTIRE cms rather than just the logo, before you can click anything

I've uploaded a video https://www.youtube.com/watch?v=CAKDL6qKE6o that shows this - bugfix1.test is the existing state where I can quickly click between the models admins.  bugfix11.test is with PR and shows how much slower it is

Perhaps if you make it so the fade affect is applied specifically to the logo and not the entire CMS we'll end up with a good result?

maxime-rainville:
The logo is in a ligthbox. Applying the fade to the logo only wouldn't do anything because we would still need to keep the lightbox there and it would block you from interacting with whatever is beneath it.

On a normal internet connection, you won't be able to interact with the UI anyway because it will still be in the process of fetching things like site trees or Page forms.

I'll go back to James and see if he's got extra feedback.

If this is controversial, I'd say let's preserve the status quo and take out the fade.

emteknetnz:
> On a normal internet connection, you won't be able to interact with the UI anyway because it will still be in the process of fetching things like site trees or Page forms.

That's not quite true. On a slow internet connect yes things do take longer to load.  But then after they've loaded you THEN still need to wait for the fade animation before you can click anything.  

> If this is controversial, I'd say let's preserve the status quo and take out the fade.

Agree.  While I don't like the flickering it is the current behaviour.  One thing that could mitigate the flickering is to have a transparent rather than white background for the large logo which should reduce how noticeable it is.  The existing large logo has a transparent background

maxime-rainville:
I removed the loading animation fade and created a follow up card. https://github.com/silverstripe/silverstripe-admin/issues/1557

emteknetnz (approved):
Great, thanks for removing the fade, I'm much happier with that.

I'll approve and leave it up to you if want to do further design review or not. If there are no further code changes then merge at your leisure.

Issue #1007, opened 2020-02-07
failed to parse SourceMap


lekoala:
a very minor issue, but still an empty dev console is a happy console... it seems that the source map are used in the .js files but are not included in resources

```
DevTools failed to parse SourceMap: /_resources/vendor/silverstripe/admin/thirdparty/bootstrap/js/dist/collapse.js.map
DevTools failed to parse SourceMap: /_resources/vendor/silverstripe/admin/thirdparty/bootstrap/js/dist/util.js.map
```

sourcemaps don't have a big value in my opinion so in LeftAndMain instead of

        Requirements::javascript('silverstripe/admin: thirdparty/bootstrap/js/dist/util.js');
        Requirements::javascript('silverstripe/admin: thirdparty/bootstrap/js/dist/collapse.js');

we could require minified files that don't use the sourcemaps

Cheddam:
Thanks for raising this @lekoala 🙂

HopStudios-Development:
Is there any solution for this problem? The call of these files in LeftAndMain is made using .js but that is rendering with .map and says 404 for them. Badly stuck in this issue. Many back-end operations aren't working.

lekoala:
@HopStudios-Development  you can run this
https://github.com/lekoala/silverstripe-base/blob/master/bin/DeleteJsMaps.php

GuySartorelli:
PR merged, it'll be automatically patched by GitHub actions.

PR #1551, opened 2023-08-17
remove sourcemap

lekoala:
it's not available and it causes issues in the dev console

fixes https://github.com/silverstripe/silverstripe-admin/issues/1007


GuySartorelli (approved):
Seems sensible. Thanks.

PR #1553, opened 2023-08-18
PHP 8.2 compatibility fix

forsdahl:
Currently silverstripe-admin in version 4 triggers a lot of deprecation errors, because of setting dynamic class properties. This can be easily fixed by declaring these properties on the classes, and should have no impact on the code otherwise.

The reason for fixing this is to be able to fix deprecation errors with PHP 8.2 in other Silverstripe modules without these errors getting buried in all the deprecation errors from silverstripe-admin, not to have version 4 of silverstripe-admin officially support PHP 8.2.

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -62,6 +62,7 @@ use SilverStripe\View\SSViewer;
  * This is essentially an abstract class which should be subclassed.
  * See {@link CMSMain} for a good example.
  */
+#[\AllowDynamicProperties]
 class LeftAndMain extends Controller implements PermissionProvider
 {
 


GuySartorelli:
You said there are a lot of deprecation warnings (note, they're not _errors_) and the description of the pr makes it sound like there are multiple changes required - but there's only one. Can you please clarify if this pr is complete?

forsdahl:
Sorry, meant deprecation notices. I meant that silverstripe-admin causes deprecation notices on all requests to /admin, hence my wording of a lot of notices. The PR is complete though, with this change (which has no impact otherwise) we can get rid of these notices, so that we can fix deprecation notices in other modules.

I see that there is a linting error from this change though, should the property perhaps be declared private in this case, since it is not used by any other code?

GuySartorelli:
It should be public, because the dynamic property was public implicitly. But in any case, the linting error is because of the underscore more than because of the visibility I think. You may need to add a phpdoc comment to disable the linting rule for that property.

GuySartorelli:
Fyi https://github.com/silverstripe/silverstripe-framework/pull/10921#issuecomment-1683612761

forsdahl:
Changed the property to public, and disabled linting for that line

code/LeftAndMain.php
```diff
@@ -297,6 +297,9 @@ class LeftAndMain extends Controller implements PermissionProvider
      */
     protected $versionProvider;
 
+    // phpcs:ignore
+    public $_cache_MainMenu;
```
```suggestion
    var $_cache_MainMenu;
```
As per https://github.com/silverstripe/silverstripe-framework/pull/10921#pullrequestreview-1592486331

code/LeftAndMain.php
```diff
@@ -297,6 +297,9 @@ class LeftAndMain extends Controller implements PermissionProvider
      */
     protected $versionProvider;
 
+    // phpcs:ignore
```
```suggestion
    // phpcs:ignore PSR2.Classes.PropertyDeclaration.Underscore
```
Only ignore that specific rule.

GuySartorelli (changes requested):

forsdahl:
I'll go ahead and change this PR to use the AllowDynamicProperties annotation instead of declaring the property public, as per https://github.com/silverstripe/silverstripe-framework/pull/10921

lozcalver (approved):

GuySartorelli (approved):

GuySartorelli:
This will be automatically tagged when GitHub Actions has finished running on the branch.
I've manually reverted it in the merge-up to 2.0

PR #1555, opened 2023-08-22
Revert "ENH Make CMSProfileController use required_permission_codes"

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/99

Reverting https://github.com/silverstripe/silverstripe-admin/pull/1027

This caused the following failures in https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/5892574918/job/15998642199 (recipe-cms builds)

1) SilverStripe\Admin\Tests\CMSProfileControllerTest::testMemberEditsOwnProfile
2) SilverStripe\Admin\Tests\LeftAndMainTest::testCanView

I didn't investigate it too deeply, though seems that the reason is that Permission::checkMember() will (I think) count any permission code that starts with `CMS_ACCESS_` e.g. `CMS_ACCESS_CMSMain` as `CMS_ACCESS`, whereas as I'm assuming `required_permission_codes` does not?

### DIFF
--- a/code/CMSProfileController.php
+++ b/code/CMSProfileController.php
@@ -18,7 +18,7 @@ class CMSProfileController extends LeftAndMain
 
     private static $menu_title = 'My Profile';
 
-    private static $required_permission_codes = 'CMS_ACCESS';
+    private static $required_permission_codes = false;
 
     private static $tree_class = Member::class;
 
@@ -59,10 +59,8 @@ class CMSProfileController extends LeftAndMain
 
     public function canView($member = null)
     {
-        $currentUser = Security::getCurrentUser();
-
         if (!$member && $member !== false) {
-            $member = $currentUser;
+            $member = Security::getCurrentUser();
         }
 
         // cms menus only for logged-in members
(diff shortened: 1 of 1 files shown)


emteknetnz:
Broken behat test is existing

GuySartorelli:
> Broken behat test is existing

Is there a PR for it? This issue is explicitly about fixing broken tests.

GuySartorelli:
> I didn't investigate it too deeply, though seems that the reason is that Permission::checkMember() will (I think) count any permission code that starts with `CMS_ACCESS_` e.g. `CMS_ACCESS_CMSMain` as `CMS_ACCESS`, whereas as I'm assuming `required_permission_codes` does not?

If you click through `parent::canView()` to `LeftAndMain::canView()`, you will see that it does in fact use `Permission::checkMember` which immediately makes me think reverting this outright is not the right move. I'll investigate further to see if I can find the actual root cause of the unit test failure.

GuySartorelli:
After investigation, it seems this is not caused by any problem with this change, but rather with inappropriate permissions checking in the subsites module (specifically in the `LeftAndMainSubsites` extension, either in `sectionSites()`, `accessible_sites()`, `hasMainSitePermission()`, or `canAccess()` - I'd probably look at `hasMainSitePermission()` first) which explains why it's failing in kitchen sink and not anywhere else.

The only reason this wasn't failing before is because there are extension hooks in `LeftAndMain` that weren't being called previously. But the `LeftAndMainSubsites` is _meant_ to work for subclasses of `LeftAndMain` so it should be made to work for this one too.

Please close this PR and resolve the problem in the subsites module.

emteknetnz:
> Is there a PR for it? This issue is explicitly about fixing broken tests.

https://github.com/silverstripe/silverstripe-frameworktest/pull/147

emteknetnz:
I'll close this PR and split off a [separate card](https://github.com/silverstripe/silverstripe-subsites/issues/535) to investigate subsites

PR #1558, opened 2023-08-24
DEP Update JS dependencies and update TinyMCE_sslink.js onunmatch()

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/90

Automated version of this PR went red in CI
- https://github.com/silverstripe/silverstripe-admin/pull/1556
- https://github.com/silverstripe/silverstripe-admin/actions/runs/5947643030/job/16130112656#step:12:234

I got the instructions for how to resolve this from the second example in https://docs.silverstripe.org/en/5/changelogs/5.0.0/#common-upgrade-patterns


GuySartorelli (approved):
At first I was concerned that `if (root)` would always evaluate falsy, but I've dug in and noticed that the `ReactRoot` property is declared and set in specific sslink implementations (e.g. email link). So this looks fine to me.

PR #1559, opened 2023-08-24
Prevent sourcemap errors in dev console

lekoala:
Two files from the tinymce skin reference sourcemap that are not published and causing errors

image


GuySartorelli:
This is the wrong approach for these files, since the next time we built out the skin they'll just be generated with the sourcemap references again. Instead, the process used to generate them will need to be amended, and then the files will need to be regenerated using that updated process.

Issue #1560, opened 2023-08-27
Source Code view in HTML Editor only shows two line of text


maxime-rainville:
![image](https://github.com/silverstripe/silverstripe-admin/assets/1168676/ae7e421e-4e0c-449a-850e-a0f5376bbedc)

This only affects CMS 5.x-dev. 5.0.x-dev looks fine.

Something has changed in TinMCE. So imply applying a new CSS rule might not be the most appropriate solution.

## PRs
 - https://github.com/silverstripe/silverstripe-admin/pull/1561
 - https://github.com/silverstripe/silverstripe-admin/pull/1566

emteknetnz:
Old working
![image](https://github.com/silverstripe/silverstripe-admin/assets/4809037/93109ab7-4461-4f11-9c9b-2fc5e1fbb37f)

New broken
![image](https://github.com/silverstripe/silverstripe-admin/assets/4809037/a995029a-d773-4a4f-bc66-5ce9b2470683)

Requires ?flush=1 to be run after changing commits for breaking/fixing to show up

first breaks in 383740a

Is still broken after updating js in https://github.com/silverstripe/silverstripe-admin/pull/1558

`tox-textarea-wrap` doesn't exist in a search of the code base of the last working version da1bddc, though it exists in 383740a in my instances under the client/dist/tinymce directory.  So I guess the regression got introduced after updating to a newer version of the tinymce library

emteknetnz:
Is still broken after updating js in https://github.com/silverstripe/silverstripe-admin/pull/1558

GuySartorelli:
PR merged

PR #1561, opened 2023-08-28
FIX TinyMCE source view editor size

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1560

I've left some investigation notes on the main issue

Regression isn't present in 2.0, so targeting 2


emteknetnz:
@dependabot rebase

client/src/styles/legacy/_tinymce.scss
ChloeSartorelli, line 1: I think this should probably be added to the silverstripe skin instead.

If this was caused by an update to tinymce, it'll be worth double checking the changelog to see if there's something specific they expected us to change.

It may be that simply rebuilding the skin the with latest tinymce version will resolve it without any customisation.

GuySartorelli (changes requested):

client/src/styles/legacy/_tinymce.scss
emteknetnz: What's a silverstripe skin?  And how is it rebuilt?

ChloeSartorelli: https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/tinymce/skins/ui/silverstripe/src/README.md

emteknetnz:
Closed in favor of https://github.com/silverstripe/silverstripe-admin/pull/1566

PR #1562, opened 2023-08-28
FIX Don't use state unnecessarily

GuySartorelli:
This `useState` was breaking a test in versioned-admin - and it wasn't necessary so I'm removing it. More details in comments on the linked issue.

## Issue
- https://github.com/silverstripe/silverstripe-versioned-admin/issues/307


sabina-talipova (approved):
Looks good to me. Test locally. Work nice.

PR #1563, opened 2023-08-29
MNT Replace deprcated function findDOMNode

sabina-talipova:
## Parent issue
- https://github.com/silverstripeltd/product-issues/issues/775


client/src/components/Search/Search.js
```diff
@@ -183,12 +182,12 @@ class Search extends Component {
   /**
    * Try to find the first form field in the advanced form and focus on it.
    */
-  focusFirstFormField(filter = 'input, textarea, select, button') {
+  focusFirstFormField(filter = 'input:not([type=hidden]), textarea, select, button') {
```
ChloeSartorelli: Why was this change made? It doesn't seem related to the main issue?

client/src/components/Search/Search.js
```diff
@@ -444,40 +444,42 @@ class Search extends Component {
...
-          tagData={this.formatTagData()}
-        >
-
-          <SearchForm
-            id={formId}
-            identifier={identifier}
-            visible={visible}
-            expanded={expanded}
-            formSchemaUrl={formSchemaUrl}
+      <div ref={node => { this.nodeRef = node; }}>
```
ChloeSartorelli: Don't add a new element to the component - this might mess with css customisations.
Instead, use [forwarding refs](https://legacy.reactjs.org/docs/forwarding-refs.html) and forward a ref from this component down to the `FocusedZone` component.

GuySartorelli (changes requested):

client/src/components/Search/Search.js
sabina-talipova: First field in "Search options" block is hidden. Because of this, when we open "Search options" block the first field is not in focus.

ChloeSartorelli: Is that somehow related to the issue this PR is for? Or is it separate? If it's separate please handle it in a separate issue and PR to make it clearer what's going on.

client/src/components/Search/Search.js
sabina-talipova: Updated.

client/src/components/Search/Search.js
sabina-talipova: It doesn't work and you cannot test focus on first field in "Search options". I can extract it to another issue if so would be better.

sabina-talipova: Updated.
PR https://github.com/silverstripe/silverstripe-admin/pull/1567

GuySartorelli (approved):
Looks good, works well locally. Tested by printing the `node` variable to the console and validating that it's the correct element. This PR doesn't change anything about what _type_ that variable is or what is done with it, so anything beyond that is out of scope.

PR #1566, opened 2023-08-29
FIX Update TinyMCE css files

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1560

I simply rebuilt the TinyMCE css files and it fixed the issue

I've updated the instructions to give a little more detail for next time


GuySartorelli (approved):
Looks good and works well locally.
Love the added context for rebuilding, that's a good idea.

Issue #1565, opened 2023-08-29
First field doesn't get focus when user opens "Search options" block.


sabina-talipova:
## Description
The first field that the query `form.querySelector('input')` receives in the "Search options" is of type hidden and we cannot get focus on it, so we must exclude this field from the filter
## PR
- https://github.com/silverstripe/silverstripe-admin/pull/1567
- https://github.com/silverstripe/silverstripe-admin/pull/1568

GuySartorelli:
PR merged. Will be automatically tagged when CI finished running on the branch.

PR #1568, opened 2023-08-29
FIX Focus on first search field

sabina-talipova:
## Description
The first field that the query `form.querySelector('input')` receives in the "Search options" is of type hidden and we cannot get focus on it, so we must exclude this field from the filter
## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1565


GuySartorelli (approved):
Looks good and works great locally.

Issue #1548, opened 2023-08-16
CMS 5: Cannot link an image in TinyMCE


axllent:
## Description
As of Silverstripe 5 it is not possible to link just an image within TinyMCE.

### Steps to reproduce:
1. Insert image into editor
2. Select image, click link (any of "Page on this site", file, external URL etc)
3. Add/select link, click "Insert link" button

All attempts fail because of `Link text is required.`. If you fill in any link text (which you shouldn't have to anyway), that text replaces the image.

## Behat tests
This fix involves multiple related pull requests - I have run a CI run with silverstripe/installer with all of the pull requests included to show that they work when they're all installed.
- **CMS 4:** https://github.com/creative-commoners/silverstripe-installer/actions/runs/6091324603
- **CMS 5:** https://github.com/creative-commoners/silverstripe-installer/actions/runs/6105659894

## PR
**CMS 4**
- https://github.com/silverstripe/silverstripe-admin/pull/1570
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1393
- https://github.com/silverstripe/silverstripe-admin/pull/1571
- https://github.com/silverstripe/silverstripe-cms/pull/2880
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1394
- https://github.com/silverstripe/silverstripe-behat-extension/pull/248
- https://github.com/silverstripe/silverstripe-admin/pull/1577
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1397
- https://github.com/silverstripe/silverstripe-cms/pull/2883

**CMS 5**
- https://github.com/silverstripe/silverstripe-admin/pull/1579
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1398
- https://github.com/silverstripe/silverstripe-cms/pull/2884
- https://github.com/silverstripe/silverstripe-behat-extension/pull/249

maxime-rainville:
I'm getting the same behaviour on CMS 4.

Interestingly, it works if you are trying to link to a file in CMS 4. In CMS5, it lets you create the link, but doesn't seem to save it.

emteknetnz:
Linked PRs have been merged

GuySartorelli:
Reopening to make sure the fix works for CMS 5 on merge up since tinyMCE changed a lot between majors

emteknetnz:
Linked PRs have been merged

PR #1570, opened 2023-08-31
FIX Cannot link an image in TinyMCE

sabina-talipova:
## Description
[tinymce.dom.Selection.getSel().toString()](https://www.tiny.cloud/docs-3x/api/dom/class_tinymce.dom.Selection.html/#getsel) returns empty string if HTML Element doesn't have inner content as ``. We should check that user selected an element and if element doesn't have inner content return outer HTML. 
## Parent issue


GuySartorelli:
Max said on the issue that he experienced this issue in CMS 4. Have you looked into that? Can this fix apply to CMS 4 as well or does that need a separate PR?

sabina-talipova:
I tested this issue in CMS 4.13, but didn't have any problem with options in "Insert link" and image element. I assume that the problem with link in CMS 5 connected to [this PR](https://github.com/silverstripe/silverstripe-admin/pull/1517)

Issue #1536, opened 2023-07-21
Regression: Allow help links to be disabled in user code


nathanbrauer:
As requested in https://github.com/silverstripe/silverstripe-cms/issues/1020, #156 added the ability to disable the `help_link`.

In the upgrade from `help_link` to `help_links`, #615 seems to have inadvertently removed this.

The following are expected to work, but now result in `[Warning] foreach() argument must be of type array|object, bool given`:
```yaml
SilverStripe\Admin\LeftAndMain:
  help_links: false
```
```yaml
SilverStripe\Admin\LeftAndMain:
  help_links: null
```

For good measure, I did try the following and it is not effective:
```yaml
SilverStripe\Admin\LeftAndMain:
  help_links: []
```

### PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1572

GuySartorelli:
PR merged

PR #1572, opened 2023-09-04
ENH Allow disabling of help_links

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1536

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -1863,6 +1863,8 @@ class LeftAndMain extends Controller implements PermissionProvider
     /**
      * The urls used for the links in the Help dropdown in the backend
      *
+     * Set this to `false` via yml config if you do not want to show any help links
+     *
      * @config
      * @var array
      */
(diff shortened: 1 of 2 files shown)


code/LeftAndMain.php
```diff
@@ -1889,13 +1889,15 @@ public function getHelpLinks()
...
         });
 
-        foreach ($helpLinks as $key => $value) {
-            $translationKey = str_replace(' ', '', $key ?? '');
-
-            $formattedLinks[] = [
-                'Title' => _t(__CLASS__ . '.' . $translationKey, $key),
-                'URL' => $value
-            ];
+        if (is_array($helpLinks)) {
```
ChloeSartorelli, line 1895: Add a comment here and in the PHPDoc for the `help_links` property that this can be set to `false` to disable it - otherwise it just looked like we're doing some type-safety stuff rather than intentionally ignoring non-array values.

GuySartorelli (changes requested):

code/LeftAndMain.php
emteknetnz: Done

GuySartorelli (approved):
Works well locally. Merge on green.

PR #1574, opened 2023-09-05
ENH Do not show loading spinner when there was a loading error

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1344

The original issue description made this sound worse than it was - there already was user feedback if the video failed to load in the form of the red message "There is currently no embeddable media available from this URL"

Just to make it a little nicer I've removed the infinite loading spinner from showing on error from the general FormBuilderLoader component

**Before**
https://www.youtube.com/watch?v=HIEI_PKx5Lk

**After**
https://www.youtube.com/watch?v=kzI02NaOf9E


GuySartorelli:
I'm not doing a proper review, just looking at this in passing - but it seems to me that when you go to edit it, you should be able to change the URL at the very least.

sabina-talipova (changes requested):
I agree with @GuySartorelli. Probable, if we know that the reason for error is wrong URL, we should provide ability to user to change wrong URL without removing element from HTML Editor.

emteknetnz:
I don't think we need to here, at least not in issue.  The user can just delete the thumbnail to the video from TinyMCE and create a new media link.

Currently that's what the user needs to do to alter the URL of a video that's functioning correctly - when you edit an existing working video the URL field is disabled:

https://www.youtube.com/watch?v=X-ePjsRt1eg

If we wanted to change this behavior so that existing videos could have their URL's changed, this this should be raised an a new enhancement card.  However I don't think this is worth doing because it's very easy to just delete existing videos and create a new one, and also I suspect that it would be a fair amount of effort to develop this functionality.

sabina-talipova:
I assume, that we should open additional ticket to make some enhancements, but only if link is broken, because I think user wants to keep additional settings that have been set up together with media URL.

sabina-talipova (approved):
LGTM. New issue will be opened to make additional improvements.

PR #1577, opened 2023-09-06
FIX Allow wrapping an image in a link

GuySartorelli:
Supercedes https://github.com/silverstripe/silverstripe-admin/pull/1571

Make sure you check the installer CI run linked in the issue to see behat passing happily

## Required for 
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1397
- https://github.com/silverstripe/silverstripe-cms/pull/2883

## Issue


client/src/legacy/TinyMCE_sslink.js
```diff
@@ -111,7 +111,53 @@ jQuery.entwine('ss', ($) => {
...
+      const node = selection.getNode();
+
+      // If there is direct selected content other than whitespace, we can wrap it.
+      if (selectionContent) {
+        return selectionContent.trim() !== '';
+      }
+
+      // If the selected node type can contain text, and we didn't find selected text above,
+      // then you haven't got a selection we can wrap in a link.
+      const x = document.createElement(node.nodeName);
```
emteknetnz, line 129: What is the point of this? Seems like it will will always return false since we're just testing an element we're creating ourselves?  Or is there some sort of difference between browsers?

client/src/legacy/TinyMCE_sslink.js
```diff
@@ -111,7 +111,53 @@ jQuery.entwine('ss', ($) => {
...
+      x.textContent = 'Check the outer HTML';
+      if (x.outerHTML.includes('Check the outer HTML')) {
+        return false;
+      }
+
+      // Check if there is a single selected node which can be wrapped in a link.
+      if (node === selection.getSel().focusNode && node === selection.getSel().anchorNode) {
+        const parsed = tinymce.activeEditor.dom.createFragment(`<a>${node.outerHTML}</a>`);
+        if (parsed.childNodes.length === 1) {
```
emteknetnz, line 138: Again this seems really odd because we're testing the length of something that we're creating ourselves that seems like it will always return the same result?

emteknetnz (changes requested):

client/src/legacy/TinyMCE_sslink.js
ChloeSartorelli: We're essentially getting the browser to check for us if this node _can_ have text. If it can have text, and we didn't have any text selected, then this selection isn't valid for wrapping a link, since we'd want the link to wrap the text.

If this element _cannot_ have text, then we go into the next set of conditional statements to see if this is an element that can be wrapped inside a link (like img), or if it's something that cannot (like iframe or another link)

client/src/legacy/TinyMCE_sslink.js
ChloeSartorelli: As above, except for this one we need tinymce to check if for us instead since tinymce may have its own parsing rules and allowed combinations.

client/src/legacy/TinyMCE_sslink.js
emteknetnz: OK sorry I totally missed the bit where `node` is variable :)

emteknetnz (approved):
Tested locally, works good

PR #1581, opened 2023-09-13
FIX Regression issue UI URLSegment buttons

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2885


client/src/styles/_fonts.scss
sabina-talipova, line 1: I think we can remove whole this block `[data-icon]: before`, since it looks like doesn't effect `[data-icon]` elements in CMS UI except "OK" and "Cancel" buttons. 
But for now, I just exclude these two elements from a list.

maxime-rainville (changes requested):
Those classes will be re-imported next time we update our icon font. There's a good chance we'll inadvertently reintroduce this bug.

My instinct would be to remove the `data-icon="accept"` and `data-icon="cancel"` attributes on those two button. They don't seem to be serving any purpose anyway.

sabina-talipova:
Close. Open https://github.com/silverstripe/silverstripe-cms/pull/2886

PR #1587, opened 2023-09-27
FIX Validation error message with ValidationResult::addError

sabina-talipova:
## Description
An array of error messages that do not relate to certain fields were not passed to the `FormBuilderLoader` component for subsequent rendering into the `Form` component, which in turn are needed to display the `FormAlert`.

## Parent issue
- https://github.com/silverstripe/silverstripe-asset-admin/issues/967

## Test Steps
Follow "Steps to reproduce" from the ticket.

## Notes
- And I think that since the a`ddError` method returns a kind of general error, you should post an error message in `FormAlert`, which is not entirely user friendly since it is not clear where the error was occurred. 
- I think that when a form is validating and the field that needs to be checked is known in advance, you should use the `addFieldError` method, which displays the error directly next to the invalid field


GuySartorelli (approved):
Works perfectly

PR #1588, opened 2023-09-28
FIX HTMLEditorField::setRows minimal height

sabina-talipova:
## Description
Two additional methods will be passed on `tinymce.init` event. They will help to adjust input field part of TyniMCE editor. Methods will set calculated height on init and reset height value on editor resize event.

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10938


client/src/legacy/HtmlEditorField.js
```diff
@@ -156,8 +156,24 @@ ss.editorWrappers.tinyMCE = (function() {
...
+          }
+        }
+        editor.on('init', (e) => {
+          let height = document.querySelector(config.selector).style.height;
+          setHeight(e, height);
+        });
+        editor.on('ResizeEditor', (e) => {
+          setHeight(e, '');
+        });
+      }
```
emteknetnz:
```suggestion
      const setHeight = (event, height) => {
        if (event.target && event.target.iframeElement) {
          event.target.iframeElement.height = height !== '' ? 'auto' : height;
          event.target.iframeElement.parentNode.parentNode.style.height = height;
        }
      }
      const initSetup = (editor) => {
        editor.on('init', (event) => {
          let height = document.querySelector(config.selector).style.height;
          setHeight(event, height);
        });
        editor.on('ResizeEditor', (event) => {
          setHeight(event, '');
        });
      }
```
I'm assuming that `e` was an event object.  If it's not then don't make this change.

Move the setHeight function outside on the initSetup function to make it clearer

Also is it possible to use css seletors here, or at least validate that the element you're selecting with parentNode.parentNode etc is what you expect? TinyMCE may well add in an extra div at some point that will break this, css selectors should be a bit more robust

emteknetnz (changes requested):
Tested locally works good, js could use a little tidy up

emteknetnz (approved):
Tested locally, works good

PR #1594, opened 2023-10-12
FIX Update Storybook logo to the 5.1 OSS logo

edwilde:
I use the [storybook pattern lib](https://silverstripe.github.io/silverstripe-pattern-lib) fairly often, the new OSS logo has rolled out across the rest of dotorg, but appears to have been missed here.

I am not sure how permanent the silverstripe.org url is for the logo, should it be included in this repo instead?


.storybook/theme/SilverstripeTheme.js
```diff
@@ -3,7 +3,7 @@ import { create } from '@storybook/theming/create';
 const SilverstripeTheme = create({
   base: 'light',
   brandTitle: 'Silverstripe CMS - Storybook',
-  brandImage: 'https://www.silverstripe.com/_resources/themes/app/dist/images/logos/silverstripe-logo.svg',
+  brandImage: 'https://www.silverstripe.org/themes/ssv3/img/global-logo-open-source.svg',
```
ChloeSartorelli: Does brandImage take a relative path? If so, I'd much rather just have the image here, either directly in this theme dir or in the .storybook/public/ dir, whichever works (if either works)

GuySartorelli (changes requested):

.storybook/theme/SilverstripeTheme.js
edwilde: Yep, you can include it from a local file. I have updated to include this, and also set a minimum width so the default sidebar width doesn't make it look too small.

Screen Shot 2023-10-27 at 5 16 13 PM

GuySartorelli (approved):
Awesome, thank you for this. I'll let ci run and then merge.

edwilde:
Quick note for any other product storybook noobs out there.

When trying to build initially I got this error.
```
info Found existing addon "@storybook/addon-controls", skipping.
info Found existing addon "@storybook/addon-actions", skipping.
info Found existing addon "@storybook/addon-controls", skipping.
info Found existing addon "@storybook/addon-actions", skipping.
info => Starting manager..
SB_CORE-COMMON_0004: Storybook could not index your stories.
Your main configuration somehow does not contain a 'stories' field, or it resolved to an empty array.

Please check your main configuration file and make sure it exports a 'stories' field that is not an empty array.

More info: https://storybook.js.org/docs/react/faq#can-i-have-a-storybook-with-no-local-stories

    at normalizeStories (./node_modules/@storybook/core-common/dist/index.js:21:2214)
    at iframe_webpack_config_default (./node_modules/@storybook/builder-webpack5/dist/presets/preview-preset.js:3:1890)
    at async starter (./node_modules/@storybook/builder-webpack5/dist/index.js:1:4658)
```

To fix this I had to adjust the `projectRoot` in `./.storybook/lib/storyTeller.js` ([line 111](https://github.com/silverstripe/silverstripe-admin/blob/2/.storybook/lib/storyTeller.js#L111)) and remove `/../../../`.

I then got this error towards the end of the build process
```
ModuleNotFoundError: Module not found: Error: Can't resolve 'constants'
```

This was fixed by adding an empty json object to the root called `./constants.json`
```sh
echo "{}" > constants.json
```

Hope this helps someone in the future 🚀

edwilde:
Thanks @GuySartorelli 👍

Issue #1609, opened 2023-10-19
"Who can edit/view pages on this site?" fields don't hide/show correctly


GuySartorelli:
There's some code in https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/legacy/LeftAndMain.EditForm.js that handles the toggle of displaying the viewer/editor groups for inherited permissions.

This used to also control the display of these fields in `SiteConfig` but doesn't anymore. It seems to be at least partly because the field types are different in siteconfig than they are in page settings - but the JS shouldn't be tied to specific field types.

## Acceptance criteria
- the display of access group fields are toggled appropriated in siteconfig
- The js logic is reasonably robust against the field type changing

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1610

GuySartorelli:
PR merged

PR #1610, opened 2023-10-19
FIX Make CanView/Edit group toggle logic more robust

GuySartorelli:
Note that behat tests will be added to siteconfig as a part of https://github.com/silverstripe/silverstripe-siteconfig/issues/145

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1609


client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -569,7 +569,23 @@ $.entwine('ss', function($){
       }
     },
     onchange: function (e) {
-      this.toggleListDisplay(e.target.value);
+      this.toggleListDisplay(e.target.value, window.matchMedia('(prefers-reduced-motion: reduce)').matches);
```
ChloeSartorelli, line 572: Technically out of scope but this just makes the animation not happen if the user as set their settings to prefer reduced motion. I can split this out if it's adding significant review overhead but I imagine it shouldn't do.

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -596,19 +612,20 @@ $.entwine('ss', function($){
     },
     showGroupsList: function (instant) {
       const holder = this.closest('.field');
-      this.showListElement(holder, holder.next().filter('.treemultiselect'), instant);
+      console.log(holder);
```
emteknetnz: Need to remove the console.log statement

emteknetnz (changes requested):

client/src/legacy/LeftAndMain.EditForm.js
ChloeSartorelli: Oops yup. Done.

emteknetnz (changes requested):
The form field looks really janky when it init's when clicking the "Only these groups" radio.  It is existing behaviour because it does do this on the current non-animated buggy implementation on page load, though it feels significantly less bad because it just looks like a regular FOUT, rather than happening on click which feels much worse

https://www.youtube.com/watch?v=E3GJURQKyxU

Though the janky FOUT is existing behaviour, is there someway to get that to happen while it's sill invisible so that when the user clicks the radio is already been rendered correctly?

GuySartorelli:
> is there someway to get that to happen while it's sill invisible so that when the user clicks the radio is already been rendered correctly?

Dunno. I guess we could put the initial hiding of the field in a delay, so it does the FOUT on page load and then hides the fields? I don't think that's really an improvement though. In any case I'd say that's out of scope probably. I'm not aware of any way to do it that doesn't require a worse on-page-load experience.

emteknetnz (approved):
Yeah probably out of scope since it's an existing issue.  Bugfix does what it's supposed to so will approve.

Issue #1544, opened 2023-08-08
ModelAdmin doesn't have toasts on error or success


GuySartorelli:
When there's a validation error, or successful creation/save of a record in a ModelAdmin, there's no toast messages.

This is _probably_ the case for all records in any GridField, since that's what ultimately powers ModelAdmin's editing capabilities, but I haven't validated that.

Only seen in CMS 4, but I haven't explicitly checked CMS 5 - still need to check if it affects that or not.

## PRs:
- https://github.com/silverstripe/silverstripe-framework/pull/11037
- https://github.com/silverstripe/silverstripe-admin/pull/1615
- https://github.com/silverstripe/silverstripe-versioned/pull/424
- https://github.com/silverstripe/silverstripe-frameworktest/pull/158
- https://github.com/silverstripe/silverstripe-userforms/pull/1250
- https://github.com/silverstripe/silverstripe-taxonomy/pull/97

GuySartorelli:
PRs merged. Patches will be automatically tagged by GitHub actions.

PR #1615, opened 2023-10-31
FIX ModelAdmin toast elements

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1544


code/LeftAndMain.php
```diff
@@ -867,6 +867,9 @@ public function redirect($url, $code = 302)
             if ($this->getRequest()->getHeader('X-Pjax') && !$response->getHeader('X-Pjax')) {
                 $response->addHeader('X-Pjax', $this->getRequest()->getHeader('X-Pjax'));
             }
+            if ($this->getRequest()->getHeader('X-Status') && !$response->getHeader('X-Status')) {
+                $response->addHeader('X-Status', $this->getRequest()->getHeader('X-Status'));
+            }
```
ChloeSartorelli: What is the intention here? I can't think of a scenario where the _request_ would ever include `X-Status`?

GuySartorelli:
Please also add some behat tests (probably just add a "and I should see" to some existing test scenarios)

code/LeftAndMain.php
sabina-talipova: We need this since we lose these headers (`X-Pjax` and `X-Status`) when we sent request for redirection from `GridFieldDetailForm_ItemRequest::redirectAfterSave`.

sabina-talipova:
Updated. Failed tests require PRs:
- https://github.com/silverstripe/silverstripe-versioned/pull/424
- https://github.com/silverstripe/silverstripe-framework/pull/11037

code/LeftAndMain.php
ChloeSartorelli: Can you please add a comment about that to the codebase directly? It could help explain what's happening if someone comes across that code in the future.

sabina-talipova: Updated

code/LeftAndMain.php
```diff
@@ -864,9 +864,17 @@ public function redirect($url, $code = 302)
         if ($this->getRequest()->isAjax()) {
             $response = $this->getResponse();
             $response->addHeader('X-ControllerURL', $url);
+
+            /** Since in some cases we need to pass additional headers,
+             * but we lose these headers (X-Pjax and X-Status) when we set new response
+             * we should explicitly add them again from received request.
+             **/
```
```suggestion
            // In some cases we need to pass additional headers,
            // but we lose these headers (X-Pjax and X-Status) when we set new response.
            // We need to explicitly add them again from the forwarded request.
```

code/LeftAndMain.php
```diff
@@ -864,9 +864,17 @@ public function redirect($url, $code = 302)
         if ($this->getRequest()->isAjax()) {
             $response = $this->getResponse();
             $response->addHeader('X-ControllerURL', $url);
+
+            /** Since in some cases we need to pass additional headers,
+             * but we lose these headers (X-Pjax and X-Status) when we set new response
+             * we should explicitly add them again from received request.
```
ChloeSartorelli: > when we set new response

You said in https://github.com/silverstripe/silverstripe-admin/pull/1615#discussion_r1379395076 that it's to do with a redirect, but here you're saying it's when we set a new response.

If it's because we're using `$this->setResponse($someNewResponse)`, then we have full control over the headers in the new response and should apply them correctly.

If it's because of a redirection, please change the wording in the comment to mention a redirection, and not to mention setting a new response.

GuySartorelli (changes requested):

code/LeftAndMain.php
sabina-talipova: I'm not sure I understand what you mean. But I think it has the right meaning. We send a redirection request with a link to where the user needs to be redirected. In doing so, we create a completely new response. But we also need to save some headers from the request.

sabina-talipova: Removed this block

client/src/legacy/GridField.js
```diff
@@ -569,6 +569,14 @@ $.entwine('ss', function($) {
       } else {
         this._super(e);
       }
+
+      const toastNotificationMessage = $(this).hasClass('action--archive')
+      ? ss.i18n._t('Admin.ARCHIVE_CONFIRM_MESSAGE', 'Archived')
+      : ss.i18n._t('Admin.DELETE_CONFIRM_MESSAGE', 'Deleted');
+
+      if (!$(this).hasClass('action--delete')) {
```
sabina-talipova: The class `action--delete` also exist on `Delete` button in GridField item itself. Check this to avoid invoking message on this event. Otherwise, we have two messages. It _does_ work for the archive button for versioned records.

tests/behat/features/profile.feature
```diff
@@ -48,16 +48,16 @@ Feature: Manage my own settings
     And I fill in "New Password" with "a"
     And I fill in "Confirm Password" with "a"
     And I press the "Save" button
-    Then I should see "Password is too short"
+    Then I should see "Please increase password strength by adding some of the following characters: uppercase, digits, punctuation"
```
sabina-talipova, line 51: Required changes after this PR : https://github.com/silverstripe/silverstripe-framework/pull/10966

sabina-talipova: Test fails since we don't have any settings for `min_test_score`. 
See: https://github.com/silverstripe/silverstripe-framework/blob/8680b6bfb93235547bfac29fcbd9edc8b13c7ab9/_config/passwords.yml#L10-L13
If we have settings then we show the last added message.

client/src/legacy/GridField.js
ChloeSartorelli: I'm not getting any message at all when clicking the "delete" button on an item in the registry modeladmin
image

sabina-talipova: Updated

GuySartorelli:
JS CI is failing - might need to be rerun or you might need to rebuild the dist files.

GuySartorelli (approved):
Looks good and works well locally.
Changes to password in behat aren't problematic so they're okay even though they're not needed.

PR #1617, opened 2023-11-03
TLN Update translations

emteknetnz:
Automated translations update generated using [silverstripe/tx-translator](https://github.com/silverstripe/silverstripe-tx-translator)

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -99,7 +99,6 @@ en:
     WONT_BE_SHOWN: 'Note: this message will not be shown to your visitors'
   SilverStripe\Admin\Navigator\SilverStripeNavigatorItem_Unversioned:
     PREVIEW: Preview
-    'PREVIEW''': Preview
   SilverStripe\Admin\SecurityAdmin:
     ACCESS_HELP: 'Allow viewing, adding and editing users, as well as assigning permissions and roles to them.'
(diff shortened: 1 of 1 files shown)


lang/en.yml
```diff
@@ -99,7 +99,7 @@ en:
     WONT_BE_SHOWN: 'Note: this message will not be shown to your visitors'
   SilverStripe\Admin\Navigator\SilverStripeNavigatorItem_Unversioned:
     PREVIEW: Preview
-    'PREVIEW''': Preview
+    "PREVIEW'": Preview
```
ChloeSartorelli: We should remove the random apostrophe instead. I'm assuming it's in the php code somewhere.
```suggestion
    PREVIEW: Preview
```

GuySartorelli (changes requested):

lang/en.yml
emteknetnz: Have removed key, we don't normally allow this though there is a valid `PREVIEW` key above it and given this is a new entry for 2.0 (was migrated from a different class in 1.13) I don't think this will cause any translation regressions. The latest patch release for 2.0 uses the correct 'PREVIEW' key

GuySartorelli (approved):

Issue #1618, opened 2023-11-05
Implement a dropdown FormField which gets its source via AJAX


GuySartorelli:
As a follow-on from https://github.com/silverstripe/silverstripe-cms/issues/2901, we need a form field that uses AJAX requests to lazy-load its data.

## Required functionality
- Initial list isn't injected into the front end - if it's a `DataList`, no DB query is initiated at all when sending the initial form to the front-end
- When user click on the field, an AJAX request is fired to get some initial results. The initial results is limited to some sane number.
- As user types into the field, AJAX requests are fired with the current field value. The response is a subset of the original list, filtered by the user input
- If user types the same value a second time, a second AJAX request is not sent - in-memory cache is used instead to fetch the results that were previously received for that search term
- in-memory cache is per instance of the field - it is not shared between even identical fields, and it is immediately discarded when the field is de-rendered.

## Acceptance criteria
- [x] Either a new field is implemented with the above functionality, or the above functionality is added to an existing field (e.g. `ListboxField`)
- [x] The above functionality works in both react and entwine contexts
- [x] The temporary workaround from https://github.com/silverstripe/silverstripe-cms/issues/2901 is removed
- [x] The new implementation is in use for all fields that fetch `Member` records by default (e.g. the "only these users" fields in cms, asset-admin, and siteconfig)
- [x] The new implementation is used instead of dropdown field for [auto-scaffolding has_one relation fields](https://github.com/silverstripe/silverstripe-framework/blob/8680b6bfb93235547bfac29fcbd9edc8b13c7ab9/src/ORM/FieldType/DBForeignKey.php#L111-L127) (NOTE: We only do this if it's 100% backwards compatible with the existing form field, in terms of its public API)
  - [ ] The [`DBForeignKey.dropdown_field_threshold`](https://github.com/silverstripe/silverstripe-framework/blob/8680b6bfb93235547bfac29fcbd9edc8b13c7ab9/src/ORM/FieldType/DBForeignKey.php#L40) configuration property is deprecated
- [x] Validate the UX of the new field with our design team
- [x] There's a field  that allows you to select a single value and there's a field that allows you to select multiple values.

## Notes
- The [silverstripe/tagfield](https://github.com/silverstripe/silverstripe-tagfield) implementation has this functionality - we can likely grab a lot of that code and reuse it in core.
- The TreeDropDown field that uses a similar approach as well.
- Might be worthwhile reviewing which fields have AJAX search capabilities.

## New issues created
[...]

PR #1625, opened 2023-11-20
NEW SearchableDropdownField

emteknetnz:

This code is adapted from [silverstripe/tagfield](https://github.com/silverstripe/silverstripe-tagfield/)

Framework PR
- https://github.com/silverstripe/silverstripe-framework/pull/11057


emteknetnz:
Design feedback:
- try and remove arrow for lazy load
- get rid of mid-grey hover state
- get rid of 'active hover' state blue, make it mid-blue 'active non-hover state'

- change padding on multi-selected options - react-select-9jq23d 
padding: 3px 10px 3px 10px;

.ss-searchable-dropdown-field__multi-value__remove
padding: 2px 5px 2px 5px;

.ss-searchable-dropdown-field__multi-value
change border color from #66afe9 (light blue) to grey #c1cad9

.ss-searchable-dropdown-field__multi-value
border-left: #c1cad9

.ss-searchable-dropdown-field__multi-value__remove:focus, .ss-searchable-dropdown-field__multi-value__remove:hover
background-color: #ffe0e0 (randomly grabbed, ideally would have an existing variable, or use calc() on $branch-danger)
color: #d40404 ($brand-danger)

emteknetnz:
Design review implementation has been approved by internal design team

client/src/bundles/bundle.js
```diff
@@ -81,6 +81,7 @@ import 'expose-loader?exposes=formatWrittenNumber!lib/formatWrittenNumber';
 import 'expose-loader?exposes=withDragDropContext!lib/withDragDropContext';
 import 'expose-loader?exposes=withRouter!lib/withRouter';
 import 'expose-loader?exposes=ssUrlLib!lib/urls';
+import 'expose-loader?exposes=SearchableDropdownField!components/SearchableDropdownField/SearchableDropdownField';
```
ChloeSartorelli, line 84: Exposing things through here needs to be coupled with a PR to the webpack config, or people can't use it.

GuySartorelli (changes requested):
Code looks broadly good.
Haven't fully tested it locally yet - will wait for the framework changes.

One thing I did notice: when I type to search, it swaps the dirty state of the form, making it look like there are unsaved changes.

IIRC we had that problem with tagfield somewhat recently as well.

client/src/bundles/bundle.js
emteknetnz: https://github.com/silverstripe/webpack-config/pull/64

emteknetnz:
I think I've fixed the issue with the fix the dirty state on search

GuySartorelli:
If I select an option and then clear that option (without saving inbetween), the page still thinks I have unsaved changes.

1. Page loads. No unsaved changes.
2. I type a search term. No unsaved changes.
3. I select an option. Page correctly identifies changes.
4. I clear the selection. Page incorrectly thinks there are still unsaved changes.

emteknetnz:
I've tried really hard to fix this though I just cannot.  There's a 'changed' class that gets added by [jquery.changetracker](https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/legacy/jquery.changetracker.js#L63) to an `<input>` that react-select added.  `jquery.changetracker` just blindly tracks virtually all inputs, which is great until you get situations like this.

The issue seems be that `jquery.changetracker` has full control over the 'changed' class being added and there's nothing you can do about it.

Setting a `no-change-track` className on the `<DynamicComponent>` will block all change tracking.  I've tried that along with also adding in a new `<input>` property as a sibling to `<DynamicComponent>`, manually setting the className on the input to 'changed' based on whether there are changes and dispatching the change event from that element. But that doesn't work, I still don't get to control whether or not the 'changed' class appears on the input

GuySartorelli (changes requested):
Please add behat tests - especially for https://github.com/silverstripe/silverstripe-framework/pull/11057#discussion_r1424715991

emteknetnz:
We don't have behat tests for FormFields in admin, instead we use jest tests.  This PR has jest tests.  For the 'clearable' functionality called out in https://github.com/silverstripe/silverstripe-framework/pull/11057#discussion_r1424715991 - that's functionality on [react-select](https://react-select.com) so all the PHP code is doing is passing a prop to it

GuySartorelli (approved):
Looks good, works great locally.

Issue #1430, opened 2023-01-06
1.12 horizontal scrollbar in split mode


tiller1010:
This PR removed a style rule that prevented horizontal scrollbars when in split mode:
https://github.com/silverstripe/silverstripe-admin/pull/1422

The scrollbar **does not** show in Firefox, but it does show in Brave and Edge.

### Firefox
![firefox](https://user-images.githubusercontent.com/42498429/211030221-5fbece45-f58b-4aa2-bdaf-518f9fa61c0d.png)

### Brave
![brave](https://user-images.githubusercontent.com/42498429/211030224-0a55e163-6173-42d0-be7d-881adbe672c1.png)

### Edge
![edge](https://user-images.githubusercontent.com/42498429/211030227-a12a161b-734d-43a3-818e-81a3fff804a8.png)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1629

StevenLantinga:
For anybody wondering: This is caused by the mode dropdown itself in the right bottom. Adding `#preview-mode-dropdown-in-content .chosen-drop { right: 0; width: 160px; }` fixed the issue.

RVXD:
I guess this would be a minor change to fix? Issue has been open since jan 6. Is there any way I can help?
One note: the issue is also visible in non-split mode, so maybe change title of issue.

lerni:
Use :fire::fox_face::exclamation: :grinning:
To make the suggested fix above also work in preview-mode, I figured
```css
/* in 5.x we need to revert #7b430d4 and add the previous snippet bellow
https://github.com/silverstripe/silverstripe-admin/commit/7b430d4171da9608918190109b85e15260e0d626 */
.chosen-container:not(.chosen-container-active) {
	overflow: hidden;
}

.preview-mode-selector .chosen-drop {
	width: inherit;
	right: 0;
}
```
If that looks acceptable I would try do make build-env working, PR and also test it a bit more. To what branch:question: If 4.x, is build than done on up-merge for 5?

RVXD:
Can this be implemented? Now all CMS's have a horizontal scrollbar. This is still visible in 5.1.

GuySartorelli:
Because this is a very minor visual bug, the CMS Squad can't dedicate time to fixing this right now. If this is causing problems for you, I recommend opening a pull request that fixes it.

RVXD:
[...]

PR #1629, opened 2023-12-01
Apply fix to prevent horizontal scrollbar in the CMS admin

lerni:
As mentioned in https://github.com/silverstripe/silverstripe-admin/issues/1430#issuecomment-1625840199 to make the suggested fix work with current 5.x 7b430d4 is manually reverted in this PR.


client/src/styles/_chosen.scss
```diff
@@ -3,3 +3,8 @@
 .chosen-container-multi .chosen-choices {
   padding: 4px 5px;
 }
+
+// Avoid chosen dropdowns overflowing when not active
```
tiller1010, line 7: Could we add a comment that this is only necessary for browsers other than Firefox? I think the original issue was that the SS devs thought that they didn't need this anymore because they were likely using firefox.

tiller1010 (changes requested):

client/src/styles/_chosen.scss
ChloeSartorelli: That's a great idea. Please add a comment to that effect.

GuySartorelli:
You mention that this undoes a commit which was part of https://github.com/silverstripe/silverstripe-admin/pull/1422 - can you please validate whether the issue that pr fixed is regressing as a result of this change?

There's also more to this change than undoing the commit you mentioned. Are the additional changes also required?

client/src/styles/_chosen.scss
lerni: I'm not sure what 7b430d4 was about (sounds similar to current issue) and if it was related to FF vs. Webkit/other browsers. Closest is https://github.com/silverstripe/silverstripe-admin/pull/604 - https://github.com/silverstripeltd/product-issues/issues/630 throws a 404.

I've made the fix before 7b430d4 was merged and noticed, the changes didn't work afterwards. So I reverted changed style from 7b430d4 and also added my previous changes, which are specific to Webkit/other browsers, so comment would make more sense in the part where `width: inherit; right: 0;` was added?

ChloeSartorelli: The comment should be written and positioned in a way that clearly covers any changes which work towards resolving this bug, since this bug as a whole only affects non-firefox.

GuySartorelli:
(fyi I've updated the PR description and your latest comment to refer to the commit hash _without_ a `#`, so that GitHub can link to it)

client/src/styles/_chosen.scss
lerni: I've added comments. Still not sure what the intend of #1422 was. Guy you mentioned it probable was to fix a bug with fluent locale switcher. I've tried to reproduce but couldn't. The fluent-switch is a native dropdown no chosen-thingy AFAICT.

GuySartorelli (changes requested):
This causes a regression - https://github.com/silverstripeltd/product-issues/issues/630

Note that the issue isn't talking about the locale dropdown in the left-hand vertical menu, it's talking about the locale dropdown when _creating_ a new locale with fluent.

Can you please see if there's a way to resolve the scrolling bug without regressing the fluent locale bug?

Also, and sorry I didn't notice this earlier, but this PR will need to target the `2.1` branch so that we can release it as a patch.

lerni:
OK finally I know what https://github.com/silverstripeltd/product-issues/issues/630 was about (private repo) and now can reproduce.

https://github.com/silverstripe/silverstripe-admin/assets/472115/b6d413ab-0d67-448e-9f92-3b59221745ed

Prepared a workaround to prevent regression, but it needs some more testing.

GuySartorelli:
@lerni How's it going with this? Have you had a chance to test your workaround yet?
Feel free to include the workaround here if you like and I can see what does/doesn't work.

lerni:
Oh my, rebase went wrong - 'll try some more git foo. Basically now `.chosen-drop` (element inside container) has `overflow: hidden` if active.
```scss
.chosen-container:not(.chosen-container-active) .chosen-drop {...}
```

GuySartorelli:
Force pushed (with no code change) to get CI to run.
I took that opportunity to update the commit message to include the `FIX` prefix as well.

GuySartorelli (approved):
Looks good to me - if CI goes green I'll merge.

GuySartorelli:
Thanks for taking the time to get this over the line 💙

PR #1630, opened 2023-12-01
fix: made CMSEditLinkExtension work with Hierarchy, added dynamic get…

### LINKED ISSUE #1628: CMSEditLinkExtension doesn't work with Hierarchy
<details>
<summary>Original context</summary>
We have a class that uses both the Hierarchy and CMSEditLinkExtension extensions. After the ParenID is changed, and the ContentLayout is saved, I get the error: `[Emergency] Uncaught LogicException: Could not produce an edit link for the passed object.`

It doesn't look like there is any support for DataObjects in a Hierarchy in the CMSEditLinkExtension. Instead, it looks like it only checks for has_many/has_one relationships: https://github.com/silverstripe/silverstripe-admin/blob/2/code/CMSEditLinkExtension.php#L62

```php
class ContentLayout extends DataObject implements CMSPreviewable
{
  private static $cms_edit_owner = 'Parent';

  private static $extensions = [
    Hierarchy::class,
    CMSEditLinkExtension::class,
  ];

  /*
    CMSEditLink
    Returns the cms edit link for the content layout object
  */
  public function CMSEditLink()
  {
    // Get the value returned by the CMSEditLinkExtension
    return $this->extend('CMSEditLink')[0];
  }

  {
    ...
    // Parent Container
    $Parent = TreeDropdownField::create('ParentID', 'Parent', ContentLayout::class)
      ->setEmptyString('Top Level');
    ...
  }
}
```
</details>

If a class has the `Hierarchy` extension it cannot rely on the `Parent` relation as the owner relation, because there is no corresponding has_many relation. `Hierarchy` relies on its own methods to get the related children records.

Developers could choose either of `Children()` or `AllChildren()` to put in the `GridField` so we'd need to check if either of these are editable, and if so return the link for it. They could also implement their own method to link back to the child records, though that would be out of scope.

Note that we also need to validate that the parent record and the child record both share the same hierarchical base class before getting a hierarchical link - otherwise you could have a `has_one` from `MyRandomDataObject` to `SiteTree`, and accidentally try to give a hierarchical link even though they don't share a hierarchical base class.

There's also no way to define what the edit owner is for the _root level_ item in the hierarchy.

e.g:

```text
some admin section
└─ root record
       └─ child record
               └─ grandchild record
```

The grandchild checks its parent and finds the child. The child checks its parent and finds the root. But the root will try to check its parent and find nothing...

tiller1010:
…CMSEditOwnerRelation method

Closes #1628 

- [ ] create a has_many relationship on a dataobject with a dataobject that uses the Hierarchy extension: https://github.com/silverstripe/silverstripe-admin/issues/1628#issuecomment-1831979928
- [ ] add a cms field to your Hierarchy dataobject to set the ParentID
- [ ] confirm saving the dataobject after changing the parentid properly redirects you to the cms edit link

### DIFF
--- a/code/CMSEditLinkExtension.php
+++ b/code/CMSEditLinkExtension.php
@@ -11,6 +11,7 @@ use SilverStripe\Forms\FieldList;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\GridField\GridFieldDetailForm;
 use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\Hierarchy\Hierarchy;
 
 /**
  * An extension that automatically generates a CMS edit link for DataObjects even if
@@ -37,6 +38,11 @@ class CMSEditLinkExtension extends Extension
 {
     private static string $cms_edit_owner = '';
 
+    public function getCMSEditOwnerRelation()
+    {
+        return $this->owner->config()->get('cms_edit_owner');
+    }
+
     /**
      * Get the ModelAdmin, LeftAndMain, or DataObject which owns this object for CMS editing purposes.
      *
@@ -44,7 +50,7 @@ class CMSEditLinkExtension extends Extension
      */
     public function getCMSEditOwner()
     {
-        $ownerType = $this->owner->config()->get('cms_edit_owner');
+        $ownerType = $this->owner->getCMSEditOwnerRelation();
         if (is_subclass_of($ownerType, LeftAndMain::class)) {
             return $ownerType::singleton();
         }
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Tha k you contributing - however based on the information in the original issue, this _does_ work correctly with the hierarchy extension, but doesn't support having different parents. I don't think we want support for that at this level. Instead, the `getCMSEditOwner()` method should be overridden dynamically return the correct owner based on project logic.

tiller1010:
> based on the information in the original issue, this does work correctly with the hierarchy extension

No, this PR calls `AllChildren()` instead of `hasMany` for Hierarchy relations.

PR #1631, opened 2023-12-03
Refactor FormBuilderModal and split up the Modal into its own component

maxime-rainville:
In the process of trying to figure out what InsertMediaModal does, I re-factored the FormBuilderModal:
- Took the bit that were just generic rendering a Modal and split them off into their own component.
- Converted FormBuilderModal to a functional component.

None of the API on the component has changed, so everything should still be working.

The extra Modal component will probably have extra usage. InsertMediaModal actually use FormBuilderModal not to render an actual form schema, but for the Modal bit.

## Parent issue
- https://github.com/silverstripe/silverstripe-asset-admin/issues/1422
- #1677


GuySartorelli (commented):
The fact that you're refactoring this whole thing into a functional component makes it really hard to see what's going on. In the future, please _at least_ do that in a separate commit, so it's clear what changes are related to that refactor vs what's related to the thing the PR is intended to actually do.

I'm not going to look at the code in this PR for exactly that reason. Instead, I'll just comment on what it seems this PR should be doing based on what you've written in the description:


> Took the bit that were just generic rendering a Modal and split them off into their own component.

What uses do you foresee for a generic modal which doesn't have a form in it? I'm struggling to see the value in this change.

emteknetnz:
+10 on the "do refactoring in a separate PR". Makes peer review so much harder all being cluttered together.

GuySartorelli:
To be clear, I'm just saying in a separate commit, not a separate PR. You can view changes by commit, which would be fine for this POC PR I think.

maxime-rainville:
I'll see if I can split the part that creates new components from the part that converts InsertMediaModal to be a functional component. Or at least, highlight what parts in the new components map to the old component.

But for the record, the PR is called *Refactor FormBuilderModal*. Refactoring is the entire point.

GuySartorelli:
> for the record, the PR is called Refactor FormBuilderModal. Refactoring is the entire point.

Fair, I suppose, but it's also intended to illustrate a point which is really hard to see with multiple types of refactor happening at once.

GuySartorelli:
No changes are needed at this point since the PR is only illustrative at this stage - but my question in https://github.com/silverstripe/silverstripe-admin/pull/1631#pullrequestreview-1763415094 is still relevant:

> What uses do you foresee for a generic modal which doesn't have a form in it? I'm struggling to see the value in this change.

maxime-rainville:
The commits are split. I added some tests.

maxime-rainville:
I think we'll need to reubil once https://github.com/silverstripe/webpack-config/pull/66 is merged.

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,47 @@
...
+      {children}
+    </ModalComponent>
+  );
+}
+
+Modal.propTypes = {
+  isOpen: PropTypes.bool,
+  className: PropTypes.string,
+  modalClassName: PropTypes.string,
+  size: PropTypes.oneOf(['', 'sm', 'lg', 'xl']),
```
maxime-rainville: After looking at the underlying Reactstrap implementation, theses are actually the only value this thing will accept.

client/src/components/Modal/tests/Modal-story.jsx
maxime-rainville, line 1: There's a slight weirdness in that once the Modal shows in the pattern library it will de displayed over the props control. I tried really heard to try to wire the onClosed handle so it toggles off the Storybook props, but it didn't want to play ball with me.

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -1,104 +1,71 @@
...
   /**
    * Call the callback for hiding this Modal
    */
-  handleHide() {
-    this.clearResponse();
-    if (typeof this.props.onClosed === 'function') {
-      this.props.onClosed();
+  const handleHide = () => {
+    clearResponse();
+    if (typeof onClosed === 'function') {
```
emteknetnz, line 65:
```suggestion
    if (onClosed) {
```
Type check via propTypes

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -1,104 +1,71 @@
-import React, { Component } from 'react';
-import i18n from 'i18n';
-import { Modal, ModalHeader } from 'reactstrap';
+import React, { useState } from 'react';
 import FormBuilderLoader from 'containers/FormBuilderLoader/FormBuilderLoader';
 import castStringToElement from 'lib/castStringToElement';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
+import Modal from 'components/Modal/Modal';
+import pick from 'lodash/pick';
```
emteknetnz:
```suggestion
```
We don't need lodash

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -108,107 +75,65 @@ class FormBuilderModal extends Component {
...
-      if (typeof title === 'object') {
-        // FormSchema title occasionaly contains html, only render text for modal title
-        const doc = new DOMParser().parseFromString(title.html, 'text/html');
-        title = doc.body.textContent || '';
+  };
+
+  const modalProps = pick(props, Object.keys(Modal.propTypes));
+  const formBuilderProps = pick(props, [
+    'schemaUrl', 'bodyClassName', 'autoFocus', 'onAction', 'identifier'
+  ]);
```
emteknetnz:
```suggestion
  const modalProps = props;
  const formBuilderProps = { schemaUrl, bodyClassName, autoFocus, onAction, identifier };
```
However we seem to be missing some props from the props declaration above - autoFocus, onAction, identifier

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -108,107 +75,65 @@ class FormBuilderModal extends Component {
...
+
+  return (
+    <Modal {...modalProps} onClosed={handleHide}>
```
emteknetnz:
```suggestion
    <Modal {...modalProps}>
```
Add onClosed to modalProps declaration above

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -108,107 +75,65 @@ class FormBuilderModal extends Component {
...
-          className="close modal__close-button"
-          onClick={this.handleHide}
-          aria-label={i18n._t('Admin.CLOSE', 'Close')}
+      {schemaUrl &&
+        <FormBuilderLoaderComponent
+          {...formBuilderProps}
+          fieldHolder={{ className: classnames('modal-body', bodyClassName) }}
+          actionHolder={{ className: 'modal-footer' }}
+          onSubmit={handleSubmit}
+          onLoadingError={handleLoadingError}
```
emteknetnz:
```suggestion
```
Add these props to formBuilderProps declaration above

client/src/components/Modal/Modal.js
```diff
@@ -0,0 +1,47 @@
...
+import PropTypes from 'prop-types';
+import ModalHeader from './ModalHeader';
+
+/**
+ * Component to render a modal
+ */
+function Modal({
+  onClosed, title, ModalHeaderComponent, showCloseButton, children,
+  ModalComponent, isOpen, className, modalClassName, size
+}) {
```
emteknetnz:
```suggestion
const Modal = ({
  children,
  className,
  isOpen,
  onClosed,
  modalClassName,
  ModalComponent,
  ModalHeaderComponent,
  showCloseButton,
  size
  title,
}) => {
```

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -1,104 +1,71 @@
...
+    responseClassGood,
+    onLoadingError,
+    showErrorMessage,
+    onClosed,
+    onSubmit,
+    schemaUrl,
+    bodyClassName,
+    FormBuilderLoaderComponent,
+    children
+  } = props;
```
emteknetnz:
```suggestion
const FormBuilderModal = ({
  bodyClassName,
  children,
  FormBuilderLoaderComponent,
  onLoadingError,
  onClosed,
  onSubmit,
  responseClassBad,
  responseClassGood,
  schemaUrl,
  showErrorMessage,
}) => {
```
While I certainly prefer just using the word `function`, it doesn't match the result of the codebase so use const + arrow instead.

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -1,104 +1,71 @@
...
+  };
+
+  // Clear the response
+  const clearResponse = () => {
+    setResponse(null);
+    setError(false);
+  };
+
+  const handleLoadingError = (schema) => {
+    const providesOnLoadingError = onLoadingError !== noop;
```
emteknetnz, line 46: Why are we doing this? Just check for the truthiness of a the onLoadingError prop

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -1,104 +1,71 @@
...
+    bodyClassName,
+    FormBuilderLoaderComponent,
+    children
+  } = props;
+
+  const [response, setResponse] = useState(null);
+  const [error, setError] = useState(null);
+
+  // Register a success message
+  const setSuccess = (message) => {
```
emteknetnz:
```suggestion
  const handleSuccess = (message) => {
```
Prefixing with `set` is confusing as it makes it sound like we're updating a state hook

Update the other 2 functions below

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -108,107 +75,65 @@ class FormBuilderModal extends Component {
...
+  const modalProps = pick(props, Object.keys(Modal.propTypes));
+  const formBuilderProps = pick(props, [
+    'schemaUrl', 'bodyClassName', 'autoFocus', 'onAction', 'identifier'
+  ]);
+
+  return (
+    <Modal {...modalProps} onClosed={handleHide}>
+      {response &&
+        <div className={error ? responseClassBad : responseClassGood}>
+          { castStringToElement('span', { html: response }) }
```
emteknetnz, line 140: What is this doing?

client/src/components/Modal/ModalHeader.js
```diff
@@ -0,0 +1,51 @@
...
+      // FormSchema title occasionally contains html, only render text for modal title
+      const doc = new DOMParser().parseFromString(title.html, 'text/html');
+      title = doc.body.textContent || '';
+    }
+    return (
+      <ModalHeaderComponent toggle={onClosed}>{title}</ModalHeaderComponent>
+    );
+  }
+
+  if (showCloseButton === true && typeof onClosed === 'function') {
```
emteknetnz:
```suggestion
  if (showCloseButton && onClosed) {
```
Rely on propTypes for type checking

client/src/components/Modal/ModalHeader.js
```diff
@@ -0,0 +1,51 @@
+import React from 'react';
+import i18n from 'i18n';
+import { ModalHeader as ReactStrapModalHeader } from 'reactstrap';
+import PropTypes from 'prop-types';
+
+/**
+ * Component to render a modal header.
+ */
+function ModalHeader({ onClosed, title, ModalHeaderComponent, showCloseButton }) {
```
emteknetnz:
```suggestion
const ModalHeader = ({ onClosed, title, ModalHeaderComponent, showCloseButton }) => {
```

client/src/components/Modal/ModalHeader.js
```diff
@@ -0,0 +1,51 @@
...
+      title = doc.body.textContent || '';
+    }
+    return (
+      <ModalHeaderComponent toggle={onClosed}>{title}</ModalHeaderComponent>
+    );
+  }
+
+  if (showCloseButton === true && typeof onClosed === 'function') {
+    return (
+      <button
```
emteknetnz: We should be using a reactstrap button?

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -108,107 +75,65 @@ class FormBuilderModal extends Component {
...
-    }
-
-    if (this.props.showCloseButton === true && typeof this.props.onClosed === 'function') {
-      return (
-        <button
-          type="button"
-          className="close modal__close-button"
-          onClick={this.handleHide}
-          aria-label={i18n._t('Admin.CLOSE', 'Close')}
+      {schemaUrl &&
```
emteknetnz: Should make schemaUrl a required prop and just assume it's there i.e. no conditional check - things won't work at all if schemaUrl is missing or wrong - up to developer to fix

client/src/components/Modal/tests/Modal-test.js
```diff
@@ -0,0 +1,89 @@
...
+      {Object
+        .entries(propsToPrint)
+        .map(([key, value]) => (<li key={key}>{key}: {value.toString()}</li>))
+      }
+    </ul>
+    {children}
+  </div>
+);
+
+beforeEach(() => {
```
emteknetnz: Change this file to use the `makeProps()` pattern to match the other jest tests e.g. https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/components/SearchableDropdownField/tests/SearchableDropdownField-test.js#L8

client/src/components/Modal/tests/ModalHeader-test.js
```diff
@@ -0,0 +1,33 @@
+/* global jest, test, describe, it, expect */
+
+import React from 'react';
+import { fireEvent, render } from '@testing-library/react';
+import ModalHeader from '../ModalHeader';
+
+let props;
+
+beforeEach(() => {
```
emteknetnz, line 9: Change this file to use makeProps()

client/src/components/Modal/tests/Modal-test.js
```diff
@@ -0,0 +1,89 @@
...
+  expect(modal.classList).toContain('modal');
+
+  const document = root.getByRole('document');
+  expect(document.classList).toContain('modal-sm');
+
+  const header = root.getByRole('heading');
+  expect(header.textContent).toContain('Hello World!');
+
+  const body = root.getByText('My Content');
+  expect(body).toBeTruthy();
```
emteknetnz:
```suggestion
  const renameThis = root.getByRole('document');
  expect(renameThis.classList).toContain('modal-sm');

  const header = root.getByRole('heading');
  expect(header.textContent).toContain('Hello World!');

  const content = root.getByText('My Content');
  expect(content).not.toBeNull();
```
Can you not use the words 'document' or 'body' for variable names here, it's really confusing as they correspond to other things in the regular DOM

Also instead of `.toBeTruthy()` use `.not.ToBeNull()` to match other tests in the codebase

client/src/components/Modal/tests/Modal-test.js
```diff
@@ -0,0 +1,89 @@
...
+  fireEvent.click(closeButton);
+
+  expect(props.onClosed).toBeCalled();
+});
+
+test('Modal is not shown', () => {
+  const root = render(
+    <Modal {...props} isOpen={false}>My Content</Modal>
+  );
+  expect(root.container.textContent).toBeFalsy();
```
emteknetnz:
```suggestion
  expect(root.container.querySelectorAll('.some-class')).toHaveLength(0);
```
This is how we do this sort of thing elsewhere

client/src/components/Modal/tests/Modal-test.js
```diff
@@ -0,0 +1,89 @@
...
+    <Modal {...props}>My Content</Modal>
+  );
+
+  expect(root.getByText('Custom Modal Component')).toBeTruthy();
+  expect(root.getByText('isOpen: true')).toBeTruthy();
+  expect(root.getByText('className: my-modal')).toBeTruthy();
+  expect(root.getByText('modalClassName: my-modal-dialog')).toBeTruthy();
+  expect(root.getByText('size: sm')).toBeTruthy();
+  expect(root.getByText('Custom Modal Header Component')).toBeTruthy();
+});
```
emteknetnz:
```suggestion
```
Seems very pointless and overkill to test that we do indeed pass some props to a child component

There's about a million instances where we don't test this and I see no reason to start now

client/src/components/Modal/tests/Modal-test.js
```diff
@@ -0,0 +1,89 @@
...
+    <div>{componentName}</div>
+    <ul>
+      {Object
+        .entries(propsToPrint)
+        .map(([key, value]) => (<li key={key}>{key}: {value.toString()}</li>))
+      }
+    </ul>
+    {children}
+  </div>
+);
```
emteknetnz, line 19:
```suggestion
);
```
I don't think we need this, I put reasons why in corresponding deleted test where it was used

emteknetnz (changes requested):
I haven't actually tested this, I basically just did a syntax check

You'll also need to create a special CI run where this PR is used as part of kitchen sink so that we run integration tests with other modules

I have a local script that automates doing this, results look like this - https://github.com/emteknetnz/recipe-kitchen-sink/pull/4 - so you should be able to copy paste the results of this and use it on your own account after making changes to this PR

client/src/components/FormBuilderModal/FormBuilderModal.js
ChloeSartorelli: propTypes doesn't actually affect runtime, I don't think?

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: I'm not 100% sure ... this part was blindly adapted from the old class component.

That's the only place `showErrorMessage` is used. If we only check the truthiness of `onLoadingError`, we're altering the behaviour of the component.

My approach was very much "the end result has to be identical".

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: `proptypes` are a soft check on types. They will throw a console warnings if the prop doesn't match the expected type.

In principal, I think it's fair to tell people "we've defined proptypes to tell you what we expect ... if you ignore them and give us something else, that's on you".

I'm also not opposed to doubling up those checks either.

In this instance, I'm simply replicated the pre-existing behaviour. So the suggestion would somewhat alter the original behaviour of the component.

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: `setSuccess` and `setFailure` are updating state hooks. They are essentially hooks.

I've moved all the `useState` logic to its own function. This as the added benefit of making a somewhat long component a bit smaller and manageable.

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: The `pick` logic was an attempt at trying to avoid a complex deconstruction of the modal props. In hindsight, it does look a bit too cute for its own good.

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: Done

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: Not sure. It was copied over from the original class component.

My guess is that it takes the response message, wraps it in a `<span>` tag and renders it. The response can probably contain HTML so you can't just render it directly or it will try to escape the HTML tag.

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: We probably should ... at some point.

The problem is that the FormSchemaModal was written with the some what silly assumption that you could choose to use it without a form schema to render a plain modal ... which is why I ended looking at this in the first place.

If we make the schema URL required, we'll probably break the InsertMediaModal in the short term ... I'll do a follow up PR to update InsertMediaModal to use the new Modal component instead.

We may break other things I don't know about.

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: This button is part of the ModalHeader component now. They are passed via the `modalProps`.

client/src/components/Modal/ModalHeader.js
maxime-rainville: Not sure ... I don't think there's any scenario where we are calling the FormBuilderModal without a title. So this bit may vary well be completely pointless.

This is what it looks like in the pattern library.
image

I have some difficulty imagining anyone out there is using this layout. I'd say leave it as is ... or maybe remove this `if` statement altogether.

client/src/components/Modal/tests/Modal-test.js
maxime-rainville: My assumption as always been that `beforeEach` is the proper way to reset props in-between test.

Is this `makeProps` pattern something official or did we make this up?

client/src/components/Modal/tests/Modal-test.js
maxime-rainville: For some reason I'm not quite sure about, `root.container.querySelectorAll` never returns anything for any of my Modal tests.

While trying to debug why it wasn't working, I came across [this official guidance](https://testing-library.com/docs/queries/about/#priority) that strongly to avoid CSS selectors to find element from the virtual dom.

> your test should resemble how users interact with your code (component, page, etc.) as much as possible
> ...
> Queries that reflect the experience of visual/mouse users as well as those that use assistive technology.

client/src/components/Modal/tests/Modal-test.js
maxime-rainville: In this specific instance, the FormBuilderModal - stupidly in my opinion - let you pass your own Modal and ModalHeader components.

This seems to at least partially imply that the signature of those components are predefined and that we can't threat them as a purely internal concern.

maxime-rainville (commented):
Answered most of the feedback.

Behat installer build can be found here: https://github.com/creative-commoners/silverstripe-installer/pull/2

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: Removed

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: Done

client/src/components/Modal/ModalHeader.js
maxime-rainville: See https://github.com/silverstripe/silverstripe-admin/pull/1631#discussion_r1487142355

client/src/components/Modal/ModalHeader.js
maxime-rainville: Done

client/src/components/Modal/tests/Modal-test.js
emteknetnz: It's made up ... though it's used heavily in our codebase so it is the standard.

makeProps() gives a fairly clean way to override any 'default' props that would have been defined in the beforeEach()

client/src/components/Modal/tests/Modal-test.js
emteknetnz: :-) That guidance also means you should be using things like `await screen.findByRole()` to sort out all the async fun present in javascript.  You're more than welcome to use that if you want, though container.querySelector() has been used throughout our codebase because it's far easier to work with.

The fact that container.querySelector() isn't working is kinda concerning, probably means that your old test wasn't actually testing what you thought it was

Testing negatives is hard. You should be able to use something like this to simulate waiting for everything to load without using a screen.findBy*() method:

```
import { render, act } from '@testing-library/react';
// ...
  await act(async () => {
    await new Promise((resolve) => setTimeout(resolve, 100));
  });
```

client/src/components/Modal/ModalHeader.js
emteknetnz: Leave as is

client/src/components/FormBuilderModal/FormBuilderModal.js
emteknetnz: Oh right it's in a custom hook

The use of a custom is honestly just confusing, it's adding complexity where it doesn't need to be added, it's not making things any simpler

For instance `setSuccess()` is called once. However there are 5 matches in the file if I search for `setSuccess`

Remove the hook and just put the state directly in the component

client/src/components/FormBuilderModal/FormBuilderModal.js
emteknetnz: I really don't understand what you're saying.  What do you mean "This button"?

These are props for the `<FormBuilderLoaderComponent` which takes `formBuilderProps`, not `modalProps`

emteknetnz (changes requested):

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -108,107 +103,80 @@ class FormBuilderModal extends Component {
...
+    modalClassName,
+    ModalComponent,
+    ModalHeaderComponent,
+    onClosed: handleHide,
+    showCloseButton,
+    size,
+    title,
+
+  };
+  const formBuilderProps = {
```
emteknetnz:
```suggestion
  const formBuilderLoaderProps = {
```
Make it consistent with the name of the component it's going into

emteknetnz (changes requested):

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: `clearResponse` was part a method on the class component.

I created the other method for symmetry's sake. But I can see the point that it's overkilled in this context.

I ditched setSuccess, setFailure and clearResponse hide and manually update the state each time.

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: Done

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: Sorry ... brain fart on my part I saw the <button> tag on the "before" side and thought the comment was about the button.

client/src/components/Modal/tests/Modal-test.js
maxime-rainville: Most of the example I found recommended using `beforeEach()` for this purpose, but those were mostly opinionated articles. Couldn't find any official. Happy to use `makeProps` if that's the thing that we've been using so far.

I find the notion that we should pass the props overrides to `makeProps` kind of weird when that feature is already built into the react syntax.

client/src/components/Modal/tests/Modal-test.js
maxime-rainville: When I run `root.debug()`, it prints out exactly the HTML I would expect.

```html
    <body
      class="modal-open"
    >
      <div />
      <div
        style="position: relative; z-index: 1050; display: block;"
        tabindex="-1"
      >
        <div
          class=""
        >
          <div
            class="modal my-modal-dialog fade"
            role="dialog"
            style="display: block;"
            tabindex="-1"
          >
            <div
              class="modal-dialog my-modal modal-sm"
              role="document"
            >
              <div
                class="modal-content"
              >
                <div
                  class="modal-header"
                >
                  <h5
                    class="modal-title"
                  >
                    Hello World!
                  </h5>
                  <button
                    aria-label="Close"
                    class="close"
                    type="button"
                  >
                    <span
                      aria-hidden="true"
                    >
                      ×
                    </span>
                  </button>
                </div>
                My Content
              </div>
            </div>
          </div>
          <div
            class="modal-backdrop fade"
          />
        </div>
      </div>
    </body>
```

I'm really not sure why `root.container.querySelector()` doesn't seem to work for this test. I'm really not sure what I'm doing different on this test compare to all the other ones.

maxime-rainville: Your comment about testing negative was prescient. The test was passing whether `isOpen` was false or true.

I've update the test to use screen instead and it's failing when isOpen is true has you would expect.
```js
const modalBody = screen.queryByText('My Content');
expect(modalBody).toBeNull();
```

client/src/components/Modal/tests/ModalHeader-test.js
maxime-rainville: Done.

client/src/components/Modal/tests/Modal-test.js
```diff
@@ -0,0 +1,90 @@
...
+
+  const modalBody = root.getByText('My Content');
+  expect(modalBody).toBeTruthy();
+
+  const closeButton = root.getByLabelText('Close');
+  expect(closeButton).toBeTruthy();
+});
+
+test('Closing the Modal', () => {
+  const props = makeProps();
```
emteknetnz:
```suggestion
  const props = makeProps({
    onClosed = jest.fn()
  });
```

client/src/components/Modal/tests/ModalHeader-test.js
```diff
@@ -0,0 +1,33 @@
...
+  props = {
+    onClosed: jest.fn(),
+    title: 'Hello World!',
+    showCloseButton: true,
+  };
+});
```
emteknetnz:
```suggestion
function makeProps(obj = {}) {
  return {
    onClosed: () => {},
    title: 'Hello World!',
    showCloseButton: true,
    ...obj
  };
});
```

client/src/components/Modal/tests/ModalHeader-test.js
```diff
@@ -0,0 +1,33 @@
...
+
+test('Modal renders', () => {
+  const root = render(<ModalHeader {...props} />);
```
emteknetnz:
```suggestion
  const root = render(<ModalHeader {...makeProps()} />);
```

client/src/components/Modal/tests/ModalHeader-test.js
```diff
@@ -0,0 +1,33 @@
...
+
+  const header = root.getByRole('heading');
+  expect(header.textContent).toContain('Hello World!');
+
+  const closeButton = root.getByLabelText('Close');
+  expect(closeButton).toBeTruthy();
+});
+
+test('Closing the Modal', () => {
+  const root = render(<ModalHeader {...props} />);
```
emteknetnz:
```suggestion
  const props = makeProps({
    onClosed: jest.fn()
  });
  const root = render(<ModalHeader {...props} />);
```

client/src/components/Modal/tests/Modal-test.js
```diff
@@ -0,0 +1,90 @@
...
+
+const makeProps = () => ({
+  isOpen: true,
+  className: 'my-modal',
+  modalClassName: 'my-modal-dialog',
+  size: 'sm',
+  onClosed: jest.fn(),
+  title: 'Hello World!',
+  showCloseButton: true,
+});
```
emteknetnz:
```suggestion
function makeProps(obj = {}) {
  return {
    isOpen: true,
    className: 'my-modal',
    modalClassName: 'my-modal-dialog',
    size: 'sm',
    onClosed: () => {},
    title: 'Hello World!',
    showCloseButton: true,
    ...obj
  };
});
```

I'm not completely sure if its required to prevent bleed over to other tests, though should define each jest.fn() in the test() that is asserting that it was called

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -1,104 +1,71 @@
...
+  ModalHeaderComponent,
+  onClosed,
+  showCloseButton,
+  size,
+  title,
+}) => {
+  /** @var {string} response Message we got back from posting the form. */
+  const [response, setResponse] = useState(null);
+  /** @var {boolean} response Whether the response was an error or not. */
+  const [error, setError] = useState(null);
```
emteknetnz, line 43:
```suggestion
  const [response, setResponse] = useState(null);
  const [error, setError] = useState(null);
```
We're not use `@var` elsewhere in the codebase

client/src/components/FormBuilderModal/FormBuilderModal.js
```diff
@@ -108,107 +75,81 @@ class FormBuilderModal extends Component {
...
+    className,
+    isOpen,
+    modalClassName,
+    ModalComponent,
+    ModalHeaderComponent,
+    onClosed: handleHide,
+    showCloseButton,
+    size,
+    title,
+
```
emteknetnz:
```suggestion
```

emteknetnz (changes requested):
Do you have a link to a CI run on silverstripe/installer that includes this PR? - I sent you script for how to set this up earlier

maxime-rainville (commented):
Sorry ... I just realised I forgot to post my reply to all your first round of feedback :facepalm:

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: It's a standard JSDoc feature which we are using in plenty of other context. 

It's useful for type hitting and most IDEs will pick it up.

client/src/components/Modal/tests/Modal-test.js
maxime-rainville: There's three ways you can avoid bleeding test when using mock functions:
- call `jest.fn()` for each test so you get a different mock
- Call `mockClear()` on the method that got returned from `jest.fn()`
- `jest.clearAllMocks()`

Normally, you would put the `mockClear` or `clearAllMocks` calls in a `beforeEach` or `afterEach` call.

client/src/components/Modal/tests/Modal-test.js
maxime-rainville: For the record, that's one of the reason why I prefer the approach I was originally using. You occasionally need to reference the props in the assertion. e.g. If I'm asserting that that "Hello word" will appear in my component, it's easier and clearer to write that assertion around the prop variable I'm passing.

```js
expect(root.getByRole('heading').textContent).toContain(props.title);
```

This makes it clear that what I'm asserting is that my title prop will be displayed in a header ... rather than any random arbitrary test value value I might have chosen for my test.

client/src/components/FormBuilderModal/FormBuilderModal.js
maxime-rainville: Done

client/src/components/Modal/tests/Modal-test.js
maxime-rainville: I really don't like this "passing test specific props" to a `makeProps` method when React is already set up to allow you to override props. So I update the test to instantiate that mock method locally instead and pass it directly to the component.

client/src/components/Modal/tests/ModalHeader-test.js
maxime-rainville: Done

maxime-rainville (commented):
I've addressed @emteknetnz feedback ... for real this time!

emteknetnz (changes requested):
We should get rid of ModalHeader.js and just put everything it has into Modal.js.  I feels a bit over abstracted wrapping the ReactStrap ModalHeader with our own component.

Things like `...ModalHeader.propTypes` in Modal.propTypes and passing through the `ModalHeaderComponent` just seem like there's adding complexity/confusion for no benefit

maxime-rainville:
> We should get rid of ModalHeader.js and just put everything it has into Modal.js. I feels a bit over abstracted wrapping the ReactStrap ModalHeader with our own component.

The original implementation had them split into two different methods and was functionnaly treating them as two distinct sub-components. That's why ended up splitting into two different components.

I do agree that converting them to functional component and splitting them out of `FormModalBulider` has simplified the code considerably to the point where treating them as two distinct things increase complexity rather than reduce it. I'll go ahead and do it now.

As a side note, I think these kind of higher level architectural concerns (how the components are split) are easier to handle at the start of the peer review rather than at the end. Although, I appreciate that these thoughts don't always occure when they are most convenient.

client/src/components/Modal/tests/Modal-test.js
```diff
@@ -0,0 +1,106 @@
...
+  expect(header.textContent).toContain('Hello World!');
+
+  const modalBody = root.getByText('My Content');
+  expect(modalBody).toBeTruthy();
+
+  const closeButton = root.getByLabelText('Close');
+  expect(closeButton).toBeTruthy();
+});
+
+test('Modal with HTML title', () => {
```
maxime-rainville, line 54: This test was added to cover the scenario where an object is passed to the title.

maxime-rainville (commented):
I folded the ModalHeader back into Modal. Functionality should still be the same.

The test for ModalHeader were all being duplicated by the Modal test, so I didn't see anything that needed to be moved around.

maxime-rainville:
I've re-triggered the build here: https://github.com/creative-commoners/silverstripe-installer/pull/2/checks

Once that's green, I'll reassigned the card to Steve.

emteknetnz (approved):
Tested locally, works good

PR #1635, opened 2023-12-14
MNT Add ID attribute to input

emteknetnz:

This is needed to support the behat test change made in https://github.com/silverstripe/silverstripe-siteconfig/pull/149


GuySartorelli (approved):
Makes the behat test pass locally

PR #1636, opened 2023-12-15
FIX Set z-index of list

emteknetnz:


GuySartorelli (approved):
For future reference:
This is fixing a bug where the dropdown isn't in front of the WYSIWYG - see https://github.com/silverstripe/silverstripe-admin/issues/1618#issuecomment-1856602291 for more context.

@emteknetnz Please remember to include appropriate details in the PR description so in the future it's easy to see why specific changes like this are being made.

Issue #1639, opened 2023-12-21
Toast and escape characters in utf8


lekoala:
### Module version(s) affected

2.1.5

### Description

This is related to https://github.com/silverstripe/silverstripe-framework/issues/11089

Currently, toast messages are escaped using Convert::raw2xml. This means that characters like ' are escaped to their utf8 equivalent (eg: X-Status: Saved Viewing "test event&#039;s ...)
But the js itself doesn't do anything to display this properly

See here the status message handler

https://github.com/silverstripe/silverstripe-admin/blob/ef1385a7fe015890e56bf0fc652dca01678efb71/client/src/legacy/LeftAndMain.js#L1511-L1514

On top of that, the method used to escape the characters (that should already be safe, but still) is not optimal (see https://stackoverflow.com/questions/1147359/how-to-decode-html-entities-using-jquery/) since adding this in a div could still execute malicious code. Ideally, that should be a standard way to escape strings across the framework without having custom bits of code like this.

### How to reproduce

- Go to model admin to a record with a ' in the name
- Click save
- Discover the utf8 encoded ' that is not displayed properly in the toast message

<img width="509" alt="image" src="https://github.com/silverstripe/silverstripe-admin/assets/250762/8109cb41-5cbe-4ea4-ac8d-8aba0f9c7f63">

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-framework/pull/11105
- https://github.com/silverstripe/silverstripe-admin/pull/1640
- https://github.com/silverstripe/silverstripe-versioned/pull/438
- https://github.com/silverstripe/silverstripe-versioned/pull/440

lekoala:
WIP

Using this works for the single quote

```js
function decodeEntities(encodedString) {
  var textArea = document.createElement('textarea');
  textArea.innerHTML = encodedString;
  return textArea.value;
}
```

but doesn't help with ö so there is more to it that it seems

lekoala:
Two pr to fix this

https://github.com/silverstripe/silverstripe-framework/pull/11105
https://github.com/silverstripe/silverstripe-admin/pull/1640

GuySartorelli:
This is mostly fixed now - but `&` is still being escaped to `&amp;`

[...]

PR #1640, opened 2023-12-21
decode entities in toast message

lekoala:
## Description

Decodes toast message entities, which are encoded before they're sent from the server. See also https://github.com/silverstripe/silverstripe-framework/pull/11105

Side bonus : this is much safer than using jquery

## Manual testing steps

Go to model admin to a record with a ' in the name or any utf8 character like ö or an emoji
Click save
Discover the utf8 encoded ' that is not displayed properly in the toast message

## Issues

- https://github.com/silverstripe/silverstripe-admin/issues/1639

## Pull request checklist

- [ ] CI is green


GuySartorelli:
Hiya, First off, thank you for this contribution.

We added a new pull request template to this repository yesterday - but it looks like your PR description isn't using it. Did the template show up when you created this pull request?

lekoala:
hey @GuySartorelli it did show but i have to admit i didn't read it :-/ it seems to me that since everything was already pretty much explained in the linked issue that felt like doing the work twice and i was already quite busy getting the two pr ready
it took me already longer than expected to pin down what was exactly happening since it's a combined issue between the data not being url encoded in X-Status and how the content is currently escaped :-)

GuySartorelli:
I've added the template back in - please add any manual testing steps, or if there aren't any that are specific to this PR, please point back at the reproduction steps in the issue.

Please also tick all of the boxes that you have done, and do whatever is still not done based on the checklist.

lekoala:
Hi @GuySartorelli. While i understand the need to raise the quality level of the PR to avoid burdening the team, sometimes it feels like a real pain to get across some changes. Ultimately, it's up to the team to determine how/where things go (eg: NOT using the current method of escaping text that is not XSS safe), or to refactor the utility to do so somewhere that makes sense so that i can be used across the whole project.
Basically, it feels like i've done my part of the job :-) i had an issue, and i fixed it and explained how i did it
the steps are described in the issue already, and it gives the wider context
i'll add them here also since i was there anyway

GuySartorelli:
The idea here is that we've added a new template to make things faster - by following the template, there's a higher degree of certainty that all of the small things have been done, and that there's enough context for someone (me in this case) to come along and review the work.

This is a new process we're trying, and I'd really appreciate your help in testing the process by following it. Frankly I'm not going to spend my time reviewing this or the related PR this side of christmas if the process isn't being followed, because if the process isn't followed it increases the amount of effort that goes into reviewing the PRs.

lekoala:
:-) yup i understand that fully

but it's pretty much the same thing for me, it's not really easy to explain to a customer that it took half a day to solve some badly escaped characters due to the open source nature of the project

=> the main issue remaining are:
-> i didn't run the build process because i didn't check yet how to do that (but should be pretty easy for anyone from the team if they want to give it a run)
-> check why the text was escaped using jquery when it's not safe
-> probably refactor the escape function into some util function that can be used elsewhere

client/src/legacy/LeftAndMain.js
```diff
@@ -1508,7 +1508,13 @@ $.entwine('ss', function($) {
   });
 });
 
+// TODO: move this to an external util
```
ChloeSartorelli: Please don't include `todo` comments - please either move it to an external util now, or just remove the comment - frankly if it's not done now it won't likely ever be done.

client/src/legacy/LeftAndMain.js
```diff
@@ -1508,7 +1508,13 @@ $.entwine('ss', function($) {
...
+function decodeEntities(encodedString) {
```
ChloeSartorelli: There's an existing `decodeURIComponent` function referenced elsewhere in this file. If they do the same thing, please remove this new function and just use the existing one instead.

client/src/legacy/LeftAndMain.js
```diff
@@ -1508,7 +1508,13 @@ $.entwine('ss', function($) {
...
+    var textArea = document.createElement('textarea');
+    textArea.innerHTML = encodedString;
+    return textArea.value;
+}
+
 var statusMessage = function(text, type) {
-  text = jQuery('<div/>').text(text).html(); // Escape HTML entities in text
-  jQuery.noticeAdd({text: text, type: type, stayTime: 5000, inEffect: {left: '0', opacity: 'show'}});
+  jQuery.noticeAdd({text: decodeEntities(text), type: type, stayTime: 5000, inEffect: {left: '0', opacity: 'show'}});
```
ChloeSartorelli: On line 191 we have a message that is being passed into a `decodeURIComponent` function - it seems likely there's a risk of double-decoding there, if that's even a thing. Should we remove the call to `decodeURIComponent` on that line?
```js
statusMessage(decodeURIComponent(msg), msgType);
```

ChloeSartorelli: The same thing is happening in https://github.com/silverstripe/silverstripe-admin/blob/4a97f0bd0f5d297684b81311353ba33573ac360c/client/src/legacy/LeftAndMain.BatchActions.js#L303

GuySartorelli (changes requested):
A few things here I'm unsure about.

Please also update the commit message to include `FIX` at the start (see [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages))

client/src/legacy/LeftAndMain.js
lekoala: as i said, i think it would be interesting to have one way to escape strings across the framework. this might be difficult since everything is split across modules.

or i just remove the todo comment and we proceed as is, that works for me too :)

client/src/legacy/LeftAndMain.js
lekoala: they don't do the same thing
- decodeURIComponent decodes what has been encoded with rawurlencode in the php side of things
- my decodeEntities function does two things : when adding the value to the textarea element, it will automatically decode any remaining utf8 things (like the single quote, emojis) AND sanitize against any html script (<script>...). Maybe I should name it better like decodeAndSanitize or something

client/src/legacy/LeftAndMain.js
lekoala: as far as i'm aware, there is only a risk of double encoding, not double decoding but i may be wrong here
at any rate, "it seems to work" on my end with my changes. i know that's not a very satisfactory answer :-)

client/src/legacy/LeftAndMain.js
ChloeSartorelli: Either of those is fine with me - but whichever way we go, it has to be done up front. If you're willing to put in the time to implement a way to escape strings across the board, great! If not (and I wouldn't blame you), please just remove the comment. We don't track todo comments, we track issues :p

client/src/legacy/LeftAndMain.js
ChloeSartorelli: Ahhh okay. Yeah please rename it and maybe add a comment as well explaining that it's both decoding utf8 entities and santizing against `<script>`.

client/src/legacy/LeftAndMain.js
ChloeSartorelli: I guess the thing I'm worried about (and it is _an extreme_ edge case so maybe we can ignore it? I'd want feedback from others in @silverstripe/core-team though) is if someone intentionally wants the encoded form.

e.g. a page called "`Are &quot; and &#39; the same thing?`"

lekoala: i'm not sure either, i just did what needed in order to fix all my use cases :)

lekoala:
hi @GuySartorelli 

so i have made a small update

- renamed the function to be more accurate about what it does (the main goal is to escape html, the added bonus is that it helps with things like single quotes)
- i don't see any risk of "double decoding", the decodeURIComponent is basically decoding what does "rawurlencode" on the php side (hence, the other PR)
- this will be safer than using jQuery for escaping html

client/src/legacy/LeftAndMain.js
ChloeSartorelli: Can you please make this change?
Edit: Ahh, I see you've changed it from `decode` to `escape` - but that to me still doesn't quite explain what it does. Can you please add a comment as per the above?

client/src/legacy/LeftAndMain.js
ChloeSartorelli: I've just tried creating a page called `Are &quot; and ' and &#39; the same thing?` both with this (and the framework) PR, and without them.
In both cases, the toast displays like this:
Screenshot from 2024-01-19 12-08-34

So this PR is still not handling some characters (`&` specifically in this case) correctly.
Can you please look into that?

GuySartorelli:
It's worth notice that I _cannot_ reproduce the original issue with `'` or `ö` being rendered incorrectly in toast messages (at least for saving a page, maybe other toasts are different?) with framework `5.1.x-dev` and admin `2.1.x-dev` - I've tried in both firefox and chromium.

I do see a problem with `&` being incorrectly escaped to `&amp;` though, so there is definitely a problem.

lekoala:
This is what i got on the latest version

image

windows/apache/chrome latest version

GuySartorelli:
Ahh, I see! By scrutinising the screenshot I realised that's a `DataObject` in some gridfield, probably in a modeladmin. I had been testing pages in the sitetree which apparently handle this differently. I can confirm that using a modeladmin, I see the same results that you do.

I think any fix for this will need to unify those two approaches - i.e. whatever code sitetree is used for decoding/sanitising/etc toast messages should be used my gridfields, and vice-versa.

lekoala:
@GuySartorelli so i finally got some time to get to the bottom of this

Here is the situation

Indeed, in the cms it's working fine. This is the current code:

https://github.com/silverstripe/silverstripe-cms/blob/cb46fd8aac90b6a1f08542b92cefac841f64a469/code/Controllers/CMSMain.php#L1779-L1795

It works, because the title is not escaped and rawurlencoded

We could do the same thing and I'll be happy to update my PR https://github.com/silverstripe/silverstripe-framework/pull/11105 accordingly and drop this one (ie: remove the escaping, and add the rawurlencode)

It does raise the question of :
- why was it deemed necessary to escape the title in the gridfield (and not in the cms) ?
- why use an unsafe escape method using JQuery().text
This could be part of another PR, while we do the quick fix (align to CMSMain behaviour) so that at least international users (and smiley lovers) can get proper toasts in the meantime

GuySartorelli:
> why was it deemed necessary to escape the title in the gridfield (and not in the cms) ?

Not sure - there's nothing in the PR that shows why it would be necessary. I suspect it was just a copy-paste from somewhere doing something different-but-similar.

> why use an unsafe escape method using JQuery().text
This could be part of another PR, while we do the quick fix (align to CMSMain behaviour) so that at least international users (and smiley lovers) can get proper toasts in the meantime

I strongly suspect it just wasn't known to be unsafe - I wouldn't have known if you hadn't mentioned it.

GuySartorelli:
I've merged the framework PR which solves the majority of the toast problems.

I'd support swapping out `JQuery().text` - whether you want to rework this PR to include that, do it separately, or just open an issue for someone else to look into, either way is fine.

For this PR, if you're up for it, could you please tackle the problem where `&` is incorrectly decoded into `&amp;`?

lekoala:
I've made a small update to use innerText, it all seem to work nicely with the latest version of the framework and the updated decode method

Even crazy input like 

image

client/src/legacy/LeftAndMain.js
```diff
@@ -1508,13 +1508,13 @@ $.entwine('ss', function($) {
   });
 });
 
-function escapeHTMLEntities(encodedString) {
-    const textArea = document.createElement('textarea');
-    textArea.innerHTML = encodedString;
-    return textArea.value;
+function decodeEntities(str) {
+    const ta = document.createElement('textarea');
+    ta.innerHTML = str;
```
ChloeSartorelli: Please don't use short variable names like this - it might be obvious to you what `ta` means, but it's even _more_ obvious if it's just clearly named `textArea` (or `element` or some other meaningful name).

GuySartorelli (changes requested):
Works great! Just one small change request.

I tested out the title `About Us &'😅<script></script>` in the pages admin and in a model admin.

Works with every scenario in the pages admin.
Works with saving in the model admin, but other actions (publishing, unpublishing, archiving) transform the emoji:
`Published Company "About Us &'ð<script></script>"`

lekoala:
@GuySartorelli yes, it seems indeed the Versioned gridfield has the same issue
PR https://github.com/silverstripe/silverstripe-versioned/pull/438

And var name updated.

GuySartorelli:
Rebased and built JS. Once CI is green I'll merge.
Thank you so much for working on this!

GuySartorelli (approved):
Works great locally. Thanks again!

PR #1645, opened 2024-01-08
DEP Update JS dependencies

GuySartorelli:
## Description

Update JS dependencies and rebuild tinymce skin (since the tinymce dep was updated).
Supercedes https://github.com/silverstripe/silverstripe-admin/pull/1644 which was failing CI

## Issues

- https://github.com/silverstripe/.github/issues/178

## Pull request checklist

- [x] tests aren't necessary for this change
- [x] No docs need to be updated
- [x] CI is green


GuySartorelli:
Jest tests fail when I try to build so need to look into that a bit - that's probably what was causing the CI to fail in the automated PR.

GuySartorelli:
Seems to be caused by https://github.com/JedWatson/react-select/releases/tag/react-select%405.8.0 - but there's more to it than just that. It seems like I should be able to just swap `aria-context` for `aria-focused` in the test.... except there seems to never be any text inside the `aria-focused` element. This seems to be the case when running it in the browser as well, so it's not just a jest wonkiness issue.

GuySartorelli:
I've opted to remove the tests that were testing the focus message - that's not functionality we control, it's part of the react-select library, so it doesn't make sense for us to be testing that anyway.

emteknetnz (approved):

PR #1652, opened 2024-01-18
ENH Remove background form description and unclickable fields

sabina-talipova:
## Description
- All styles that were applied to readonly fields also were applied to description string. Exclude `.form__field-description` from list to exclude styles.
- Readonly input fields is not clickable .

Before:
Screenshot 2024-01-18 at 12 09 46 PM

After:
Screenshot 2024-01-19 at 11 25 26 AM


client/src/components/Form/Form.scss
```diff
@@ -147,7 +147,8 @@
 .form__field-holder p.readonly,
 .form__field-holder input.readonly,
 .form__field-holder span.readonly,
-.readonly .form__field-holder > div {
+.readonly .form__field-holder > div:not(.form__field-description) {
+  pointer-events: none;
```
emteknetnz:
```suggestion
```
I'm don't think we should be setting this to none in the context of this PR, because I really don't know what all the possible scenarios are, and I don't even know if it's desirable. I did a quick search for this and a couple of comments [here](https://ux.stackexchange.com/questions/77163/should-disabled-elements-ever-have-pointer-events) were negative about setting this to `none`

Somewhat related I've raised a [new card](https://github.com/silverstripe/silverstripe-linkfield/issues/181) to handle `->setDisabled(true);` and when a TextField is disabled it has no mouse-pointer, though when it's readonly it does have a mouse pointer.

If we're going to remove the mouse-pointer from all readonly form fields in the CMS then it should really be done in a separate issue, not one that's specifically for linkfield

emteknetnz (changes requested):
Should change the PR title to "ENH Remove background from disabled form field descriptions"

client/src/components/Form/Form.scss
sabina-talipova: Updated

client/src/components/Form/Form.scss
```diff
@@ -161,6 +161,10 @@
     border-color: $input-focus-border-color;
     outline: none;
   }
+
+  &:hover {
+    cursor: not-allowed;
```
emteknetnz: This shouldn't be the cursor state.  It implies they're doing something wrong by mousing over it, though they may just want to select the text so they can copy-paste it which is perfectly fine.

emteknetnz (changes requested):

client/src/components/Form/Form.scss
sabina-talipova: Updated

emteknetnz (approved):

emteknetnz (changes requested):
Needs a rebase + `yarn build` to resolve merge conflicts

emteknetnz (approved):

PR #1653, opened 2024-01-19
FIX Add tinymce link menuitems per editor instance

GuySartorelli:
## Description

Fixes an issue where multiple tinymce instances with different configs don't correctly register the link menus.

There is (or will be) associated PRs for asset-admin and cms as well, to update their tinymce link plugins.

Note that this change *is* backwards compatible with older versions of cms or asset-admin, but the PRs for those modules are *not* backwards compatible with older versions of this module.

## Manual testing steps

See issue for reproduction steps

## Issues

- https://github.com/silverstripe/silverstripe-elemental/issues/1120

## Pull request checklist

- [ ] CI is green


client/src/legacy/TinyMCE_sslink-email.js
```diff
@@ -9,19 +9,20 @@ import { loadComponent } from 'lib/Injector';
...
+    // Add "Link to email address" to link menu for this editor
+    TinyMCEActionRegistrar.addAction(
+      'sslink',
+      {
+        text: i18n._t('Admin.LINKLABEL_EMAIL', 'Link to email address'),
+        onclick: (editorInst) => editorInst.execCommand(commandName),
+        priority: 51,
+      },
+      editor.settings.editorIdentifier,
+    ).addCommandWithUrlTest(commandName, /^mailto:/);
```
ChloeSartorelli, line 23: Moving this into here means we have access to the actual editor instance, so we can find its unique editor identifier.
Note the change from `editorIdentifier` to `editor.settings.editorIdentifier`.

If multiple editors use the same config (and therefore have the same identifier), the action is not duplicated - there's deduping logic in `TinyMCEActionRegistrar` to handle that.

client/src/legacy/TinyMCE_sslink.js
```diff
@@ -19,25 +19,6 @@ const plugin = {
...
-      title: titleWithShortcut,
-      type: 'menubutton',
-      menu: actions,
-    });
-
-    editor.addMenuItem('sslink', {
-      icon: 'link',
-      text: title,
-      menu: actions,
-    });
```
ChloeSartorelli, line 40: The additional link type plugins' `init` function hasn't been called yet, so we have to defer this.

emteknetnz (approved):
Tested locally using replication steps on issue and reconfirmed on vanilla install that replication steps are valid. PR works correctly.

Issue #1656, opened 2024-01-23
Broken builds


GuySartorelli:
https://github.com/silverstripe/silverstripe-admin/actions/runs/7618883901/job/20750916959

BulkLoader is being initialised with the tab name instead of the class name when a custom tab name is provided.

https://github.com/silverstripe/silverstripe-admin/actions/runs/7618820966/job/20753517379

There's ambiguity resulting in the wrong item being deleted

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1657
- https://github.com/silverstripe/silverstripe-admin/pull/1659

GuySartorelli:
PRs merged

PR #1659, opened 2024-01-23
MNT Fix ambiguous behat test

GuySartorelli:
Fixes https://github.com/silverstripe/silverstripe-admin/actions/runs/7618820966/job/20753517379

The failure was caused by an ambiguity between "Employee A" in the name column and "Employee A" in the colleagues column - which meant we were clicking on "Employee A" in the colleagues column of "Employee C" which meant we were then editing (and ultimately deleting/archiving) "Employee C" (which is not who we though we were getting rid of).

Resolved by moving "Employee A" so it doesn't have any colleagues.

gridfield-toast-messages feature_57

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1656


sabina-talipova (approved):
LGTM

PR #1660, opened 2024-01-24
FIX HTMLEditorField::setRows with Elemental

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-elemental/issues/1132


client/src/legacy/HtmlEditorField.js
```diff
@@ -83,6 +83,7 @@ ss.editorWrappers.tinyMCE = (function() {
     getConfig: function() {
       var selector = "#" + editorID,
         config = jQuery(selector).data('config'),
+        height = config.row_height ? config.row_height : '400px',
```
ChloeSartorelli: Where is that `400px` value coming from? Seems fairly arbitrary? I assume this is replacing what the deleted line below used to do?

GuySartorelli (changes requested):

client/src/legacy/HtmlEditorField.js
sabina-talipova: `400px` is the default value that we calculate on the server side if the setRows method was not applied. Yes, you're right, I'll change the logic here.

sabina-talipova: Updated

client/src/legacy/HtmlEditorField.js
```diff
@@ -83,6 +86,7 @@ ss.editorWrappers.tinyMCE = (function() {
     getConfig: function() {
       var selector = "#" + editorID,
         config = jQuery(selector).data('config'),
+        height = config.row_height ? config.row_height : 'undefined',
```
sabina-talipova, line 89: Set to `undefined` if `row_height` is not set, `null` or `0`.

client/src/legacy/HtmlEditorField.js
```diff
@@ -66,6 +66,9 @@ ss.editorWrappers.tinyMCE = (function() {
     },
 
     setHeight: function (event, height) {
+      if (typeof height === 'undefined') {
+        return;
```
sabina-talipova, line 70: If `height` is `undefined` we should apply default TinyMCE configuration.

client/src/legacy/HtmlEditorField.js
ChloeSartorelli, line 89:
```suggestion
        height = config.row_height ? config.row_height : undefined,
```
The `undefined` type must not have quotes, or else it'll be `typeof string`

GuySartorelli (changes requested):

client/src/legacy/HtmlEditorField.js
sabina-talipova: Updated

GuySartorelli (approved):
LGTM, works as expected.

PR #1663, opened 2024-01-25
FIX Type error

sabina-talipova:
## Description
The undefined type must not have quotes, or else it'll be typeof string.


GuySartorelli (approved):
Merge on green

Issue #1668, opened 2024-01-31
ModelAdmin defined the custom url segment approach broken


pack254:
### Module version(s) affected

1.13.19

### Description

 I have found an issue in `silverstripe/admin` version `1.13.19` when I did the upgrade security patches. this version it breaks the ModelAdmin which is defined using the custom url segment approach e.g.
```
private static $managed_models = [
  'product-category' => [
    'dataClass' => Category::class,
    'title' => 'Product categories',
   ],
]
```
This patch 1.13.19 appears to break potentially caused from line 662 [ModelAdmin.php](https://github.com/silverstripe/silverstripe-admin/compare/1.13.18...1.13.19#diff-d6ec8df426c5d673f7390f98f9bcf37dee0828848957d1903fd10e1b9e99d96dL662) is changed from `modelClass` to `modelTab`.
In a nuthshell, any module that uses this approach will break i.e. [SiteBanner module](https://github.com/silverstripe/silverstripe-sitebanner/blob/master/src/Models/Admin.php#L13)

### How to reproduce

This can be reproduced by using the `silverStripe/admin` version `1.13.19` and `nzta/silverstripe-sitebanner` version `3.0.1` after running `composer install && composer dev-build flush=all` logged in to CMS and then go to Site Banner tab at the left panel, so you will see this error.

![Screen Shot 2024-01-31 at 2 02 12 PM](https://github.com/silverstripe/silverstripe-admin/assets/17997208/e2e97021-7552-4fad-b744-abab76c920bb)

### Possible Solution

revert code back to `$importer = new $importerClass($this->modelClass);`

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1670

emteknetnz:
@pack254 linked PR has been merged, will be auto-tagged shortly

Note I left a comment on https://github.com/silverstripe/silverstripe-admin/pull/1670#pullrequestreview-1857624066 that the site banner possibly isn't working correctly, so you'll want to validate this on your local

Fixing that would be a totally separate issue and it's not a module that the open-source team look after

emteknetnz:
[...]

PR #1657, opened 2024-01-23
FIX Use valid class for model importers

GuySartorelli:
## Description

BulkLoader is being initialised with the tab name instead of the class name when a custom tab name is provided.
This is related to the wider issue https://github.com/silverstripe/silverstripe-admin/issues/1364

## Issues

- https://github.com/silverstripe/silverstripe-admin/issues/1656

## Pull request checklist

- [x] The target branch is correct
    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
- [x] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [x] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] This change is covered with tests (or tests aren't necessary for this change)
- [ ] CI is green

PR #1670, opened 2024-02-01
FIX Use valid class for model importers

GuySartorelli:
Basically the same fix as https://github.com/silverstripe/silverstripe-admin/pull/1657, which I should have realised at the time would affect the 1.x line as well.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1668


emteknetnz (approved):
Tested locally, fixes the bug in the description

Note that that the site banner module didn't appear to actually work correctly e.g. I couldn't save a start date, the banner did not actually show, though presumably that's a completely different issue

Issue #1672, opened 2024-02-03
Empty import modal - Users/Groups import


HeyImPhil:
### Module version(s) affected

2.1.14

### Description

When opening the import users or groups from the security admin shows an empty modal.

![image](https://github.com/silverstripe/silverstripe-admin/assets/4695076/23e18f1d-7de8-404a-8e48-be4c1eb4ac9e)

### How to reproduce

1. Fresh install using silverstripe/installer
2. Go to the security page in the admin
3. Click Import CSV from the Users or Groups tab
4. An empty modal will appear with only the Title showing for each type i.e "Import users" or "Import groups"

### Possible Solution

The ModelAdmin ImportForm method checks $importers[$this->modelTab] but the getModelImporters only returns array of import model classes so the isset check is incorrect later.

` if (!$importers || !isset($importers[$this->modelTab])) {`
could be changed to this
` if (!$importers || !isset($importers[$this->modelClass])) {`

Also the next line would need to change from
`$importer = $importers[$this->modelTab];`
to
`$importer = $importers[$this->modelClass];`

If this looks to be the right path I am happy to make a PR but don't fully understand all of the use cases that I need to account for here.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1676

michalkleiner:
Hi @HeyImPhil. I believe this issue has been fixed in [2.1.15](https://github.com/silverstripe/silverstripe-admin/releases/tag/2.1.15), please give it a go. If that's the case, feel free to close the issue.

HeyImPhil:
Hiya @michalkleiner, still same behavior after upgrading to 2.1.15 on my fresh silverstripe/installer site

![image](https://github.com/silverstripe/silverstripe-admin/assets/4695076/f5658966-cf50-4752-93a7-93a6abfb1941)

michalkleiner:
And you ran `?flush` in the admin area after the upgrade?

GuySartorelli:
I can reproduce this locally. Probably the same thing happens for CMS 4 modeladmins that are set up in a similar way.
[...]

PR #1676, opened 2024-02-08
FIX Set the CSV importer based on the tab, not the model

GuySartorelli:
## Description

Confirmed the bug affects CMS 4 as well, so fixing it there.

## Manual testing steps

Copy the `SecurityAdmin` from CMS 5 and plonk it into CMS 4, then try to import a member. If you add it as a separate class you'll want to set a different `url_segment` and `menu_title` so you know what one you're looking at. Note that there's syntax in there that needs to run on PHP 8.1 - but that doesn't affect the actual code fix which as per CI is compatible with 7.4

## Issues

- https://github.com/silverstripe/silverstripe-admin/issues/1672

## Pull request checklist

- [x] CI is green

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -621,6 +621,7 @@ abstract class ModelAdmin extends LeftAndMain
 
         $importers = [];
         foreach ($importerClasses as $modelClass => $importerClass) {
+            $tab = $modelClass;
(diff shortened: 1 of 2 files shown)


emteknetnz (approved):
Tested locally, works correctly

PR #1683, opened 2024-02-21
FIX Remove current active tab from session storage after use

satrun77:
## Description
The PR fixes the code that attempts to remember the tab state in the page edit screen. Currently, the tab state is stored in browser session storage but never removed, so the next time a CMS author edits the same page, it loads another tab instead of the main tab.

The fix is to remove the tab state after navigating away from the page. Note that the state is not removed if the URL from the address bar is changed manually. 

https://github.com/silverstripe/silverstripe-admin/assets/166450/44141843-cc67-498d-b254-e982f9cf5a18

## Manual testing steps

- Go to a CMS page with tabs.
- Access a tab different from Main.
- Save the page.
- You stay on the same tab.
- Navigate to a different page in the CMS.
- Come back to your original page.

## Issues

- #1684

## Pull request checklist

- [x] CI is green

## Parent issue
- https://github.com/silverstripe/silverstripe-cms/issues/2924


client/src/legacy/LeftAndMain.js
```diff
@@ -52,11 +52,20 @@ window.ss.debounce = function (func, wait, immediate) {
  * The URL to use for saving and loading tab state
  */
 window.ss.tabStateUrl = function() {
-  return window.location.href
+  return window.ss.formatTabStateUrl(window.location.href);
+};
+/**
```
michalkleiner, line 57:
```suggestion

/**
```

michalkleiner:
The change looks ok to me, haven't tested it.

For the situation when the URL is changed manually, could we add a TTL to the saved value and if it's expired, reset the saved tab to the default? Could be quite short lived I'd say.

Alternatively, could it be safe to assume there should only be one tab stored for the current session, so any other URL can remove the saved state for other pages? Do we need to index the storage by the URL at all?

GuySartorelli:
In your issue you've indicated that this issue affects "basically every version ever made" - which includes CMS 4. Can you please retarget this PR to the `1.13` branch so we can fix it there at the same time?

You'll need to [reset your commits](https://gist.github.com/GuySartorelli/74627843aeb75e12b777b1f44cb22d2f) and rebuild the js after retargetting it.

satrun77:
> In your issue you've indicated that this issue affects "basically every version ever made" - which includes CMS 4. Can you please retarget this PR to the `1.13` branch so we can fix it there at the same time?
> 
> You'll need to [reset your commits](https://gist.github.com/GuySartorelli/74627843aeb75e12b777b1f44cb22d2f) and rebuild the js after retargetting it.

I copied Max's issue 🤣 but the code is a little different there

client/src/legacy/LeftAndMain.js
```diff
@@ -891,7 +906,7 @@ $.entwine('ss', function($) {
         // Allow opt-out via data element or entwine property.
         if($(el).data('ignoreTabState') || $(el).getIgnoreTabState()) return;
 
-        selectedTabs.push({id:id, selected:$(el).tabs('option', 'active')});
+        selectedTabs.push({id:id, selected:resetTab? 0 : $(el).tabs('option', 'active')});
```
```suggestion
        selectedTabs.push({id:id, selected:resetTab ? 0 : $(el).tabs('option', 'active')});
```

client/src/legacy/LeftAndMain.js
```diff
@@ -878,11 +889,15 @@ $.entwine('ss', function($) {
...
+     * Parameters:
+     *  (String) url used for session storage key
+     *  (Boolean) resetTab true force selected tab to first, else current active
      */
-    saveTabState: function() {
+    saveTabState: function(url, resetTab) {
       if(typeof(window.sessionStorage)=="undefined" || window.sessionStorage === null) return;
 
-      var selectedTabs = [], url = window.ss.tabStateUrl();
+      var selectedTabs = [];
```
ChloeSartorelli, line 906:
```suggestion
      var selectedTabs = [];
      if (url === undefined) {
        const url = window.ss.tabStateUrl();
      }
```
Someone might be using this for their own purposes, so we should aim to retain backwards compatibility.

GuySartorelli (changes requested):

GuySartorelli:
> I copied Max's issue 🤣 but the code is a little different there

Nevertheless, if it is a bug in CMS 4 (which should be pretty quick to verify), we should be fixing it there.

GuySartorelli:
Can you please force push so CI reruns correctly? It's stuck thinking you're trying to push to `2.1` because you pushed changes before swapping the target branch 😅

GuySartorelli:
Looks good and says what it does on the tin. Have you put any thought into [Michal's comment](https://github.com/silverstripe/silverstripe-admin/pull/1683#issuecomment-1958564519) about ways we could have this work for accessing the page by URL?

satrun77:
> Looks good and says what it does on the tin. Have you put any thought into [Michal's comment](https://github.com/silverstripe/silverstripe-admin/pull/1683#issuecomment-1958564519) about ways we could have this work for accessing the page by URL?

Yes, It all depends on what the almighty James wants. if he wants me to fix that, I will work on it.

GuySartorelli:
Fair enough, please let us know whether you'll work on that or not.
If you don't, I'd still be inclined to merge this since it is a clear improvement on the previous behaviour. But I'll wait to know if you'll continue working on it or not.

satrun77:
> Fair enough, please let us know whether you'll work on that or not. If you don't, I'd still be inclined to merge this since it is a clear improvement on the previous behaviour. But I'll wait to know if you'll continue working on it or not.

He is happy with the current state of the change, this is all I can do now. Thanks

GuySartorelli (approved):
This is a clear improvement. Thanks!

PR #1685, opened 2024-02-22
FIX Enable bypassing entwine onsubmit handler

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/329


client/src/components/Toasts/Toast.js
```diff
@@ -47,11 +47,12 @@ const Toast = ({ type, text, onDismiss, dismissed, actions }) => {
...
 Toast.propTypes = {
   ...toastShape,
-  onDismiss: PropTypes.func.isRequired
+  onDismiss: PropTypes.func
 };
 
 Toast.defaultProps = {
-  actions: []
+  actions: [],
+  onDismiss: () => {},
```
emteknetnz, line 55: This is to fix an unrelated console warning after running `yarn dev`

"Warning: Failed prop type: The prop `toasts[0].onDismiss` is marked as required in `Toasts`, but its value is `undefined`."

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -260,6 +260,11 @@ $.entwine('ss', function($){
      * Suppress submission unless it is handled through ajaxSubmit().
      */
     onsubmit: function(e, button) {
+      // If the submit button a child of on an element with a .bypass-entwine class on it then
+      // skip handling. Usually the element in question will be used by react e.g. FormBuilder
+      if ($(e.target.activeElement).closest('.bypass-entwine').length) {
```
ChloeSartorelli:
```suggestion
      if ($(e.target.activeElement).closest('.bypass-entwine-submission').length) {
```
Make the css class more specific - even this may not be specific enough but at least it gives a hint about what is actually being bypassed.

ChloeSartorelli: Why are we using `closest`?
IIRC this will check all elements up and down the hierarchy in the DOM - feels a bit too broad a check?

GuySartorelli (changes requested):

client/src/legacy/LeftAndMain.EditForm.js
emteknetnz: Updated

closest() only goes up - `For each element in the set, get the first element that matches the selector by testing the element itself and traversing up through its ancestors in the DOM tree.`
https://api.jquery.com/closest/

GuySartorelli (approved):
Looks okay. Will merge when the elemental PR is ready.

maxime-rainville:
I've rebased the PR to solved the merge conflict and will merge on green.

Issue #1690, opened 2024-02-25
`SearchableDropdownField` doesn't work in form modal


GuySartorelli:
### Module version(s) affected

5.x-dev (unreleased feature for 5.2)

### Description

In an upload field, if you select "choose existing" and then click on a file, the "permissions" tab of the form on the right doesn't show the "viewer groups" and "editor groups" fields. These are `SearchableDropdownField` in 5.x-dev.

With `yarn dev` in silverstripe/admin, you get the following console output:

```text
Uncaught exception TypeError: schema is undefined
    onmatch SearchableDropdownFieldEntwine.js:16
```

When you insert the file, you get this console output:
```text
Uncaught exception TypeError: Root is undefined
    onunmatch SearchableDropdownFieldEntwine.js:27
```

### How to reproduce

1. In any upload field, click "choose existing"
1. Click on a file, but don't insert it yet
1. Look in the "permissions" tab in the form on the right.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PR
- https://github.com/silverstripe/silverstripe-admin/pull/1698
- https://github.com/silverstripe/silverstripe-admin/pull/1699

GuySartorelli:
PR merged.

As per https://github.com/silverstripe/silverstripe-admin/pull/1699#issuecomment-1979549915 I'll leave it up to you to decide whether to close this issue and raise a separate one about `TreeDropdownField` not displaying, or whether you want to work on that as part of the existing issue.

sabina-talipova:
New issue for TreeDropdownField was open. https://github.com/silverstripe/silverstripe-admin/issues/1701
Close this ticket.

PR #1698, opened 2024-03-04
FIX JS Console errors and warnings

sabina-talipova:
## Description
`SearchableDropdownField` was tested with LinkField, Elemental and simply on a Page. There are not any related errors or warnings in dev or prod.

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1690


client/src/legacy/SearchableDropdownField/SearchableDropdownFieldEntwine.js
```diff
@@ -13,18 +13,23 @@ window.jQuery.entwine('ss', ($) => {
...
       const context = cmsContent ? { context: cmsContent } : {};
       const schema = this.data('schema');
-      this.data('entwine-value', schema.value);
-      const Root = createRoot(this[0]);
-      const ReactField = loadComponent(schema.component, context);
-      this.setRoot(Root);
-      this.setComponent(ReactField);
-      this._super();
-      this.refresh();
+      if (schema) {
```
sabina-talipova, line 16: In first load Entwine `this.data('schema')` method returns null.

client/src/components/SearchableDropdownField/SearchableDropdownField.js
```diff
@@ -166,12 +166,12 @@ SearchableDropdownField.propTypes = {
   multi: PropTypes.bool.isRequired,
   name: PropTypes.string.isRequired,
   placeholder: PropTypes.string.isRequired,
-  onChange: PropTypes.func.isRequired,
+  onChange: PropTypes.func,
```
sabina-talipova, line 169: I don't think that this method is always required. [This recommendation](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/require-default-props.md) gives advice to use `defaultProps` instead of `isRequired` if we are not sure that this value will be provided.

client/src/components/SearchableDropdownField/SearchableDropdownField.js
```diff
@@ -166,12 +166,12 @@ SearchableDropdownField.propTypes = {
...
   options: PropTypes.arrayOf(PropTypes.object),
   optionUrl: PropTypes.string,
   passRef: PropTypes.bool.isRequired,
   searchable: PropTypes.bool.isRequired,
-  value: PropTypes.any.isRequired,
```
sabina-talipova, line 174: Here as well. We don't have any values in a new field and pass `value` as `null`.

GuySartorelli (changes requested):
1. Please retarget to `2.2` since this is a bug that should be resolved prior to tagging 5.2.0 stable
1. There is still no formfield displaying. A field should be displayed (that's the main thing the issue is about - the console errors were just added to help diagnose what's causing the problem)

I haven't looked at the actual code yet, just tried locally and saw that the form field still doesn't display.

sabina-talipova:
Close in favour of https://github.com/silverstripe/silverstripe-admin/pull/1699

PR #1705, opened 2024-03-09
ENH Move spinner to a seperate template so other module can include it

maxime-rainville:
This PR moves the spinner to its own template so other module can include it.

## Parent issue
https://github.com/silverstripe/silverstripe-linkfield/issues/242


michalkleiner (approved):
The actual change is fine, but can this be done in a patch version? 
I'm sometimes lost in how strict we are about this but if it was up to me it's fine.

emteknetnz (approved):

emteknetnz:
2.2 is currently beta so I don't think it really matters.  Also it's a template not a class so if it was outside of a beta it's also probably fine IMO

PR #1721, opened 2024-04-11
FIX Redirecting on preview loading

sabina-talipova:
## Description
The `'.cms-content:input[name=ID]'` element doesn't exist in another CMS section, only in "Pages" section. User should not be redirected to Pages section from another section when preview page is loaded.
## Parent issue
- https://github.com/silverstripe/silverstripe-versioned-admin/issues/335


emteknetnz (approved):
Tested locally, there are less network requests when using split view

Issue #1723, opened 2024-04-17
"Uncaught TypeError: can't convert undefined to object" in browser console


GuySartorelli:
I'm seeing this in the broswer developer console with a fresh install of Silverstripe CMS 5.2 in Firefox:

> Uncaught TypeError: can't convert undefined to object

I don't see any issues in the browser directly that I can trace as being caused by this, but we should resolve it in case it's causing unexpected flow-on effects that we just haven't noticed yet.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1724

PR #1724, opened 2024-04-17
FIX Ensure message events are handled correctly

GuySartorelli:
Resolves an error in the browser developer console.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1723


emteknetnz (approved):
Makes sense

PR #1728, opened 2024-04-22
MNT Use correct permission

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1729


GuySartorelli (approved):
Yup good catch

Issue #1733, opened 2024-04-30
SearchableDropdownField breaks grid field filtering


UndefinedOffset:
### Module version(s) affected

2.2.2

### Description

In Silverstripe Admin 2.2.0+ the new `SearchableDropdownField` breaks grid field filtering, if you select any option from the dropdown and click search `[object Object]` is sent as the value and displayed in the filter tags. This appears to be caused by [onChange](https://github.com/silverstripe/silverstripe-admin/blob/2.2.2/client/src/components/SearchableDropdownField/SearchableDropdownField.js#L89C9-L96) receiving an object (`{value: 1, label: "Option Label", selected: false }` for example) instead of a string/numeric value.

This is exaggerated because `SearchableDropdownField` now appears to be the scaffold default for relationships.

### How to reproduce

1. Add `ParentID` to `SilverStripe\Security\Group`'s `searchable_fields`:
   ```yml
   SilverStripe\Security\Group:
     searchable_fields:
       - 'ParentID'
   ```
2. ~Attempt to search for the default `Administrators` Group~
3. ~Note the tag in the filter bar and the request made to the server has the same value~
4. Create a group and make it the child of a group
5. In `/admin/security/groups` click the search magnifying glass, then click the "search options" dropdown.
6. In the "Parent ID" dropdown, select the parent group (or any group really) and click "Search"
7. Note that there are no results, and both the tag in the filter bar and the value in the request say `[object Object]`

### Possible Solution

A possible solution to this is to change the `handleChange` function to call `onChange` like `SingleSelectField` does passing the actual value as the second argument.

```javascript
const handleChange = (val) => {
  setHasChanges(false);
  if (JSON.stringify(value) !== JSON.stringify(val)) {
    setHasChanges(true);
    setJustChanged(true);
  }
  onChange({}, val);
};
```

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1734
- https://github.com/silverstripe/silverstripe-admin/pull/1736

sabina-talipova:
I've made several attempts to work around these changes, but due to the variability of how data is stored, how we send data to the server and how it is displayed, this requires more careful study than simply fixing a tag display problem.

[...]

PR #1736, opened 2024-05-06
FIX SearchableDropdownField pass incorrect value on onChange event

sabina-talipova:
## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1733


client/src/components/SearchableDropdownField/SearchableDropdownField.js
```diff
@@ -133,6 +133,15 @@ const SearchableDropdownField = ({
...
 
+  if (Number.isInteger(parseInt(val, 10))) {
+    options.forEach(option => {
+      // eslint-disable-next-line eqeqeq
+      if (option.value == value) {
+        val = option;
+      }
+    });
+  }
+
```
sabina-talipova, line 144: `react-select` expect to get value as an array of objects of ValueType in the following format `{ label: 'label', value: 'value' }`. See discussion [here](https://github.com/JedWatson/react-select/issues/2674#issuecomment-401318026).
`FormBuilder.buildComponent()` set `onChange` props as function with 2 arguments (event and payload), but handle payload value first and redux-form set this payload value in the state. [See](https://github.com/silverstripe/silverstripe-admin/blob/a3aa41cea4c4df82050eef65ad5efcfae7bfde69/client/src/components/FormBuilder/FormBuilder.js#L124-L156)
After we render `SearchableDropdownField` again we get integer value from state instead of object. So, if we pass only value that is returned back from `react-select` ([See](https://github.com/silverstripe/silverstripe-admin/blob/a3aa41cea4c4df82050eef65ad5efcfae7bfde69/client/src/components/FormBuilder/FormBuilder.js#L124-L156)) then we process it and store it as an `Object.toString()` and get `[object.Object]` in view. If we pass two params then we get integer or any other value, but not object, after processing.

client/src/components/SearchableDropdownField/SearchableDropdownField.js
```diff
@@ -133,6 +133,15 @@ const SearchableDropdownField = ({
...
 
+  if (typeof val !== 'object') {
+    options.forEach(option => {
+      // eslint-disable-next-line eqeqeq
+      if (option.value == value) {
+        val = option;
+      }
+    });
+  }
+
```
sabina-talipova: `react-select` expect to get value as an array of objects of `ValueType` in the following format `{ label: 'label', value: 'value' }`. See discussion [here](https://github.com/JedWatson/react-select/issues/2674#issuecomment-401318026).
`FormBuilder.buildComponent()` set onChange props as function with 2 arguments (event and payload), but handle payload value first and redux-form set this payload value in the state. [See](https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/components/FormBuilder/FormBuilder.js#L124-L156)
After we render SearchableDropdownField again we get integer value from state instead of object. So, if we pass only value that is returned back from react-select ([See](https://react-select.com/advanced#action-meta)) then we process it and store it as an Object.toString() and get [object.Object] in view. If we pass two params then we get integer or any other value, but not object, after processing.

sabina-talipova:
I left comment in the issue: https://github.com/silverstripe/silverstripe-admin/issues/1733#issuecomment-2097160219

GuySartorelli:
This PR seems to work without any problems for me.

In https://github.com/silverstripe/silverstripe-admin/issues/1733#issuecomment-2097160219 you mentioned "this requires more careful study than simply fixing a tag display problem" - but I don't understand what problems are left?
Does this PR introduce new regressions? Or are there other separate problems which also need to be fixed separately?

sabina-talipova:
That's right, this PR solves the current problem and, as far as I was able to test, it does not cause regression problems.

I will try to slightly reformulate my doubts about the correct operation of `SearchableDropdownField` expressed in the comments.

`SearchableDropdownField` works great everywhere except for search or where we need to pass a specific value (number or string).
Since `react-select` accepts an object or an array of objects as a “props.value” and return specific object onChange event, `SearchableDropdownField` through `FormBuilder` stores the object in redux state, from where we later retrieve this data before sending it to the server.

Everything works fine if we process the data before sending it to the server, as for example in the case of an asset-admin and users who are allowed to view. But not in the case of searching, since the data is not processed in any way before sending and we simply transfer the object.

The problem occurs in `$.ajax`, where I assume `obj.toString()` is applied to the object instead of `JSON.stringify(obj)`. Therefore, we send `"[objectObject]"` to the server instead of `"{id: 1, value: something}"`. But in any case, the server expects to receive some specific value (a number or a string), but not an object. Therefore, we should make changes to the search function on the backend.

Also, after the server returns the response, we will store the value `“[objectObject]”` in state, which in turn will later be displayed in the CMS elements.

So I believe the problem is a little deeper than just `SearchableDropdownField`. That is, I think that we should analyse where and how this field can be used and what data it will contain. You should also consider what type of data react-select accepts. Since in the future we plan to change the approach to working using Redux, this should also be given attention.

This PR simply stores in state only the value for a single value field and the entire array for a multiple choice field. That is, I think if someone decides to use` SearchableDropdownField` with multiple choice as a filter for searching, then it will not work.

GuySartorelli:
Sorry, that went a bit too deep into the weeds (too much specific technical detail) for me to be able to easily understand what problems remain. 

> But not in the case of searching, since the data is not processed in any way before sending and we simply transfer the object.
>
> The problem occurs in `$.ajax`, where I assume `obj.toString()` is applied to the object instead of `JSON.stringify(obj)`. Therefore, we send `"[objectObject]"` to the server instead of `"{id: 1, value: something}"`. But in any case, the server expects to receive some specific value (a number or a string), but not an object. Therefore, we should make changes to the search function on the backend.

The above suggests there's still some problem related to `SearchableDropdownField` in relation to searching - but that's exactly what this PR fixes, isn't it?

> You should also consider what type of data ReactSelect accepts. Since in the future we plan to change the approach to working using Redux, this should also be given attention.

Are there any specific problems related to this that you're aware of?

> This PR simply stores in state only the value for a single value field and the entire array for a multiple choice field. That is, I think if someone decides to use `SearchableDropdownField` with multiple choice as a filter for searching, then it will not work.

That's a good point. I've just tested that using the following code:
```php
<?php

namespace App\Extension;

use SilverStripe\Core\Extension;
use SilverStripe\Forms\SearchableMultiDropdownField;
use SilverStripe\Security\Member;

class GroupSearchableFieldsExtension extends Extension
{
    public function updateSearchableFields(array &$fields)
    {
        $fields['Members'] = [
            'title' => 'Users',
            'field' => SearchableMultiDropdownField::create('Members', 'Users', Member::get()),
        ];
    }
}
```
```yml
SilverStripe\Security\Group:
  extensions:
    - App\Extension\GroupSearchableFieldsExtension
```

Searching with that field does indeed give a bad result - specifically if I select two records I get `[object+Object],[object+Object]` in the request, which results in an exception. That should be easy to remedy in this PR.

client/src/components/SearchableDropdownField/SearchableDropdownField.js
```diff
@@ -92,7 +92,11 @@ const SearchableDropdownField = ({
       setHasChanges(true);
       setJustChanged(true);
     }
-    onChange(val);
+    if (multi) {
+      onChange(val);
```
ChloeSartorelli, line 96: Please update the case where the value is multi, the same as the non-multi case.
There's code to test this in https://github.com/silverstripe/silverstripe-admin/pull/1736#issuecomment-2099568462

GuySartorelli:
Are there any specific problems outstanding beyond that? As in specific problems that cause errors or bad UX, which can be easily reproduced?

sabina-talipova:
@GuySartorelli , how I said before, the current PR fixes initial issue. For another problems that were discussed in comments I've opened new ticket https://github.com/silverstripe/silverstripe-admin/issues/1746  

> Are there any specific problems outstanding beyond that? As in specific problems that cause errors or bad UX, which can be easily reproduced?

No, I couldn't find any problems. I would be good to test this field created by FormBuilder.

GuySartorelli:
I've updated that issue to reflect what we discussed in slack - the main problem is that `many_many` relations can't be searched against, and fixing that is necessary for using multiple values in `SearchableMultiDropdownField` to be useful.

GuySartorelli (approved):
LGTM, works well locally for single values and there's a separate issue to deal with multi values since that's a more involved problem.

PR #1739, opened 2024-05-07
FIX Rebuild bundle

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/235

Fixes [broken build](https://github.com/silverstripe/silverstripe-admin/actions/runs/8908721417/job/24464857467#step:12:605)
```
Running git diff
git diff found modified files when it should not have:
M js/bundle.js
sha1sum of files that are different:
c2229f73bc5aa2fdbafa28e1ec313b041bc96940  client/dist/js/bundle.js
```


GuySartorelli (approved):
LGTM

PR #1757, opened 2024-05-28
FIX return 404 instead of 500 for invalid model class, fixes #570

### LINKED ISSUE #570: ModelAdmin should fail gracefully when provided a non-sense modelclass
If you try to manage to access a `ModelAdmin` with a non-sensical model (e.g.: `http://example.com/admin/my-model-admin/EditForm/field/Acme-Web-Some-Models-That-Doesnt-exists`) you get some pretty scary 500 error.

This should return either a 400 or 404 error.

image

xini:
## Description

When a ModelAdmin is called with a model that doesn’t exist, the fact that it doesn’t exist is exposed in a 500 error if the site is in dev.
 
Affected: silverstripe/silverstripe-admin 1 and 2 (Silverstripe 4 and 5)
 
In the init() method of ModelAdmin (https://github.com/silverstripe/silverstripe-admin/blob/2/code/ModelAdmin.php#L165) a RuntimeException is thrown including the class information. This is also the case when there is no user logged into the CMS. 

This is the case for v1 and v2 of this module!

## Issues
- #570

## Pull request checklist
- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [x] CI is green

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Admin;
 
 use InvalidArgumentException;
+use Psr\Log\LoggerInterface;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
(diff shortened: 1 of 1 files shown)


GuySartorelli:
I'm not convinced this is desirable. As @dhensby points out in the issue, the existing error is useful for debugging. At a minimum I'd want to keep throwing the exception in dev mode.

Maybe we want a 404 in live and maybe test mode? Though I'd want @silverstripe/core-team to weigh in with their thoughts.
In my opinion it's currently working as expected.

GuySartorelli:
In the meantime, please retarget this to the `2.2` branch, and either add tests or check the relevant checkbox if tests aren't required.

michalkleiner:
I'd leave it as is. The fact the model doesn't exist would still be exposed, just with a different HTTP code. It might be more confusing, as a developer, to get 404 instead of 500.

xini:
I disagree. When an URL is incorrect, like https://www.silverstripe.org/foo, you expect a 404. This should also be the case for a model admin URL like https://www.silverstripe.org/admin/my-admin/foo. There is no logical reason for this to be a 5xx response and not a 404.

Issue #1755, opened 2024-05-27
Replace redux-form with react-hook-form


emteknetnz:
Implementation of the [spike](https://github.com/silverstripe/silverstripe-admin/issues/1747) which recommends that we continue replacing redux-form with react-hook-form

In the linked spire there are a few PRs that have successfully replaced redux-form with react-hook-form, there's a number of loose ends to tidy up e.g. no longer using redux-form helpers to update the redux store

## Acceptance criteria
- [ ] redux-form dependency is removed entirely
- [ ] Any references to redux-form, including code comments, are removed or updated as appropriate
- [ ] redux store continues to be updated as needed, without redux-form helpers
- [ ] Any API change to FormBuilder component is explicitly noted in the CMS 6 changelog and an upgrade path is provided
- [ ] Kitchen-sink CI is run using all PRs and Behat enabled
- [ ] Form UI remains unchanged.
- [ ] yarn warning about providerValue in silverstripe/admin is resolved
- [ ] Elemental TextCheckboxGroupField renders correctly
- [ ] Elemental handleFormInit / onFormInit functionality works correctly

## Notes
- We may end up ditching redux altogether for form management, but not now.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1758
- https://github.com/silverstripe/silverstripe-elemental/pull/1195

## POC PRs
- https://github.com/silverstripe/webpack-config/pull/74
- https://github.com/silverstripe/silverstripe-admin/pull/1754

emteknetnz:
In the PRs above I've swapped out redux-form for react-hook-form and run it through a [CI run](https://github.com/emteknetnz/recipe-kitchen-sink/actions/runs/9297968319) of kitchen-sink including all of the behat tests, and there were a fairly large number and wide range of failures. I had a decent look at one of the failures and it was because the [AnchorSelectorField](https://github.com/silverstripe/silverstripe-cms/blob/5/client/src/components/AnchorSelectorField/AnchorSelectorField.js) relies on the redux state set by redux-form of an entirely different react component in order to function.

Currently, the form state is updated by reducer provided by redux-form, which updates window.ss.store.getState().form.formState. This reducer is connected in [registerReducers.js](https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/boot/registerReducers.js#L19). 

[...]

PR #1758, opened 2024-05-29
NEW Use react-hook-form in FormBuilder

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1755


emteknetnz:
This will probably not be completed

PR #1771, opened 2024-05-30
FIX Allow reloading react forms based on changes to a prop

GuySartorelli:
We could just add garbage getvars to the schema URL to get the same effect - but that's gross.

## Issue
- https://github.com/silverstripe/silverstripe-linkfield/issues/256


GuySartorelli:
Behat failure is unrelated - it's happening on the `2.2` branch as well. See https://github.com/silverstripe/silverstripe-admin/actions/runs/9310224470

emteknetnz (approved):

Issue #1778, opened 2024-06-11
Lazyloaded SearchableDropdownField does not work in elemental


emteknetnz:
Lazyloaded SearchableDropdownField does not function correctly when using in elemental.  Selecting a value does not visually update the field, though when you save the value is saved.  I tested using an autoscaffolded field, meaning I had over 100 DataObject which I create via requireDefaultFields(). When setting ->setIsClearable(true) on DBForeignKey.php, when I clicked the clear button I got console warnings (I had run yarn dev in admin)

Discovered when implementing https://github.com/silverstripe/silverstripe-framework/pull/11273

Note this isn't an issue when using a SearchableDropdownField in a regular form

Test setup:

You'll need dnadesign/silverstripe-elemental installed

**MyBlock.php**
```php
<?php

use DNADesign\Elemental\Models\BaseElement;

class MyBlock extends BaseElement
{
    private static $db = [
        'MyField' => 'Varchar',
    ];

    private static $has_one = [
        'MySubDataObject' => MySubDataObject::class,
    ];
    private static $table_name = 'MyBlock';

    private static $singular_name = 'My Block';

    private static $plural_name = 'My Blocks';

    private static $description = 'This is my block';

    private static $icon = 'font-icon-block-content';

    public function getType()
    {
        return 'My Block';
    }
}
```

**MySubDataObject.php**
```php
<?php

use SilverStripe\ORM\DataObject;

class MySubDataObject extends DataObject
{
    private static $table_name = 'MySubDataObject';

    private static $db = [
        'Title' => 'Varchar'
    ];

    private static $has_many = [
        'MyBlock' => MyBlock::class,
    ];

    public function requireDefaultRecords()
    {
        parent::requireDefaultRecords();
        if (self::get()->count() < 110) {
            for ($i = 0; $i < 110; $i++) {
                $record = self::create();
                $record->Title = 'MySubDataObject ' . $i;
                $record->write();
            }
        }
    }
}
```

## New issues created
- https://github.com/silverstripe/silverstripe-admin/issues/1791

## Shared CI run
- https://github.com/emteknetnz/silverstripe-installer/actions/runs/9589272730

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1786
- https://github.com/silverstripe/silverstripe-frameworktest/pull/188
- https://github.com/silverstripe/silverstripe-elemental/pull/1209

GuySartorelli:
PRs merged

PR #1786, opened 2024-06-18
FIX Use correct argument for FormBuilder context

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1778


client/src/components/SearchableDropdownField/SearchableDropdownField.js
```diff
@@ -92,11 +92,9 @@ const SearchableDropdownField = ({
...
       setJustChanged(true);
     }
-    if (multi) {
-      onChange(val);
-    } else {
-      onChange(val, { value: val.value });
-    }
+    // The first param is for handleChange() SearchableDropdownFieldEntwine.js
+    // The second param is for onChange in FormBuilder::buildComponent()
+    onChange(val, { value: val });
```
ChloeSartorelli: This change has resulted in a regression of https://github.com/silverstripe/silverstripe-admin/issues/1733
Probably still needs to be `val.value`, at least for single fields. Please read the comments in the issue and the PR 'cause I recall there was discussion about the difference between single and multi fields, as well.

GuySartorelli (changes requested):

GuySartorelli:
I can't de-select a record from the single `SearchableDropdownField` in an elemental block anymore.
I can select a _new_ record, but not deselect one that was already saved into the field.

emteknetnz:
I think this might just a be bad default when lazy-loading

Things are "clearable" by default when not lazy-loading because there's a default blank option that's selectable. This option is not there however when you lazy-load. I can solve this by setting `->setIsClearable(true)` alongside `->setIsLazyLoaded(true)` then I can clear the value.  

I think the solution here is that we just update the field to be clearable by default? This won't be a breaking change, just a behavioural change we could ship in 5.3

Note that clicking the little 'x' when it's clearable will clear ALL selected options for the mutli version of the field, though this seems like it's still a good default to me

I'll make a new issue for this

emteknetnz:
New issue to ensure field is always clearable https://github.com/silverstripe/silverstripe-admin/issues/1791

GuySartorelli (approved):
LGTM

Issue #1789, opened 2024-06-19
Regression caused by scss linting fixes


GuySartorelli:
[This PR](https://github.com/silverstripe/silverstripe-admin/pull/1774) which fixes scss linting warnings has caused a regression in symbiote/silverstripe-gridfieldextensions.

When using the `GridFieldNestedExtensions` component, there should be a dropdown arrow next to gridfield items:

![image](https://github.com/silverstripe/silverstripe-admin/assets/36352093/b3575b72-0e96-44dd-a608-c3aa2146418f)

instead, the dropdown arrow is missing:

![image](https://github.com/silverstripe/silverstripe-admin/assets/36352093/1fc830b3-8736-4eb8-804f-a989f02af369)

I ran a `git bisect` and found the above-linked PR is what caused this bug.

You can reproduce this by using the steps in the description of [this PR](https://github.com/symbiote/silverstripe-gridfieldextensions/pull/384) to set up a nested gridfield.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1790
- https://github.com/silverstripe/silverstripe-elemental/pull/1213

PR #1790, opened 2024-06-19
FIX Restore correct margin for gridfield

GuySartorelli:
This was being incorrectly interpreted as a mathematical expression with the values being subtracted, rather than a bunch of negative numbers.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1789


emteknetnz (approved):
Tested locally works good

PR #1792, opened 2024-06-24
NEW Wait for promises before submitting forms

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1215


client/src/legacy/LeftAndMain.js
```diff
@@ -460,77 +460,93 @@ $.entwine('ss', function($) {
...
+        promises: [],
+        // callbacks that are called on ajax success after submitted the form
+        onAjaxSuccessCallbacks: [],
+      };
+      form.trigger('beforesubmitform', beforeSubmitFormEventData);
 
-      // set button to "submitting" state
-      $(button).addClass('btn--loading loading');
-      $(button).prop('disabled', true);
+      Promise.all(beforeSubmitFormEventData.promises).then(function(results) {
```
emteknetnz, line 473: The diff in this PR looks confusing, though basically what I've done is nest the old logic instead the `Promise.all(...).then(...)` block

client/src/legacy/LeftAndMain.js
```diff
@@ -460,77 +460,93 @@ $.entwine('ss', function($) {
...
+        jQuery.ajax(jQuery.extend({
+          headers: {"X-Pjax" : "CurrentForm,Breadcrumbs,ValidationResult"},
+          url: form.attr('action'),
+          data: formData,
+          type: 'POST',
+          complete: function() {
+            clearButton()
+          },
+          success: function(data, status, xhr) {
+            beforeSubmitFormEventData.onAjaxSuccessCallbacks.forEach(fn => fn());
```
emteknetnz, line 564: This is a new line added

client/src/legacy/LeftAndMain.js
```diff
@@ -460,77 +460,93 @@ $.entwine('ss', function($) {
...
+      form.trigger('beforesubmitform', beforeSubmitFormEventData);
 
-      // set button to "submitting" state
-      $(button).addClass('btn--loading loading');
-      $(button).prop('disabled', true);
+      Promise.all(beforeSubmitFormEventData.promises).then(function(results) {
 
-      if($(button).is('button')) {
-        $(button).data('original-text', $(button).text());
+        for (const result of results) {
```
emteknetnz, line 477: This block is newly added - if any of the promises resolve(false) then the form will not submit

emteknetnz:
Broken --prefer-lowest build is unrelated and will be fixed in https://github.com/silverstripe/gha-ci/issues/132

client/src/legacy/LeftAndMain.js
ChloeSartorelli: Diff is much more sensible with whitespace ignored 👌

client/src/legacy/LeftAndMain.js
```diff
@@ -460,77 +460,93 @@ $.entwine('ss', function($) {
...
+        promises: [],
+        // callbacks that are called on ajax success after submitted the form
+        onAjaxSuccessCallbacks: [],
+      };
+      form.trigger('beforesubmitform', beforeSubmitFormEventData);
+
+      Promise.all(beforeSubmitFormEventData.promises).then(function(results) {
+
+        for (const result of results) {
+          if (!result) {
```
```suggestion
          if (result === false) {
```
I'd prefer this since that's [what you've said](https://github.com/silverstripe/silverstripe-admin/pull/1792#discussion_r1653845328) is the condition for this. Means we don't have to care about weird edge cases of other falsy values getting in here either with our own changes down the line or in peoples' customisations.

GuySartorelli (changes requested):

GuySartorelli (approved):
LGTM, will merge when elemental PR is approved

client/src/components/HtmlEditorField/HtmlEditorField.js
```diff
@@ -54,7 +54,7 @@ class HtmlEditorField extends TextField {
    */
   registerChangeListener() {
     const target = this.getEditorElement();
-    this.getEditor().on('change keyup setcontent', () => {
+    this.getEditor().on('change keyup', () => {
```
emteknetnz, line 57: This change was made to resolve the "Blocks get double saved - which turns them back to "modified" issue noted [here](https://github.com/silverstripe/silverstripe-elemental/pull/1214#issuecomment-2218753147). The reason this was happening is that form-schema value of a blank HTMLText field (or Varchar field for that matter), starts of as `null`.

However as part of the HTMLEditorField initialising it has a `setcontent` call made to it, which in turns tells the redux that "this form has changes" and therefore `formDirty` is always set for the element, which is obviously wrong. `formDirty` is used in the new logic that determines if the element form should be inline submitted when clicking a page/parent save button.

I have regression tested this change in sink using this [PR](https://github.com/silverstripe/silverstripe-admin/pull/1796) that only contains this change.
That [run of sink](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/9983689111) has the same failures as a [vanilla sink run](https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/9984678122)

GuySartorelli (approved):
LGTM, will merge when elemental PR is approved

client/src/boot/applyTransforms.js
```diff
@@ -65,6 +65,14 @@ const applyTransforms = () => {
       updater.form.addValidation(
         '*',
         (values, Validation, schema) => {
+          // Hardcoded exclusion for elemental inline forms
+          // This is done so that client-side validation doesn't run so that all validation is done
+          // on the server. This is done so that invalid, closed elements will popup open and
+          // also trigger a toast notification.
+          // Note that this has no effect on non-inline forms as they do not use form schema
+          if (schema.name.indexOf('ElementForm_') === 0) {
```
emteknetnz, line 73: This is to resolve https://github.com/silverstripe/silverstripe-elemental/pull/1214#issuecomment-2237727252

I can't think of a clean way to abstract that isn't basically window.ss.excludeClientSideValidation = ['ElementForm_']

I think hardcoding may actually be better here as there's a nice big inline comment explaining what's going on.

ChloeSartorelli: Hardcoding isn't ideal but it gets the job done. Would be better if we did something like `if (field.parents('.skipClientSideFormValidation').length > 0)` (if any parent of the field has that css class, then we can skip this client-side validation) and add a `skipClientSideFormValidation` css class to the elemental form.

I'll let you decide if you want to do that - if not I'm okay with merging as is.

emteknetnz: I think merge as.

I've had a quick look at doing this "properly", would involved adding new API to formschema rather than just a class on a form.  Also because it deals with validation I think we'd possibly want to add it to the CompositeValidator, rather than directly on the Form e.g. Form->setClientSideValidation(false);  I'm not sure where it would be best.

Regardless it would be a proper standalone PR.  We could easily add this later should the need arise and remove the hack added here, so I think the hack here is OK

ChloeSartorelli: Ahhh it's react nonsense. That explains why this doesn't happen for validation of fields directly on the page. I thought this was jquery nonsense but honestly it being react both makes sense and is hilarious.
No worries.

GuySartorelli (approved):
LGTM, will merge when elemental PR is approved

GuySartorelli:
CI failure will be resolved as part of https://github.com/silverstripe/.github/issues/285

GuySartorelli:
Will rerun CI after https://github.com/silverstripe/silverstripe-admin/pull/1792 is merged

GuySartorelli:
Can you please rebase this? Looks like it needs a rebase to pick up the missing `@job` tags in behat.

emteknetnz:
Done

GuySartorelli (approved):
LGTM

PR #1798, opened 2024-07-28
FIX Correctly set img tag attributes for WYSIWYG

GuySartorelli:
The same fix as https://github.com/silverstripe/cwp/pull/340 but for the default "cms" WYSIWYG config.

Fixes https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/10102090864/job/27937027096 and https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/10102090864/job/27937027255 in tandem with https://github.com/silverstripe/silverstripe-cms/pull/2978

In this case for some reason the behat test was explicitly expecting _no_ alt tag, though it's clear that the empty alt tag should be displayed.

## Issue
- https://github.com/silverstripe/.github/issues/287


GuySartorelli:
behat failures are pre-existing - see https://github.com/silverstripe/silverstripe-admin/actions/runs/10123308730

emteknetnz (approved):

Issue #1799, opened 2024-07-29
Remove GraphQL from the CMS


emteknetnz:
GraphQL is used inconsistently in the CMS for some of the XHR requests made by react components. The majority of XHR requests are currently made to "regular Silverstripe controller endpoints", which also includes "Form Schema" endpoints used to process Form requests.

This lack of standardisation has lead to inefficiencies in development an maintenance for the CMS Squad. There has also been a fair bit of negative feedback from several developers in the community who think they're supposed to do all admin XHR requests using GraphQL, which has proven to be very difficult for people to learn and use.

This card is about removing GraphQL from the CMS entirely and be replaced with "regular Silverstripe controller endpoints" in the name of standardisation.

Removing GraphQL from the CMS also means that projects no longer need to have the `.graphql-generated` directory by default.

## Related

- https://github.com/silverstripe/silverstripe-admin/issues/1805

## Acceptance criteria
- [x] Functionality that GraphQL enabled is retained and replaced with regular Silverstripe controller endpoints
- [x] Overall test coverage is not reduced
- [x] All GraphQL code and config is removed (excluding the actual silverstripe/graphql module)
- [x] GraphQL dependency is not included in any commericially supported composer.json files
- [x] The .graphql-generated folder is not created when creating a project with silverstripe/installer ~~or silverstripe/recipe-kitchen-sink~~ (will probably still be created for recipe-sink because we're including the graphql module)
- [x] silverstripe/graphql is still included in silverstripe/recipe-kitchen-sink
- [x] PRs target CMS 6
- [x] Removed classes and methods are deprecated in CMS 5
- [ ] In docs.silverstripe.org, the "Developer Guides > GraphQL" section is moved to "Optional features" (have raised new card for this)
- [x] Any other relevant documentation referring to graphql in docs.silverstripe.org is updated accordingly
- [ ] Create follow up PRs to use the correct deprecation version in CMS 5 e.g. for asset-admin it should be deprecated in 2.3.0, not 5.3.0. After merging ensure merge-up is done correctly as the deprecated methods will have been removed

## New issues created
- https://github.com/silverstripe/doc.silverstripe.org/issues/304

## CMS 5 PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1484
- https://github.com/silverstripe/silverstripe-cms/pull/2986
- https://github.com/silverstripe/silverstripe-elemental/pull/1235
[...]

PR #1802, opened 2024-07-30
API Remove GraphQL

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1799


code/LeftAndMain.php
```diff
@@ -53,6 +53,7 @@
 use SilverStripe\View\ArrayData;
 use SilverStripe\View\Requirements;
 use SilverStripe\View\SSViewer;
+use stdClass;
```
```suggestion
```
Doesn't seem to be used (if it is, please say which line it's on 'cause I missed it)

code/LeftAndMain.php
```diff
@@ -429,12 +430,16 @@ public function schema(HTTPRequest $request): HTTPResponse
     /**
      * Creates a successful json response
      */
-    protected function jsonSuccess(int $statusCode, array $data = []): HTTPResponse
+    protected function jsonSuccess(int $statusCode, ?array $data = null): HTTPResponse
```
ChloeSartorelli, line 443: Why not just use empty array as empty value?

code/LeftAndMain.php
```diff
@@ -429,12 +430,16 @@ public function schema(HTTPRequest $request): HTTPResponse
...
     {
         if ($statusCode < 200 || $statusCode >= 300) {
             throw new InvalidArgumentException("Status code $statusCode must be between 200 and 299");
         }
-        $body = empty($data) ? '' : json_encode($data, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
+        if (is_null($data)) {
+            $body = '';
+        } else {
+            $body = json_encode($data, JSON_UNESCAPED_SLASHES | JSON_UNESCAPED_UNICODE);
+        }
```
ChloeSartorelli, line 452: This change doesn't seem to actually do anything? Doesn't it give the exact same result it did before?

code/LeftAndMain.php
```diff
@@ -479,6 +484,23 @@ public function jsonError($errorCode, $errorMessage = null)
...
+        $securityToken = SecurityToken::inst();
+        $name = $securityToken->getName();
+        if ($name === 'SecurityID') {
+            // SecurityID is a non-standard HTTP header, it should be prefixed with 'X-'
+            $name = "X-SecurityID";
+        }
+        $value = $request->getHeader($name);
+        if (!$securityToken->check($value)) {
+            $this->jsonError(400);
+        }
```
ChloeSartorelli: Why not just `SecurityToken::inst()->checkRequest($request)`? Seems like this whole method is unnecessary?

tests/php/LeftAndMainTest.php
```diff
@@ -155,6 +155,11 @@ public function testCanView()
         $menuItems = $menuItems->column('Code');
         sort($menuItems);
 
+        // filter out any menus from other modules that may be installed
+        $menuItems = array_filter($menuItems, function ($item) {
+            return strpos($item, 'SilverStripe-Admin-') === 0;
+        });
```
ChloeSartorelli: Why is this needed now when it wasn't before?

GuySartorelli (changes requested):

code/LeftAndMain.php
emteknetnz: Removed

code/LeftAndMain.php
emteknetnz: Sometimes you actually want to return an empty JSON array

code/LeftAndMain.php
emteknetnz: It's to allow JSON encoding an empty array. Previously it would return an empty body instead.

code/LeftAndMain.php
emteknetnz: That is correct. Will update.

tests/php/LeftAndMainTest.php
emteknetnz: No longer required I think, will remove

code/LeftAndMain.php
ChloeSartorelli: Discussed in person - the reason reason is so that you can have an empty body, which seems sensible to allow.

code/LeftAndMain.php
ChloeSartorelli: See https://github.com/silverstripe/silverstripe-admin/pull/1802#discussion_r1726207432

GuySartorelli (approved):
LGTM, will merge when the other PRs are ready

GuySartorelli (approved):
re-approved most recent changes

GuySartorelli (approved):
re-approved most recent changes

GuySartorelli:
Needs rebase

emteknetnz:
Done

GuySartorelli (approved):
LGTM

PR #1810, opened 2024-08-18
API Standardise extension hooks

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11258

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -784,7 +784,7 @@ class LeftAndMain extends Controller implements PermissionProvider
             }
         }
 
-        $this->extend('init');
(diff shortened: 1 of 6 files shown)


code/LeftAndMain.php
```diff
@@ -784,7 +784,7 @@ protected function init()
             }
         }
 
-        $this->extend('init');
+        $this->extend('onInit');
```
ChloeSartorelli, line 787: Intentionally not changing to `onAfterInit` because this isn't at the end of the method.
Also not `updateInit` because it's not updating the value of some argument.

emteknetnz (approved):

PR #1814, opened 2024-08-20
Fix TinyMCE edit link when link contains html

satrun77:
## Description

If you have a link in tinymce like `<a href=""><span>Title</span></a>`, then you can edit the link. The HTML inside the link tag can be any tag. 

The issue is that the selected node is the inside HTML tag instead of the link tag.

I'm not 100% keen on the solution using `while` but this is legacy code, so might be ok! Alternative code is welcomed :) 

The PR https://github.com/silverstripe/silverstripe-cms/pull/2993 depends on this PR.

## Manual testing steps

1. Install fresh CMS
2. Edit a page
3. Add a link (internal or external)
4. Edit HTML code in TinyMCE, add any HTML tags inside the link
5. Close the HTML code view in TInyMCE
6. Try to edit the link, it won't show the popup form.

## Issues

- https://github.com/silverstripe/silverstripe-cms/issues/2439

## Pull request checklist

- [x] CI is green


GuySartorelli (changes requested):
The comments in https://github.com/silverstripe/silverstripe-cms/pull/2993#pullrequestreview-2249599581 apply to this PR as well.

satrun77:
I have updated the fix by moving the selection to the link element. This does not need the other PR. I have tested this with internal and external linking, it worked for me as expected.

client/src/legacy/TinyMCE_sslink.js
```diff
@@ -35,6 +35,9 @@ const plugin = {
 
     // Callback for opening the edit link dialog form
     function openLinkDialog() {
+      // Find "a" node, issue https://github.com/silverstripe/silverstripe-cms/issues/2439
```
ChloeSartorelli:
```suggestion
      // Find "a" node (we might have clicked on a child element e.g. "span")
```
There's no need to list the issue - git blame will inevitably lead us back to this PR and then to that issue if we really need that level of context on this.

GuySartorelli (changes requested):
Thanks, this works great locally. Just one tiny change to make.

client/src/legacy/TinyMCE_sslink.js
satrun77: Done updated

GuySartorelli (approved):
LGTM, works well locally. Thanks for this!

Issue #1829, opened 2024-09-24
ModelAdmin executes init code even when there is no authenticated user


lekoala:
### Module version(s) affected

5.x

### Description

I recently discovered this unexpected error in my logs...

```
RuntimeException
ModelAdmin::init(): Invalid Model class lib
```

for the following url: /admin/events/lib/external/responsive_filemanager/filemanager/dialog.php

That's expected, "lib" is not a valid model class. But what's really odd, is that this is the case for anonymous users... meaning any bot can basically hammer your website and create tons of errors logs due to this.

The issue is that the auth check does not interrupt the init() process in subclasses (there is simply a return statement in the init parent class)

### How to reproduce

Visit any /admin/security|modeladmin_segment/invalid_modal/xxx url on a ss website and get a server error

eg:
https://some.domain.com/admin/security/lib/external/responsive_filemanager/filemanager/dialog.php

### Possible Solution

Always add a redirectedTo check in any ModelAdmin subclasses... (not great, because you have to think about it)

```php
<?php
    protected function init()
    {
        parent::init();

        if ($this->redirectedTo()) {
            return;
        }
?>
```

Better long term solution:

Throw a RedirectException (this does not exist, but I think it really should be added to the core) in the init method to avoid any further processing instead of what's currently in place. This would make the whole thing much simpler and avoid issues for unsuspecting developers.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1830

GuySartorelli:
A `RedirectException` (similar to the existing `ValidationException`) sounds like a good long-term solution that could be introduced in CMS 6. Would you be interested in implementing that?

I'll review the CMS 5 bugfix PR in the meantime.

lekoala:
> A `RedirectException` (similar to the existing `ValidationException`) sounds like a good long-term solution that could be introduced in CMS 6. Would you be interested in implementing that?

I don't have much time at the moment :-) I can easily draft up some basics, but if it needs to be a proper PR (ie : adding tests, documentation etc etc) then it's clearly to time consuming

GuySartorelli:
[...]

PR #1830, opened 2024-09-24
FIX Don't trigger init if already redirected

lekoala:
Fixes https://github.com/silverstripe/silverstripe-admin/issues/1829

See issue for context

## Issues

- https://github.com/silverstripe/silverstripe-admin/issues/1829

## Pull request checklist

- [x] CI is green

### DIFF
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -152,7 +152,8 @@ abstract class ModelAdmin extends LeftAndMain
 
         // security check for valid models
         if ($this->modelTab && !$this->isManagedModel($this->modelTab)) {
-            // No need to throw exceptions since we are already redirected
+            // No need to check model tab since we are already redirected
+            // This happens when there's a permission failure in LeftAndMain
             if ($this->redirectedTo()) {
                 $this->modelTab = null;
             } else {
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -147,6 +147,11 @@ abstract class ModelAdmin extends LeftAndMain
     {
         parent::init();
 
+        // Don't do anything if already redirected
+        if ($this->redirectedTo()) {
+            return;
+        }
+
         $models = $this->getManagedModels();
         $this->modelTab = $this->getRequest()->param('ModelClass');
 
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -147,11 +147,6 @@ abstract class ModelAdmin extends LeftAndMain
     {
         parent::init();
 
-        // Don't do anything if already redirected
-        if ($this->redirectedTo()) {
-            return;
-        }
-
         $models = $this->getManagedModels();
         $this->modelTab = $this->getRequest()->param('ModelClass');
 
@@ -159,6 +154,11 @@ abstract class ModelAdmin extends LeftAndMain
         if ($this->modelTab === null) {
             reset($models);
             $this->modelTab = key($models ?? []);
+        }
+
+        // Don't do anything if already redirected
+        if ($this->redirectedTo()) {
+            return;
         }
 
         // security check for valid models
--- a/code/ModelAdmin.php
+++ b/code/ModelAdmin.php
@@ -150,25 +150,25 @@ abstract class ModelAdmin extends LeftAndMain
         $models = $this->getManagedModels();
         $this->modelTab = $this->getRequest()->param('ModelClass');
 
+        // security check for valid models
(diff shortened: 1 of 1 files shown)


GuySartorelli (commented):
Code looks good to me.
I'm just running a [CI run against `2.2`](https://github.com/silverstripe/silverstripe-admin/actions/runs/11022117621) to double check if the CI failures are pre-existing.

GuySartorelli:
The CI failures aren't happening without this PR. Can you please take a look and see if you can resolve them?

lekoala:
The CI errors are a bit strange honestly

EDIT : ok I see the classes expect to have a default tab class even with an empty session, i'll fix this

I think these tests assume too much and should probably be fixed but I haven't had a deeper look at it

```
1) SilverStripe\Admin\Tests\ModelAdminTest::testExportFieldsDefaultIsSummaryFields
InvalidArgumentException: singleton() Called without a class

/home/runner/work/silverstripe-admin/silverstripe-admin/vendor/silverstripe/framework/src/includes/functions.php:28
/home/runner/work/silverstripe-admin/silverstripe-admin/code/ModelAdmin.php:377
/home/runner/work/silverstripe-admin/silverstripe-admin/tests/php/ModelAdminTest.php:81
/home/runner/work/silverstripe-admin/silverstripe-admin/vendor/bin/phpunit:122

2) SilverStripe\Admin\Tests\ModelAdminTest::testGetGridField
InvalidArgumentException: Call <classname>::get() instead of DataObject::get()

/home/runner/work/silverstripe-admin/silverstripe-admin/vendor/silverstripe/framework/src/ORM/DataObject.php:3320
/home/runner/work/silverstripe-admin/silverstripe-admin/code/ModelAdmin.php:401
/home/runner/work/silverstripe-admin/silverstripe-admin/code/ModelAdmin.php:2[94](https://github.com/silverstripe/silverstripe-admin/actions/runs/11013516566/job/30582087994?pr=1830#step:12:97)
/home/runner/work/silverstripe-admin/silverstripe-admin/tests/php/ModelAdminTest/OverridenModelAdmin.php:50
/home/runner/work/silverstripe-admin/silverstripe-admin/code/ModelAdmin.php:266
/home/runner/work/silverstripe-admin/silverstripe-admin/tests/php/ModelAdminTest.php:119
/home/runner/work/silverstripe-admin/silverstripe-admin/vendor/bin/phpunit:122
```

lekoala:
@GuySartorelli CI is green

code/ModelAdmin.php
```diff
@@ -150,22 +150,27 @@ protected function init()
         $models = $this->getManagedModels();
         $this->modelTab = $this->getRequest()->param('ModelClass');
 
+        // security check for valid models
+        if ($this->modelTab && !$this->isManagedModel($this->modelTab)) {
+            // No need to throw exceptions since we are already redirected
```
ChloeSartorelli:
```suggestion
            // No need to check model tab since we are already redirected
            // This happens when there's a permission failure in LeftAndMain
```
Just a small change to make it really clear why this is here - otherwise I could see this being removed if this code is refactored in the future.

GuySartorelli (changes requested):
Thanks for updating so the tests pass!
It's a little less clear what's going on now though, so I've requested a change to the comment to make up for that.

michalkleiner:
I'm not sure if the `// if we've hit the "landing" page` code block now makes sense in the place where it is now, _after_ the new check.

lekoala:
> I'm not sure if the `// if we've hit the "landing" page` code block now makes sense in the place where it is now, _after_ the new check.

i can update to "if we've hit the "landing" page or if the model is not valid"

i agree it still a bit odd, because you would expect and invalid model to do some kind of errors, but since the model admin class can be instantiated by itself, without a valid user (as designed in the unit tests at the moment), it should still properly return default summary fields for instance

GuySartorelli:
Probably updating that comment would do IMO, or else adding a 'reset so we use the "landing page"' comment when setting it to `null` in the new condition.
What do you think @michalkleiner?

GuySartorelli:
Given the change freeze corresponding with the 5.3.0-rc1 release is coming in the next few days (if all goes well), @lekoala can you please update the comment as mentioned above? I'll then merge it in if Michal doesn't respond before I click the button.

lekoala:
@GuySartorelli done

michalkleiner (approved):
All good from my perspective now, thanks Thomas!

GuySartorelli (approved):
LGTM, thanks for making that change.

PR #1832, opened 2024-10-03
NEW Give feedback of password strength

emteknetnz:
~~Issue https://github.com/silverstripe/silverstripe-framework/issues/11413~~

~~This PR is part of a POC to give UX feedback of the new password~~

Provides JS and CSS for functionality that displays feedback when entering a password into a `ConfirmedPasswordField` which uses Symfony's entropy-based password validation.

Requires https://github.com/silverstripe/silverstripe-framework/pull/11417

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11475


GuySartorelli:
You might want to add a bit of a description about what you're actually doing here

emteknetnz:
~Updated description, this PR is just a POC~

GuySartorelli:
Note that this is no longer just a POC, despite all the descriptions saying it is. See https://github.com/silverstripe/silverstripe-framework/issues/11475.
I'll update the PR description as well.

client/src/legacy/ConfirmedPasswordField.js
```diff
@@ -12,3 +13,61 @@ $(document).on('click', '.confirmedpassword .showOnClick a', function () {
...
+  strengthEl.innerHTML = `<p class="alert alert-${level}" role="alert">${message}</p>`;
+}
+
+const debouncedEvaulatePassword = debounce((event) => evaulatePassword(event), 500, {
+  'leading': true,
+  'trailing': true,
+});
+
+// Use event delegation so that this still works as the user navigates around using pjax
+document.addEventListener('input', (event) => {
```
ChloeSartorelli, line 78: Is there a reason we're using raw JS instead of jQuery or entwine?
I'm not bothered just wondering since it's not the style our JS tends to use.

client/src/legacy/ConfirmedPasswordField.js
```diff
@@ -12,3 +13,61 @@ $(document).on('click', '.confirmedpassword .showOnClick a', function () {
...
+  if (!strengthEl) {
+    return;
+  }
+  const response = await fetch(url, {
+    method: 'POST',
+    body: JSON.stringify({
+      password: passwordEl.value,
+    }),
+  });
+  const json = await response.json();
```
ChloeSartorelli, line 54: If there's an error (either non-200 response or unexpected body), we should clear any existing message in `strengthEl`

GuySartorelli (changes requested):
When typing, it's a bit jarring that the message can skip over strength types, which is a result of the debounce (e.g. I might see it go from "low" straight to "very strong" - I keep typing waiting for it to tell my my password is stronger, and get surprised that it isn't strong enough yet.... until I stop typing)

Can we add some indicator that it's updating while typing (e.g. a spinner)? Might be something to quickly buzz past the UX team?

Note that any JS changes here should also be reflected in login-forms

client/src/legacy/ConfirmedPasswordField.js
emteknetnz: So that it also works on silverstripe/login-forms (without needing to add jquery to package.json)

emteknetnz:
Looks like I didn't have the debounce configured correctly which is why it was so jarring. I've added additional config `'maxWait': 300,` as well as just changing the debounce from 500 to 300.

300 seems to be a common value online for debounce it's related to human reaction time.  There will be a server lag which will increase this time, though I tried throttling the response in the network panel to GPRS, which kind of simulates a slow server, and the experience actually still seemed perfectly fine.  I don't think we need a loading indicator in its current state (which would probably increase complexity quite a lot)

client/src/legacy/ConfirmedPasswordField.js
emteknetnz: Done

client/src/legacy/ConfirmedPasswordField.js
ChloeSartorelli: This isn't the login-forms module, though?
It's probably fine, it's just a bit weird.

GuySartorelli (approved):
LGTM, will merge after the framework PR

PR #1833, opened 2024-10-07
ENH Update code to reflect changes in template layer

GuySartorelli:
Relies and reflects on changes in https://github.com/silverstripe/silverstripe-framework/pull/11405

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11322

### DIFF
--- a/code/Forms/GridFieldDetailFormPreviewExtension.php
+++ b/code/Forms/GridFieldDetailFormPreviewExtension.php
@@ -35,7 +35,6 @@ class GridFieldDetailFormPreviewExtension extends Extension
     {
         $navigator = SilverStripeNavigator::create($this->owner->getRecord());
         $templates = SSViewer::get_templates_by_class(LeftAndMain::class, '_SilverStripeNavigator', LeftAndMain::class);
-        $renderWith = SSViewer::chooseTemplate($templates);
-        return $navigator->renderWith($renderWith);
+        return $navigator->renderWith($templates);
     }
 }
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -901,16 +901,12 @@ class LeftAndMain extends AdminController implements PermissionProvider
     }
 
     /**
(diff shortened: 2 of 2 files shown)


code/Forms/GridFieldDetailFormPreviewExtension.php
```diff
@@ -35,7 +35,6 @@ private function getSilverStripeNavigator(): DBHTMLText
     {
         $navigator = SilverStripeNavigator::create($this->owner->getRecord());
         $templates = SSViewer::get_templates_by_class(LeftAndMain::class, '_SilverStripeNavigator', LeftAndMain::class);
-        $renderWith = SSViewer::chooseTemplate($templates);
-        return $navigator->renderWith($renderWith);
+        return $navigator->renderWith($templates);
```
ChloeSartorelli, line 38: This gives the _exact_ same result. There was no need for the `chooseTemplate()` there inbetween (which is good 'cause we can't do that anymore)

code/LeftAndMain.php
```diff
@@ -901,16 +901,12 @@ public function MenuCurrentItem()
...
-     * @param string $suffix
-     * @return string|array
      */
-    public function getTemplatesWithSuffix($suffix)
+    public function getTemplatesWithSuffix(string $suffix): array
     {
-        $templates = SSViewer::get_templates_by_class(get_class($this), $suffix, __CLASS__);
-        return SSViewer::chooseTemplate($templates);
+        return SSViewer::get_templates_by_class(get_class($this), $suffix, __CLASS__);
     }
```
ChloeSartorelli, line 910: This used to give specific absolute file paths to templates.
We can't do that anymore, because a) some template engines might not support that, and b) we don't want to pass a .ss template path to a twig template engine for example.

Thankfully, everywhere that used this doesn't actually _need_ full absolute file paths. They just need some template candidates, and the ability to ask the configured engine if it can find files from those candidates or not.

emteknetnz (approved):

Issue #1761, opened 2024-05-30
Refactor `admin/*` routing logic from `LeftAndMain` into its own abstract class


GuySartorelli:
We often subclass `LeftAndMain` just to get controllers routed in the admin section, and then hide them because they're not actually left and main sections.

We should instead have an abstract controller class we can use as the superclass for those which doesn't include all of the complexity of `LeftAndMain`.

## Related
- https://github.com/silverstripe/silverstripe-cms/issues/2933

## Notes
- The exact name of the class isn't important - if you think of a better name, feel free.
  - I'm going with `AdminController` for now, as it's less likely to be confused with `AdminRootController`.

## Acceptance criteria
- [x] A new abstract `AdminRouteController` class is created
- [x] `AdminRootController` checks for (non-abstract) subclasses of the new class instead of subclasses of `LeftAndMain` to set up routing rules
- [x] Methods, properties, and configuration required for routing or permissions are moved from `LeftAndMain` to the new class (see [this comment](https://github.com/silverstripe/silverstripe-cms/issues/2933#issuecomment-2111557980) for recommendations about this)
- [x] `LeftAndMain` subclasses the new class
- [x] Any controller in supported modules that currently subclasses `LeftAndMain` but only needs the routing and permissions logic is made to subclass the new class instead.
- [x] Any controller in supported modules that uses an admin route but which doesn't currently subclass `LeftAndMain` is made a subclass of the new class instead.
- [x] Documentation is updated to indicate best practice for controllers using a `/admin/` route

## PRs

**[kitchen sink run](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/11393064475)**
Compare with [CI run in silverstripe/recipe-kitchen-sink](https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/11318774193)

- https://github.com/silverstripe/silverstripe-admin/pull/1836
- https://github.com/silverstripe/silverstripe-externallinks/pull/151
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1502
- https://github.com/silverstripe/silverstripe-elemental/pull/1262
- https://github.com/tractorcow-farm/silverstripe-fluent/pull/896
- https://github.com/silverstripe/silverstripe-frameworktest/pull/207
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/335
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/366
[...]

PR #1836, opened 2024-10-14
NEW AdminController as superclass for /admin/* routed controllers

GuySartorelli:
There are two commits here:

1. Mostly-unrelated refactor to use config instead of runtime code to remove the `CMSProfileController` menu item
2. Add the new `AdminController` class and wire it up

I've also added a bunch of strong typing to the new class and to `AdminRouteController`.
I have _not_ added strong typing to `LeftAndMain` though that may happen as part of a follow-up card in this epic.
I have _not_ added strong typing to methods which would required extensive trickle-down effects (e.g. `init()` and `canView()`).

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1761

### DIFF
--- a/_config.php
+++ b/_config.php
@@ -31,5 +31,3 @@ $editorConfig
     ]);
 // enable ability to insert anchors
 $editorConfig->insertButtonsAfter('sslink', 'anchor');
-
-CMSMenu::remove_menu_class(CMSProfileController::class);
--- a/code/AdminController.php
+++ b/code/AdminController.php
@@ -1,0 +1,359 @@
+<?php
+
+namespace SilverStripe\Admin;
+
+use BadMethodCallException;
+use InvalidArgumentException;
+use SilverStripe\Control\ContentNegotiator;
+use SilverStripe\Control\Controller;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Control\HTTPResponse_Exception;
+use SilverStripe\Control\Middleware\HTTPCacheControlMiddleware;
+use SilverStripe\Forms\HTMLEditor\HTMLEditorConfig;
+use SilverStripe\i18n\i18n;
+use SilverStripe\Security\Permission;
+use SilverStripe\Security\Security;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\View\SSViewer;
+
+/**
+ * The base class for all controllers routed using the /admin/* route.
+ *
+ * This class is automatically routed via the AdminRootController.
+ * It's responsible for ensuring permissions are respected.
+ */
+abstract class AdminController extends Controller
+{
+    /**
+     * The current url segment attached to the controller
+     */
+    private static ?string $url_segment = null;
+
+    /**
+     * Used by {@link AdminRootController} to augment Director route rules for subclasses of AdminController
+     */
+    private static string $url_rule = '/$Action/$ID/$OtherID';
+
+    /**
+     * Priority order for routing rules. If two controllers match a given request, the one with a higher
+     * priority will handle the request.
+     */
+    private static int $url_priority = 50;
+
+    /**
+     * Codes which are required from the current user to view this controller.
+     *
+     * If multiple codes are provided, all of them are required.
+     * All CMS controllers require "CMS_ACCESS_LeftAndMain" as a baseline check,
+     * and fall back to "CMS_ACCESS_<class>" if no permissions are defined here.
+     * See {@link canView()} for more details on permission checks.
+     */
+    private static string|array $required_permission_codes = [];
+
+    /**
+     * The configuration passed to the supporting JS for each CMS section includes a 'name' key
+     * that by default matches the FQCN of the current class. This setting allows you to change
+     * the key if necessary (for example, if you are overloading CMSMain or another core class
+     * and want to keep the core JS - which depends on the core class names - functioning, you
+     * would need to set this to the FQCN of the class you are overloading).
+     *
+     * See getClientConfig()
+     */
+    private static ?string $section_name = null;
+
+    /**
+     * Get list of required permissions for accessing this controller.
+     * If false, no permission is required.
+     */
+    public static function getRequiredPermissions(): array|string|false
+    {
+        if (static::class === AdminController::class) {
+            throw new BadMethodCallException('getRequiredPermissions should be called on a subclass');
+        }
+        // If the user is accessing LeftAndMain directly, only generic permissions are required.
+        if (static::class === LeftAndMain::class) {
+            return 'CMS_ACCESS';
+        }
+        $codes = static::config()->get('required_permission_codes');
+        // allow explicit FALSE to disable subclass check
+        if ($codes === false) {
+            return false;
+        }
+        if ($codes) {
+            return $codes;
+        }
+        // Fallback if no explicit permission was declared
+        return 'CMS_ACCESS_' . static::class;
+    }
+
+    public function canView($member = null)
+    {
+        if (!$member && $member !== false) {
+            $member = Security::getCurrentUser();
+        }
+
+        // don't allow unauthenticated users
+        if (!$member) {
+            return false;
+        }
+
+        // alternative extended checks
+        if ($this->hasMethod('alternateAccessCheck')) {
+            $alternateAllowed = $this->alternateAccessCheck($member);
+            if ($alternateAllowed === false) {
+                return false;
+            }
+        }
+
+        // Check for "Access to all CMS sections" permission
+        if (Permission::checkMember($member, 'CMS_ACCESS_LeftAndMain')) {
+            return true;
+        }
+
+        // Check for permission to access this specific controller
+        $codes = static::getRequiredPermissions();
+        // allow explicit FALSE to disable subclass check
+        if ($codes === false) {
+            return true;
+        }
+        foreach ((array) $codes as $code) {
+            if (!Permission::checkMember($member, $code)) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    public function Link($action = null)
+    {
+        // LeftAndMain methods have a top-level uri access
+        if (static::class === LeftAndMain::class) {
+            $segment = '';
+        } else {
+            // Get url_segment
+            $segment = static::config()->get('url_segment');
+            if (!$segment) {
+                throw new BadMethodCallException(
+                    sprintf('AdminController subclasses (%s) must have url_segment', static::class)
+                );
+            }
+        }
+
+        $link = Controller::join_links(
+            AdminRootController::admin_url(),
+            $segment,
+            "$action"
+        );
+        $this->extend('updateLink', $link);
+        return $link;
+    }
+
+    /**
+     * Overloaded redirection logic to trigger a fake redirect on ajax requests.
+     * While this violates HTTP principles, its the only way to work around the
+     * fact that browsers handle HTTP redirects opaquely, no intervention via JS is possible.
+     * In isolation, that's not a problem - but combined with history.pushState()
+     * it means we would request the same redirection URL twice if we want to update the URL as well.
+     * See LeftAndMain.js for the required jQuery ajaxComplete handlers.
+     */
+    public function redirect(string $url, int $code = 302): HTTPResponse
+    {
+        if ($this->getRequest()->isAjax()) {
+            $response = $this->getResponse();
+            $response->addHeader('X-ControllerURL', $url);
+            if ($this->getRequest()->getHeader('X-Pjax') && !$response->getHeader('X-Pjax')) {
+                $response->addHeader('X-Pjax', $this->getRequest()->getHeader('X-Pjax'));
+            }
+            $newResponse = new LeftAndMain_HTTPResponse(
+                $response->getBody(),
+                $response->getStatusCode(),
+                $response->getStatusDescription()
+            );
+            foreach ($response->getHeaders() as $k => $v) {
+                $newResponse->addHeader($k, $v);
+            }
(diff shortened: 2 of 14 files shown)


_config.php
```diff
@@ -31,5 +31,3 @@
     ]);
 // enable ability to insert anchors
 $editorConfig->insertButtonsAfter('sslink', 'anchor');
-
-CMSMenu::remove_menu_class(CMSProfileController::class);
```
ChloeSartorelli, line 35: That's handled by config in the class itself now.

code/AdminController.php
ChloeSartorelli, line 1: Basically everything in here was just lifted straight from `LeftAndMain` - though some small code quality improvements were made here-and-there.

code/AdminController.php
```diff
@@ -0,0 +1,359 @@
...
+
+    /**
+     * Get list of required permissions for accessing this controller.
+     * If false, no permission is required.
+     */
+    public static function getRequiredPermissions(): array|string|false
+    {
+        if (static::class === AdminController::class) {
+            throw new BadMethodCallException('getRequiredPermissions should be called on a subclass');
+        }
```
ChloeSartorelli, line 73: Just to avoid it being called here, since this is explicitly abstract there's no permission for this class (or reason to want one)

code/AdminController.php
```diff
@@ -0,0 +1,359 @@
...
+
+        return true;
+    }
+
+    public function Link($action = null)
+    {
+        // LeftAndMain methods have a top-level uri access
+        if (static::class === LeftAndMain::class) {
+            $segment = '';
```
ChloeSartorelli, line 133: A follow-up card will make `LeftAndMain` abstract after removing the methods on it that are used from singleton - but for now this has to stay.

code/AdminController.php
```diff
@@ -0,0 +1,359 @@
...
+                $newResponse->addHeader($k, $v);
+            }
+            $newResponse->setIsFinished(true);
+            $this->setResponse($newResponse);
+            // Actual response will be re-requested by client
+            return $newResponse;
+        } else {
+            return parent::redirect($url, $code);
+        }
+    }
```
ChloeSartorelli, line 184: I'm not 100% sure exactly what all of the ramifications of this are - but since it's dealing explicitly with AJAX requests and that's mostly what this method is for, I decided to lift this out of LeftAndMain and into this class.

code/AdminController.php
```diff
@@ -0,0 +1,359 @@
...
+        // and use in routing definitions.
+        $url = trim($this->Link(), '/');
+        $clientConfig = [
+            'name' => $name,
+            'url' => $url,
+            'reactRoutePath' => preg_replace('/^' . preg_quote(AdminRootController::admin_url(), '/') . '/', '', $url),
+        ];
+        $this->extend('updateClientConfig', $clientConfig);
+        return $clientConfig;
+    }
```
ChloeSartorelli, line 203: This is lifted over here so that we can inject links for these controllers into the config that JS gets a hold of.

code/AdminController.php
```diff
@@ -0,0 +1,359 @@
...
+        // set language
+        $member = Security::getCurrentUser();
+        if (!empty($member->Locale)) {
+            i18n::set_locale($member->Locale);
+        }
+
+        // Don't allow access if the request hasn't finished being handled and the user can't access this controller
+        if (!$this->canView() && !$this->getResponse()->isFinished()) {
+            // Allow subclasses and extensions to redirect somewhere more appropriate
+            $this->invokeWithExtensions('onInitPermissionFailure');
```
ChloeSartorelli, line 223: This lets LeftAndMain do some redirecting it was doing before without making the method abstract (and therefore required on classes that don't need it) and without doing a `method_exists()` check which always feels like the wrong move.

code/AdminController.php
```diff
@@ -0,0 +1,359 @@
...
+                ),
+                'alreadyLoggedIn' => _t(
+                    __CLASS__ . '.PERMALREADY',
+                    "I'm sorry, but you can't access that part of the CMS."
+                ),
+                'logInAgain' => _t(
+                    __CLASS__ . '.PERMAGAIN',
+                    "You have been logged out of the CMS."
+                ),
+            ];
```
ChloeSartorelli, line 248: Shortened these messages to suit the fact that most of the time these controllers will be handling AJAX requests and therefore there won't be a menu "below" like the old messages suggest.

I considered finding a way to use the longer messages in LeftAndMain specifically, but the context of "there's a form below" is unnecessary because... well, there's a form below. You can see it (or tab to it if you're on a screen reader)

code/AdminRootController.php
```diff
@@ -10,24 +10,41 @@
...
     /**
      * Fallback admin URL in case this cannot be infered from Director.rules
+     */
+    private static string $url_base = 'admin';
+
+    /**
+     * The LeftAndMain child that will be used as the initial panel to display if none is selected (i.e. if you
+     * visit /admin)
+     */
+    private static string $default_panel = SecurityAdmin::class;
```
ChloeSartorelli, line 30: These are just moved from below per our coding standards (i.e. properties before methods)

code/AdminRootController.php
```diff
@@ -36,55 +53,31 @@ public static function get_admin_route()
...
      */
-    public static function rules()
+    public static function rules(): array
     {
         if (AdminRootController::$adminRules === null) {
             AdminRootController::$adminRules = [];
 
             // Map over the array calling add_rule_for_controller on each
-            $classes = CMSMenu::get_cms_classes(null, true, CMSMenu::URL_PRIORITY);
+            $classes = CMSMenu::get_cms_classes(AdminController::class, true, CMSMenu::URL_PRIORITY);
```
ChloeSartorelli, line 71: `null` defaults to `LeftAndMain` - we want the new controller as the base class.

code/AdminRootController.php
```diff
@@ -118,7 +111,7 @@ public function handleRequest(HTTPRequest $request): HTTPResponse
         // Otherwise
         $rules = AdminRootController::rules();
         foreach ($rules as $pattern => $controller) {
-            if (($arguments = $request->match($pattern, true)) !== false) {
+            if ($request->match($pattern, true) !== false) {
```
ChloeSartorelli, line 114: Unrelated refactoring since that variable is never used.

code/CMSProfileController.php
```diff
@@ -22,6 +22,8 @@ class CMSProfileController extends LeftAndMain
 
     private static $tree_class = Member::class;
 
+    private static $ignore_menuitem = true;
```
ChloeSartorelli, line 25: This is the config that lets us remove the code from `_config.php`

code/LeftAndMain.php
```diff
@@ -281,16 +244,35 @@ class LeftAndMain extends Controller implements PermissionProvider
...
+     */
+    private static $application_link = '//www.silverstripe.org/';
+
+    /**
+     * The application name
+     *
+     * @config
+     * @var string
+     */
+    private static $application_name = 'Silverstripe';
```
ChloeSartorelli: Moved from below, along with `searchFilterCache` as well. These were below some methods.

code/LeftAndMain.php
```diff
@@ -312,7 +301,7 @@ class LeftAndMain extends Controller implements PermissionProvider
     public function getCombinedClientConfig()
     {
         $combinedClientConfig = ['sections' => []];
-        $cmsClassNames = CMSMenu::get_cms_classes(LeftAndMain::class, true, CMSMenu::URL_PRIORITY);
+        $cmsClassNames = CMSMenu::get_cms_classes(AdminController::class, true, CMSMenu::URL_PRIORITY);
```
ChloeSartorelli, line 304: The config itself can be on any `AdminController` now.

code/SudoModeController.php
```diff
@@ -91,7 +89,7 @@ public function activate(HTTPRequest $request): HTTPResponse
             return $this->jsonResponse([
                 'result' => false,
                 'message' => _t(__CLASS__ . '.INVALID', 'Incorrect password'),
-            ]);
+            ], 400);
```
ChloeSartorelli: Unrelated refactoring - this is an error, so should have an error response code!  There's no context in the original PR that created this class for why it was a 200. It looks like it was an oversight.

tests/php/AdminControllerTest.php
```diff
@@ -0,0 +1,236 @@
...
+        foreach ($allowed as $allowedClass) {
+            $this->assertTrue($allowedClass::singleton()->canView(), 'Tried to access ' . $allowedClass);
+        }
+        $disallowed = array_diff($allControllers, $allowed);
+        foreach ($disallowed as $disallowedClass) {
+            $this->assertFalse($disallowedClass::singleton()->canView(), 'Tried to access ' . $disallowedClass);
+        }
+    }
+
+    public static function provideJsonSuccess(): array
```
ChloeSartorelli, line 98: This and everything below it was pulled from `LeftAndMainTest`.

tests/php/LeftAndMainTest.php
```diff
@@ -80,8 +74,7 @@ protected function tearDown(): void
 
     public function testExtraCssAndJavascript()
     {
-        $admin = $this->objFromFixture(Member::class, 'admin');
-        $this->logInAs($admin);
+        $this->logInWithPermission('ADMIN');
```
ChloeSartorelli, line 77: The admin fixture and the `logInWithPermission('ADMIN')` user have the same permissions, so no need for the fixture.

tests/php/LeftAndMainTest.php
```diff
@@ -135,65 +128,6 @@ public function testLeftAndMainSubclasses()
         $this->assertMatchesRegularExpression('/<body[^>]*>/i', $response->getBody(), "$link should contain <body> tag");
     }
 
-    public function testCanView()
```
ChloeSartorelli, line 138: This was testing the wrong thing - it was trying to test `canView()` checks by seeing what's in the menu. But the menu actually _shouldn't_ contain `CMSProfileController` so this test was passing when it should have failed.

Moved canView check to the new test class, and now actually testing canView instead of the menu.

The menu itself is already tested in this class.

code/SudoModeController.php
emteknetnz: This should be 401, context for this is getting password wrong when trying to activate sudo mode.

https://stackoverflow.com/a/32752617

code/AdminRootController.php
```diff
@@ -10,24 +10,41 @@
...
+
+    /**
+     * The LeftAndMain child that will be used as the initial panel to display if none is selected (i.e. if you
+     * visit /admin)
+     */
+    private static string $default_panel = SecurityAdmin::class;
+
+    /**
+     * Holds an array of url_pattern => controller k/v pairs, the same as Director::rules. However this is built
+     * dynamically from introspecting on all the classes that derive from LeftAndMain.
```
emteknetnz:
```suggestion
     * dynamically from introspecting on all the classes that derive from AdminController.
```

code/AdminController.php
```diff
@@ -0,0 +1,359 @@
...
+            Security::permissionFailure($this, $messageSet);
+            return;
+        }
+
+        // Don't continue if there's already been a redirection request.
+        if ($this->getResponse()->isRedirect()) {
+            return;
+        }
+
+        $this->extend('onInit');
```
emteknetnz, line 260: Weird place for an 'onInit' hook - seems like it should be at the bottom?  Any reason why it's here? Does the implementation in LeftAndMain not work right if this hook is at the bottom?

code/LeftAndMain.php
```diff
@@ -281,16 +244,35 @@ class LeftAndMain extends Controller implements PermissionProvider
...
+     * Set this to `false` via yml config if you do not want to show any help links
      *
      * @config
-     * @var string|null
+     * @var array
      */
-    private static $section_name = null;
+    private static $help_links = [
+        'CMS User help' => 'https://userhelp.silverstripe.org/en/5',
+        'Developer docs' => 'https://docs.silverstripe.org/en/5/',
```
emteknetnz:
```suggestion
        'CMS User help' => 'https://userhelp.silverstripe.org/en',
        'Developer docs' => 'https://docs.silverstripe.org/en/',
```

emteknetnz: Or /6/

emteknetnz (changes requested):

code/SudoModeController.php
ChloeSartorelli: I hate HTTP response codes lol.
Updated to 401

code/AdminRootController.php
ChloeSartorelli: Done

code/AdminController.php
ChloeSartorelli: I'm just keeping it in the same logic place it was with `LeftAndMain`. I don't know if it would break anything if we move it, and I don't particularly want to find out. Chances are if it breaks anything to move it, we won't find out for a year or so.
I'm _guessing_ it's intentionally placed and that it's specifically not called `onAfterInit()` for that reason, though I can't confirm that.

Safer to just leave it where it is. But I can move it if you feel strongly.

code/LeftAndMain.php
ChloeSartorelli: We'll need to update a bunch of references like this, we should probably just do them all together. But I'll make this change to avoid ping pong.
Going with an explicit `6` so people don't get docs for `7` in their `6` sites when we get to that point.

code/AdminController.php
emteknetnz: Oh right, hook was called 'init' in CMS 5, sorry didn't see that it was existing. Leave as is

emteknetnz (approved):

PR #1837, opened 2024-10-16
API Stop refering to "pages" when dealing with generic records

GuySartorelli:
Two commits here:

1. Updates some i18n keys that should live in `LeftAndMain`. Found this because I had to review all i18n strings in `CMSMain` anyway.
1. Updates some API that currently refers to "pages" when it's really dealing with any `DataObject` class, and adds a new `getModelClass()` method to `LeftAndMain` (standardises that method between `ModelAdmin` and `CMSMain` and will be used extensively in the follow-up refactor cards in this epic)

## Issue


code/SecurityAdmin.php
```diff
@@ -141,11 +141,11 @@ public function providePermissions()
         return [
             "CMS_ACCESS_SecurityAdmin" => [
                 'name' => _t(
-                    CMSMain::class . '.ACCESS',
+                    LeftAndMain::class . '.ACCESS',
```
ChloeSartorelli, line 144: CMSMain doesn't even exist if you're installing silverstripe/admin without silverstripe/cms. `LeftAndMain` is the obvious choice here.

code/SecurityAdmin.php
```diff
@@ -141,11 +141,11 @@ public function providePermissions()
...
                     "Access to '{title}' section",
                     ['title' => $title]
                 ),
-                'category' => _t(Permission::class . '.CMS_ACCESS_CATEGORY', 'CMS Access'),
+                'category' => _t(LeftAndMain::class . '.CMS_ACCESS_CATEGORY', 'CMS Access'),
```
ChloeSartorelli, line 148: This belongs on `LeftAndMain` as it goes hand-in-hand with the one above.

code/CMSMenu.php
```diff
@@ -379,7 +379,7 @@ public static function get_cms_classes($root = null, $recursive = true, $sort =
         if (!$root) {
             $root = LeftAndMain::class;
         }
-        $abstractClasses = [LeftAndMain::class, CMSMain::class];
+        $abstractClasses = [LeftAndMain::class];
```
ChloeSartorelli, line 382: This change is needed so that `CMSMain` can be included in the left menu and have its route built.
Without this, it was the `CMSPagesController` which was a very-nearly-useless subclass of `CMSMain` that was used.
That class has been merged into `CMSMain` so we need its menu item back

GuySartorelli:
Replaced with https://github.com/silverstripe/silverstripe-admin/pull/1867

PR #1838, opened 2024-10-17
MNT Update behat tests

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/320

Update password used in behat tests because we increased password validator required strength when switching to symfony


GuySartorelli (approved):
LGTM

Issue #1764, opened 2024-05-30
Refactor logic for sections that deal with only a single record out of `LeftAndMain` into its own class


GuySartorelli:
Most subclasses of `LeftAndMain` deal with more than one record at a time - be it `ModelAdmin` which handles multiple records in a `GridField`, or `CMSMain` which deals with multiple records in a hierarchical structure.

The logic in `LeftAndMain` which is intended for sections that only edit a single record should be abstracted out into its own abstract class.

## Related
- https://github.com/silverstripe/silverstripe-cms/issues/2933

## Notes
- The exact name of the class isn't important - if you think of a better name, feel free.
- Refer to https://github.com/silverstripe/silverstripe-cms/issues/2933#issuecomment-2111603655 for recommendations on some implementation details

## Acceptance criteria
- [x] A new abstract `SingleRecordAdmin` class is created, which is a subclass of `LeftAndMain`
- ~~`AdminRootController` checks for (non-abstract) subclasses of the new class instead of subclasses of `LeftAndMain` to set up routing rules~~ - already handled by https://github.com/silverstripe/silverstripe-admin/issues/1761
- [x] Methods, properties, and configuration used only for editing single records are moved from `LeftAndMain` to the new class
- [x] `CMSProfileController`, `SiteConfigLeftAndMain`, and any other controllers in supported modules which only edit a single record are made subclasses of this new class
- [x] Functionality that's reimplemented in `CMSProfileController`, `SiteConfigLeftAndMain`, etc are made generic and are implemented in the new class directly

### CMS 5 PRs

- https://github.com/silverstripe/silverstripe-admin/pull/1844
- https://github.com/silverstripe/silverstripe-siteconfig/pull/181
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/339
- https://github.com/silverstripe/silverstripe-reports/pull/201
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1506
- https://github.com/silverstripe/silverstripe-cms/pull/3021
- https://github.com/silverstripe/developer-docs/pull/616

### CMS 6 PRs

**[kitchen sink ci](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/11695642970)**

- https://github.com/silverstripe/silverstripe-admin/pull/1842
- https://github.com/silverstripe/silverstripe-siteconfig/pull/179
- https://github.com/silverstripe/silverstripe-cms/pull/3019
[...]

Issue #1763, opened 2024-05-30
Make `LeftAndMain` abstract and other small refactoring in this class


GuySartorelli:
There's no good reason to instantiate `LeftAndMain`. It should be abstract.

## Related
- https://github.com/silverstripe/silverstripe-cms/issues/2933

## Acceptance criteria
- `LeftAndMain` is made abstract
- Anything that requires instantiating `LeftAndMain` is refactored to use a concrete class instead.
- A new abstract `getModelClass()` method is created
- The `tree_class` configuration property is removed - anything that used this configuration is updated to call `getModelClass()` instead.
- Methods which used that config are updated according to the "`tree_class` config" section in [this comment](https://github.com/silverstripe/silverstripe-cms/issues/2933#issuecomment-2111557980) unless doing so causes unforeseen problems
- Recommendations in the "Additional recommendations" section in [this comment](https://github.com/silverstripe/silverstripe-cms/issues/2933#issuecomment-2111557980) are applied unless doing so causes unforeseen problems

PR #1842, opened 2024-10-25
NEW SingleRecordAdmin class for editing one record at a time

GuySartorelli:
Note that there may be some methods still on `LeftAndMain` that will be migrated into this class, such as `save()` which isn't used by `CMSMain` or `ModelAdmin` - but moving those will be handled by https://github.com/silverstripe/silverstripe-admin/issues/1763 once the bulk of the other refactoring has been completed. This is to reduce the chance of regressions in areas that still need to be refactored.

## TODO

1. CMS 5 PR to deprecate `SiteConfigLeftAndMain::save_siteconfig()` and `LeftAndMain.tree_config`

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1764

### DIFF
--- a/code/CMSProfileController.php
+++ b/code/CMSProfileController.php
@@ -2,17 +2,13 @@
 
 namespace SilverStripe\Admin;
 
-use SilverStripe\CMS\Controllers\CMSMain;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Forms\Form;
 use SilverStripe\Forms\HiddenField;
-use SilverStripe\Forms\FormAction;
-use SilverStripe\Model\List\ArrayList;
 use SilverStripe\Security\Member;
-use SilverStripe\Security\Permission;
 use SilverStripe\Security\Security;
 
-class CMSProfileController extends LeftAndMain
+class CMSProfileController extends SingleRecordAdmin
 {
     private static $url_segment = 'myprofile';
 
@@ -20,98 +16,61 @@ class CMSProfileController extends LeftAndMain
 
     private static $required_permission_codes = 'CMS_ACCESS';
 
-    /**
-     * @deprecated 5.4.0 Will be renamed to model_class
-     */
-    private static $tree_class = Member::class;
+    private static $model_class = Member::class;
 
     private static $ignore_menuitem = true;
 
-    public function getEditForm($id = null, $fields = null)
+    private static bool $restrict_to_single_record = false;
+
+    private static bool $allow_new_record = false;
+
+    protected function init()
     {
-        $this->setCurrentPageID(Security::getCurrentUser()->ID);
+        parent::init();
+        if (!$this->getResponse()->isRedirect()) {
+            $this->setCurrentPageID(Security::getCurrentUser()->ID);
+        }
+    }
 
+    public function getEditForm($id = null, $fields = null): Form
+    {
         $form = parent::getEditForm($id, $fields);
-
-        if ($form instanceof HTTPResponse) {
-            return $form;
-        }
-
-        $form->Fields()->removeByName('LastVisited');
-        $form->Fields()->push(new HiddenField('ID', null, Security::getCurrentUser()->ID));
-        $form->Actions()->push(
-            FormAction::create('save', _t(CMSMain::class . '.SAVE', 'Save'))
-                ->addExtraClass('btn-primary font-icon-save')
-                ->setUseButtonTag(true)
-        );
-
-        $form->Actions()->removeByName('action_delete');
-
-        if ($member = Security::getCurrentUser()) {
-            $form->setValidator($member->getValidator());
-        } else {
-            $form->setValidator(Member::singleton()->getValidator());
-        }
-
-        if ($form->Fields()->hasTabSet()) {
-            $form->Fields()->findOrMakeTab('Root')->setTemplate('SilverStripe\\Forms\\CMSTabSet');
-        }
-
-        $form->addExtraClass('member-profile-form root-form cms-edit-form center fill-height');
-
+        $form->Fields()->push(HiddenField::create('ID', null, Security::getCurrentUser()->ID));
         return $form;
     }
 
     public function canView($member = null)
     {
         $currentUser = Security::getCurrentUser();
-
         if (!$member && $member !== false) {
             $member = $currentUser;
         }
-
-        // cms menus only for logged-in members
+        // cms menus are only for logged-in members
         if (!$member) {
             return false;
         }
-
         // Check they are trying to edit themselves and they have permissions
         return $member->ID === $currentUser->ID && parent::canView($member);
     }
 
     public function save(array $data, Form $form): HTTPResponse
     {
-        $member = Member::get()->byID($data['ID']);
-        if (!$member) {
-            $this->httpError(404);
+        // Make sure the ID is a positive number
+        $id = $data['ID'] ?? null;
+        if ((!is_int($id) && !ctype_digit($id)) || $id < 1) {
+            $this->httpError(400);
         }
-        $origLocale = $member->Locale;
-
-        if (!$member->canEdit()) {
-            $form->sessionMessage(_t(__CLASS__.'.CANTEDIT', 'You don\'t have permission to do that'), 'bad');
-            return $this->redirectBack();
-        }
-
+        // Get the current member locale so we can see if it changes
+        $member = Member::get()->byID($id);
+        $origLocale = $member?->Locale;
+        // actual save (along with error handling e.g. if there's no member with that ID) is handled by the parent class
         $response = parent::save($data, $form);
-
+        // If the member locale changes, reload the full CMS so localisation can kick in.
+        // Otherwise only the edit form will reload.
         if (isset($data['Locale']) && $origLocale != $data['Locale']) {
             $response->addHeader('X-Reload', true);
             $response->addHeader('X-ControllerURL', $this->Link());
         }
-
         return $response;
-    }
-
-    /**
-     * Only show first element, as the profile form is limited to editing
-     * the current member it doesn't make much sense to show the member name
-     * in the breadcrumbs.
-     *
-     * @param bool $unlinked
-     */
-    public function Breadcrumbs($unlinked = false)
-    {
-        $items = parent::Breadcrumbs($unlinked);
-        return new ArrayList([$items[0]]);
     }
 }
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -100,7 +100,7 @@ class LeftAndMain extends AdminController implements PermissionProvider
      * @var string
      * @deprecated 5.4.0 Will be renamed to model_class
      */
-    private static $tree_class = null;
+    private static $model_class = null;
 
     /**
      * @var array
@@ -939,7 +939,7 @@ class LeftAndMain extends AdminController implements PermissionProvider
      */
     public function getRecord($id)
     {
-        $className = $this->config()->get('tree_class');
+        $className = $this->config()->get('model_class');
         if (!$className) {
             return null;
         }
@@ -1005,7 +1005,7 @@ class LeftAndMain extends AdminController implements PermissionProvider
     public function save(array $data, Form $form): HTTPResponse
     {
         $request = $this->getRequest();
-        $className = $this->config()->get('tree_class');
+        $className = $this->config()->get('model_class');
 
         // Existing or new record?
         $id = $data['ID'];
@@ -1018,7 +1018,7 @@ class LeftAndMain extends AdminController implements PermissionProvider
                 $this->httpError(404, "Bad record ID #" . (int)$id);
             }
         } else {
-            if (!singleton($this->config()->get('tree_class'))->canCreate()) {
+            if (!singleton($this->config()->get('model_class'))->canCreate()) {
                 return Security::permissionFailure($this);
             }
             $record = $this->getNewItem($id, false);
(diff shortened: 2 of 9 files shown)


code/CMSProfileController.php
```diff
@@ -2,113 +2,75 @@
...
+class CMSProfileController extends SingleRecordAdmin
 {
     private static $url_segment = 'myprofile';
 
     private static $menu_title = 'My Profile';
 
     private static $required_permission_codes = 'CMS_ACCESS';
 
-    private static $tree_class = Member::class;
+    private static $model_class = Member::class;
```
ChloeSartorelli, line 19: See change in LeftAndMain

code/LeftAndMain.php
```diff
@@ -99,7 +99,7 @@ class LeftAndMain extends AdminController implements PermissionProvider
      * @config
      * @var string
      */
-    private static $tree_class = null;
+    private static $model_class = null;
```
ChloeSartorelli, line 103: Most subclasses of `LeftAndMain` which use this config are _not_ trees - and even with `CMSMain` which is the only tree structure, this is getting refactored to be more generic anyway.

Makes sense to update this to be named something that _all_ the classes using it can respect.

code/LeftAndMain.php
```diff
@@ -1223,24 +1223,7 @@ public function getEditForm($id = null, $fields = null)
...
-        } elseif ($record->hasMethod('getCMSValidator')) {
-            // BC support for framework < v4.7
-            $validator = $record->getCMSValidator();
-
-            // The clientside (mainly LeftAndMain*.js) rely on ajax responses
-            // which can be evaluated as javascript, hence we need
-            // to override any global changes to the validation handler.
-            if ($validator) {
-                $form->setValidator($validator);
-            }
```
ChloeSartorelli, line 1238: All of this boils down to `$form->setValidator($record->getCMSCompositeValidator())`, because all `DataObject` classes have that method, and `$record` will always be a `DataObject`.

code/LeftAndMain.php
```diff
@@ -1223,24 +1223,7 @@ public function getEditForm($id = null, $fields = null)
...
-        } else {
-            $form->unsetValidator();
-        }
```
ChloeSartorelli, line 1241: That method just sets it null anyway so even if somehow `getCMSCompositeValidator()` returned null (which it can't) the result would be the same.

code/CMSProfileController.php
ChloeSartorelli, line 1: Anything removed from this class was either not necessary (likely related to legacy CMS 3 or older code that no longer exists) or is already being done in one of the superclasses.

code/CMSProfileController.php
```diff
@@ -2,113 +2,75 @@
...
-        }
+    private static bool $allow_new_record = false;
 
-        if ($form->Fields()->hasTabSet()) {
-            $form->Fields()->findOrMakeTab('Root')->setTemplate('SilverStripe\\Forms\\CMSTabSet');
+    protected function init()
+    {
+        parent::init();
+        if (!$this->getResponse()->isRedirect()) {
+            $this->setCurrentPageID(Security::getCurrentUser()->ID);
```
ChloeSartorelli, line 31: Moving this into `init` instead of `getEditForm` so it takes affect for _all_ actions on this controller (including any that might be added through extensions).

Skipping for redirects because at that point we're not doing anything in this controller anyway.

code/SecurityAdmin.php
```diff
@@ -58,7 +58,7 @@ class SecurityAdmin extends ModelAdmin implements PermissionProvider
 
     private static $menu_priority = 0;
 
-    private static $tree_class = Group::class;
+    private static $model_class = Group::class;
```
ChloeSartorelli, line 61: I strongly suspect this config doesn't actually do anything for this class, but checking that is out of scope for this PR. I'm just updating it to the config that `LeftAndMain` expects.

code/SingleRecordAdmin.php
```diff
@@ -0,0 +1,51 @@
...
+ */
+abstract class SingleRecordAdmin extends LeftAndMain
+{
+    /**
+     * Determines if there should be a single record in the database that this admin edits.
+     *
+     * If this is not true, you need to provide a mechanism to tell the form which record it should be editing.
+     * This could be an action (e.g. edit/$ID), or could be based on something about the current member, etc.
+     */
+    private static bool $only_one_record = true;
```
ChloeSartorelli: This is for `SiteConfig` style functionality - but since `CMSProfileController` is explicitly for the _current_ member and not _the one member in the database_, it needs to be configurable.

code/SingleRecordAdmin.php
```diff
@@ -0,0 +1,51 @@
...
+
+    /**
+     * If no record exists, allow a new one to be created
+     */
+    private static bool $allow_new_record = true;
```
ChloeSartorelli, line 24: For almost all controllers, we'll want to be able to create a record if there isn't one. `CMSProfileController` is the exception. Even with `SiteConfig`, if the record in the DB gets somehow deleted, you'll want to be able to create a new one.

code/LeftAndMain.php
emteknetnz: Is `$model_class` always a `DataObject`? If so this should probably be called `$data_class` instead which as far as I can tell the most common way refer to the "class of the DataObject" i.e. $dataClass.  $model_class seems like it could be any model, including something that's not a DataObject

emteknetnz: `$model_class` is always a DataObject correct? If so this should probably be called `$data_class` instead which as far as I can tell the most common way refer to the "class of the DataObject" i.e. `$dataClass`. `$model_class` seems like it could be any model, including something that's not a `DataObject`

code/SingleRecordAdmin.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+abstract class SingleRecordAdmin extends LeftAndMain
+{
+    /**
+     * Determines if there should be a single record in the database that this admin edits.
+     *
+     * If this is not true, you need to provide a mechanism to tell the form which record it should be editing.
+     * This could be an action (e.g. edit/$ID), or could be based on something about the current member, etc.
+     */
+    private static bool $only_one_record = true;
```
emteknetnz:
```suggestion
    private static bool $single_record = true;
```
`$only_one_record` sounds weird to me, almost implying that it's a bad thing :-)

code/SingleRecordAdmin.php
emteknetnz: Seems like these two new configurable properties shouldn't be configurable? I can't think of a scenario where they would be better of simply being protected non-static properties instead? There's a miniscule performance benefit, though more importantly it prevents people from doing things they shouldn't via config, like changing the the config values of SecurityAdmin

code/SingleRecordAdmin.php
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * The base class for admin sections that exist for editing a single record.
+ */
+abstract class SingleRecordAdmin extends LeftAndMain
+{
+    /**
+     * Determines if there should be a single record in the database that this admin edits.
+     *
+     * If this is not true, you need to provide a mechanism to tell the form which record it should be editing.
```
emteknetnz:
```suggestion
     * If false, you need to provide a mechanism to tell the form, usually via getEditForm(), which record it should be editing.
```

code/CMSProfileController.php
```diff
@@ -2,113 +2,75 @@
...
         }
-        $origLocale = $member->Locale;
-
-        if (!$member->canEdit()) {
-            $form->sessionMessage(_t(__CLASS__.'.CANTEDIT', 'You don\'t have permission to do that'), 'bad');
-            return $this->redirectBack();
-        }
-
+        // Get the current member locale so we can see if it changes
+        $member = Member::get()->byID($data['ID']);
```
emteknetnz:
```suggestion
        $member = Member::get()->byID($id);
```

code/CMSProfileController.php
```diff
@@ -2,113 +2,75 @@
...
     }
 
     public function save(array $data, Form $form): HTTPResponse
     {
-        $member = Member::get()->byID($data['ID']);
-        if (!$member) {
-            $this->httpError(404);
+        // Make sure the ID is a positive number
+        $id = $data['ID'] ?? null;
+        if (!is_numeric($id) || $id < 1) {
```
emteknetnz: Should validate that it's an int (or string int), is_numeric() will allow a float

code/SingleRecordAdmin.php
```diff
@@ -0,0 +1,56 @@
...
+
+    public function getRecord($id): ?DataObject
+    {
+        if (!$id) {
+            return $this->getSingleRecord();
+        }
+        return parent::getRecord($id);
+    }
+
+    protected function getSingleRecord(): ?DataObject
```
emteknetnz, line 44:
```suggestion
    private function getSingleRecord(): ?DataObject
```
If we're providing `only_one_record` and `allow_new_record` and implementing logic to do stuff based on how they're set, why would anyone need to override this in a subclass? Things should be private unless there's an immediate need to make it protected.

code/SingleRecordAdmin.php
```diff
@@ -0,0 +1,56 @@
...
+        }
+        return $form;
+    }
+
+    public function getRecord($id): ?DataObject
+    {
+        if (!$id) {
+            return $this->getSingleRecord();
+        }
+        return parent::getRecord($id);
```
emteknetnz, line 41: What's the scenario where $id is truthy?  Seems like we should be throwing an exception here instead?

emteknetnz (changes requested):

code/LeftAndMain.php
ChloeSartorelli: I chose this name to complement `getModelClass()` which currently only exists in `ModelAdmin` but will become more ubiquitous in other cards in the epic. 

In cmsmain and likely others the `getModelClass()` method will return the value held in this config.

If you're okay with `getModelClass()` returning the value of a `data_class` config property I can make the requested change - though that would be a little unintuitive.

code/SingleRecordAdmin.php
ChloeSartorelli: The pattern in these controllers is to use configuration for this sort of thing, so I was just following that pattern.
If you feel strongly about it, I can change them to protected member properties. 

`SecurityAdmin` isn't relevant, I assume you mean the profile admin. People can do what they like with it in their own projects.

code/SingleRecordAdmin.php
ChloeSartorelli: Feel free to suggest an alternative name.

code/SingleRecordAdmin.php
ChloeSartorelli: Will make the change, but won't add "usually via getEditForm()" because the only scenario we've actually implemented doesn't do it that way.

GuySartorelli:
I've responded to the bits that are relevant for the CMS 5 PRs. I'll make the requested changes that are agreed upon after the CMS 5 PRs have been merged and I can rebase on top of them.

code/LeftAndMain.php
emteknetnz: Yeah, I guess ModelAdmin does use $modelClass everywhere right now, so I guess if that's the local standard we should be using that

code/SingleRecordAdmin.php
emteknetnz: $single_record

ChloeSartorelli: How about `only_single_record`? Or `restrict_to_single_record`? I want the name to be fairly clear that it's saying "there should only ever be one record of this type" when set to true.

Or maybe `dynamic_current_record`? As in "this will dynamically fetch the current record without you telling it what the ID is"?

code/SingleRecordAdmin.php
emteknetnz: Yup, I meant CMSProfileController

You're probably right it is something of a pattern, which it probably shouldn't be though we should respect what's already there.

code/SingleRecordAdmin.php
emteknetnz: $restrict_to_single_record sounds good

ChloeSartorelli: Done

code/SingleRecordAdmin.php
ChloeSartorelli: Done

code/CMSProfileController.php
ChloeSartorelli: Done

code/CMSProfileController.php
ChloeSartorelli: Done

code/SingleRecordAdmin.php
ChloeSartorelli: `$id` is truthy if an ID is passed in. So we either want the "single record" based on the configuration (where a `null` or `0` is passed in), or we want the specified record if a specific ID is passed in.

code/SingleRecordAdmin.php
ChloeSartorelli: Please see the docs where overriding this method is explicitly mentioned as a way to provide more complex behaviour for getting this record.

code/SingleRecordAdmin.php
emteknetnz: Why would be specify an $id though in a SingleRecordAdmin? Isn't the whole point to only have a single record?

emteknetnz (approved):

emteknetnz (approved):

PR #1844, opened 2024-10-30
API Deprecate API that will be renamed

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1764

### DIFF
--- a/code/CMSProfileController.php
+++ b/code/CMSProfileController.php
@@ -20,6 +20,9 @@ class CMSProfileController extends LeftAndMain
 
     private static $required_permission_codes = 'CMS_ACCESS';
 
+    /**
+     * @deprecated 5.4.0 Will be renamed to model_class
+     */
     private static $tree_class = Member::class;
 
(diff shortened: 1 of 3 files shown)


code/CMSProfileController.php
```diff
@@ -20,6 +20,9 @@ class CMSProfileController extends LeftAndMain
 
     private static $required_permission_codes = 'CMS_ACCESS';
 
+    /**
+     * @deprecated 5.4.0 Will be renamed model_class
```
emteknetnz:
```suggestion
     * @deprecated 5.4.0 Will be renamed to $model_class
```
Update here and in all other files in all all the PRs

emteknetnz (changes requested):

code/CMSProfileController.php
ChloeSartorelli: Done

emteknetnz (approved):

Issue #1845, opened 2024-11-01
ignore_menuitem is not documented because is not declared


catharsisjelly:
### Description

This is already implemented but not documented in the API `CMSMenu` references a var that can be used called `ignore_menuitem` which is a boolean field. This var is not declared in `LeftAndMain` and therefore not documented. This was implemented by #211 but there is no documentation on it which has been confusing for me

### Validations

- [X] You intend to implement the feature yourself
- [X] You have read the [contributing guide](https://docs.silverstripe.org/en/contributing/code/)
- [X] You strongly believe this feature should be in core, rather than being its own community module
- [X] You have checked for existing issues or pull requests related to this feature (and didn't find any)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1846

GuySartorelli:
PR closed

PR #1846, opened 2024-11-01
Adding `ignore_menuitem` into LeftAndMain

catharsisjelly:
## Description
Adds the configuration `ignore_menuitem` into LeftAndMain so that anyone looking at the class will knows that this configuration exists. This will also now show up in the API docs (I think).

## Issues
- #1845 

## Pull request checklist
- [x] CI is green

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -116,6 +116,12 @@ class LeftAndMain extends Controller implements PermissionProvider
     private static $url_priority = 50;
 
     /**
+     * @config
+     * @var string
+     */
+    private static bool $ignore_menuitem = false;
+
+    /**
      * A subclass of {@link DataObject}.
      *
      * Determines what is managed in this interface, through
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -116,7 +116,7 @@ class LeftAndMain extends Controller implements PermissionProvider
     private static $url_priority = 50;
 
     /**
-     * Set this to true not display the menu item 
+     * When set to true, this controller isn't given a menu item in the left panel in the CMS. 
      */
     private static bool $ignore_menuitem = false;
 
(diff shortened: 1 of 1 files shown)


catharsisjelly:
This is aimed to fix it in the main branch, happy to rebase it to another if needed. It could be also implemented in `1.x`

code/LeftAndMain.php
```diff
@@ -115,6 +115,12 @@ class LeftAndMain extends Controller implements PermissionProvider
      */
     private static $url_priority = 50;
 
+    /**
+     * @config
+     * @var string
+     */
+    private static bool $ignore_menuitem = false;
```
```suggestion
     */
    private static bool $ignore_menuitem = false;
```
`string` is obviously not the correct var type - but we don't need any `@var` in the PHPDoc since this has a strong typehint anyway.

`@config` doesn't actually do anything, so removing that too.

Please also add a PHPDoc description of what this configuration does.

GuySartorelli (changes requested):
Thanks for doing this. You've got the correct branch - new API definitions belong in minor releases, and the `2` branch represents the April 2025 minor release currently.

code/LeftAndMain.php
catharsisjelly: Done, thanks for the input

code/LeftAndMain.php
```diff
@@ -115,6 +115,11 @@ class LeftAndMain extends Controller implements PermissionProvider
      */
     private static $url_priority = 50;
 
+    /**
+     * Set this to true not display the menu item
```
```suggestion
     * When set to true, this controller isn't given a menu item in the left panel in the CMS.
```

GuySartorelli (changes requested):

GuySartorelli (approved):
LGTM, thanks for doing this.
Will merge once CI has finished running and goes green.

GuySartorelli:
There's a linting failure related to your change. Can you please resolve that?

GuySartorelli (approved):
LGTM, thanks for this.

Issue #1762, opened 2024-05-30
Refactor formschema logic from `LeftAndMain` into its own abstract class


GuySartorelli:
Currently `LeftAndMain` is bloated with logic for setting up formschema, but that isn't needed by most of its subclasses. This should be moved into its own class so the logic is only available where it's needed, and to make `LeftAndMain` easier to maintain.

## Related
- https://github.com/silverstripe/silverstripe-cms/issues/2933

## Notes
- The exact name of the class isn't important - if you think of a better name, feel free.

## Acceptance Criteria
- A new abstract `FormSchemaController` class is created, and is a subclass of `AdminRouteController` (see https://github.com/silverstripe/silverstripe-admin/issues/1761)
- Methods, properties, and configuration required for formschema functionality are moved from `LeftAndMain` to the new class (see [this comment](https://github.com/silverstripe/silverstripe-cms/issues/2933#issuecomment-2111557980) for recommendations about this)
- `ModalController` is a subclass of the new controller
- `ModalController` handles the routes which currently require creating a singleton of `LeftAndMain` for formschemas
- Any controller in supported modules that currently subclasses `LeftAndMain` but only needs the formschema logic is made to subclass the new class instead (e.g. `LinkFieldController`)
- Documentation is updated to indicate best practice for form schema controllers

### CMS 5 PRs

- https://github.com/silverstripe/silverstripe-admin/pull/1851
- https://github.com/silverstripe/silverstripe-subsites/pull/613
- https://github.com/silverstripe/developer-docs/pull/629

After merging reassign to Guy so they can merge up and rebase the CMS 6 PRs

### CMS 6 PRs

[**Kitchen sink CI**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/11904419675)

- https://github.com/silverstripe/silverstripe-admin/pull/1850
- https://github.com/silverstripe/silverstripe-subsites/pull/611
- https://github.com/silverstripe/silverstripe-linkfield/pull/348
- https://github.com/silverstripe/silverstripe-mfa/pull/573
- https://github.com/silverstripe/silverstripe-userforms/pull/1348
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/369
- https://github.com/silverstripe/silverstripe-elemental/pull/1273
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1509
- https://github.com/silverstripe/silverstripe-contentreview/pull/267
[...]

PR #1850, opened 2024-11-12
NEW Move form schema logic into FormSchemaController

GuySartorelli:
In a perfect world `LeftAndMain` would _not_ be a subclass of the new `FormSchemaController` class - but so many of its subclasses rely directly on that logic, and refactoring those to spin off sub-controllers looks like way more work than it's really worth.

I thought of making the new class a trait instead, but if a trait has `private static` properties the subclass can't also have those private static properties.
We could implement `getExtraMethodConfig()` in the trait to work around that, but that's an ugly solution imo.

This PR still takes us closer to making `LeftAndMain` abstract (since it doesn't need to route `Modals` anymore) and makes it easier to have non-`LeftAndMain` form schema controllers, and those are the main goals for this card.

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1762

### DIFF
--- a/code/FormSchemaController.php
+++ b/code/FormSchemaController.php
@@ -1,0 +1,156 @@
+<?php
+
+namespace SilverStripe\Admin;
+
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\HTTPResponse;
+use SilverStripe\Core\Validation\ValidationResult;
+use SilverStripe\Forms\Form;
+use SilverStripe\Forms\HTMLEditor\HTMLEditorConfig;
+use SilverStripe\Forms\HTMLEditor\TinyMCEConfig;
+use SilverStripe\Forms\Schema\FormSchema;
+use SilverStripe\Security\Security;
+
+/**
+ * The base class for controllers in the CMS that need access to form schema functionality.
+ */
+abstract class FormSchemaController extends AdminController
+{
+    /**
+     * Form schema header identifier
+     */
+    public const SCHEMA_HEADER = 'X-Formschema-Request';
+
+    private static array $allowed_actions = [
+        'schema',
+    ];
+
+    private static array $url_handlers = [
+        'GET schema/$FormName/$ItemID/$OtherItemID' => 'schema',
+    ];
+
+    private static array $dependencies = [
+        'FormSchema' => '%$' . FormSchema::class,
+    ];
+
+    /**
+     * Current form schema helper
+     */
+    private ?FormSchema $schema = null;
+
+    protected function init()
+    {
+        $this->beforeExtending('onInit', function () {
+            $this->prepareTinyMce();
+        });
+        parent::init();
+    }
+
+    /**
+     * Get the form schema helper for this controller
+     */
+    public function getFormSchema(): FormSchema
+    {
+        if (!$this->schema) {
+            $this->schema = FormSchema::singleton();
+        }
+        return $this->schema;
+    }
+
+    /**
+     * Set the form schema helper for this controller
+     */
+    public function setFormSchema(FormSchema $schema): static
+    {
+        $this->schema = $schema;
+        return $this;
+    }
+
+    /**
+     * Gets a JSON schema representing the current edit form.
+     */
+    public function schema(HTTPRequest $request): HTTPResponse
+    {
+        $formName = $request->param('FormName');
+        $itemID = $request->param('ItemID');
+
+        if (!$formName) {
+            $this->jsonError(400, 'Missing request params');
+        }
+
+        // Most usages of this method (e.g. AssetAdmin) have two methods - one method is
+        // named the same as the form, and does not accept a parameter - the other prefixed
+        // with "get" and accepts an $itemID parameter.
+        // Some usages exclude the getter method, so we need to check for both.
+        $formMethod = "get{$formName}";
+        if (!$this->hasMethod($formMethod)) {
+            $formMethod = $formName;
+            if (!$this->hasMethod($formMethod)) {
+                $this->jsonError(404, 'Form not found');
+            }
+        }
+
+        if (!$this->hasAction($formName)) {
+            $this->jsonError(401, 'Form not accessible');
+        }
+
+        if ($itemID) {
+            $form = $this->{$formMethod}($itemID);
+        } else {
+            $form = $this->{$formMethod}();
+        }
+        $schemaID = $request->getURL();
+        return $this->getSchemaResponse($schemaID, $form);
+    }
+
+    /**
+     * Check if the current request has a X-Formschema-Request header set.
+     * Used by conditional logic that responds to validation results
+     */
+    protected function getSchemaRequested(): bool
+    {
+        $parts = $this->getRequest()->getHeader(static::SCHEMA_HEADER);
+        return !empty($parts);
+    }
+
+    /**
+     * Generate schema for the given form based on the X-Formschema-Request header value
+     *
+     * @param string $schemaID ID for this schema. Required.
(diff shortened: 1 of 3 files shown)


code/FormSchemaController.php
```diff
@@ -0,0 +1,161 @@
...
+    private ?FormSchema $schema = null;
+
+    /**
+     * Get form schema helper
+     */
+    public function getFormSchema(): FormSchema
+    {
+        if (!$this->schema) {
+            $this->schema = FormSchema::singleton();
+        }
```
ChloeSartorelli, line 56: This wasn't there in `LeftAndMain` but it's a little protection in case the dependency doesn't get injected (e.g. not using injector to instantiate this controller for some reason)

code/FormSchemaController.php
ChloeSartorelli, line 1: The code in this class is mostly directly copied in from `LeftAndMain`.

Note that there were no existing unit tests explicitly for the methods in this class - instead, this functionality is fairly heavily tested in behat (linkfield, elemental, history controller, WYSIWYG link modals all use it)

code/FormSchemaController.php
```diff
@@ -0,0 +1,162 @@
...
+
+        // The getter can accept an ID where the main form action wouldn't
+        $formMethod = "get{$formName}";
+        if (!$this->hasMethod($formMethod)) {
+            $formMethod = $formName;
+            if (!$this->hasMethod($formMethod)) {
+                $this->jsonError(404, 'Form not found');
+            }
+        }
```
ChloeSartorelli: Most usages of this method right now (e.g. see AssetAdmin) have two methods - one method is named the same as the form, and does not accept a parameter - the other prefixed with `get` and accepts a parameter.

The usages of `ModalController` used to go through `methodSchema` on `LeftAndMain` - but now we just go directly through `schema` on `ModalController`. Those methods do _not_ have `get` prefixes.

I'm not going through and refactoring all of those - that's out of scope. Instead, I've just updated this method to account for both scenarios.

code/LeftAndMain.php
```diff
@@ -330,141 +307,19 @@ public function getClientConfig(): array
...
-        }
-        $schemaID = $request->getURL();
-        return $this->getSchemaResponse($schemaID, $form);
-    }
-
-    /**
-     * Get the form schema from a given method.
-     * The method must return a Form.
-     */
-    public function methodSchema(HTTPRequest $request): HTTPResponse
```
ChloeSartorelli, line 407: This method was just a bizarre way to avoid directly linking to `ModalController`.
It's not used anywhere now that we're linking to `ModalController` directly. If anyone's using it directly in their own projects or modules, the recommendation will be to directly link to the `schema` for the `FormSchemaController` that handles that form, instead of this indirection.

code/FormSchemaController.php
```diff
@@ -0,0 +1,148 @@
...
+            // file because insufficient information exists when that is being processed
+            $htmlEditorConfig = HTMLEditorConfig::get_active();
+            $htmlEditorConfig->setOption('language', TinyMCEConfig::get_tinymce_lang());
+            $langUrl = TinyMCEConfig::get_tinymce_lang_url();
+            if ($langUrl) {
+                $htmlEditorConfig->setOption('language_url', $langUrl);
+            }
+        });
+        parent::init();
+    }
```
ChloeSartorelli: Needed for controllers such as `ElementalAreaController` which may include WYSIWYG in their forms.

code/FormSchemaController.php
```diff
@@ -0,0 +1,148 @@
...
+    /**
+     * Current form schema helper
+     */
+    private ?FormSchema $schema = null;
+
+    protected function init()
+    {
+        $this->beforeExtending('onInit', function () {
+            // Set the members html editor config
+            if (Security::getCurrentUser()) {
```
emteknetnz: Looks kind of messy having all of this here around the top of the file since it's all TinyMCE related which is weird in a FormSchema class. Move it to a private method down the bottom of the class

code/FormSchemaController.php
```diff
@@ -0,0 +1,148 @@
...
+            $langUrl = TinyMCEConfig::get_tinymce_lang_url();
+            if ($langUrl) {
+                $htmlEditorConfig->setOption('language_url', $langUrl);
+            }
+        });
+        parent::init();
+    }
+
+    /**
+     * Get form schema helper
```
emteknetnz:
```suggestion
     * Get the form schema helper for this controller
```

code/FormSchemaController.php
```diff
@@ -0,0 +1,148 @@
...
+    public function getFormSchema(): FormSchema
+    {
+        if (!$this->schema) {
+            $this->schema = FormSchema::singleton();
+        }
+        return $this->schema;
+    }
+
+    /**
+     * Set form schema helper for this controller
```
emteknetnz:
```suggestion
     * Set the form schema helper for this controller
```

code/FormSchemaController.php
```diff
@@ -0,0 +1,148 @@
...
+        $formMethod = "get{$formName}";
+        if (!$this->hasMethod($formMethod)) {
+            $formMethod = $formName;
+            if (!$this->hasMethod($formMethod)) {
+                $this->jsonError(404, 'Form not found');
+            }
+        }
+
+        if (!$this->hasAction($formName)) {
+            $this->jsonError(401, 'Form not accessible');
```
emteknetnz, line 94:
```suggestion
            $this->jsonError(404, 'Form not accessible');
```
It's an invalid route, rather than an issue with a user not being logged in

code/LeftAndMain.php
```diff
@@ -104,7 +94,6 @@ class LeftAndMain extends AdminController implements PermissionProvider
      *
      * @config
      * @var string
-     * @deprecated 2.4.0 Will be renamed to model_class
```
ChloeSartorelli, line 107: unrelated - was missed in a merge-up

code/FormSchemaController.php
emteknetnz: Could you put some the github comment info into the inline comment - by itself the existing comment "The getter can accept an ID where the main form action wouldn't" is pretty confusing

emteknetnz (changes requested):

code/FormSchemaController.php
ChloeSartorelli: Changing error codes is not in scope for this PR. This isn't new code.

code/FormSchemaController.php
ChloeSartorelli: Happy to do so, but I'm not sure what you find confusing about it so I don't know what info you want in there. What would you like the comment to say? Or what about the existing comment is confusing?

code/FormSchemaController.php
ChloeSartorelli: TinyMCE is in forms, so form schema having form stuff doesn't seem weird to me. Done anyway though.

code/FormSchemaController.php
ChloeSartorelli: Done

code/FormSchemaController.php
ChloeSartorelli: Done

code/FormSchemaController.php
emteknetnz: Most usages of this method (e.g. AssetAdmin) have two methods - one method is named the same as the form, and does not accept a parameter - the other prefixed with "get" and accepts an `$itemID` parameter.

^ Maybe just the comment to this

ChloeSartorelli: Done

emteknetnz (approved):

PR #1851, opened 2024-11-14
API Deprecate API that will be removed as a result of refactoring

GuySartorelli:
See https://github.com/silverstripe/silverstripe-admin/pull/1850

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1762

### DIFF
--- a/code/CMSProfileController.php
+++ b/code/CMSProfileController.php
@@ -21,7 +21,7 @@ class CMSProfileController extends LeftAndMain
     private static $required_permission_codes = 'CMS_ACCESS';
 
     /**
-     * @deprecated 5.4.0 Will be renamed to model_class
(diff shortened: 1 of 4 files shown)


code/CMSProfileController.php
```diff
@@ -21,7 +21,7 @@ class CMSProfileController extends LeftAndMain
     private static $required_permission_codes = 'CMS_ACCESS';
 
     /**
-     * @deprecated 5.4.0 Will be renamed to model_class
+     * @deprecated 2.4.0 Will be renamed to model_class
```
ChloeSartorelli, line 24: Unrelated, along with a few other of these where I noticed the version number was wrong

emteknetnz (approved):

Issue #1852, opened 2024-11-14
Saving does not submit FileFields


lekoala:
### Module version(s) affected

4.x, 5.x

### Description

Using FileField in the cms is not possible because files are not submitted due to how the form is submitted through ajax.

This is an issue because it means it's not always practical to "send files" not meant to be saved as assets, but for example as import templates or whatever.

Issue for reference:
https://github.com/lekoala/silverstripe-pure-modal/issues/8

### How to reproduce

Have a model admin instance and a model

Have this

```php
    public function getCMSFields()
    {
         $fields = parent::getCMSFields();

        $fields->addFieldsToTab('Root.Upload', new FileField("TestFile"));
...

  public function onBeforeWrite()
    {
        parent::onBeforeWrite();

        dd($_FILES)
```

Select a file, click save and close, and see how $_FILES is empty

### Possible Solution

i believe it's due to how the form is submitted by LeftAndMain see here

https://github.com/silverstripe/silverstripe-admin/blob/0517656dbb9c6ba38522a11c379a64c11f640162/client/src/legacy/LeftAndMain.js#L539-L546

it should really be replaced by a simple FormData that should work fine instead

https://stackoverflow.com/questions/41684385/how-to-add-file-value-with-jquery-serialize

This should be really easy to fix, I'm happy to give it a try

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1854

GuySartorelli:
Thanks for reporting this.
`FileField` is a weird one - in the past I've given up and made my own complex uploading nonsense based around how `UploadField` works so it'd be nice to see this fixed.

> This should be really easy to fix, I'm happy to give it a try

I'd love to review a PR

lekoala:
hey @GuySartorelli 

here you go https://github.com/silverstripe/silverstripe-admin/pull/1854

seems to work fine on my end!

i targeted 5.3 i don't think anyone expected this to work anyway, but this could also be easily back ported to older versions

GuySartorelli:
@emteknetnz Why was this closed? There's a PR attached that isn't merged which indicates this hasn't been resolved yet.

emteknetnz:
Sorry, unintentional

PR #1854, opened 2024-11-15
ENH leverage FormData to support file upload

lekoala:
## Description

Currently, ajax forms are submitted using serializeArray. This doesn't allow sending more advanced content, like a file upload

## Manual testing steps

- Add a FileField in the cms
- Check that $_FILES is empty on save

With this change, files are submitted properly. It also offers the opportunity to do more advanced things (eg: sending binary canvas data as a blob) and, all things considered, should probably work better because it uses a native js feature instead of jquery serializeArray

## Issues

https://github.com/silverstripe/silverstripe-admin/issues/1852

## Pull request checklist
- [ ] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x] CI is green


GuySartorelli:
The JS CI is failing as you haven't built the dist files. Can you please build that and include it in the PR?
Instructions are at https://docs.silverstripe.org/en/5/contributing/build_tooling/

lekoala:
from what i see in the tests, it's not my changes that creates error but i've compiled the files

on windows, it requires this to work 
module.exports.rules['linebreak-style'] = ['error', 'windows'];

otherwise you get Expected linebreaks to be 'LF' but found 'CRLF' linebreak-style

could be nice to add that as a default for https://github.com/silverstripe/eslint-config/blob/1/.eslintrc.js to make it windows friendly

GuySartorelli:
> could be nice to add that as a default for https://github.com/silverstripe/eslint-config/blob/1/.eslintrc.js to make it windows friendly

I've created https://github.com/silverstripe/eslint-config/issues/28 to track that. That's a very good idea.

Sorry this has taken so long for me to look at. There was a merge conflict so I've rebased and re-compiled the assets. I'll try to review this properly shortly.

GuySartorelli (changes requested):
This doesn't quiiite seem to work the way I'd expect. It does upload the file, but:

1. If the file is for a relation (only tried with `has_one` so far), the file doesn't get stored against that relation.
1. If the file is an image, a thumbnail isn't included. Going to the asset admin section, there is no thumbnail for the image.

2 is probably out of scope for this, so I'd be okay with that being handled separately. But 1 definitely seems like it is part of the general problem this PR is intended to resolve.

The code I used is:

```php
<?php

namespace {

    use SilverStripe\Assets\File;
    use SilverStripe\Forms\FileField;

    class Page extends SiteTree
    {
        private static $db = [];

        private static $has_one = [
            'MyFile' => File::class,
        ];

        {
            // I also tried with the field being named `MyFile` without the "ID" suffix
            $fields->addFieldToTab('Root.Main', FileField::create('MyFileID'));
            return $fields;
        }
    }
}
```

lekoala:
hi @GuySartorelli it does not seem related to my PR itself

first, FileField don't need the ID suffix
https://github.com/silverstripe/silverstripe-framework/blob/b662f03b75a7f1e14075f3068a5c787197724501/src/Forms/FileField.php#L142-L143

from what i can see, the upload part itself works fine

if you set the relation as an image, thumbnail generation should work as far as i can tell (but it needs to be published in order to work, so maybe it's an ownership issue)

relation seem to work
image

image

so all good as far as it goes for this module i would say

PR #1855, opened 2024-11-18
DEP Update JS dependencies

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/308


package.json
```diff
@@ -50,6 +50,7 @@
     "detect-browser": "^5.3.0",
     "es6-promise": "^4.2.8",
     "events-polyfill": "^2.1.2",
+    "immutable": "^4.3.7",
```
emteknetnz, line 53: This is a required change to lock immutable to ^4 rather than allowing it to upgrade to ^5 which will cause an issue in redux-form when sorting a gridfield header

`Fixes error when clicking sortable headers in a GridField
Uncaught TypeError: _immutable.Iterable.isIterable is not a function
Search.js:545`

ChloeSartorelli: I really wish we could add comments to json files. We're gonna have no idea why this is here in a year's time lol.

GuySartorelli (approved):
LGTM

PR #1856, opened 2024-11-20
ENH Allow status flags to be used generically

GuySartorelli:
Status flags are made generic in https://github.com/silverstripe/silverstripe-framework/pull/11460

This PR generalises the styling for them across the board, and ensures `LeftAndMain` classes always show badges in the breadcrumbs.

Styling in `CMSMain` looks like this:
Screenshot from 2024-11-20 14-56-20

Custom flags look like this (but custom styling can be easily added):
Screenshot from 2024-11-20 14-57-07

Status flags in gridfields look like this:
Screenshot from 2024-11-20 14-56-03
Screenshot from 2024-11-20 14-55-48

## Issue


client/src/styles/legacy/_badges.scss
```diff
@@ -0,0 +1,16 @@
+.badge {
+  display: inline-block;
+  background-color: $color-notice;
+  border-color: transparent;
+  color: $body-color-dark;
```
ChloeSartorelli, line 5: These colours are fairly arbitrary but look nice as a default IMO. Only custom flags will use this styling, all of the versioned/fluent flags still use the colours they used to use.

client/src/styles/legacy/GridField.scss
```diff
@@ -158,27 +158,6 @@
...
-            color: #636363;
-            border: 1px solid #E49393;
-            background-color: #F2DADB;
-          }
-
-          span.badge.status-workflow-approval {
-            color: #56660C;
-            border: 1px solid #7C8816;
-            background-color: #DAE79A;
-          }
```
ChloeSartorelli, line 181: These were moved into the relevant modules - versioned-admin and advanced-workflow.

client/src/styles/legacy/_badges.scss
```diff
@@ -0,0 +1,16 @@
...
+  // Badges with no text aren't displayed
+  &:empty {
+    display: none;
+  }
+}
```
ChloeSartorelli, line 10: This already gets inherited from bootstrap, but it surprised me at first. Adding it here so it's explicitly intended behaviour.
Fluent relies on this for something that probably shouldn't be using status flags but I don't want to widen the scope to unpicking that right now.

client/src/styles/legacy/_style.scss
```diff
@@ -119,8 +119,10 @@ html, body {
     min-height: $toolbar-height;
     min-width: 250px;
 
-    * {
-      display: inline-block; // align back button and breadcrumbs
+    // align back button and breadcrumbs
+    // but specifically do NOT override badge display
+    *:not(.badge) {
```
ChloeSartorelli, line 124: Required to respect the `display:none` on empty badges.

michalkleiner:
Is it just me or is the text in the tag like 1 or 2 pixels too high?

GuySartorelli:
I wondered that too at first but that's where it's always been vertically.
Can adjust it if that's the consensus, of course.

emteknetnz:
Yeah it probably needs to be adjusted

It absolutely needs to be centered in chrome, and preferably also firefox

emteknetnz (approved):

PR #1857, opened 2024-11-20
API Remove silverstripe/campaign-admin integration support

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/342


GuySartorelli (changes requested):
`i18n._t('CampaignAdmin.DELETECAMPAIGN', 'Are you sure you want to delete this record?')` in `client/src/components/GridField/GridField.js` should not use a localisation key from campaign admin. There's a few others in that file too.
1. Double check if that is actually at all related to campaign functionality and if so, update the logic as needed
1. Use a localisation key that doesn't have anything to do with campaigns

emteknetnz:
I don't actually know what this react component for GridField even does, or if it's even used (I suspect it isn't)

Perhaps this one should be a new card to work out if we should just delete the file, along with the seemingly associated GridFieldActions.js?

GuySartorelli:
Looks like `Gridfield.js` is specifically the component for the gridfield-like UI in campaign admin. Almost certainly not used anywhere else in core.

That said, if you want to remove that in a separate card I'm okay with that. Up to you.

Either way, this needs a rebase.

emteknetnz:
Will do as [new card](https://github.com/silverstripe/silverstripe-admin/issues/1862)

GuySartorelli (approved):
LGTM

Issue #1765, opened 2024-05-30
Refactor `ModalController`


GuySartorelli:
Currently registering forms from `FormFactory` classes as form schemas requires applying an extension to `ModalController` - this should be doable via YAML config directly instead.

## Related
- https://github.com/silverstripe/silverstripe-cms/issues/2933

## Acceptance Criteria

- [x] `ModalController` has an API to register form factories via YAML
- [x] There is a single action for getting a modal form - it accepts a unique identifier for the form as a route parameter, and returns the form (or returns an HTTP error response if no form is available for that identifier)
- ~~getvars are passed into form factories as needed (this will require either modifying existing factories, modifying the `getForm()` method signature, or modifying the names of the getvars currently used)~~ Not relevant
- [x] All existing form factories that are routed through `ModalController` are registered using the new YAML config instead of as extensions or hardcoded actions

### CMS 5

- https://github.com/silverstripe/silverstripe-admin/pull/1861
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1516
- https://github.com/silverstripe/silverstripe-cms/pull/3030
- https://github.com/silverstripe/developer-docs/pull/637

Reassign to Guy after merging so they can merge up to 6 and rebase

### CMS 6

[**Kitchen sink CI**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/12041329005)

Note there's currently no documentation about setting up modal forms. There is some limited information about form schema but that's focused more on setting up the REST(ish) endpoints.
Because of that I've opted not to include documentation for this, as it would effectively require adding docs for the JS that results in building the modal on the front-end as well and I think that's a fair amount of work that doesn't feel super in-scope.

Any custom modal forms should continue working exactly as they were (extensions applied to `ModalController` still work), any customisations to the forms will continue working (the form factories aren't changed at all except for one change to the anchor factory, but that's just changing how it accepts the page ID and doesn't affect the actual form)... this doesn't break anything anyone is likely to be using in any way. I think the only docs we need are the auto-generated "this deprecated method was removed" in the changelog.

That said, if you want docs, let me know the scope of the docs you want and I'll see what I can do.

- https://github.com/silverstripe/silverstripe-admin/pull/1860
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1515
- https://github.com/silverstripe/silverstripe-cms/pull/3029
- https://github.com/silverstripe/silverstripe-subsites/pull/616
[...]

PR #1860, opened 2024-11-25
API Provide a way to register link forms dynamically for ModalController

GuySartorelli:
Reduces boilerplate required to set up an insert link modal form for the WYSIWYG

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1765


code/FormSchemaController.php
```diff
@@ -121,7 +111,7 @@ protected function getSchemaRequested(): bool
      * @param ValidationResult $errors Required for 'error' response
      * @param array $extraData Any extra data to be merged with the schema response
      */
-    protected function getSchemaResponse(string $schemaID, ?Form $form = null, ValidationResult $errors = null, array $extraData = []): HTTPResponse
+    public function getSchemaResponse(string $schemaID, ?Form $form = null, ValidationResult $errors = null, array $extraData = []): HTTPResponse
```
ChloeSartorelli, line 114: Needs to be public to avoid code duplication. See https://github.com/silverstripe/silverstripe-asset-admin/pull/1515

code/FormSchemaController.php
```diff
@@ -103,16 +103,6 @@ public function schema(HTTPRequest $request): HTTPResponse
...
 
-    /**
-     * Check if the current request has a X-Formschema-Request header set.
-     * Used by conditional logic that responds to validation results
-     */
-    protected function getSchemaRequested(): bool
-    {
-        $parts = $this->getRequest()->getHeader(static::SCHEMA_HEADER);
-        return !empty($parts);
-    }
```
ChloeSartorelli, line 114: This has just been moved further down the file so it's not above the public method

client/src/containers/InsertLinkModal/InsertLinkModal.js
```diff
@@ -93,7 +93,7 @@ const createInsertLinkModal = (sectionConfigKey, formName) => {
 
     // get the schemaUrl to use as a key for overrides
     const schemaUrl = `${sectionConfig.form[formName].schemaUrl}${requireTextFieldUrl}`
-      .replace(/:pageid/, ownProps.currentPageID);
+      .replace(/:pageid/, ownProps.currentPageID || 0);
```
ChloeSartorelli, line 96: Most link forms don't actually need the current page, so the JS plugins don't pass it through. In those cases `0` is sufficient.

Without this it'd be `undefined` which gets included as a string in the URL which causes errors.
The alternative would be to update the other link plugins to include the actual ID but a) the way the anchor link plugin grabs it right now feels a little flakey and b) they don't need it so there's no need to add that functionality for them.

code/ModalController.php
```diff
@@ -13,40 +18,77 @@ class ModalController extends FormSchemaController
...
     {
+        $modalName = $request->param('ModalName');
+        $itemID = $request->param('ItemID');
+        if ($modalName === null || $itemID === null) {
+            $this->jsonError(400, 'Missing request params');
+        }
+        $modalForms = static::config()->get('link_modal_form_factories');
+        if (!array_key_exists($modalName, $modalForms)) {
+            $this->httpError(400);
+        }
```
ChloeSartorelli, line 57: Using `400` for consistency with the `schema()` method on the superclass.

code/ModalController.php
```diff
@@ -13,40 +18,77 @@ class ModalController extends FormSchemaController
...
+        /** @var LinkFormFactory $factory */
+        $factory = Injector::inst()->get($class);
+        $form = $factory->getForm(
             $this,
-            __FUNCTION__,
-            [ 'RequireLinkText' => isset($showLinkText) ]
+            'linkModalForm/'.$modalName,
+            [
+                'RequireLinkText' => isset($showLinkText),
+                'ItemID' => $itemID,
```
ChloeSartorelli, line 73: `ItemID` is currently only used by the anchor link form, but a) adding it here allows us to include that form via the config instead of having a whole extension for it and b) this means future forms and custom project forms can use it if they want.

code/FormSchemaController.php
```diff
@@ -67,7 +67,7 @@ public function setFormSchema(FormSchema $schema): static
     }
 
     /**
-     * Gets a JSON schema representing the current edit form.
+     * Gets a JSON schema representing a form.
```
ChloeSartorelli, line 70: Unrelated change but the form isn't necessarily (or even usually) the "current edit form" (implying main record edit form which isn't in a modal most of the time).

emteknetnz (approved):

Issue #1865, opened 2024-12-03
Save button loading state is missing


hitaishi19:
### Module version(s) affected

2.3

### Description

There is a delay in the "loading" of the Save button when clicked.

### How to reproduce

1. Edit the existing text block on a page
2. Click the page’s “Save” button

**Expected:**
The “Save” button is loading/processing

**Actual:**
No change to the "Save" button after it is clicked multiple times.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1866

GuySartorelli:
PR merged. This will be automatically tagged by GitHub Actions

PR #1866, opened 2024-12-03
Change button state before event process

hitaishi19:
## Description

Save button does not show loading state when clicking on it. Instead it remains still for sometime and then shows loading state. This is due to event being fully processed before setting the loading class and disabling the button.

## Issues

- https://github.com/silverstripe/silverstripe-admin/issues/1865

## Pull request checklist

- [x] CI is green


GuySartorelli (changes requested):
I've added the pull request template back into the PR description. Can you please fill it in, including ticking all of the checkboxes that apply? This will help you identify things that need to be done before this is ready to be reviewed (for example [build the dist javascript](https://docs.silverstripe.org/en/contributing/build_tooling/)).

Please also retarget this PR to the `2.3` branch so that it can be included as a patch release rather than waiting for the next minor release.
You may need to [reset your commits](https://gist.github.com/GuySartorelli/74627843aeb75e12b777b1f44cb22d2f) after retargetting the PR.


I have also tested this locally and found two problems:
1. Instead of swapping to the "loading" appearance, the button goes completely blank:
    image
    It only uses the full loading styling after the `beforesubmitform` stuff has finished.
    An easy way to test this is to add a `sleep(5)` into the `validate()` method of an elemental block, then make sure to edit the block before hitting save.
1. If an elemental block has a validation error, the button stays disabled forever.

GuySartorelli:
To fix 1, probably you need to move everything in the `if($(button).is('button')) {` conditional block.

To fix 2, I suspect the `clearButton()` function needs to be called in the `if (!success) {` conditional block just before it returns `false`.

client/src/legacy/LeftAndMain.js
```diff
@@ -499,15 +517,12 @@ $.entwine('ss', function($) {
...
           return false;
         }
 
         self.trigger('submitform', {form: form, button: button});
 
-        // set button to "submitting" state
-        $(button).addClass('btn--loading loading');
-        $(button).prop('disabled', true);
-
         if($(button).is('button')) {
```
ChloeSartorelli: This condition and the code in this block is now duplicated - does it _need_ to be in both places, or can this second block be removed?

If it does need to be in both places separately, please abstract it to a function call, and add some in-line code comments explaining _why_ it needs to happen both here _and_ above.

GuySartorelli (changes requested):

client/src/legacy/LeftAndMain.js
```diff
@@ -460,6 +460,24 @@ $.entwine('ss', function($) {
...
+              "</div>"
+          )
+        );
+
+        $(button).css($(button).outerWidth() + "px");
+      }
+
+      // set button to "submitting" state
+      $(button).addClass('btn--loading loading');
+      $(button).prop('disabled', true);
```
ChloeSartorelli: This used to happen before the conditional block. That should probably remain the case.

client/src/legacy/LeftAndMain.js
```diff
@@ -460,6 +460,24 @@ $.entwine('ss', function($) {
       // default to first button if none given - simulates browser behaviour
       if(!button) button = this.find('.btn-toolbar :submit:first');
 
+      if ($(button).is("button")) {
```
```suggestion
      if ($(button).is('button')) {
```

client/src/legacy/LeftAndMain.js
```diff
@@ -460,6 +460,24 @@ $.entwine('ss', function($) {
...
+          $(
+            '<div class="btn__loading-icon">' +
+              '<span class="btn__circle btn__circle--1"></span>' +
+              '<span class="btn__circle btn__circle--2"></span>' +
+              '<span class="btn__circle btn__circle--3"></span>' +
+              "</div>"
+          )
+        );
+
+        $(button).css($(button).outerWidth() + "px");
```
```suggestion
        $(button).css($(button).outerWidth() + 'px');
```

GuySartorelli:
Problem 2 from https://github.com/silverstripe/silverstripe-admin/pull/1866#pullrequestreview-2483336309 is still happening. Did you test that scenario?

hitaishi19:
> Problem 2 from [#1866 (review)](https://github.com/silverstripe/silverstripe-admin/pull/1866#pullrequestreview-2483336309) is still happening. Did you test that scenario?

This is not reproducing to me, can you please share the reproduction steps?

GuySartorelli:
1. Create a block
1. Add some validation to an elemental block which will fail
    ```php
    public function validate()
    {
        $result = parent::validate();
        $result->addFieldError('Title', 'FAILURE');
        return $result;
    }
    ```
1. Make a change to the block (or else validation won't trigger)
1. Click "save" on the page.

client/src/legacy/LeftAndMain.js
```diff
@@ -460,6 +460,24 @@ $.entwine('ss', function($) {
...
+      $(button).prop('disabled', true);
+
+      if ($(button).is('button')) {
+        $(button).append(
+          $(
+            '<div class="btn__loading-icon">' +
+              '<span class="btn__circle btn__circle--1"></span>' +
+              '<span class="btn__circle btn__circle--2"></span>' +
+              '<span class="btn__circle btn__circle--3"></span>' +
+              "</div>"
```
```suggestion
              '</div>'
```
Sorry, didn't notice this one before

GuySartorelli (approved):
LGTM, works well locally. Thanks for taking the time to implement this fix.

PR #1809, opened 2024-08-15
API Change public CMSEditLink to protected updateCMSEditLink

GuySartorelli:
Relies on changes made in https://github.com/silverstripe/silverstripe-framework/pull/11338

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11258

PR #1868, opened 2024-12-08
API Deprecate API which will be removed in CMS 6

GuySartorelli:
## Issue

### DIFF
--- a/code/CMSEditLinkExtension.php
+++ b/code/CMSEditLinkExtension.php
@@ -71,6 +71,7 @@ class CMSEditLinkExtension extends Extension
 
     /**
      * Get a link to edit this DataObject in the CMS.
+     * @deprecated 2.4.0 Will be replaced with SilverStripe\ORM\DataObject::getCMSEditLink() and updateCMSEditLink()
      */
     public function CMSEditLink(): string
     {
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -171,6 +171,7 @@ class LeftAndMain extends Controller implements PermissionProvider
      * Current pageID for this request
      *
      * @var null
+     * @deprecated 2.4.0 Will be renamed to recordID.
      */
     protected $pageID = null;
 
@@ -1029,7 +1030,7 @@ class LeftAndMain extends Controller implements PermissionProvider
     public function show(HTTPRequest $request): HTTPResponse
     {
         if ($request->param('ID')) {
-            $this->setCurrentPageID($request->param('ID'));
+            $this->setCurrentRecordID($request->param('ID'));
         }
         return $this->getResponseNegotiator()->respond($request);
     }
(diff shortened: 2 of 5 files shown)


code/CMSEditLinkExtension.php
```diff
@@ -71,6 +71,7 @@ public function getCMSEditLinkForManagedDataObject(DataObject $obj, string $reci
 
     /**
      * Get a link to edit this DataObject in the CMS.
+     * @deprecated 2.4.0 Will be replaced with SilverStripe\ORM\DataObject::getCMSEditLink() and updateCMSEditLink()
```
ChloeSartorelli, line 74: This should have been done along-side https://github.com/silverstripe/silverstripe-admin/pull/1809

code/LeftAndMain.php
```diff
@@ -1710,30 +1716,42 @@ public function getSilverStripeNavigator(?DataObject $record = null)
...
 
         if (is_numeric($this->getRequest()->param('ID'))) {
-            return $this->getRequest()->param('ID');
+            return (int) $this->getRequest()->param('ID');
         }
 
-        /** @deprecated */
+        // Using session for this is deprecated - see https://github.com/silverstripe/silverstripe-admin/pull/19
         $session = $this->getRequest()->getSession();
         return $session->get($this->sessionNamespace() . ".currentPage") ?: null;
```
ChloeSartorelli, line 1754: Will remove this and from `setCurrentRecordID()` in the related CMS 6 PRs.

code/LeftAndMain.php
```diff
@@ -1745,12 +1763,23 @@ public function currentPageID()
...
         $id = (int)$id;
-        /** @deprecated */
+        $this->setCurrentRecordID($id);
+    }
+
+    /**
+     * Sets the ID for the current record which can be retrieved later through {@link currentRecordID()}.
+     * Keep in mind that setting an ID through GET/POST or as a URL parameter will overrule this value.
+     */
+    public function setCurrentRecordID(?int $id): void
```
ChloeSartorelli, line 1777: Note `CMSMain` explicitly passes in `null`, so we have to allow that.

code/LeftAndMain.php
```diff
@@ -1759,22 +1788,43 @@ public function setCurrentPageID($id)
...
-        return $this->getRecord($this->currentPageID());
+        Deprecation::notice('5.4.0', 'use currentRecord() instead.');
+        return $this->currentRecord();
+    }
+
+    /**
+     * Uses {@link getRecord()} and {@link currentRecordID()}
+     * to get the currently selected record.
+     */
+    public function currentRecord(): ?DataObject
```
ChloeSartorelli, line 1801: This will be `null` before an edit page is opened, e.g. in `CMSMain` before selecting a page.

code/LeftAndMain.php
```diff
@@ -171,9 +171,15 @@ class LeftAndMain extends Controller implements PermissionProvider
...
      *
      * @var null
+     * @deprecated 2.4.0 use recordID instead.
      */
     protected $pageID = null;
 
+    /**
+     * ID of the current record for this request
+     */
+    protected ?int $recordID = null;
```
emteknetnz: Won't the fact we're suddenly no longer updating $pageID lead to it potentially breaking any subclasses that rely on it?

Seems like we should continue to use $pageID for now, and then rename that to $recordID in CMS 6

emteknetnz (changes requested):

code/LeftAndMain.php
ChloeSartorelli: Updated

emteknetnz (approved):

Issue #1859, opened 2024-11-22
Empty GridField without actions creates a 'No items found' row hardcoded to span only two columns


Kevinn1109:
### Module version(s) affected

2.1.16

### Description

When a GridField has no items, normally a row spanning all columns is inserted with the text 'No items found'. However, when the GridField is given no actions, this row is fixed to span only two columns.
![Screenshot 2024-11-22 at 11 20 19](https://github.com/user-attachments/assets/65cd5e50-12b0-43ac-8c0d-4cd3eb510174)

This is caused by the 'hacky' fix for https://github.com/silverstripe/silverstripe-framework/issues/7610, which makes clear that this code exists so that an additional column could be added. This works fine if there's one cell for each column, since only the last cell is changed to span two columns, but doesn't in this particular situation.

The colspan change happens in https://github.com/silverstripe/silverstripe-admin/blob/8c1861c12aae9341da4b898cbb933181799c0c5a/client/src/legacy/GridField.js#L175-L188

### How to reproduce

```php
public function getCMSFields() {
  $fields = parent::getCMSFields();
  $gridField = GridField::create("EmptyRelation", "EmptyRelation", DataObject2::get()->where(["ID" => 0]);
  $gridField->getConfig()->removeComponentsByType([GridFieldAddNewButton::class, GridFieldAddExistingAutocompleter::class]);
  $fields->addFieldToTab("Root.EmptyRelation", $gridField);
  return $fields;
}
```

### Possible Solution

This code seems to have been added to create space for a search button in SS3. This search button doesn't seem to exist anymore, so complete removal of this function may do the trick without consequences.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1869

GuySartorelli:
PR merged. Will be automatically tagged by GitHub Actions

PR #1869, opened 2024-12-09
FIX Add one to current colspan in fixColumns instead of resetting to 2

Kevinn1109:
## Description
In the old situation, the final column in each row would be hardcoded to colspan="2" if the function 'fixColumns' was called. This is required because this function adds an additional column to the table that needs to be filled in. This change removes the hardcoded number and instead adds one to the current colspan of the last cell. This fixes the bug that the 'No items found' row would be capped at two columns even if the table has more.

## Manual testing steps
1. Manually create a GridField for an has_many relation that contains no records, this should fullfill the check in needsColumnFix by default. The relation class should have at least 3 summary fields.
2. Without this change, the 'No items found' row gets a colspan of 2. With this change, it should be 4.

## Issues

- #1859 

## Pull request checklist

- [ ] CI is green


client/src/legacy/GridField.js
```diff
@@ -176,7 +176,8 @@ $.entwine('ss', function($) {
       this.find('.sortable-header').append('<th class="main col-Actions" />');
       this.find('tbody tr').each(function () {
         var cell = $(this).find('td:last');
-        cell.attr('colspan', 2);
+        var colspan = cell.attr("colspan") ?? 1;
```
ChloeSartorelli:
```suggestion
        var colspan = cell.attr('colspan') ?? 1;
```
What is the `?? 1` there for? What scenarios result in there not being a `colspan` attribute?

client/src/legacy/GridField.js
```diff
@@ -176,7 +176,8 @@ $.entwine('ss', function($) {
...
+        cell.attr('colspan', colspan + 1);
```
ChloeSartorelli: I assume the `+1` is what this is referring to from the PR description:

> This is required because this function adds an additional column to the table that needs to be filled in.

Can you please add a comment about that to the code? As is it's not immediately clear just by glancing at the code why we're doing this.

GuySartorelli (changes requested):
Thanks! Couple of small changes

client/src/legacy/GridField.js
Kevinn1109: The colspan attribute is not set by default, I've not seen it in use other than in the case of an empty gridfield.

client/src/legacy/GridField.js
Kevinn1109: It's not clear to me either why this still exists. It comes from the SS3 era, and by the look of it the reason for this fix is no longer relevant. Removing it altogether might be the way to go, but that comes with a larger impact. For now I have chosen to only fix the bug that this code causes to ensure that only the context of this bug is affected.

client/src/legacy/GridField.js
```diff
@@ -176,7 +177,8 @@ $.entwine('ss', function($) {
       this.find('.sortable-header').append('<th class="main col-Actions" />');
       this.find('tbody tr').each(function () {
         var cell = $(this).find('td:last');
-        cell.attr('colspan', 2);
+        var colspan = cell.attr('colspan') ?? 1;
```
ChloeSartorelli, line 181:
```suggestion
        // Note we need to add 1 to the current column span, because we're going to add a new column
        var colspan = cell.attr('colspan') ?? 1;
```

client/src/legacy/GridField.js
ChloeSartorelli: Would colspan `2` always be correct when `colspan` wasn't already defined?

GuySartorelli (changes requested):

client/src/legacy/GridField.js
```diff
@@ -12,6 +12,7 @@ import '../../../thirdparty/jquery-entwine/jquery.entwine.js';
 $.entwine('ss', function($) {
   $('.grid-field').entwine({
     onmatch: function () {
+      // Adds an actions column for the search action if it does not exist already
```
ChloeSartorelli: Please remove this comment from here and add it where this actually happens - probably on the line just before `    fixColumns: function (visible) {` would be perfect.

client/src/legacy/GridField.js
```diff
@@ -175,8 +176,10 @@ $.entwine('ss', function($) {
     fixColumns: function (visible) {
       this.find('.sortable-header').append('<th class="main col-Actions" />');
       this.find('tbody tr').each(function () {
+        // Note we need to add 1 to the current column span, because we're going to add a new column
```
ChloeSartorelli: Please move this comment down one as per the original suggestion in https://github.com/silverstripe/silverstripe-admin/pull/1869#discussion_r1887574735

GuySartorelli (changes requested):
https://github.com/silverstripe/silverstripe-admin/pull/1869#discussion_r1887577258 hasn't been responded to. We're assuming if the colspan wasn't defined, it is already 1 and should become 2. Does that always hold true? Can we do something like count the number of columns in the table or something?

Kevinn1109:
> [#1869 (comment)](https://github.com/silverstripe/silverstripe-admin/pull/1869#discussion_r1887577258) hasn't been responded to. We're assuming if the colspan wasn't defined, it is already 1 and should become 2. Does that always hold true? Can we do something like count the number of columns in the table or something?

The calling function adds one column to the table, so a correctly set up table will always have rows with the number of columns minus that new column. My change changes a hardcoded 2 that disregards the existing colspan to something that takes it into account. Counting the number of columns isn't needed, the entire purpose of this function is to add one column. If this leads to a wrong number of columns, there's a near certainty that said table already has a wrong number of columns in the current release. As far as I can tell, the only native integration that's affected by this is the 'No items found' row.

In my eyes this is legacy code that should be removed altogether, but I'm not entirely sure if that comes with complications or not.

client/src/legacy/GridField.js
```diff
@@ -172,11 +172,14 @@ $.entwine('ss', function($) {
...
 
+    // Adds an actions column for the search action if it does not exist already
     fixColumns: function (visible) {
       this.find('.sortable-header').append('<th class="main col-Actions" />');
       this.find('tbody tr').each(function () {
         var cell = $(this).find('td:last');
-        cell.attr('colspan', 2);
+        // Note we need to add 1 to the current column span, because we're going to add a new column
+        var colspan = cell.attr('colspan') ?? 1;
+        cell.attr('colspan', colspan + 1);
```
ChloeSartorelli: In my local testing, I found `colspan` is a string, so this concatenates the values into a new string.
e.g. if `colspan` is `'3'`, we end up setting the new value to `31` instead of the expected `4`.

GuySartorelli (changes requested):

client/src/legacy/GridField.js
Kevinn1109: As it turns out, it can be both an int and a string depending on how the cell is created. I've added support for strings so that that isn't an issue anymore.

client/src/legacy/GridField.js
```diff
@@ -172,11 +172,14 @@ $.entwine('ss', function($) {
...
 
+    // Adds an actions column for the search action if it does not exist already
     fixColumns: function (visible) {
       this.find('.sortable-header').append('<th class="main col-Actions" />');
       this.find('tbody tr').each(function () {
         var cell = $(this).find('td:last');
-        cell.attr('colspan', 2);
+        // Note we need to add 1 to the current column span, because we're going to add a new column
+        var colspan = cell.attr('colspan') ?? 1;
+        cell.attr('colspan', +colspan + 1);
```
ChloeSartorelli: That syntax seems a bit weird - if this is just to cast to number, please use `Number` instead.
```suggestion
        cell.attr('colspan', Number(colspan) + 1);
```

GuySartorelli (changes requested):

GuySartorelli (approved):
Looks good, thanks! Just letting CI run and I'll merge if it's green.

PR #1872, opened 2024-12-11
FIX Treat value not in SingleSelectField options as blank

GuySartorelli:
This makes the react dropdown fields behave more like the entwine ones when the DB value is not in the set of dropdown options.

Note that the reproduction steps in the issue rely on the DB field being `Int` (or perhaps other numeric fields as well - but specifically not a string type). This is because its default value is `0`, while the default for `Varchar` for example is `null` which wouldn't exhibit the bug.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11502


emteknetnz (changes requested):
This feels like a pretty messy way to handle this.  I think we're better off handling this on the PHP side if possible.  SingleSelectField::getDefaultValue() seems like the place where we should be checking if the value is in the source values and if not, then set the value to null.  The results of that will be passed to the form schema.

GuySartorelli:
Good point. Closing in favour of https://github.com/silverstripe/silverstripe-framework/pull/11510

PR #1880, opened 2025-01-06
API Remove deprecated extension hooks

GuySartorelli:
## Issue
- https://github.com/silverstripe/.github/issues/311

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -45,7 +45,6 @@ use SilverStripe\SiteConfig\SiteConfig;
 use SilverStripe\Model\ArrayData;
 use SilverStripe\View\Requirements;
 use SilverStripe\View\SSViewer;
-use SilverStripe\Dev\Deprecation;
 
 /**
  * LeftAndMain is the parent class of all the two-pane views in the CMS.
--- a/code/ModalController.php
+++ b/code/ModalController.php
@@ -10,7 +10,6 @@ use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Forms\Form;
-use SilverStripe\Dev\Deprecation;
 
 /**
  * Parent controller for all CMS-global modals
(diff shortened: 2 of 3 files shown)


code/LeftAndMain.php
```diff
@@ -45,7 +45,6 @@
 use SilverStripe\Model\ArrayData;
 use SilverStripe\View\Requirements;
 use SilverStripe\View\SSViewer;
-use SilverStripe\Dev\Deprecation;
```
ChloeSartorelli, line 48: Unused

code/ModalController.php
```diff
@@ -10,7 +10,6 @@
 use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Forms\Form;
-use SilverStripe\Dev\Deprecation;
```
ChloeSartorelli, line 13: Unused

code/ModelAdmin.php
```diff
@@ -321,19 +321,6 @@ protected function getGridFieldConfig(): GridFieldConfig
...
-        $config->removeComponentsByType(GridFieldFilterHeader::class);
-        $config->addComponent(Injector::inst()->createWithArgs(GridFieldFilterHeader::class, [
-            false,
-            function ($context) {
-                $this->extend('updateSearchContext', $context);
-            },
-            function ($form) {
-                $this->extend('updateSearchForm', $form);
-            }
-        ]));
```
ChloeSartorelli, line 335: We don't really have a way to deprecate these per-se, but they were removed in other admin sections already so it seems sensible to remove them from here as well.

This code was added in https://github.com/silverstripe/silverstripe-admin/pull/665/ which moved the search context from `ModelAdmin` into `GridFieldFilterHeader` - it seems clear the intention was to completely shift responsibility of managing it out of the controller and into the gridfield coponent.

Developers who need to update these can either:
1. update the component from their own subclass of `ModelAdmin`
2. update the search context in their `DataObject` subclass

EDIT:
These don't work anyway lol. The code for them was [deprecated](https://github.com/silverstripe/silverstripe-framework/pull/8420) in CMS 4 and then [subsequently removed](https://github.com/silverstripe/silverstripe-framework/pull/10594) in CMS 5.

GuySartorelli: That said, I don't feel strongly about this - if there's any push back at all I'm happy to just keep these and remove the comment that suggests they should be deprecated.

emteknetnz (approved):

PR #1896, opened 2025-01-30
NEW SudoModePasswordField component

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1898


code/CMSProfileController.php
```diff
@@ -57,6 +63,18 @@ public function getEditForm($id = null, $fields = null)
 
         $form->addExtraClass('member-profile-form root-form cms-edit-form center fill-height');
 
+        // tree_class must be chanaged to model_class for CMS 6
+        $dataClass = $this->config()->get('tree_class');
```
emteknetnz, line 67: Scenario is editing your own profile

code/LeftAndMain.php
```diff
@@ -1540,6 +1540,12 @@ public function getEditForm($id = null, $fields = null)
             $readonlyFields = $form->Fields()->makeReadonly();
             $form->setFields($readonlyFields);
         }
+
+        // Sudo mode
```
emteknetnz: Scenario is editing an individual record that is a record managed by a gridfield e.g. CMSProfileController will use this when editing the logged in users profile

code/ModelAdmin.php
```diff
@@ -274,6 +274,13 @@ public function getEditForm($id = null, $fields = null)
         $form->setFormAction($editFormAction);
         $form->setAttribute('data-pjax-fragment', 'CurrentForm');
 
+        // Sudo mode
```
emteknetnz: Scenario is a managed model on a model admin, for instance Member on SecurityAdmin (which extends ModelAdmin)

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -0,0 +1,112 @@
...
+import Config from 'lib/Config';
+import backend from 'lib/Backend';
+import PropTypes from 'prop-types';
+import qs from 'qs';
+import React, { createRef, useState } from 'react';
+
+function SudoModePasswordField(props) {
+  const {
+    className,
+    panelPadded,
```
emteknetnz: The "panelPadded" functionality in [this PR](https://github.com/silverstripe/silverstripe-cms/pull/3053) is related to making the field looking pretty "inline" which is only used in that PR which I think we should close. If we close that, then the panelPadded functionality should be removed from this PR.

client/src/components/SudoModePasswordField/SudoModePasswordField.js
ChloeSartorelli, line 1: Feels like this is duplicating a lot of what the `SudoMode` HOC is doing in `renderSudoModeNotice()` and `renderSudoModeVerification()` (not to mention actually handling verification) - should that HOC be updated to render this component instead?

client/src/components/SudoModePasswordField/SudoModePasswordField.scss
```diff
@@ -0,0 +1,55 @@
...
+  min-height: 108px;
+
+  @include media-breakpoint-up(lg) {
+    min-height: 140px;
+  }
+
+  .form__field-holder input {
+    display: none;
+  }
+}
```
ChloeSartorelli, line 13: We tend to use a loading component for this - why are we doing it differently here?

client/src/components/SudoModePasswordField/SudoModePasswordField.scss
```diff
@@ -0,0 +1,55 @@
...
+    margin-left: 3px;
+  }
+
+  &__verify {
+    margin-top: 1rem;
+  }
+
+  &__verify-form-group.form-group {
+    margin: 0;
+  }
```
ChloeSartorelli: Don't use `&__` because that makes it hard to search for e.g. `sudo-mode-password-field__inner`. I'm never going to think to break up a css class to search for it.

Either include the css class names in full, or rename them to something shorter if you think they'd be too long here in full.

client/src/components/SudoModePasswordField/SudoModePasswordField.scss
```diff
@@ -0,0 +1,55 @@
...
+
+  &__notice-help {
+    margin-left: 3px;
+  }
+
+  &__verify {
+    margin-top: 1rem;
+  }
+
+  &__verify-form-group.form-group {
```
ChloeSartorelli: Why does this need to specify both class names? Is `sudo-mode-password-field_verify-form-group` ever used without `form-group`? And are there ever any other `form-group` inside `sudo-mode-password-field`?

If no to either of those, update to just use the one class name here in the selector. No need to be more specific than is required, that just makes overriding styles harder.

client/src/components/SudoModePasswordField/SudoModePasswordField.scss
```diff
@@ -0,0 +1,55 @@
...
+    margin: 0;
+  }
+
+  // Reactstrap requires form feedback to be places in the same input group as the field
```
ChloeSartorelli:
```suggestion
  // Reactstrap requires form feedback to be placed in the same input group as the field
```

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
```diff
@@ -0,0 +1,63 @@
...
+  $('.js-injector-boot .SudoModePasswordField').entwine({
+    Component: null,
+    ReactRoot: null,
+
+    onmatch() {
+      // onmatch will match both the field holder and the field
+      // we only want to run this on the field holder
+      if (this.is('input')) {
+        return;
+      }
```
ChloeSartorelli: Just use `:not(:input)` or a css class that is only on the field holder in the selector above instead.

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
```diff
@@ -0,0 +1,63 @@
...
+      this._super();
+      const cmsContent = this.closest('.cms-content').attr('id');
+      const context = (cmsContent)
+        ? { context: cmsContent }
```
ChloeSartorelli:
```suggestion
      const cmsContentID = this.closest('.cms-content').attr('id');
      const context = (cmsContentID)
        ? { context: cmsContentID }
```

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
```diff
@@ -0,0 +1,63 @@
...
+      this._super();
+      const root = this.getReactRoot();
+      if (root) {
+        root.unmount();
+        this.setReactRoot(null);
+      }
+    },
+
+    /**
+     * This is used by the react component to reload the page after sudo mode is activated
```
ChloeSartorelli:
```suggestion
     * This is used by the react component to reload the page after sudo mode is activated.
     * This is here instead of inside the react component because in a pure react form the success handler
     * should result in re-rendering the react form with editable fields, instead of reloading the whole page.
```

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
```diff
@@ -0,0 +1,63 @@
...
+     */
+    reloadPage() {
+      // Add a ?reload=1 query parameter to the URL to force the browser to reload the page
+      // window.location.reload() does not work as expected
+      const query = qs.parse(window.location.search, { ignoreQueryPrefix: true });
+      const reload = query.reload ? parseInt(query.reload, 10) + 1 : 1;
+      const hrefNoQuery = window.location.href.split('?')[0];
+      window.location.href = hrefNoQuery + qs.stringify({ ...query, reload }, { addQueryPrefix: true });
+    },
```
ChloeSartorelli: This is probably not the right way to handle this - we only need to reload the form/gridfield, not the whole page. We should use one of the `refresh()`, `redraw()`, or `reloadCurrentPanel()` entwine functions - especially when it's for specifically just a gridfield on its own.

For example on a GridField use `refresh()`.
For whole forms, form submission uses `redraw()` which is a shortcut for `refresh()` with a something-extra that it does - I dunno if that something extra is necessary or not - or when `reloadCurrentPanel()` should be used instead. I'd try one of the first two first, and `reloadCurrentPanel()` as a last resort.

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
```diff
@@ -0,0 +1,63 @@
...
+        return;
+      }
+      this._super();
+      const cmsContent = this.closest('.cms-content').attr('id');
+      const context = (cmsContent)
+        ? { context: cmsContent }
+        : {};
+      const SudoModePasswordField = loadComponent('SudoModePasswordField', context);
+      this.setComponent(SudoModePasswordField);
+      this.refresh();
```
ChloeSartorelli, line 22: Why do we refresh on match? What other scenarios call refresh on this component?

code/CMSProfileController.php
```diff
@@ -57,6 +57,10 @@ public function getEditForm($id = null, $fields = null)
 
         $form->addExtraClass('member-profile-form root-form cms-edit-form center fill-height');
 
+        // We need to call requireSudoMode() here again, despite it being called in LeftAndMain::getEditForm(),
+        // so that the newly added Action goes into readonly mode
+        $form->requireSudoMode();
```
ChloeSartorelli, line 62: This suggests to me that the mechanism for `requireSudoMode()` isn't really suitable - there are plenty of scenarios where project-specific-code is also likely to add their own fields to a form, and they shouldn't have to call this method.

Perhaps `requireSudoMode()` should simply be a setter, and when the form is _used_ is when the transformation happens.

GuySartorelli (changes requested):

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: I don't this so, the HOC is basically redundant at this point because it's only used in a single place on the Member for, whose form is now protected by the SudoModePasswordField so you'll never actually see the HOC.

I'd be more inclined to just delete the HOC in CMS 6, though that said we may end up using it to protect another field at some point

client/src/components/SudoModePasswordField/SudoModePasswordField.scss
emteknetnz: I'm not sure what you are referring to here

client/src/components/SudoModePasswordField/SudoModePasswordField.scss
emteknetnz: Updated

client/src/components/SudoModePasswordField/SudoModePasswordField.scss
emteknetnz: It needs to be this specific to override `.cms-edit-form .form-group`

client/src/components/SudoModePasswordField/SudoModePasswordField.scss
emteknetnz: Updated

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: Updated

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: Updated

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: Updated

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: It would always be reload the entire form, because the sudo password fields are added to the form, i.e. when a relation gridfield is sudo protected you still need to click on a record to get an edit form with a sudo password field

What supposed to be the syntax for doing this? I tried `this.closest('.cms-edit-form').reloadCurrentPanel();` and the same for `refresh` and `redraw` but none of this actually did anything

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: refresh() is a method defined in this file. I based this file I think from LinkField.js which has the same thing.

code/CMSProfileController.php
emteknetnz: The problem is that getEditForm() is the last "central" place where we can do this i.e. several unrelated code-paths will call this

client/src/components/SudoModePasswordField/SudoModePasswordField.js
ChloeSartorelli: The HOC is not redundant (discussion about that in https://github.com/silverstripe/developer-docs/pull/684#discussion_r1958826344)

We should reduce code duplication, instead of increasing it which this PR seems to be doing.

client/src/components/SudoModePasswordField/SudoModePasswordField.scss
ChloeSartorelli: NEvermind, not worth arguing over

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
ChloeSartorelli: > It would always be reload the entire form, because the sudo password fields are added to the form

Fair enough. However, reloading the form is not the same as reloading the _whole page_ which the current code is doing.

I'll have a quick play and see if I can get a working example of using those functions.

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
ChloeSartorelli: If you're only intending for it to be called on match, just move the code straight into the onmatch event handler.
This makes it look like `refresh()` will be called from other places in some scenarios.

code/CMSProfileController.php
ChloeSartorelli: That makes `getEditForm()` a great place to call a _setter_ method. But then somewhere in `Form` itself the actual transformation could happen, before the form itself is used in the front-end (and by frontend here I mean the client-side of the CMS).

If there are multiple methods in `Form` that are used to present it in the frontend, then they could all call a private `performSudoTransformation()` method for example.

That way you only need to call `setRequireSudo()` once - and developers don't have to be aware that adding fields to a form might require them to call it themselves, because they wouldn't have to anymore.

GuySartorelli:
Sorry, forgot GitHub doesn't let me start a review from this view.
New comments:
- https://github.com/silverstripe/silverstripe-admin/pull/1896#discussion_r1958831400
- https://github.com/silverstripe/silverstripe-admin/pull/1896#discussion_r1958834369
- https://github.com/silverstripe/silverstripe-admin/pull/1896#discussion_r1958835049
- https://github.com/silverstripe/silverstripe-admin/pull/1896#discussion_r1958836047

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
ChloeSartorelli: Looks like this is the way:
```js
this.closest('.cms-container').reloadCurrentPanel();
```

There is a bug in the `reloadCurrentPanel()` implementation though, so the following change needs to be done in https://github.com/silverstripe/silverstripe-admin/blob/9bc9af892de849ecdcd1c0a8fbb58d7ce49ed4df/client/src/legacy/LeftAndMain.js#L438-L440

```diff
 reloadCurrentPanel: function() {
-  this.loadPanel(document.URL, null, null, true);
+  this.loadPanel(document.URL, '', {}, true);
 },
```

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: I think it means the component will rerender in some scenarios, like if the large fragment of html gets pjaxed in. Maybe?

It's following an existing pattern will use for other things e.g. LinkField. I don't know why we have that pattern, maybe it does nothing, though I see no downside in just following it for better standardization.

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: It's nicer this way, though doesn't quite work in all scenarios. When activating sudo mode from a record being edited from a gridfield e.g. a regular member record in security admin, on reload the form is still in read-only mode

I've added this as an inline comment explaining why `reloadCurrentPanel()` is not being used

code/CMSProfileController.php
emteknetnz: I did try to find such a locations originally, though I ended up at getEditForm() because that was the best location I could find. If you know of a better location then let me know, otherwise I don't know where else this could go.

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: Leaving aside the HOC's utiltity (which is now very low IMO, all it does it stop a react component from rendering, it doesn't protect a FormField, sorry should have made that clearer)

The HOC and the new component have several differences
- different messages
- different styling
- the HOC has an inner `<WrappedComponent>` which the password field does not
- the HOC has uses a `<Loading>` component which may create a fout for no benefit
- the HOC is a class component, password field is a functional component

If we did a refactor we need to create a third shared component that would need to be moderately complex to work in both scenarios.

I just don't think it's worthwhile. Maybe create a new card to refactor later?

client/src/legacy/LeftAndMain.js
```diff
@@ -436,7 +436,7 @@ $.entwine('ss', function($) {
      * Nice wrapper for reloading current history state.
      */
     reloadCurrentPanel: function() {
-      this.loadPanel(document.URL, null, null, true);
+      this.loadPanel(document.URL, '', {}, true);
```
emteknetnz, line 439: Unrelated fix from https://github.com/silverstripe/silverstripe-admin/pull/1896/files#r1958854697

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
ChloeSartorelli: That pattern is in other places explicitly because `refresh()` is intended to be called from various scenarios on some components.

It's probably in linkfield erroneously, and causes confusion having this method which is used elsewhere for specific purposes but is here just to be here, because it was seen somewhere else. This method is causing confusion and from what I can tell actually breaking standardisation by making it unclear when the `refresh()` method is serving a purpose and when it's not.

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
ChloeSartorelli: For me even the current reload page doesn't work for that same scenario.

The reason is that when you click on a record in the gridfield, because the gridfield is set to be "readonly" it uses the `GridFieldViewButton` instead of the `GridFieldEditButton` - so the URL looks like `https://sudowudo.ddev.site/admin/security/users/EditForm/field/users/item/93/view?gridState-users-0=%7B%22Readonly%22%3Atrue%7D` instead of like `https://sudowudo.ddev.site/admin/security/users/EditForm/field/users/item/93/?gridState-users-0=%7B%22Readonly%22%3Atrue%7D` (note the explicit `/view` suffix)

One way to fix this would be to keep the edit button (so we get the correct URL) and trust `GridFieldDetailForm_ItemRequest` to ensure the form itself is in readonly mode.

code/CMSProfileController.php
ChloeSartorelli: One option is:
- Everything currently calling `requireSudoMode()` instead calls `setRequireSudoMode()` which just sets a boolean property.
- A new `performSudoModeTransformation()` or similar method is created which has all this logic - with an early exit if sudo mode isn't required.
- Call the new method from the following places:
  - `FormRequestHandler::httpSubmission()`
  - `Form::forTemplate()`
  - `Form::forAjaxTemplate()`
  - `Form::renderWithoutActionButton()`
  - `FormSchema::getSchema()` (which also covers things like the CMS 6 `FormSchemaController` and its CMS 5 counterpart in `LeftAndMain`

There are some trade offs here and I'm happy to discuss them.

The main benefit is that you only ever have to call `setRequireSudo()` once. This means developers adding new fields to the form don't have to even think about whether their new field is protected by sudo mode or not, which IMO is what makes this worth while.

The main drawback is that if developers are using the form in a way that doesn't go through one of the above mentioned methods, they will need to know to call `performSudoModeTransformation()` on the form to ensure sudo mode protects the form.

Looks like for this to work we also need to clear the `obj` cache for both the form and the fields during the transformation - something is trying to render the form too early, which results in building a non-sudomode-cache for the form and for the fields, which means even though the form becomes readonly and has an extra field, the cached result used in templates doesn't reflect that.

client/src/components/SudoModePasswordField/SudoModePasswordField.js
ChloeSartorelli: I do think it's worthwhile, and I'm hesitant to merge work that we intend to completely redo later - seems sensible to just reduce the code reuse right now rather than adding duplicate code now only to remove it later.

> different messages

That can be a prop, or we can just have wording that serves both purposes simultaneously.

> different styling

They look basically the same to me? If there's different CSS to achieve the same result, that just proves my point.
image
image

> the HOC has an inner `<WrappedComponent>` which the password field does not

What does that do? Is it needed? Does it do harm if added to the password field?

> the HOC has uses a `<Loading>` component which may create a fout for no benefit

That brings me back to https://github.com/silverstripe/silverstripe-admin/pull/1896#discussion_r1955340888 - but ultimately if we've got something there that isn't beneficial we should remove it. The HOC using a loading component to do what your CSS does, so pick one and run with it.

> the HOC is a class component, password field is a functional component

That's fine, the HOC wouldn't be outright _replaced_ with the new component, it would internally render it instead of `renderSudoModeNotice()` and `renderSudoModeVerification()`

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: Have updated the framework PR so that `/view` is not suffixed, so now we can use `reloadCurrentPanel();`

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: :shrug: Honestly it's doing no harm so let's just keep as is, I don't think we need to go back and forth over this

code/CMSProfileController.php
emteknetnz: > Call the new method from the following places:

Yeah that's my point, you've listed five places we need to make the call, and it's easy to miss in any future code paths. The call isn't being made centrally

Since sudo mode is deactivated by default, if fields have been added in say an extension hook, it makes it real easy to spot that there's say one writable field in a sea of read-only fields and that something needs remedied, as opposed sudo mode not being activated is a potentially backend only code path e.g. form submission endpoint, where it's hard to see

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: > we intend to completely redo later

I simply don't think this is worthwhile. I don't think there's a real world use case for the HOC. I cannot think of a single react component that we'd protect with it right now.

I don't want to any spend time refactoring something where there's no intended use for the component

ChloeSartorelli: Discussed offline - I don't understand your reasoning for not reducing the amount of code duplication but since you utterly refuse to do it, I'll allow the PR to be merged with the duplication under the condition that the beta isn't released until I have been given a fair chance to refactor it myself.

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
ChloeSartorelli: I think it is doing harm, as I've already mentioned, but you seem to be adamant to not make changes so I'll include this in my refactoring after this PR gets merged.

code/CMSProfileController.php
ChloeSartorelli: > if fields have been added in say an extension hook, it makes it real easy to spot that there's say one writable field in a sea of read-only fields

It requires the developer to notice something that isn't providing a warning. I don't think it would be that easy to notice in a spot check especially when that's not something you'd think you have to check for.

> as opposed sudo mode not being activated is a potentially backend only code path e.g. form submission endpoint, where it's hard to see

I think a whole form not being readonly is _way_ easier to see than a single field not being readonly.

Regardless it seems you're pretty over this review cycle, which is understandable. If you insist on keeping it as is, the following must be done:
Add a clear `[!WARNING]` block to the docs that adding new fields to the form may mean having to call `requireSudoMode()` again on the form

emteknetnz: Have updated docs PR to add `[!WARNING]`

GuySartorelli (approved):
LGTM with the acknowledged caveats. Will merge when other PRs are approved

PR #1900, opened 2025-02-12
TLN Update translations

emteknetnz:
Automated translations update generated using [silverstripe/tx-translator](https://github.com/silverstripe/silverstripe-tx-translator)


GuySartorelli (approved):
LGTM
There's some garbage in that yml but it looks like it's coming from Transifex, and we don't police what people put in there.

PR #1902, opened 2025-02-18
API Remove code which is being migrated to a new TinyMCE module

GuySartorelli:
## Issue
- https://github.com/silverstripe/silverstripe-htmleditor-tinymce/issues/2


code/FormSchemaController.php
```diff
@@ -33,14 +30,6 @@ abstract class FormSchemaController extends AdminController
      */
     private ?FormSchema $schema = null;
 
-    protected function init()
-    {
-        $this->beforeExtending('onInit', function () {
-            $this->prepareTinyMce();
-        });
-        parent::init();
-    }
```
GuySartorelli, line 42: Doesn't need to be deprecated since the method still exists in the parent class.

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -300,21 +300,7 @@ $.entwine('ss', function($){
...
-        var self = this,
-          field = $(e.target).closest('.field.htmleditor'),
-          editor = field.find('textarea.htmleditor').getEditor().getInstance();
-
-        // TinyMCE 4 will add a focus event, but for now, use click
-        editor.onClick.add(function(e){
-          self.saveFieldFocus(field.attr('id'));
-        });
-      }
-    },
```
ChloeSartorelli, line 317: This wasn't being hit - I assume that even doesn't exist in TinyMCE 6.

client/src/legacy/jquery.changetracker.js
```diff
@@ -90,21 +90,9 @@ import debounce from 'lodash.debounce';
...
-
-              var serializer = tinymce.html.Serializer(config);
-              var parser = tinymce.html.DomParser(config);
-
-              value = serializer.serialize(parser.parse(value));
-              break;
-          }
+        if ($field && $field.is('textarea.htmleditor')) {
+          const editor = $field.entwine('ss').getEditor();
+          value = editor.prepValueForChangeTracker(value);
```
ChloeSartorelli, line 95: Moved the TinyMCE specific code into the TinyMCE module and generalised it on this end so other editors can be given the same treatment if needed.

code/FormSchemaController.php
```diff
@@ -35,9 +34,11 @@ abstract class FormSchemaController extends AdminController
...
     protected function init()
     {
-        $this->beforeExtending('onInit', function () {
-            $this->prepareTinyMce();
-        });
+        // Set the members html editor config
+        $currentUser = Security::getCurrentUser();
+        if ($currentUser) {
+            HTMLEditorConfig::set_active_identifier($currentUser->getHTMLEditorConfigForCMS());
+        }
```
ChloeSartorelli, line 41: Since only this tiny part is needed here now, there's no need to have the extra method.

emteknetnz (approved):

Issue #1895, opened 2025-01-26
CMS 6 REGRESSION - Archiving a page with bulk actions results in server error


emteknetnz:
CMS 6.0

When using bulk actions on a page and selecting "Unpublish and Archive" the following server error can be seen in the network panel

`ERROR [Warning]: Attempt to read property "TreeTitle" on null IN POST /admin/pages/batchactions/archive Line 114 in /var/www/vendor/silverstripe/admin/code/CMSBatchAction.php`

## PR
- https://github.com/silverstripe/silverstripe-admin/pull/1903

GuySartorelli:
PR merged

PR #1903, opened 2025-02-19
FIX Ensure record exists before trying to update TreeTitle

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1895

Regression introduced [here](https://github.com/silverstripe/silverstripe-admin/commit/b42c820c44a90b10136d9ce78629c1119eb74c65#diff-d9091c39d47fdfe4f05392bdd1ada56cc04b05009a1c7005c8a0215abb33f023R114)

Issue is that while the page exists at the start of [SilverStripe\Admin\CMSBatchAction::batchaction()](https://github.com/silverstripe/silverstripe-admin/blob/2/code/CMSBatchAction.php#L95), the call to `call_user_func_array([$obj, $helperMethod], $arguments ?? [])` will call `doArchive()` meaning that the future assignment to `$publishedRecord` will be `null`, meaning that `$publishedRecord->TreeTitle;` will fail

### DIFF
--- a/code/CMSBatchAction.php
+++ b/code/CMSBatchAction.php
@@ -108,12 +108,12 @@ abstract class CMSBatchAction
 
             // Now make sure the tree title is appropriately updated
             $publishedRecord = DataObject::get_by_id($this->managedClass, $id);
-            if ($publishedRecord instanceof SiteTree) {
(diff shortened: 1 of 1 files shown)


GuySartorelli (approved):
LGTM

Issue #1904, opened 2025-02-19
Refactor Sudo mode components


emteknetnz:
Following on from https://github.com/silverstripe/silverstripe-admin/pull/1896#discussion_r1960691831 and https://github.com/silverstripe/silverstripe-admin/pull/1896#discussion_r1960545480

## Acceptance criteria
- Refactor the sudo mode components so that they share code and styles
- Do not use the the refresh() method in SudoModePasswordFieldEntwine.js
- ~~Ensure this is run through a kitchen sink CI run~~

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1905

GuySartorelli:
Not doing a sink run because nothing in sink uses the HOC, so there's no point.

PR #1905, opened 2025-02-20
ENH Refactor sudo mode components to reduce code reuse

GuySartorelli:
Tested the HOC locally by reinstating the HOC to MFA (i.e. undo the JS changes from https://github.com/silverstripe/silverstripe-mfa/pull/587) and removing the calls to `requireSudoMode()`.

Comparison of visual appearance is below. Functionality should also be identical to what it was without this PR.

If anyone customised the HOC styling or was relying on those selectors they'll be out of luck - I can't be assed updating the new component to use the old css class names so I used its css instead of what was there prior to the recent work.

## OLD look

Screenshot from 2025-02-19 11-37-40
Screenshot from 2025-02-19 11-37-57

## NEW look

Screenshot from 2025-02-20 15-05-16
Screenshot from 2025-02-20 15-05-34

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1904


client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -80,12 +80,16 @@ function SudoModePasswordField(props) {
    */
   function renderConfirm() {
     const helpLink = clientConfig.helpLink;
+    let { verifyMessage } = props;
```
emteknetnz: At the top of the component onSuccess and autocomplete props are assigned to variables, do the same with this prop, and use a different variable for verifyMessage below

emteknetnz (changes requested):

client/src/components/SudoModePasswordField/SudoModePasswordField.js
ChloeSartorelli: > At the top of the component onSuccess and autocomplete props are assigned to variables, do the same with this prop

... It.... *is* assigned to a variable. What do you mean?

ChloeSartorelli: Reading again this morning... I *think* you mean you want this variable assignment to be done in the same place as the other two? That's a very subjective change, which to my eye is not as good as what I've done, since this is only used right here, not elsewhere in the file.
But in the interest of unlocking the beta, I'll make that change.

ChloeSartorelli: Done. I don't like having two variables for the price of one, and I couldn't think of a good name for the new variable because the name of the variable should be `verifyMessage` - but I've made the requested change.

emteknetnz (approved):
Seems to work well

PR #1906, opened 2025-02-24
FIX Handle 401 from server

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/368

Requires https://github.com/silverstripe/silverstripe-frameworktest/pull/228

Fixes https://github.com/silverstripe/silverstripe-admin/actions/runs/13491490322/job/37690345812#step:12:163

tests/behat/features/form-sudo-mode.feature:61

CMS 6 controller endpoints now returns a 401 instead of a 200 https://github.com/silverstripe/silverstripe-admin/compare/2...3#diff-038d2edcd2449c0dc4bf092c2aae85c43ee7c6300769e194d98787aacc70a178R92


client/src/components/SudoModePasswordField/tests/SudoModePasswordField-test.js
```diff
@@ -16,11 +16,29 @@ window.ss.config = {
...
 
 let doResolve;
+let doReject;
+
+beforeEach(() => {
+  doResolve = undefined;
+  doReject = undefined;
+});
+
+function createJsonError(message) {
```
emteknetnz, line 26: Copied from approach used in https://github.com/silverstripe/silverstripe-elemental/blob/6/client/src/components/ElementActions/tests/ArchiveAction-test.js#L82

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -52,12 +52,16 @@ function SudoModePasswordField(props) {
...
-    if (responseJson.result) {
-      onSuccess();
-    } else {
-      setResponseMessage(responseJson.message);
-    }
+    fetcher(data, headers)
+      .then(responseJson => {
+        if (responseJson.result) {
+          onSuccess();
+        }
```
ChloeSartorelli: What is the condition where this is false? Is that some sort of uncaught error?

GuySartorelli:
Can you please check if the failing behat is relevant?

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: Actually the `result` bool is actually kind of redundant as in CMS 6 it's fully use status codes - https://github.com/silverstripe/silverstripe-admin/blob/3.0/code/SudoModeController.php#L97 - this wasn't the case in CMS 5

I'll simply remove the check for 'result' and remove it from the endpoint

code/SudoModeController.php
```diff
@@ -79,22 +79,20 @@ public function activate(HTTPRequest $request): HTTPResponse
 
         if (!SecurityToken::inst()->checkRequest($request)) {
             return $this->jsonResponse([
-                'result' => false,
```
emteknetnz, line 82: Remove 'result' from data as we should simply rely on the status code instead

GuySartorelli (approved):
LGTM

PR #1908, opened 2025-02-25
API Deprecate API being removed in CMS 6

GuySartorelli:
Note that this PR requires https://github.com/silverstripe/silverstripe-framework/pull/11623 to be merged first, since that adds the new constant.

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2949

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -67,8 +67,9 @@ class LeftAndMain extends Controller implements PermissionProvider
 
     /**
      * Form schema header identifier
+     * @deprecated 2.4.0 use SilverStripe\Forms\Schema\FormSchema::SCHEMA_HEADER instead.
      */
-    const SCHEMA_HEADER = 'X-Formschema-Request';
+    const SCHEMA_HEADER = FormSchema::SCHEMA_HEADER;
 
     /**
      * Enable front-end debugging (increases verbosity) in dev mode.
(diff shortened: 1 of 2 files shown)


code/LeftAndMain.php
```diff
@@ -67,8 +67,9 @@ class LeftAndMain extends Controller implements PermissionProvider
 
     /**
      * Form schema header identifier
+     * @deprecated 2.4.0 use SilverStripe\Forms\Schema\FormSchema::SCHEMA_HEADER instead.
      */
-    const SCHEMA_HEADER = 'X-Formschema-Request';
+    const SCHEMA_HEADER = FormSchema::SCHEMA_HEADER;
```
emteknetnz, line 72: This looks really weird, can we just keep the existing string. There's pretty much zero risk of it getting changed.

emteknetnz: Also shouldn't we update the CMS 5 references to this that to use the new const on FormSchema instead?

emteknetnz (changes requested):

code/LeftAndMain.php
ChloeSartorelli: > This looks really weird, can we just keep the existing string. 

I don't think it looks weird - we're setting the value to the correct constant. It's no different than when we make a deprecated method return the value of the new method.

> Also shouldn't we update the CMS 5 references to this that to use the new const on FormSchema instead?

Yes, sorry, I missed that. Done.

emteknetnz (approved):

Issue #1907, opened 2025-02-24
CMS 5.4 REGRESSION: Sudo protected relation gridfield with no records has no indication its protected


emteknetnz:
Applying the frameworktest [MemberRelationsPageExtension](https://github.com/silverstripe/silverstripe-frameworktest/blob/1/code/sudo-mode/MemberRelationsPageExtension.php) to a Page will look like this - there's no indication that the model the gridfield is managing can be added to or edited when you disable sudo mode

![Image](https://github.com/user-attachments/assets/199a1e59-f30a-4fd1-83d2-269d536a64c8)

We need some sort of UX saying that you need to activate sudo mode in order to use the gridfield

## Acceptance criteria
- [x] A sudo mode password like component is rendered above each GridField that is protected by sudo mode
- [x] It starts off in a collapsed state e.g. you click on it to expand it to see the full component
- [x] The UX is clear that an action needs to be taken and this is how you do it
- [x] When sudo mode is activated, all GridFields that are protected in the form refresh (not non-sudo protected ones)
- [x] This new thing doesn't show on forms already protected by sudo mode e.g. the Member GridField in SecurityAdmin
 
## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/13934758553

## PRs 
- https://github.com/silverstripe/silverstripe-admin/pull/1912
- https://github.com/silverstripe/silverstripe-framework/pull/11631

GuySartorelli:
@emteknetnz The only PRs are still in draft

emteknetnz:
Opps, set ready for review now

GuySartorelli:
PRs merged

PR #1912, opened 2025-03-10
ENH Show collapsed sudo message on GridFields

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1907


GuySartorelli (changes requested):
It's not visually obvious to me that the "section" mentioned is the gridfield.

image

Can we update this to say something like "The {$field->getName()} field is protected and is in read-only mode"?

It's also confusing to me that the message changes after clicking "verify to continue":

image

image

For context, here's the non-collapsed version in a profile page:

image
image

I honestly think the "collapsed" functionality is adding confusion rather than reducing it - and isn't much smaller than the regular non-collapsed field prior to clicking the button.

emteknetnz:
The collapsed fields are a big visual improvement over non collapsed IMO 

image

vs

image

I'll see what I can do re messaging etc

emteknetnz:
The collapsed components are a big visual improvement over non collapsed IMO in terms of less annoyance / better vertical real estate, and we should use them.

image

**VS**

image

I'll see what I can do re messaging etc

emteknetnz:
Have discussed offline, we'll get it to look like this

image

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
```diff
@@ -47,8 +49,14 @@ jQuery.entwine('ss', ($) => {
      * This is here instead of inside the react component because in a pure react form the success handler
      * should result in re-rendering the react form with editable fields, instead of reloading the whole page.
      */
-    reloadPage() {
-      this.closest('.cms-container').reloadCurrentPanel();
+    reloadSection() {
+      const gridField = this.closest('.ss-gridfield');
+      if (gridField.length) {
```
ChloeSartorelli: I'm not convinced this is a robust way to check if the field is for a gridfield or not. Probably better to add a `data-gridfield="true"` or similar attribute when it's for a gridfield instead.
You can set attributes on the field using `$field->setAttribute()`

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -15,14 +15,41 @@ function SudoModePasswordField(props) {
...
 
+  let verifyMessageValue = verifyMessage;
+  if (!verifyMessageValue) {
+    if (sectionTitle) {
+      verifyMessageValue = i18n.inject(
+        i18n._t(
+          'Admin.SUDO_MODE_PASSWORD_FIELD_VERIFY_SECTION_TITLE',
+          '{sectionTitle} is protected and is in read-only mode. Before editing please verify that it\'s you first.'
+        ),
+        { sectionTitle, }
```
ChloeSartorelli:
```suggestion
        { sectionTitle }
```

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -15,14 +15,41 @@ function SudoModePasswordField(props) {
...
 
   const clientConfig = Config.getSection('SilverStripe\\Admin\\SudoModeController');
 
+  let verifyMessageValue = verifyMessage;
+  if (!verifyMessageValue) {
+    if (sectionTitle) {
+      verifyMessageValue = i18n.inject(
+        i18n._t(
+          'Admin.SUDO_MODE_PASSWORD_FIELD_VERIFY_SECTION_TITLE',
+          '{sectionTitle} is protected and is in read-only mode. Before editing please verify that it\'s you first.'
```
ChloeSartorelli:
```suggestion
          '"{sectionTitle}" is protected and is in read-only mode. Before editing please verify that it\'s you first.'
```

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -15,14 +15,41 @@ function SudoModePasswordField(props) {
   const {
     onSuccess,
     autocomplete,
+    collapsed,
```
ChloeSartorelli: Rename the prop to "initiallyCollapsed" - otherwise `collapsed` and `expanded` can both be true at the same time, which is wrong.

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -15,14 +15,41 @@ function SudoModePasswordField(props) {
...
   const passwordFieldRef = createRef();
   const [responseMessage, setResponseMessage] = useState('');
   const [showVerify, setShowVerify] = useState(false);
+  // setting to null initially to prevent a fout as useEffect()
+  // below will cause a re-render when collapsed
+  const [expanded, setExpanded] = useState(null);
+
+  useEffect(() => {
+    setExpanded(!collapsed);
+  }, [collapsed]);
```
ChloeSartorelli: [Screencast from 12-03-25 17:37:35.webm](https://github.com/user-attachments/assets/389cfbe8-cec3-4b3b-8a3e-9b8e5a66c212)

This is what it looks like for me when I make the following change:
```suggestion
  const [expanded, setExpanded] = useState(!collapsed);
```

There is the barest almost-imperceptible moment while the field isn't visible - I think that's fine. It would only be a problem if it was rendering something else first, which it isn't.

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -145,19 +170,38 @@ function SudoModePasswordField(props) {
     </div>;
   }
 
+  // Prevent a fout on initial render when rendering collapsed
+  if (expanded === null) {
+    return null;
+  }
+
```
ChloeSartorelli:
```suggestion
```
As per https://github.com/silverstripe/silverstripe-admin/pull/1912#discussion_r1990545659

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
```diff
@@ -47,8 +49,14 @@ jQuery.entwine('ss', ($) => {
...
+        // Reload all sudo protected gridfields (not just this one)
+        gridField.closest('.cms-container').find('.ss-gridfield').reload();
```
ChloeSartorelli:
```suggestion
        this.closest('.cms-container').find('.ss-gridfield').reload();
```

GuySartorelli (changes requested):
Alignment doesn't seem quite right - in split view it's hard against the preview panel on the right

image

The font is tiny - `11.999px` vs `13px` for the non-gridfield version.

Finally - and I know I keep harping on about this, but the "collapsed" concept seems a bit weird to me.
- Both versions have some text, a link, and a button.
- For both versions, clicking the button hides the button and renders a text field and a new button.

Seems like we don't need separate js logic, just separate CSS? Even if a prop is the way to determine which to use (`isCondensed` might be appropriate), the only difference really is just where that button lives (under or next to the text)

GuySartorelli:
Also just realised the "what's this?" link is missing in the gridfield version

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: OK, looked a lot worse on my local.  I believe it's also necessary for storybook so that you can use the controls to toggle between collapsed and not collapsed, so will leave as is

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: Updated

client/src/legacy/SudoModePasswordField/SudoModePasswordFieldEntwine.js
emteknetnz: Updated

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: Will leave as is

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: Updated

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: Updated

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: Updated

emteknetnz:
> Also just realised the "what's this?" link is missing in the gridfield version

Have added in

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -15,14 +15,50 @@ function SudoModePasswordField(props) {
...
   const passwordFieldRef = createRef();
   const [responseMessage, setResponseMessage] = useState('');
   const [showVerify, setShowVerify] = useState(false);
+  // setting to null initially to prevent a fout as useEffect()
+  // below will cause a re-render when initiallyCollapsed
+  const [expanded, setExpanded] = useState(null);
+
+  useEffect(() => {
+    setExpanded(!initiallyCollapsed);
+  }, [initiallyCollapsed]);
```
ChloeSartorelli: As per https://github.com/silverstripe/silverstripe-admin/pull/1912#pullrequestreview-2676837781 I don't think this additional state is necessary.

This PR effectively moved the "verify" button from below the text to be next to it, right? (There are other changes like adding `sectionTitle` but they aren't affected by this state).
The button does the same thing in both cases - hides the "verify" button and displays a text field and a submit button.

I don't understand why this state is needed for that given it's the same functionality in both cases.

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -145,19 +173,38 @@ function SudoModePasswordField(props) {
...
   // Render the component
   return <div className="sudo-mode-password-field">
-    <div className="sudo-mode-password-field-inner alert alert-info panel panel--padded">
+    { !expanded && <div className="sudo-mode-password-field__expander alert alert-info panel">
+      <div className="sudo-mode-password-field__expander-text-container">{ verifyMessageValue }</div>
+      <div className="sudo-mode-password-field__expander-button-container">
+        <Button
+          className="sudo-mode-password-field__expander-button font-icon-lock"
+          color="info"
+          onClick={() => handleExpand()}
```
ChloeSartorelli:
```suggestion
          onClick={() => handleConfirmClick()}
```
Would this not work? And make this button render depend on the value of `showVerify`?
i.e `if !showVerify && condensed` where `condensed` replaces `initiallyCollapsed` and no longer has any state associated with it?

client/src/components/SudoModePasswordField/SudoModePasswordField.js
```diff
@@ -145,19 +173,38 @@ function SudoModePasswordField(props) {
...
+        >
+          { i18n._t('Admin.VERIFY', 'Verify') }
+        </Button>
+      </div>
+    </div>}
```
ChloeSartorelli: This should probably be in a function, like the other `render<SomethingSpecific>` functions - though given how many review cycles this has had and that this doesn't affect the end result other than making it a little easier to visually parse the code, I won't hold off merging if this doesn't get done.

GuySartorelli (changes requested):

GuySartorelli:
When trying this with multiple sudo-mode gridfields on a page, verifying seemed to refresh all of the sudo mode gridfields, but only the one I verified on was changed from being read-only.

emteknetnz:
> When trying this with multiple sudo-mode gridfields on a page, verifying seemed to refresh all of the sudo mode gridfields, but only the one I verified on was changed from being read-only.

Are you sure? Works fine on my local.  I authenticated using protected gridfield a), and then was able to delete a record on protected gridfield b) via the little button on the gridfield row

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: Have removed expanded state

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: Updated

client/src/components/SudoModePasswordField/SudoModePasswordField.js
emteknetnz: Leaving as is

GuySartorelli:
>> When trying this with multiple sudo-mode gridfields on a page, verifying seemed to refresh all of the sudo mode gridfields, but only the one I verified on was changed from being read-only.
> Are you sure? Works fine on my local. I authenticated using protected gridfield a), and then was able to delete a record on protected gridfield b) via the little button on the gridfield row

After further testing it turns out it's because I sorted the gridfield first. If you sort a gridfield (likely other actions like filtering too) before verifying, then it will remain in read-only mode after being refreshed.

emteknetnz:
> After further testing it turns out it's because I sorted the gridfield first.

Have added new functionality to GridField.js to allow removing the gridifield readonly value from the gridstate stored in the url

GuySartorelli (approved):
LGTM, works as expected. Will re-run CI after the framework PR is merged.

PR #1918, opened 2025-03-27
FIX Debounce on SearchableDropdownField

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11652


GuySartorelli (approved):
LGTM, works as expected

Issue #1920, opened 2025-03-30
CMS 6 REGRESSION: Loading ellipses are hard to see in modal edit form buttons


GuySartorelli:
Discovered when regression testing CMS 6

In a modal edit form, after clicking a button that results in loading, the ellipses on the button are hard to see:

![ellipses are dark grey on a dark green background](https://github.com/user-attachments/assets/8ba6eefe-b804-4e80-9910-2e70469d36d2)

Note that these are the same ellipses used for the save/publish buttons in regular CMS edit forms, but the background for those turns white while loading.

Check if CMS 5 and CMS 6 are the same before starting to implement a fix.

## Steps to reproduce

> [!WARNING]
> These steps rely on having `silverstripe/linkfield` installed as an easy way to set up a modal edit form.

1. Log into CMS as a content author
1. Go to a page that has a `LinkField`
1. Click "Add link" and pick any link type
1. Fill in the form however you choose
1. Click "Create link"

> [!NOTE]
> The loading can happen very quickly in a local dev environment. Consider enabling throttling in your browser's dev console.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1924

GuySartorelli:
PR merged

PR #1924, opened 2025-04-07
FIX Use white ellipses

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1920

This style was previously in CMS 5 - https://github.com/silverstripe/silverstripe-admin/blob/2/client/src/components/FormAction/FormAction.scss#L151


GuySartorelli (approved):
LGTM, works as expected

PR #1932, opened 2025-05-13
TLN Update translations

GuySartorelli:
Automated translations update generated using [silverstripe/tx-translator](https://github.com/silverstripe/silverstripe-tx-translator)

PR #1933, opened 2025-05-14
FIX Fix localisation entities for text collection

GuySartorelli:
Resolves a problem with collected localisation strings identified when running tx-translator:

- https://github.com/silverstripe/silverstripe-admin/pull/1932#discussion_r2085924448

## Issue
- https://github.com/silverstripe/.github/issues/400

### DIFF
--- a/code/AdminController.php
+++ b/code/AdminController.php
@@ -12,6 +12,7 @@ use SilverStripe\Control\HTTPResponse_Exception;
 use SilverStripe\Control\Middleware\HTTPCacheControlMiddleware;
 use SilverStripe\Forms\HTMLEditor\HTMLEditorConfig;
 use SilverStripe\i18n\i18n;
+use SilverStripe\i18n\i18nEntityProvider;
 use SilverStripe\Security\Permission;
 use SilverStripe\Security\Security;
 use SilverStripe\Versioned\Versioned;
@@ -23,7 +24,7 @@ use SilverStripe\View\SSViewer;
  * This class is automatically routed via the AdminRootController.
  * It's responsible for ensuring permissions are respected.
  */
-abstract class AdminController extends Controller
+abstract class AdminController extends Controller implements i18nEntityProvider
 {
     /**
      * The current url segment attached to the controller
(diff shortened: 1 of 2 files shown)


code/CMSMenu.php
```diff
@@ -429,7 +429,7 @@ public function provideI18nEntities()
         foreach ($cmsClasses as $cmsClass) {
             $defaultTitle = LeftAndMain::menu_title($cmsClass, false);
             $ownerModule = ClassLoader::inst()->getManifest()->getOwnerModule($cmsClass);
-            $entities["{$cmsClass}.MENUTITLE"] = [
+            $entities["{$cmsClass}.MENUTITLE_CMS6"] = [
```
ChloeSartorelli: I think it's important in this case to distinguish specifically what version the new key becomes used. I'm open to other ways to do that, but I don't want to just use e.g. `MENUTITLE2` because it doesn't make it clear enough what the distinction is between the old and the new keys.

ChloeSartorelli: That said - the old localised values (e.g. "edit pages" for `CMSPageEditController`) isn't actually used in the UI, because it was using a hardcoded value for the title in CMS 5.

I think honestly the best thing to do is _not_ change this key, and let the old values be overridden in CMS 5 knowing they're not actually used for the controllers where it's changing.

That is an exception to our sort-of-policy about how to handle these cases, but I honestly think it will result in the least amount of disruption.

code/AdminController.php
ChloeSartorelli, line 1: Note that the changes in this file _are_ needed, regardless of what happens with the other bits.

code/AdminController.php
```diff
@@ -325,17 +326,13 @@ protected function jsonError(int $errorCode, string $errorMessage = ''): void
...
-                default => 'Error',
-            };
-            /** @phpstan-ignore translation.key (we need the key to be dynamic here) */
-            $error['value'] = _t(__CLASS__ . ".ErrorMessage{$errorCode}", $messageDefault);
+            $messageDefault = $this->getErrorCodeDefaultMessages()[$errorCode] ?? null;
+            if ($messageDefault) {
+                /** @phpstan-ignore translation.key (we need the key to be dynamic here) */
+                $error['value'] = _t(__CLASS__ . ".ErrorMessage{$errorCode}", $messageDefault);
+            } else {
+                $error['value'] = _t(__CLASS__ . ".ErrorMessage", 'Sorry, it seems there was a {errorcode} error.', $errorCode);
```
ChloeSartorelli: This fallback was missing when the code was ported over from `LeftAndMain`. The localisation key was already there, just forgot to keep the message itself lol.

code/CMSMenu.php
emteknetnz: If you're completely confident it's not used in the CMS 5, then updating the old key is probably better, so go with that.

Note that `_CMS6` wouldn't be a good key name, as it introduces confusion one we get to CMS 7

ChloeSartorelli: Double checked and the affected strings aren't used in CMS 5.
Removing these bits.

emteknetnz (approved):

Issue #1839, opened 2024-10-20
Re-login doesn't work as expected


jules0x:
### Module version(s) affected

2.2.14

### Description

When a user has unsaved changes, and is idle for a period, the screen prompts for the users password.
When the user enters the password, nothing happens.
Clicking the link "refresh this page" loses the users unsaved changes,
It also shows an X at the top right, which does't do anything.

The interaction here should seamlessly refresh the session without interruption, allowing the user to continue editing or save.

Or is this the intended state?

Tested on a fresh install of silverstripe.

### How to reproduce

Edit some content (e.g. HTMLEditorField).
Wait for idle timeout.
Press save.

### Validations

- [X] Check that there isn't already an issue that reports the same bug
- [X] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

### Acceptance criteria
- [x] Modal closes after successful submission
- [x] Users session is restored

## PRs

[**Kitchen sink CI**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/15057055199)

- https://github.com/silverstripe/silverstripe-admin/pull/1934

Cambis:
Done some investigation into this issue, it seems that the reauthentication is successful but the login modal doesn't close

### How to reproduce

1. Login to the cms
2. Edit any page
3. Invalidate your session (easiest way to is to delete your php session cookie)
4. Click save
5. See the login modal prompt, fill in details and submit
6. See that the login modal does not close

At this point if I remove the modal from the DOM manually, I can edit and save the page as normal.

### Possible solution

I'm not that familiar with how this library works but here's some ideas 😅 

1. Check to see if there are multiple react root elements? Perhaps this line references a different react root when the modal is opened vs closed? https://github.com/silverstripe/silverstripe-admin/blob/8c1861c12aae9341da4b898cbb933181799c0c5a/client/src/legacy/LeftAndMain.js#L1070
2. Investigate using `onUnmountClose` and `onClosed` props to help remove the modal see https://reactstrap.github.io/?path=/docs/components-modal--modal#props

### Additional context

I can confirm that the close function is being called as expected https://github.com/silverstripe/silverstripe-admin/blob/8c1861c12aae9341da4b898cbb933181799c0c5a/client/src/legacy/LeftAndMain.js#L1055

GuySartorelli:
[...]

PR #1934, opened 2025-05-15
FIX Close re-authentication dialog

GuySartorelli:
## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1839


client/src/legacy/LeftAndMain.js
```diff
@@ -1076,7 +1089,9 @@ $.entwine('ss', function($) {
       let root = this.getReactRoot();
       if (!root) {
         root = createRoot(this[0]);
+        this.setReactRoot(root);
```
ChloeSartorelli, line 1092: Not setting the property, and not adding it in line 1057 in the first place, are ultimately the root cause of this bug.

client/src/legacy/LeftAndMain.js
```diff
@@ -1054,13 +1054,26 @@ $.entwine('ss', function($) {
...
+    ReactRoot: null,
+
+    onunmatch() {
+      this._super();
+      const root = this.getReactRoot();
+      if (root) {
+        root.unmount();
+        this.setReactRoot(null);
+      }
+    },
```
ChloeSartorelli, line 1066: Removing the root in `onunmatch` is just good housekeeping.

client/src/legacy/LeftAndMain.js
```diff
@@ -1085,6 +1100,7 @@ $.entwine('ss', function($) {
           iframeId="login-dialog-iframe"
           iframeClassName="login-dialog__body__iframe"
           isOpen={isOpen}
+          onClosed={handleClose}
```
ChloeSartorelli, line 1103: Handling the react close button this way is a pattern we do elsewhere, e.g. with the tinymce link modals.

tests/behat/features/reauthenticate.feature
ChloeSartorelli, line 1: Changes to behat are:

1. Assert we see the actual modal element, and its title
2. Assert we see the reason the user was logged out inside the modal's iframe
3. Assert we don't see the modal element after authenticating

emteknetnz (changes requested):
Should be perfectly fine, though could you run this one through the kitchen sink CI just in case there was anything that breaks because of this i.e. essentially relied on things not being done completely correctly

GuySartorelli:
This reauthentication dialog is very stand-alone.... nothing else relies on this. But sure.

emteknetnz (approved):

Issue #1899, opened 2025-02-12
Front-end validation errors display on some search fields


GuySartorelli:
For some `FormField` types (noticed specifically with `CurrencyField` but bound to be with others as well), there's front-end validation errors that occur on-blur in the search filters in gridfield.

The example screenshot below comes from https://github.com/silverstripe/silverstripe-framework/pull/11602#pullrequestreview-2610601037

![example](https://github.com/user-attachments/assets/6f82fb6d-d58c-48ed-8f72-ebdb36e785fc)

## Steps to reproduce
- Set up a `DataObject` with a `Currency` db field.
- Add that field as a searchable field
- make a model admin for the new class
- Click "Search options", click on the currency field, and then click off it.
- Behold, a message saying the value isn't valid.

## Notes
- It's unclear whether we should allow some validation here or not. Perhaps engage the UX team?
  - Removing the frontend validation resolves the problem, technically, but reduces the usefulness of these specific form fields if they're explicitly used
  - Keeping the frontend validation could be useful for some scenarios but isn't backed up with backend validation at all
- You can still submit the form - the error is purely visual
- None of the default scaffolded search fields should have this problem anymore - but it still happens on explicitly set fields like `CurrencyField` and probably `EmailField`

## Acceptance criteria
- [ ] Remove JS validation of form fields in the search box
- [ ] Check 5.4 to see current status of validation, if it's completely broken probably makes sense to target 5.4
- [ ] This most likely requires product owner approval, whether it targets 5.4 or 6.0

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1939

GuySartorelli:
No owner approval needed because of the very minimal scope of the change that got implemented.

PR merged.
Reassigning to Steve to check it works when merged up to `6.0` since there was no behat or other test added which would catch it in CI.

emteknetnz:
CI on 3.0 is green after merge-up

PR #1939, opened 2025-05-22
FIX Remove frontend validation for currency

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1899

I tried a variety of field types in CMS 5.4, only currency field seemed to have an issue with validation

CMS 5.4 will [scaffold a currency field](https://github.com/silverstripe/silverstripe-framework/blob/5.4/src/ORM/FieldType/DBCurrency.php#L80) for the search form, it has the issue described in the original issue

CMS 6 will [scaffold a numeric field](https://github.com/silverstripe/silverstripe-framework/blob/6.0/src/ORM/FieldType/DBCurrency.php#L73) instead so it won't validate on the front end 

Due to localisation, validating currency is very hard because of the large number of combinations that valid currency could take. We use [validator.js](https://www.npmjs.com/package/validator) for validation which does have an `isCurrency()` function available, however it take an options argument, not a locale string e.g. `en-NZ`. We'd need a way to turn a locale string into the options argument, which would be probably a fairly big bit of work to ensure it actually works. I couldn't find an existing npm library to do this.

Because of this complexity, I think it's best to simply not attempt to validate currency on the frontend

The original issue can be seen in CMS 5.4 with the following code

```php
// app/src/MyDataObject.php

{

    private static $db = [
        'MyCurrency' => 'Currency',
    ];

    private static $searchable_fields = [
        'MyCurrency'
    ];
}
```

```php
// app/src/MyModelAdmin.php

{

        MyDataObject::class,
    ];
}
```


client/src/lib/Validator.js
```diff
@@ -66,6 +66,11 @@ class Validator {
       case 'email': {
         return validator.isEmail(value);
       }
+      case 'currency': {
+        // do not attempt to validate currency, due to localisation it
+        // is very difficult to handle every possible scenario
```
ChloeSartorelli:
```suggestion
        // do not attempt to validate currency, due to localisation it
        // is very difficult to handle every possible scenario so rely on
        // server-side validation
```

GuySartorelli (changes requested):
Very tiny change but just avoids the implication that we're not validating currency field _at all_

client/src/lib/Validator.js
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #700, opened 2018-10-18
SearchForm's autoload and cause lots of XHR requests


ScopeyNZ:
**Steps to reproduce**

- Spin up a copy of `cwp/cwp-recipe-kitchen-sink`
- Open your network tab
- Browse to an edit page
- Observe the SearchForm requests:

![lotsofsearchforms](https://user-images.githubusercontent.com/3260989/47127841-43b6c880-d2eb-11e8-97c5-a76e69344282.gif)

In this case, there's a bunch of searchable gridfields added by the comments module.

Ideally these are only loaded if you click on the advanced filters dropdown. I imagine there's a very small number of times people browse to edit a page and intend to use advanced search on a gridfield.

## Acceptance criteria

- [x] The search form for filtering records (be it in a gridfield or for pages in `CMSMain`) is not loaded until after the user clicks the button which would display said form
- [x] If the form is hidden, it doesn't need to be fetched a second time
- [ ] If you enter something into a field in the search form, then hide it, then show it again, whatever you entered is still there.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1942

maxime-rainville:
The file search doesn't do that, because the search component isn't actually rendered until you click the search toggle. 

The GridField and SiteTree search always render their search components, but hide them with CSS. A simple fix could be to do the same thing as the File Search.

A side effect of this is that if you untoggle the search in File, the search component is completely destroyed and you loose whatever search term you had in there. You'll also get a Form Schema request each time you toggle the search. https://github.com/silverstripe/silverstripe-admin/issues/627 Should fix this.

The GridField search remember its term when you untoggle it.

Another solution could be to only render `<SearchForm />` when you actually toggle the advanced search. But that would mean you would have fetch the form schema each time you open the advanced search, until we implement schema caching.

ScopeyNZ:
> You'll also get a Form Schema request each time you toggle the search. #627 Should fix this.

CCers recently added a `refetchSchemaOnMount` prop for FormBuilderLoader. Have you tried toggling that?

I think the attached issue was more along the lines of caching schema per object type - so separate forms can use the same cache. Redux-form should remember individual(ly named) forms.

ScopeyNZ:
> > You'll also get a Form Schema request each time you toggle the search. #627 Should fix this.
> 
> CCers recently added a `refetchSchemaOnMount` prop for FormBuilderLoader. Have you tried toggling that (to false)?

[...]

PR #1942, opened 2025-06-03
ENH Only load SearchForm schema if parent is visible

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/700

Note that in CMS 5 is requests from `/schema/SearchForm`, in CMS 6 it requests from `/SearchForm/schema`

This will "preload" the schema when you click the magnifying glass icon

The other option was to load the schema even later when the search form is "expanded" (click the "Search options" button), however this means there's a delay rendering the form as the schema is XHR'd into form holder component


client/src/components/Search/SearchForm.js
```diff
@@ -26,11 +26,10 @@ function onEnter(callback) {
  * Display an advanced filter form.
  */
 function SearchForm({
-  visible, expanded, onSearch, onClear, formSchemaUrl, id, identifier, clearable
+  expanded, forceLoadForm, onSearch, onClear, formSchemaUrl, id, identifier, clearable
 }) {
   const handleKeyDown = onEnter(onSearch);
-  const loadForm = (visible || expanded);
```
emteknetnz, line 32: The "visible" state does not do what we need it to do. It's true by default which is what's causing all the unnecessary calls to the endpoint in the first place

It's not "visible" by default to the end user because it's hidden and toggled visible by css and jquery

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -64,9 +64,9 @@ class Focusedzone extends Component {
     this.wasClicked = true;
   }
 
-  handleDocumentClick() {
+  handleDocumentClick(evt) {
```
ChloeSartorelli, line 67:
```suggestion
  handleDocumentClick(event) {
```
There's no need to shorten this, `event` is already short and is more immediately clear what it is.

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -64,9 +64,9 @@ class Focusedzone extends Component {
...
     if (!this.wasClicked) {
-      this.props.onClickOut();
+      this.props.onClickOut(evt);
```
ChloeSartorelli, line 69:
```suggestion
      this.props.onClickOut(event);
```

client/src/components/Search/Search.js
```diff
@@ -267,6 +269,30 @@ class Search extends Component {
     setTimeout(() => this.focusFirstFormField(selector), 50);
   }
 
+  onClickOut(evt) {
```
ChloeSartorelli, line 272:
```suggestion
  onClickOut(event) {
```

client/src/components/Search/Search.js
```diff
@@ -267,6 +269,30 @@ class Search extends Component {
...
+        doShow = true;
+      }
+    });
+    // Match on target class
+    ['font-icon-search', 'filter-open'].forEach(className => {
+      if (el.classList.contains(className)) {
+        doShow = true;
+      }
+    });
+    if (doShow) {
```
ChloeSartorelli:
```suggestion
    if (el.matches('#filters-button, .font-icon-search, .filter-open')) {
```
https://developer.mozilla.org/en-US/docs/Web/API/Element/matches

client/src/components/Search/SearchForm.js
```diff
@@ -63,16 +63,17 @@ function SearchForm({
 SearchForm.propTypes = {
   onSearch: PropTypes.func,
   onClear: PropTypes.func,
-  visible: PropTypes.bool,
```
ChloeSartorelli, line 66: I would normally say we shouldn't remove props in a patch - but since this prop wasn't really doing anything I'm okay with it.

GuySartorelli (changes requested):
If I go to `/admin/test/SilverStripe-FrameworkTest-Model-Company` (included with `silverstripe/frameworktest`) and click the magnifying glass, it only loads one search form schema which is good.

If I go to that route (refresh the page if you were there already), then click on the "Employees" tab, and then click the magnifying glass, it loads _two_ search form schemas.

I assume it's loading the schema for the form for each of those two gridfields? But it should only load the one I'm clicking the form for.

I assume this is because the conditions in `onClickOut` are only checking if _any_ search magnifying glass is being clicked, not if the one for _this gridfield_ is being clicked.

client/src/components/Focusedzone/Focusedzone.js
emteknetnz: Technically you can use `event`, though I don't like it because of confusion of whether it clashes with `window.Event`. Easier to avoid the confusion

client/src/components/Search/Search.js
emteknetnz: Updated

emteknetnz:
Have updated the onClickOut logic to check if the relevant component is visible

Will fix it for the scenario you tested above, also for two gridfields on the same edit form

```php
// app/src/Page.php

class Page extends SiteTree
{
    private static $has_one = [
        'HOA' => Page::class,
        'HOB' => Page::class,
    ];

    private static $has_many = [
        'MyPagesA' => Page::class . '.HOA',
        'MyPagesB' => Page::class . '.HOB',
    ];

    {
        $gf1 = new GridField('MyPagesA', 'My Pages A', $this->MyPagesA());
        $gf1->setConfig(GridFieldConfig_RelationEditor::create());
        $gf2 = new GridField('MyPagesB', 'My Pages B', $this->MyPagesB());
        $gf2->setConfig(GridFieldConfig_RelationEditor::create());
        $fields->addFieldsToTab('Root.Main',[$gf1, $gf2]);
        $fields->removeByName('Content');
        return $fields;
    }
}

```

client/src/components/Focusedzone/Focusedzone.js
ChloeSartorelli: Discussed offline - while `event` is _way_ more common in our codebase, a few `evt` have already snuck in so we may as well allow this here too.

client/src/components/Search/Search.js
```diff
@@ -267,6 +270,36 @@ class Search extends Component {
...
 
+  isVisible(el) {
+    const rect = el.getBoundingClientRect();
+    return rect.width > 0 &&
+      rect.height > 0 &&
+      rect.top >= 0 &&
+      rect.left >= 0 &&
+      rect.bottom <= (window.innerHeight || document.documentElement.clientHeight) &&
+      rect.right <= (window.innerWidth || document.documentElement.clientWidth);
+  }
```
ChloeSartorelli: Since we're using jQuery, we may as well use `$(el).is(:visible)`
- https://api.jquery.com/is/#is-selector
- https://api.jquery.com/visible-selector/

client/src/components/Search/Search.js
```diff
@@ -267,6 +270,36 @@ class Search extends Component {
...
+    // This is a fairly rudimentory way to detect if the user clicked on an entirely different
+    // DOM element, which probably uses jQuery
+    // It's important that we only call show() when a relevent element is clicked, otherwise we'll
+    // end up making lots of unnecessary AJAX requests to the form schema endpoint
+    const buttonEl = evt.target;
+    const componentEl = this.nodeRef.container;
+    const filterButtonClicked = buttonEl.matches('#filters-button, .font-icon-search, .filter-open');
+    if (filterButtonClicked && this.isVisible(componentEl)) {
+      const buttonGridField = window.jQuery(buttonEl).closest('.grid-field');
+      const componentGridField = window.jQuery(componentEl).closest('.grid-field');
```
ChloeSartorelli: We're explicitly checking for gridfield here - is this component not used for the CMSMain search form?

client/src/components/Search/Search.js
```diff
@@ -267,6 +270,36 @@ class Search extends Component {
...
+
+  onClickOut(evt) {
+    // This is a fairly rudimentory way to detect if the user clicked on an entirely different
+    // DOM element, which probably uses jQuery
+    // It's important that we only call show() when a relevent element is clicked, otherwise we'll
+    // end up making lots of unnecessary AJAX requests to the form schema endpoint
+    const buttonEl = evt.target;
+    const componentEl = this.nodeRef.container;
+    const filterButtonClicked = buttonEl.matches('#filters-button, .font-icon-search, .filter-open');
+    if (filterButtonClicked && this.isVisible(componentEl)) {
```
ChloeSartorelli: What is this visibility check actually for? I tried removing it to see what would change, but it seems to work as expected without it?

GuySartorelli (changes requested):
For `CMSMain` the search form isn't loaded immediately when the magnifying glass is clicked, but instead only when I click on something inside the search bar (either the input or the "search options" dropdown). Is that intentional?

It seems like the difference might be that for gridfield, the input field immediately gets focus, but for CMS main it doesn't (and probably should).

client/src/components/Search/Search.js
emteknetnz: Updated

client/src/components/Search/Search.js
emteknetnz: CMSMain search is not contained in a gridfield

This check is for multiple gridfields on a single form

client/src/components/Search/Search.js
emteknetnz: It's for the scenario where you go to `/admin/test`, then click the Employees tab, which will xhr in another gridfield and hide the first one

emteknetnz:
> For CMSMain the search form isn't loaded immediately when the magnifying glass is clicked

Have fixed this

client/src/components/Search/Search.js
ChloeSartorelli: > CMSMain search is not contained in a gridfield

... Yes, I'm aware of that 😅
Let me rephrase my question.

1. is this component used for the CMSMain search form?
2. If so, are these checks for explicit gridfield elements being skipped when this component is used in `CMSMain`?

ChloeSartorelli: Nevermind, it looks like that's what was causing https://github.com/silverstripe/silverstripe-admin/pull/1942#pullrequestreview-2894576071 and you've since separated out to have two separate `show()` calls.

client/src/components/Search/Search.js
```diff
@@ -267,6 +269,27 @@ class Search extends Component {
...
 
+  onClickOut(evt) {
+    // This is a fairly rudimentory way to detect if the user clicked on an entirely different
+    // DOM element, which probably uses jQuery
+    // It's important that we only call show() when a relevent element is clicked, otherwise we'll
+    // end up making lots of unnecessary AJAX requests to the form schema endpoint
+    const buttonEl = evt.target;
+    const $componentEl = window.jQuery(this.nodeRef.container);
+    const filterButtonClicked = buttonEl.matches('#filters-button, .font-icon-search, .filter-open');
+    if (filterButtonClicked && $componentEl.is(':visible')) {
```
ChloeSartorelli, line 281:
```suggestion
    // The visibility check is needed to avoid loading the form for hidden tabs (e.g. in ModelAdmin)
    if (filterButtonClicked && $componentEl.is(':visible')) {
```

client/src/components/Search/Search.js
```diff
@@ -267,6 +269,27 @@ class Search extends Component {
...
+      const buttonGridField = window.jQuery(buttonEl).closest('.grid-field');
+      const componentGridField = $componentEl.closest('.grid-field');
+      if (buttonGridField.length > 0 && componentGridField.length > 0) {
+        if (buttonGridField[0] === componentGridField[0]) {
+          this.show();
+        }
+      } else {
+        this.show();
```
ChloeSartorelli, line 291:
```suggestion
        // This is not a GridField, e.g. for CMSMain
        this.show();
```

client/src/components/Search/Search.js
```diff
@@ -267,6 +269,27 @@ class Search extends Component {
...
+    // It's important that we only call show() when a relevent element is clicked, otherwise we'll
+    // end up making lots of unnecessary AJAX requests to the form schema endpoint
+    const buttonEl = evt.target;
+    const $componentEl = window.jQuery(this.nodeRef.container);
+    const filterButtonClicked = buttonEl.matches('#filters-button, .font-icon-search, .filter-open');
+    if (filterButtonClicked && $componentEl.is(':visible')) {
+      const buttonGridField = window.jQuery(buttonEl).closest('.grid-field');
+      const componentGridField = $componentEl.closest('.grid-field');
+      if (buttonGridField.length > 0 && componentGridField.length > 0) {
+        if (buttonGridField[0] === componentGridField[0]) {
```
ChloeSartorelli, line 286:
```suggestion
        // Only show the form for the relevant GridField
        if (buttonGridField[0] === componentGridField[0]) {
```

GuySartorelli (changes requested):
Works well locally. Just adding a few comments for the things I had asked questions about because it's not immediately obvious just from looking at the code.

client/src/components/Search/Search.js
emteknetnz: Updated

client/src/components/Search/Search.js
emteknetnz: Updated

client/src/components/Search/Search.js
emteknetnz: Updated

GuySartorelli (approved):

PR #1954, opened 2025-08-05
FIX Ensure changetracker notes first change in WYSIWYG

GuySartorelli:
CMS 6 will need its own separate fix, since the affected code was moved to https://github.com/silverstripe/silverstripe-htmleditor-tinymce/

https://www.tiny.cloud/docs/tinymce/6/events/ for docs about the `dirty` event.

## Issue

- https://github.com/silverstripe/silverstripe-elemental/issues/1358


emteknetnz (approved):
Works well

PR #1974, opened 2025-08-12
FIX Let admin sections have same url-segment-prefix

GuySartorelli:
React admin section routing shouldn't kick in when the URL segment for an admin section starts with the same string, unless it's actually part of that same section.

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1459


client/src/boot/BootRoutes.js
```diff
@@ -98,7 +98,7 @@ class BootRoutes {
 
       // Check if the beginning of the route is the same as the current location.
       // Since we haven't decided on a router yet, we can't use it for route matching.
-      return currentPath.match(route);
+      return currentPath.match(`${route}(/|$)`);
```
ChloeSartorelli, line 101: i.e. `/admin/assets` will match and `/admin/assets/some-sub-route` will match, but `/admin/assets-test` and `/admin/assetstinger` won't.

tests/behat/features/single-page-routing.feature
ChloeSartorelli, line 1: I couldn't find a suitable place for this in an existing behat test, and this can't be tested with unit tests.

I made this feature intentionally quite open in its naming so we can add other tests related to single-page-routing here (e.g. for https://github.com/silverstripe/silverstripe-admin/issues/1975)

tests/behat/features/single-page-routing.feature
```diff
@@ -0,0 +1,15 @@
+@retry @job2
+Feature: Single-page routing works as expected
```
emteknetnz: Weirdly named file and feature, I don't quite understand what it's supposed to mean

emteknetnz (changes requested):

tests/behat/features/single-page-routing.feature
ChloeSartorelli: Named after "single page apps" which the CMS routing kinda tries to mimic.

emteknetnz: That's non-obvious

Call it model-admin.feature or something like that

emteknetnz (changes requested):

tests/behat/features/single-page-routing.feature
ChloeSartorelli: `model-admin.or-something-like-that.newnamehere.feature` :p
`model-admin` isn't indicative of what this is for. I'll give it a new name that's as good as the current one and you can tell me to change it again lol.

emteknetnz (approved):
Cheers

Issue #1972, opened 2025-08-11
Accessibility: Sudo mode focus needs to be updated


GuySartorelli:
In sections protected by sudo mode, clicking "verify to continue" should move the focus to the password field `input` element.

This means keyboard users can expand the section and then immediately type their password.
It also means for screen reader users they don't suddenly lose their place.

## Acceptance criteria

- [ ] For sudo mode, activating the "verify to continue" button should move focus to the newly displayed input element

**Do not merge until CMS 6.1 beta has been released**

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/16924623442

## PRs
- https://github.com/silverstripe/silverstripe-behat-extension/pull/310
- https://github.com/silverstripe/silverstripe-admin/pull/1976

PR #1976, opened 2025-08-13
ENH Focus the sudo mode field after clicking button

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1972

Requires https://github.com/silverstripe/silverstripe-behat-extension/pull/310


GuySartorelli (changes requested):
The `autoFocus` attribute should be used instead.
https://react.dev/reference/react-dom/components/input
> [autoFocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#autofocus): A boolean. If true, React will focus the element on mount.

emteknetnz:
Updated

GuySartorelli (approved):
LGTM - so much nicer to use.
Just waiting on behat to pass.

PR #1977, opened 2025-08-13
FIX Do not over extend icon hover area

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1597

Fixes issue where the mouse hover area extends to the left past the icon, which is on the :before pseudo element


GuySartorelli (approved):
LGTM

Issue #1956, opened 2025-08-06
Accessibility: left menu and toggle-collapse aria attributes


GuySartorelli:
[WCAG 2.2 A F68](https://www.w3.org/TR/WCAG20-TECHS/F68.html) 4.1.2
The [`aria-expanded` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-expanded) is supposed to be "applied to [a] focusable, interactive control that toggles the visibility of [an] object".

This attribute is currently applied to the `div#cms-menu` element, which is the "object" in the above quote, i.e. it's the element that gets hidden or displayed.

The attribute should instead be applied to the element that controls the visibility, which is `a.toggle-collapse` and `a.toggle-expand`.
That element should also be a `<button>` rather than a `<a>` tag, should have an `aria-controls` attribute, and should be a single element instead of two elements that swap out.

## Acceptance Criteria

- [x] `div#cms-menu` element does not have `aria-expanded` attribute
- [x]  `a.toggle-collapse` and `a.toggle-expand` elements that controls the visibility of `div#cms-menu`:
    - [x] are swapped out for a single `<button>` element
    - [x] have the `aria-controls` attribute added, pointing to `cms-menu`
    - [x] have the `aria-expanded` attribute added
- [x] The changelog clearly calls out any changes to markup selectors so people can adjust their custom css/js as necessary

## Kitchen sink run
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/16929127593

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1978
- https://github.com/silverstripe/silverstripe-cms/pull/3102
- https://github.com/silverstripe/developer-docs/pull/793

GuySartorelli:
PRs merged

PR #1978, opened 2025-08-13
ENH Use button and aria attrs for cms menu toggle

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1956


client/src/legacy/LeftAndMain.Panel.js
```diff
@@ -172,37 +183,18 @@ $.entwine('ss', function($) {
     }
   });
 
-  $('.cms-panel.collapsed .cms-panel-toggle').entwine({
```
emteknetnz, line 175: This was pretty pointless as the only thing this container has inside of it is the `<button>`, which has its own click handler

client/src/legacy/LeftAndMain.Panel.js
```diff
@@ -96,27 +106,38 @@ $.entwine('ss', function($) {
...
       // Create default controls unless they already exist.
       if(!this.find('.cms-panel-toggle').length) {
-        container = $("<div class='toolbar toolbar--south cms-panel-toggle'></div>")
-          .append('<a class="toggle-expand" href="#" data-bs-toggle="tooltip" title="'+
-              i18n._t('Admin.EXPANDPANEL', 'Expand Panel') +'"><span>&raquo;</span></a>')
-          .append('<a class="toggle-collapse" href="#" data-bs-toggle="tooltip" title="'+
-              i18n._t('Admin.COLLAPSEPANEL', 'Collapse Panel') +'"><span>&laquo;</span></a>');
-
-        this.append(container);
+        const expandedTitle = i18n._t('Admin.COLLAPSEPANEL', 'Collapse Panel');
```
emteknetnz, line 116: This whole thing isn't actually used in practice as far as I can tell, there are 2x ".cms-panel-toggle" menus - one for the left hand menu that holds all of the model admins, the other for silverstripe/cms for the sitetree (down the bottom right of the panel). Both of those have their menus defined is templates LeftAndMain_MenuToggle.ss and CMSMain_Tools.ss

I was tempted to just remove this whole thing though it's pretty trivial to keep this functionality the off chance there's customised cms out there that relies on this

client/src/legacy/LeftAndMain.Panel.js
```diff
@@ -96,27 +106,38 @@ $.entwine('ss', function($) {
...
     onadd: function() {
-      var collapsedContent, container;
-
-      if(!this.find('.cms-panel-content').length) throw new Exception('Content panel for ".cms-panel" not found');
+      if(!this.find('.cms-panel-content').length) {
+        throw new Exception('Content panel for ".cms-panel" not found');
+      }
 
       // Create default controls unless they already exist.
       if(!this.find('.cms-panel-toggle').length) {
```
ChloeSartorelli, line 115:
```suggestion
      // This is here for backwards compatibility reasons and should be removed in a future major release
      if(!this.find('.cms-panel-toggle').length) {
```

client/src/legacy/LeftAndMain.Panel.js
ChloeSartorelli: Fair enough. I've made a suggestion to add a comment with the words "backwards compatibility" in so we'll find and remove it in a major release.

GuySartorelli (changes requested):
The `&laquo;` and `&raquo;` characters are probably going to be read to screen readers.
They should be put inside a `span` with the `aria-hidden="true"` attribute.

client/src/legacy/LeftAndMain.Panel.js
emteknetnz: Updated

client/src/legacy/LeftAndMain.Panel.js
emteknetnz: Done

client/src/legacy/LeftAndMain.Panel.js
```diff
@@ -157,6 +179,18 @@ $.entwine('ss', function($) {
 
       this.trigger('toggle', doExpand);
       this.trigger(doExpand ? 'expand' : 'collapse');
+
+      // Update decendant button
+      const button = this.find('.cms-panel-toggle__button');
+      if (doExpand) {
+        button.attr('aria-expanded', 'true');
+        button.attr('title', button.attr('data-expanded-title'));
+        button.html(button.attr('data-expanded-label'));
```
ChloeSartorelli, line 190: Need spans with `aria-hidden="true"` as per https://github.com/silverstripe/silverstripe-admin/pull/1978#pullrequestreview-3153168151

The alternative is for this change in the templates:
```diff
- data-expanded-label="&laquo;"
+ data-expanded-label="<span aria-hidden="true">&laquo;</span>"
```

GuySartorelli (changes requested):

client/src/legacy/LeftAndMain.Panel.js
emteknetnz: Added aria-label instead e.g. "Collapse Panel" or "Expand Panel"

GuySartorelli (approved):
LGTM

PR #1984, opened 2025-08-20
FIX Correctly route from non-react to react section.

GuySartorelli:
Needs https://github.com/silverstripe/silverstripe-frameworktest/pull/243 for new behat test to go green

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1975


tests/behat/features/gridfield-inline-delete.feature
```diff
@@ -16,7 +16,6 @@ Feature: GridField inline delete
     And I go to "/admin/test"
 
   Scenario: I can see toast message when I successfully delete a record by clicking the Delete button in action menu
-  Given I take a screenshot after every step
```
ChloeSartorelli, line 19: This shouldn't be there so cleaning that up while I'm here.

Noticed it because I was copying some of the background from this feature.

emteknetnz (changes requested):
Tested locally with replication steps and fix works.

However I tried running the new behat scenario locally with the frameworktest PR installed, and it said that it find something for `And I click "behat link to file"`

Maybe I'm doing something wrong? Please do a kitchen-sink CI run with the two PRs to demonstrate it working.

GuySartorelli:
Kitchen sink is excessive for something that can just be merged one at a time but I'll do a minimal installer run to avoid ping pong.
I just got the exact wording of how to click a link wrong.

Usually for these low-risk PRs we merge the framework test PR and then rerun CI instead of requiring a separate sink or installer CI run. Not sure when you started requiring those but we don't need them.

emteknetnz (approved):

Issue #1971, opened 2025-08-11
Accessibility: Left menu should be navigation


GuySartorelli:
The left menu should either be wrapped in a semantic `<nav>` element, or (to preserve existing selectors for CSS and JS) should have[ `role="navigation"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/navigation_role) applied.

See https://www.w3.org/WAI/ARIA/apg/practices/landmark-regions/ to learn more about navigation and other landmarks

## Acceptance criteria

- [ ] The left menu is identified as a "navigation" landmark and has an appropriate label

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1987
- https://github.com/silverstripe/developer-docs/pull/807

GuySartorelli:
PRs merged

PR #1987, opened 2025-08-26
ENH Use nav element for cms-menu

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1971

I couldn't find any CSS or JSS that prefixed the selector with `div` i.e. `div.cms-menu` or `div#cms-menu`, so I think it's safe to just change the tag

### DIFF
--- a/templates/SilverStripe/Admin/Includes/LeftAndMain_Menu.ss
+++ b/templates/SilverStripe/Admin/Includes/LeftAndMain_Menu.ss
@@ -1,6 +1,6 @@
 <div class="cms-mobile-menu-toggle-wrapper"></div>
 
-<div class="fill-height cms-menu cms-panel cms-panel-layout" id="cms-menu" data-layout-type="border" aria-expanded="false">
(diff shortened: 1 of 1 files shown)


templates/SilverStripe/Admin/Includes/LeftAndMain_Menu.ss
```diff
@@ -1,6 +1,6 @@
 <div class="cms-mobile-menu-toggle-wrapper"></div>
 
-<div class="fill-height cms-menu cms-panel cms-panel-layout" id="cms-menu" data-layout-type="border" aria-expanded="false">
+<nav class="fill-height cms-menu cms-panel cms-panel-layout" id="cms-menu" data-layout-type="border" aria-label="CMS menu" aria-expanded="false">
```
ChloeSartorelli: Put the label through i18n.
It should also probably be something like "main menu". Any menu in the CMS is a CMS menu.

GuySartorelli (changes requested):
Since you've opted to change the HTML tag, we need to note this change in the changelog.

templates/SilverStripe/Admin/Includes/LeftAndMain_Menu.ss
emteknetnz: Updated

GuySartorelli (approved):
LGTM

Issue #1989, opened 2025-08-27
Accessibility: Wrong element had `aria-expanded` removed.


GuySartorelli:
In https://github.com/silverstripe/silverstripe-admin/pull/1978 the `aria-expanded` attribute should have been removed from the `nav#cms-menu` but was removed from the `button.cms-help__menu` instead.

## PRs

- https://github.com/silverstripe/silverstripe-admin/pull/1990

PR #1990, opened 2025-08-27
FIX Keep aria-expanded on the button and remove from the nav

GuySartorelli:
## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1989


client/src/legacy/LeftAndMain.MobileMenuToggle.js
```diff
@@ -21,8 +21,7 @@ $.entwine('ss', function($){
         const state = store.getState();
         const isOpen = state.mobileMenu.isOpen;
         menu
-          .toggleClass('cms-menu--open', isOpen)
-          .attr('aria-expanded', isOpen);
```
emteknetnz, line 25: Shouldn't there be some corresponding code which adds `.attr('aria-expanded', isOpen);` to update `.cms-help__menu`?  Or does that already existt?

emteknetnz (changes requested):

client/src/legacy/LeftAndMain.MobileMenuToggle.js
ChloeSartorelli: If you click the help menu button you'll see the `aria-expanded` attribute is already being updated correctly.
It's not related to this JS which is specifically for toggling the main nav when at a mobile-width breakpoint.

emteknetnz (approved):

Issue #1963, opened 2025-08-07
Accessibility: `ReadOnlyField` is missing `role` attribute


GuySartorelli:
[ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/#namefromprohibited) Cannot use aria-label or aria-labelledby on elements and roles that prohibit naming.

`ReadOnlyField` is represented by a `<p>` tag which has an implicit role `paragraph`. This role cannot be named/labelled.

We should add `role="input"` and `aria-readonly="true"` attributes to that element

## Notes

- We could alternatively use an actual `<input readonly>` element - though that will mean:
    - changing markup which may break some customisations
    - we can no longer use _(None)_ as a way to denote empty fields (they'll just be empty instead, which is probably better)

## Acceptance criteria

- [ ] `ReadOnlyField` markup has a role that allows naming and uses the `aria-readonly` attribute (or is replaced with a real readonly `input` element)
- [ ] Check both entwine and react forms

[Kitchen sink CI](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/17280489579)

## PRs
- https://github.com/silverstripe/silverstripe-framework/pull/11844
- https://github.com/silverstripe/silverstripe-admin/pull/1991
- https://github.com/silverstripe/silverstripe-advancedworkflow/pull/609

emteknetnz:
I've implemented as:

```html
<tag
  role="textbox"
  aria-readonly="true"
  tabindex="0"
>
```

aria-readonly is not supported by role="input" and will trigger the following js lint errors:

```
673:9   error  The attribute aria-readonly is not supported by the role input     jsx-a11y/role-supports-aria-props
675:11  error  Elements with ARIA roles must use a valid, non-abstract ARIA role  jsx-a11y/aria-role
```

So I've used textbox instead which does allow aria-readonly="true"

tabindex="0" is required to make this non-interactive element focusable. Without it, keyboard and screen reader users cannot navigate to the element.

I've had a good look through all fields in the codebase and have updated a few fields beyond just ReadOnlyField which uses non `<input>` tags when in a readonly state

GuySartorelli:
PRs merged

PR #1991, opened 2025-08-27
ENH Add accessible attributes to readonly fields

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1963


themes/cms-forms/templates/SilverStripe/Forms/ReadonlyField.ss
```diff
@@ -1,6 +1,10 @@
-<p id="$ID" tabIndex="0" class="form-control-static<% if $extraClass %> $extraClass<% end_if %>" <% include SilverStripe/Forms/AriaAttributes %>>
-	$FormattedValue
-</p>
+<p id="$ID"
+   tabindex="0"
+   class="form-control-static<% if $extraClass %> $extraClass<% end_if %>"
+   <% include SilverStripe/Forms/AriaAttributes %>
```
emteknetnz: SilverStripe/Forms/AriaAttributes will end up being one of these
https://github.com/silverstripe/silverstripe-admin/blob/3/templates/SilverStripe/Forms/Includes/AriaAttributes.ss
https://github.com/silverstripe/silverstripe-admin/blob/3/themes/cms-forms/templates/SilverStripe/Forms/Includes/AriaAttributes.ss

Seems like a confusing template pattern that's very inconsistently used, so I don't want to use it for this set of PRs

On the off chance that a project overrides these templates with something that does include either role or aria-readonly - the LAST html attributes in the element will have precedence

ChloeSartorelli: I don't understand your hesitance to use a template that's very obviously intended for specifically this purpose but I'll put that aside, we can debate that another day.

However, given the last attributes will take precedence and we're considering that people may have overridden that template to include their own explicit role and area-readonly attribute, why are our attributes last?

Seems like the ones in the template should take precedent (i.e. be last) if someone has taken the time to override it to add these?

GuySartorelli (changes requested):

themes/cms-forms/templates/SilverStripe/Forms/ReadonlyField.ss
emteknetnz: Have move the include to be below the new attrs

GuySartorelli (approved):
LGTM

Issue #1964, opened 2025-08-10
Accessibility: `OptionSetField` and `CheckboxSetField` options missing `role="option"`


GuySartorelli:
[ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/) [WCAG 2.2 A 1.3.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html) Elements with role=listbox must contain or own an element with role=option and must not contain elements with other roles

The parent `div` for checkboxes in a `CheckboxSetField` and radio buttons in a `OptionSetField` have a `role="listbox"` attribute.
The individual checkboxes and radio buttons are _supposed_ to have a `role="option"` attribute but that is missing.
The PHP code is setting those attributes, so something is likely wrong with the templates.

## Acceptance criteria

- [ ] Checkboxes in a  `CheckboxSetField` have a `role="option"` attribute
- [ ] Radio buttons in a  `OptionSetField` have a `role="option"` attribute
- [ ] Check both entwine and react forms

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1992
- https://github.com/silverstripe/silverstripe-framework/pull/11847

GuySartorelli:
PRs merged

PR #1992, opened 2025-08-27
ENH Add role attribute to CheckboxSet and Optionset fields

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1964


GuySartorelli:
Will hold off reviewing pending https://github.com/silverstripe/silverstripe-framework/pull/11847#pullrequestreview-3162146723

GuySartorelli (changes requested):
In asset admin, when you select a file, the individual radio buttons now have the `option` role, but there's no `role="listbox"` above them.

The listbox role needs to be added to a parent container, I think.

I presume the checkboxSetField has the same problem.

emteknetnz:
Updated, have changed `<div>` to `<div role="listbox">` in the 2 react components

GuySartorelli (approved):
LGTM

Issue #1958, opened 2025-08-06
Accessibility: Do not use the meta viewport tag to disable zoom


GuySartorelli:
[WCAG 2.2 AA 1.4.4](https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-scale.html)
In the `templates/SilverStripe/Admin/LeftAndMain.ss` template, the `viewport` meta tag looks like this:

`<meta name="viewport" content="width=device-width, maximum-scale=1.0" />`

That prevents users from zooming on mobile devices. According to MDN any `maximum-scale` value less than 3 is an automatic fail for accessibility.

The `maximum-scale` should be removed, and most likely `initial-scale` should be added and set to `1.0`:

`<meta name="viewport" content="width=device-width, initial-scale=1.0" />`

See [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Viewport_meta_element) for more details about the viewport meta element and its options.

## Acceptance criteria

- [ ] `maximum-scale` is not used
- [ ] If appropriate, `initial-scale` is added

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1993
- https://github.com/silverstripe/developer-docs/pull/810

GuySartorelli:
PRs merged

PR #1993, opened 2025-08-28
ENH Use initial-scale instead of maximum-scale

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1958

This can be tested on desktop by using developer tools in chrome, selecting "Toggle device toolbar", holding down shift and left click while moving the mouse up or down


GuySartorelli (approved):
LGTM

Issue #1962, opened 2025-08-07
Accessibility: `ReadonlyField` has invalid `aria-labelledby` attribute value in the CMS


GuySartorelli:
[WCAG 2.2 A 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value.html)
`ReadonlyField` has invalid `aria-labelledby` attribute value in the CMS. For example:

```html
<p id="Form_ItemEditForm_IntentionallyEmpty" tabindex="0" class="form-control-static readonly" aria-labelledby="title-Form_ItemEditForm_IntentionallyEmpty ">
	<i>('none')</i>
</p>
```
Note the whitespace in the `aria-labelledby` attribute - it should be `"title-Form_ItemEditForm_IntentionallyEmpty"` but it’s `"title-Form_ItemEditForm_IntentionallyEmpty "`

`aria-labelledby` allows having multiple elements provide a label by having a space between the ID of each element (e.g. `aria-labelledby="element1 element2"` but apparently some systems get confused by whitespace _after_ the ID list.

The template that handles this is `themes/cms-forms/templates/SilverStripe/Forms/Includes/AriaAttributes.ss`

## Acceptance criteria

- [x] The `aria-labelledby` element only has whitespace between element ID references, but not at the beginning or end of the overall value
- [x] Check both entwine and react forms

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1994
- https://github.com/silverstripe/silverstripe-userforms/pull/1407

emteknetnz:
I've searched the supported codebase for templates, php or js that could potentially have trailing or leading spaces added for either "aria-labelledby" or "aria-describedby" - I've raised PRs to fix the only instances of it that I could fine

PR #1994, opened 2025-09-01
FIX Trim blank spaces from aria attribute values

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1962

While it's not in the originally ACs, aria-describedby may have the same issue as aria-labelledby, so I've updated it to also ensure there's no leading/trailing spaces


GuySartorelli (approved):
LGTM

Issue #1968, opened 2025-08-11
Accessibility: Add a link to skip navigation


GuySartorelli:
[WCAG2.2 2.4.1](https://www.w3.org/TR/WCAG22/#bypass-blocks)

We should make it easy for screen-reader and keyboard-navigation users to skip the left menu.

## Acceptance Criteria

- [x] The first focusable element in the CMS is a "skip navigation" link
- [x] The "skip navigation" link is invisible when it doesn't have focus
- [x] The "skip navigation" link moves focus to the main container on the page (usually the element with class `.cms-content`)
- [x] The changelog calls out any changes developers need to make to their template customisations (e.g. adding an id to a specific element) to make this functionality work.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1995
- https://github.com/silverstripe/developer-docs/pull/812

GuySartorelli:
PRs merged

PR #1995, opened 2025-09-01
ENH Add main nav skip link

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1968


templates/SilverStripe/Admin/LeftAndMain.ss
```diff
@@ -1,20 +1,29 @@
 <!DOCTYPE html>
 <html lang="$Locale.RFC1766">
-	<head>
```
emteknetnz, line 3: I changed tabs to 4x spaces

templates/SilverStripe/Admin/LeftAndMain.ss
```diff
@@ -1,20 +1,29 @@
...
+    <a class="cms-container-skip-link"
+       href="#cms-container-skip-link-target"
+       tabindex="0"
+    ><%t SilverStripe\\Admin\\LeftAndMain.SkipLink "Skip to main content" %></a>
+
+    <div class="cms-container" data-layout-type="custom">
+        $Menu
+        <div id="cms-container-skip-link-target"
+             class="cms-container-skip-link-target"
+             tabindex="-1"
```
emteknetnz, line 23: Adding tabindex="-1" to the skip link's target element makes it programmatically focusable, without being tabbable

ChloeSartorelli: Why `role="main"` and not use a `<main>` tag?

client/src/styles/legacy/_style.scss
```diff
@@ -48,6 +48,30 @@ html, body {
   }
 }
 
+.cms-container-skip-link {
```
ChloeSartorelli, line 51:
```suggestion
.skip-link {
```
Seems like it makes sense for it to be quite loosely named? That way developers can safely reuse this CSS class for any skip link they might want to add to the CMS.

If you don't want the JS to apply to other skip links, you could give the element both classes in the markup (one for the JS hook and one for CSS) - but it looks like the JS is written in a way that should work for any custom skip link as well.

templates/SilverStripe/Admin/LeftAndMain.ss
```diff
@@ -1,20 +1,29 @@
...
+    <% include SilverStripe\\Admin\\CMSLoadingScreen %>
 
-	<div class="cms-container" data-layout-type="custom">
-		$Menu
-		$Content
-		$PreviewPanel
+    <a class="cms-container-skip-link"
+       href="#cms-container-skip-link-target"
+       tabindex="0"
+    ><%t SilverStripe\\Admin\\LeftAndMain.SkipLink "Skip to main content" %></a>
```
ChloeSartorelli: I think "Skip main navigation" would be more appropriate - "main content" is really more of a front-end concept IMO. The CMS doesn't have "content" in the same way - it just has a bunch of forms for editing content.

Take `/admin/pages` for example. If you select "skip to main content" it feels a bit weird to be confronted by another menu, in which you have to select a page before seeing [an edit form containing] any actual content.

GuySartorelli (changes requested):

client/src/styles/legacy/_style.scss
emteknetnz: The styles are really not designed to be reusable e.g. top/left: 3px;

> but it looks like the JS is written in a way that should work for any custom skip link as well.

I hadn't really thought of that, I guess it does, though at the same time do we really need to think about making this reusable? I don't think we're planning on adding any more skip links?

templates/SilverStripe/Admin/LeftAndMain.ss
emteknetnz: Updated

templates/SilverStripe/Admin/LeftAndMain.ss
emteknetnz: Updated

client/src/styles/legacy/_style.scss
ChloeSartorelli: > do we really need to think about making this reusable? I don't think we're planning on adding any more skip links?

I had thought it was already reusable - and so changing the name would be an easy way to open it up for reuse (e.g. if an external accessibility review indicates we should use them to skip the gridfield table to get to pagination or something) or if other developers want to add skiplinks for whatever purposes they may have.

But you're right the position is very specific. We can always refactor some of this to be reusable if we find ourselves adding more skiplinks later.

GuySartorelli (approved):
LGTM

PR #1996, opened 2025-09-02
FIX Don't let one element changetrack override another.

GuySartorelli:
<details>
<summary>Original PR description</summary>

> The original changetracker debounce was added in https://github.com/silverstripe/silverstripe-admin/pull/340 after repeated events firing on each key press were found to be causing problems. I added a new debounce to the change events fired from `HtlmEditorField` specifically since these were firing more often than they needed to - but that doesn't actually fix the problem on its own.
> 
> For some reason (and I couldn't figure out what that reason was but see https://github.com/silverstripe/silverstripe-elemental/issues/821#issuecomment-3243589532 for my musings), the 250ms debounce for detecting changes was essentially throwing away the change event on the `form` element which happens when `input.element-form-dirty-state` is created (see https://github.com/silverstripe/silverstripe-elemental/pull/520). In my testing I found that 200ms made it intermittent and 150ms resolved the problem altogether.
> 
> We could swap out the debounce for a throttle (see https://css-tricks.com/debouncing-throttling-explained-examples/ for an explanation of the differences) which would mean it would fire more often - but I don't think that would ultimately be useful.
> 
> One thing we can look at if this PR is considered too flakey is see if we can implement our own debounce that is unique to each target. I suspect what's happening is the change event from the textarea is replacing the change event for the form because of the way debounce works. This wouldn't be a problem if the debounce was unique to each event target.
> That would mean implementing our own custom debounce function and would only be a valid solution if it doesn't reintroduce the performance problems that the original debounce was added to resolve.

</details>

I'm using a new approach now - as per the original PR description since changing the debounce timing was too flakey, I've implemented a way to debounce per element. This means the change tracking detection for the change event from the text area doesn't override the change track detection from the change event for the form, which was the underlying cause of the changetracker not updating.

## Issue

- https://github.com/silverstripe/silverstripe-elemental/issues/821


client/src/legacy/HtmlEditorField.js
```diff
@@ -92,13 +93,21 @@ ss.editorWrappers.tinyMCE = (function() {
       // Add instance specific data to config
       config.selector = selector;
 
+      const change = debounce(function () {
+        var instance = self.getInstance();
+        jQuery(instance.getElement()).trigger("change");
+        instance.getElement().dispatchEvent(new Event('input', { bubbles: true }));
+      }, 250, { leading: true, trailing: true });
```
ChloeSartorelli: This isn't strictly necessary - but given the WYSIWYG is where most key events will be happening it seems sensible to debounce this to offset the fact that the debounce is shorter in the changetracker.

GuySartorelli:
I can't think of a way to test this with behat but let me know if you have any thoughts.

emteknetnz (changes requested):
Solution is flakey - on my local when editing WYSIWYG in elemental blocks it's 50/50 whether or not the "unsaved changes" popup shows

This issue doesn't seem to happen when using WYSIWYG on regular Page edit forms, which also uses the code changed in this PR - does elemental have some an extra layer for change tracking that might be causing these issues?

GuySartorelli:
> does elemental have some an extra layer for change tracking that might be causing these issues?

Yes. See the second paragraph in the (original) PR description.

> Solution is flakey - on my local when editing WYSIWYG in elemental blocks it's 50/50 whether or not the "unsaved changes" popup shows

It's been a while since I worked on this - I'll see what it does for me locally and see if I can make it less "flakey".

client/src/lib/debounceByElement.js
```diff
@@ -0,0 +1,31 @@
...
+ * Using lodash.debounce directly may otherwise result in the debounce for one element overriding the debounce for another,
+ * so the callback would only be called for one of the elements instead of for each of them.
+ *
+ * See lodash.debounce documentation for usage of arguments.
+ * @returns debouncable function that takes an element as an argument.
+ * @example
+ * // Ensure `batchLog` is invoked once after 1 second of debounced calls with a separate debounce for each event target.
+ * var debounced = debounceByElement(batchLog, 250, { 'maxWait': 1000 });
+ * var source = new EventSource('/stream');
+ * jQuery(source).on('message', (e) => debounced(e.target)(e));
```
ChloeSartorelli, line 15: This example is modified from the lodash.debounce doc: https://github.com/lodash/lodash/blob/8a26eb42adb303f4adc7ef56e300f14c5992aa68/dist/lodash.js#L10364-L10367

emteknetnz (changes requested):
New solution seems much better

Just need to add a behat test - you can find out how to test the alert here - https://github.com/silverstripe/silverstripe-admin/blob/3/tests/behat/features/manage-users.feature#L112

GuySartorelli:
There's already a behat test in elemental which theoretically tests that the changetracking works (see https://github.com/silverstripe/silverstripe-elemental/blob/6/tests/Behat/features/change-tracking.feature) - the problem is this was a race condition so I don't think we can reliably test for it.

emteknetnz (approved):

Issue #2000, opened 2025-09-25
Accessibility: Add aria attributes to site tree


emteknetnz:
Split from - https://github.com/silverstripe/silverstripe-cms/issues/3101

[WAI-ARIA Tree View Pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/)

The CMS page site tree lacks aria attributes

## Context

The site tree in `/admin/pages/` had almost no keyboard interactivity even though there were various functions possible with mouse interaction.

The relevant keyboard interaction will have been added through the "related" issues below, but screen readers probably don't give necessary hints to blind users.

### Main Container
- The main container should have `role="tree"`. This tells screen readers that the element contains a hierarchical structure.

### Pages
- Each page item within the tree should have `role="treeitem"`.
- To indicate nesting levels, use `aria-level` (e.g., `aria-level="1"` for top-level pages, `aria-level="2"` for their children).
- For pages with nested children, use `aria-expanded="false"` when the list is collapsed and `aria-expanded="true"` when it's open.

### Expand/Collapse Arrow
- Icons controlling the expand/collapse action should have a dynamic and descriptive `aria-label`, such as "Toggle child pages".

### Mouse Drag-and-Drop
- ~The page being dragged should have `aria-grabbed="true"` while in motion.~
- ~The invisible drop zones where the page can be inserted should have `aria-dropeffect="move"`.~
- [aria-grabbed](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-grabbed) and [aria-dropeffect](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-dropeffect) are both deprecated in ARIA 1.1 and instead the current recommendation is to provide a non-dragging alternative i.e. "keyboard move mode" - https://www.w3.org/WAI/WCAG22/Understanding/dragging-movements.html

### Notes
- The `<ins>` elements with `tabindex="0"` added in https://github.com/silverstripe/silverstripe-cms/pull/3116 will need an appropriate role added

## Related
- https://github.com/silverstripe/silverstripe-admin/issues/1957
- https://github.com/silverstripe/silverstripe-cms/issues/3101
- https://github.com/silverstripe/silverstripe-cms/issues/3115

## Acceptance criteria
- [ ] Screen readers give sensible instructions and labels where appropriate
  - [ ] Whether nested tree is expanded or not, and how to expand/collapse it
[...]

Issue #1957, opened 2025-08-06
Accessibility: Decorative icons should not be read by screen readers


GuySartorelli:
[WCAG 2.2 A F96](https://www.w3.org/WAI/WCAG21/Techniques/failures/F96) 2.5.3
In the left menu (and other places) we’re using a font to represent icons, and psuedo `::before` elements with CSS like `content: "C"; font-family: "silverstripe";`.

That way, the text "C" is styled to look like an icon - but obviously this could represent a problem for screen readers.

The `speak:none;` css property ([spec](https://drafts.csswg.org/css-speech-1/#speaking-props-speak)) has also been applied, though [that isn’t widely supported](https://caniuse.com/?search=speak).

We will need to change the way these icons are applied so that there's no chance of decorative icons being read as text by screen readers.

## Places I know for sure we're using fonts for icons

- in the left menu for each section
- in the "previous/next" buttons in an edit form in a gridfield
- in buttons (e.g. "save", "publish")
- in the site tree (to indicate page types)
- asset admin (e.g. edit folder button in breadcrumbs)

There's probably a ton more

## Some options

If the element with the `font-icon-*` CSS class _only_ contains an icon, it's a lot easier to deal with than an element that has an icon _and other text_ inside it.

1. Replace `::before` pseudo element with a `background-image` or other CSS that displays the icon and doesn't inject text into the DOM
2. Add `aria-hidden="true"` property to any element that has a `font-icon-*` CSS class

If the element with the `font-icon-*` CSS class contains other content beyond just the icon, the content should be moved to be _next to_ the icon element, so that one of the above (or other) mitigations can be applied.

## Notes

- The existing functionality needs to keep working in case other developers are relying on it for their own icon purposes.

## Acceptance criteria

- [ ] When purely presentational icons are displayed in the CMS, screen readers ignore them entirely
- [ ] When non-presentational icons are displayed in the CMS, screen readers read appropriate text to describe them
- [ ] The changelog clearly calls out any changes that might affect custom css/js

## PRs - Templates and simple CSS

[**kitchen sink CI run**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/17366430816)

> [!NOTE]
> This is the first load of PRs - I'm breaking it up so it's not as much of a pain to review.
> This set of PRs includes any trivial `content:` in css that sets font icons, and any font icons set in templates.

[...]

Issue #1377, opened 2022-10-05
Update jstree from the thirdparty folder


GuySartorelli:
`jstree` in the thirdparty folder should be removed, and [jstree](https://github.com/vakata/jstree) should be added via yarn (yarn add jstree).

**NOTE:** There are some massively breaking changes going from the version in thirdparty to the most recent version. The chance for regressions was too high to do this upgrade in CMS 4.

GuySartorelli:
Closing this separate card - the jQuery update has surfaced some regressions which necessitate upgrading this as part of #1351

GuySartorelli:
Note: This didn't end up being updated in that card - but the number of regressions caused by _trying_ to update it were overwhelming and with very little benefit. I don't think there's sense in re-opening this card.

PR #2001, opened 2025-09-26
ENH Keyboard navigation for site tree

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/3101

Requires https://github.com/silverstripe/silverstripe-cms for behat to pass

Note that I've updated the original issue which was too large IMO so that keyboard nav, moving pages with keyboard (not even sure how to technically do this) and aria attributes have their own separate issues


composer.json
```diff
@@ -30,6 +30,9 @@
         "silverstripe/standards": "^1",
         "phpstan/extension-installer": "^1.3"
     },
+    "conflict": {
+        "silverstripe/cms": "<6.2.0"
```
emteknetnz: This isn't stricely necessary, it's here so that the `tabindex="0"` is on the `<ins>` elements - https://github.com/silverstripe/silverstripe-cms/pull/3116/files#diff-18770ed5912c530e92998ae0990dd4c39701fffbc73e4e55fec6d5fff6e4151eR1

If the conflict was not here, it would theoretically fail the behat test added in this PR, if we ran it as a --prefer-lowest (which we don't)

ChloeSartorelli: Please remove this since it isn't needed.

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -130,6 +130,111 @@ $.entwine('ss.tree', function($){
             // Clear out namespace before triggering so entwine handlers are called
             e.namespace = '';
             $(document).triggerHandler(e, data);
+          })
+          .on('keydown', function(e) {
+            var key = e.key;
```
ChloeSartorelli, line 135: I'm going to ignore use of `var` since I presume you're matching the style of this file rather than using our current standards.

tests/behat/features/accessibility.feature
```diff
@@ -65,3 +65,106 @@ Feature: Accessibility
...
+    When I press the "Tab" key globally
+    When I press the "Enter" key globally
+
+    # Tab to the site tree
+    When I press the "Tab" key globally
+    When I press the "Tab" key globally
+    When I press the "Tab" key globally
+    When I press the "Tab" key globally
+    When I press the "Tab" key globally
+    When I press the "Tab" key globally
```
ChloeSartorelli: Probably makes more sense to use `When I focus on the "a.page-view-link" element` (the interactive element just before the site tree) and just add a comment before that line something like `Move focus to the element before the site tree so we have a known starting point`

Then a single `I press the "Tab" key globally` will put focus on the first interactive element in the tree.

We don't need to explicitly test that there are x number of tabs before you get to the site tree - that will just be a pain if we add/remove interactive elements above the tree.

GuySartorelli (changes requested):
Works well locally with one exception - the `ins` that collapses the whole tree can't be accessed with keyboard.
I honestly don't know why you'd ever want to collapse the whole tree, but since it can be done with the mouse we should make it doable with keyboard too.

Note that against my better judgment I have opted to not test this using a screen reader, since screen reader compatibility has been split of into https://github.com/silverstripe/silverstripe-admin/issues/2000

## Question

Did you look at the docs for jstree, which is the underlying lib we use to control this site tree?
According to https://www.jstree.com/docs/interaction/ (and in their examples it's true) most of this keyboard interaction should come out of the box. We shouldn't need to add event handlers for all this.

Maybe we just need to update our version of jstree (it's in the `thirdparty/` folder) or fix something about how we're using it? Or if we're using the old version, we might need to include the [hotkeys plugin](https://old.jstree.com/documentation/hotkeys)

emteknetnz:
I originally did look at the jstree in the thirdparty library and hotkeys plugin - https://github.com/jeresig/jquery.hotkeys - it's unmaintained at this point (last commit was 10 years ago), so while we probably could add it, we're just adding tech-debt.

Instead I choose to just implement this manually since it's not hard and it gives us maximum control.

client/src/legacy/LeftAndMain.Tree.js
emteknetnz: Yup, am matching style

composer.json
emteknetnz: Removed

emteknetnz:
> Works well locally with one exception - the ins that collapses the whole tree can't be accessed with keyboard.
I honestly don't know why you'd ever want to collapse the whole tree, but since it can be done with the mouse we should make it doable with keyboard too.

Have added tabindex for this el

tests/behat/features/accessibility.feature
emteknetnz: Updated

tests/behat/features/accessibility.feature
```diff
@@ -65,3 +65,100 @@ Feature: Accessibility
...
+    And the "Page" "Page 2bi" is a child of a "Page" "Page 2b"
+    And the "Page" "Page 2bii" is a child of a "Page" "Page 2b"
+    And the "Page" "Page 2biii" is a child of a "Page" "Page 2b"
+    And the "Page" "Page 2c" is a child of a "Page" "Page 2"
+    And a "Page" "Page 3"
+    And I am logged in with "ADMIN" permissions
+    And I go to "/admin/pages"
+
+    # Move focus to the element before the site tree so we have a known starting point
+    When I focus on the ".font-icon-list" element
```
ChloeSartorelli: Don't use `.font-icon-*` css class as a behat selector for interactive elements, since https://github.com/silverstripe/silverstripe-admin/issues/1957 is moving these into child elements instead.

GuySartorelli:
I don't personally see how using an old but functional version of the plugin is any worse than having a chunk of new code we have to maintain that only works with our old but functional version of jstree :sweat_smile: 
Either way when we either upgrade or replace jstree we're gonna have to rip out the tech debt - removing a plugin seems easier to me than removing a big chunk of custom code that may have expanded by the time we do that.

But probably not worth arguing about. I found https://github.com/silverstripe/silverstripe-admin/issues/1377 which indicates we have tried to upgrade jstree in the past but ran into trouble, so upgrading for this card is off the table.

GuySartorelli (changes requested):

tests/behat/features/accessibility.feature
emteknetnz: Done

GuySartorelli (approved):
Works well locally

PR #2007, opened 2025-10-01
FIX Hide main content when preview mode is active

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/454

Fixes multiple issues where the preview panel is completely busted after merging the [skip link PR](https://github.com/silverstripe/silverstripe-admin/pull/1995), because the newly added `<main>` element also needs have `display: none` in order for the preview to display properly.

e.g. fixes https://github.com/silverstripe/silverstripe-cms/actions/runs/18144418597/job/51643320374#step:12:512


client/src/styles/legacy/_style.scss
```diff
@@ -85,6 +85,7 @@ html, body {
 }
 
 .cms-container--preview-mode {
+  .cms-container-skip-link-target,
   .cms-content {
```
emteknetnz, line 89: While we don't need technically the old `.cms-content` selector here in cms 6.2 and beyond because it's a child of .cms-container-skip-link-target, it's worth keeping so that we don't need to add any new composer conflicts

GuySartorelli:
> Fixes multiple issues where the preview panel is completely busted after merging the https://github.com/silverstripe/silverstripe-admin/pull/1995

Are all of the "multiple issues" represented in the failing CI, or are there other things I need to test too?

emteknetnz:
By multiple issues I meant there are multi CI failures, there only seems to be one actual issue which is that the preview isn't working

GuySartorelli:
Width of the preview panel in split mode is too narrow.

CMS 6.1
image


CMS 6.2
image

emteknetnz:
Looks like thats an existing regression on 6.x-dev, I've opened a new card for that https://github.com/silverstripe/silverstripe-admin/issues/2008

GuySartorelli:
I think it should be handled as part of this (it seems related - yes it exists on `6.x-dev`, that's the branch this PR is fixing) but it's not worth arguing with you about it.

GuySartorelli (approved):
Works fine except for the problem I pointed out that you're gonna handle separately.

PR #1988, opened 2025-08-27
ENH Hide font icons from screen readers

GuySartorelli:
- Updates CSS to support icons in inner `span` elements
- Replaces any `<i>` being used as an icon with `span` for consistency and semantic correctness
- Hides decorative icons from screen readers (usually by adding an inner `span`)
- In a few places, adds `aria-label` (will be noted with PR comments)

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1957

PR #2009, opened 2025-10-01
ENH Hide JS font icons from screen readers

GuySartorelli:
- Replaces any `<i>` being used as an icon with `span` for consistency and semantic correctness
- Hides decorative icons from screen readers (usually by adding an inner `span`)
- In a few places, adds `aria-label` (will be noted with PR comments)

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1957


client/src/components/Breadcrumb/Breadcrumb.js
ChloeSartorelli, line 1: Changes here allow other modules to still pass in `font-icon-*` css classes in the `className` prop - we then parse them out here and add them as a `span` child instead.

client/src/components/Button/Button.js
```diff
@@ -2,23 +2,26 @@ import React from 'react';
 import { Button as BaseButton } from 'reactstrap';
 import PropTypes from 'prop-types';
 import classnames from 'classnames';
-import IconHOC from './IconHOC';
 
-const Button = ({ className, noText, children, ...props }) =>
+const Button = ({ icon, className, noText, children, ...props }) =>
```
ChloeSartorelli, line 6: This prop used to be added through the `IconHOC`.
We can't rework that HOC to add the child span in a way that's super generic, but we can easily do it here for buttons specifically - so that's what we do.

client/src/components/Button/IconHOC.js
```diff
@@ -5,6 +5,8 @@ import classnames from 'classnames';
 /**
  * Higher order component that adds a `icon` props to another component and translate
  * it to a matching class.
+ *
+ * @deprecated 3.2.0 Use an inner span with aria-hidden="true" instead.
  */
 const IconHOC = (Component) => {
```
ChloeSartorelli, line 11: This HOC is only used in the `Button` component in core/supported modules - but deprecating it instead of outright removing it means there's less upgrade pain in case someone is using it in their own project/module.

client/src/components/FileStatusIcon/FileStatusIcon.js
```diff
@@ -28,7 +28,7 @@ class FileStatusIcon extends PureComponent {
         'SilverStripe\\Admin\\FileStatusIcon.TRACKED_FORM_UPLOAD_UNRESTRICTED',
         'Form submission, unrestricted access'
       );
-    return { className, 'data-title': dataTitle };
+    return { className, 'data-title': dataTitle, 'aria-label': dataTitle };
```
ChloeSartorelli, line 31: Adds an explicit `aria-label` which both hides the font icon from screen readers and ensures screen readers have the same textual context as sighted users.

Same as the one added below.

client/src/components/FormAction/FormAction.scss
ChloeSartorelli, line 1: Changes here allow for child elements other than `span` to hold font-icons (fixing an assumption I made in https://github.com/silverstripe/silverstripe-admin/pull/1988)

client/src/components/GridFieldActions/GridFieldActions.js
```diff
@@ -28,7 +28,6 @@ class GridFieldActions extends PureComponent {
       'btn',
       'btn--no-text',
       'btn-sm',
-      'font-icon-dot-3'
```
ChloeSartorelli, line 31: Added by default in `ActionMenu` now instead.

client/src/components/PopoverField/PopoverField.js
```diff
@@ -102,6 +114,7 @@ PopoverField.propTypes = {
   container: PropTypes.any,
   className: PropTypes.string,
   buttonClassName: PropTypes.string,
+  buttonIcon: PropTypes.string,
```
ChloeSartorelli, line 117: New prop is used by sharedraft module

client/src/font/styles.css
```diff
@@ -9,7 +9,6 @@
     url("fonts/silverstripe.svg#silverstripe") format("svg");
   font-weight: normal;
   font-style: normal;
-
```
ChloeSartorelli, line 12: unrelated code style fix

client/src/components/ActionMenu/ActionMenu.js
```diff
@@ -52,7 +54,9 @@ class ActionMenu extends PureComponent {
         toggle={this.toggle}
         {...restProps}
       >
-        <DropdownToggle className={toggleClassName} aria-label={i18n._t('Admin.ACTIONS', 'View actions')} {...dropdownToggleProps} />
+        <DropdownToggle className={toggleClassName} title={toggleText} aria-label={toggleText} {...dropdownToggleProps} >
```
ChloeSartorelli, line 57: Adding a `title` attribute here because:
1. There used to be a `.visually-hidden` text that isn't there anymore. That text was used in behat to identify buttons - now the `title` attribute is used instead
2. This gives sighted users the same context that screen reader users have.

emteknetnz (approved):

PR #2011, opened 2025-10-06
ENH Add tabIndex props

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/1374


client/src/components/InputField/InputField.js
```diff
@@ -113,6 +115,7 @@ InputField.defaultProps = {
   className: '',
   type: 'text',
   attributes: {},
+  tabIndex: 0,
```
ChloeSartorelli: Should probably be `null` by default i.e. explicitly don't add the attribute unless an override value is added. Please apply to all of the affected react components including in other PRs.

client/src/components/TreeDropdownField/TreeDropdownField.js
```diff
@@ -670,7 +670,10 @@ class TreeDropdownField extends Component {
 
     return (
       <div className={className}>
-        <span className="treedropdownfield__title">{title}</span>
+        <span
+          className="treedropdownfield__title"
+          tabIndex={this.props.tabIndex}
```
ChloeSartorelli: Why is a `span` getting a tab index?

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -447,10 +447,12 @@ FormBuilderLoader.propTypes = Object.assign({}, basePropTypes, {
   onFetchingSchema: PropTypes.func,
   onReduxFormInit: PropTypes.func,
   loadingComponent: PropTypes.elementType.isRequired,
+  fieldTabIndex: PropTypes.number,
```
ChloeSartorelli, line 450: Where is this prop used?

GuySartorelli (changes requested):

GuySartorelli:
Overall I don't think this is a robust solution. Even in our own modules an interactive element was forgotten about - it's likely many developers won't be updating their own custom react form fields to add `tabIndex` (and yes they do exist).

Maybe we can use [`inert`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/inert) on the block form's container instead? A quick test shows that still allows the form to correctly submit and the browser support is pretty good.

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
emteknetnz: In FormBuilder.js - see line 430 and how its props is constructed (uses this.props)

client/src/components/TreeDropdownField/TreeDropdownField.js
emteknetnz: Removed

client/src/components/InputField/InputField.js
emteknetnz: Updated

emteknetnz:
Closing as using `inert` attr instead

Issue #1966, opened 2025-08-11
Accessibility: Cannot open form tabs with keyboard


GuySartorelli:
WCAG 2.1.1

The interactivity for tabs in a form is incorrect.
https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ provides correct keyboard interaction for tabs.

We correctly use home/end/tab/arrow keys, but there is currently no way to actually open a tab panel. Pressing either space or enter should open the tab panel - or else the tab panel should open automatically when its tab is given focus (the latter is recommended under certain circumstances, see link above).

## Notes

- Tab panels _can_ be made to lazy-load in the CMS, which makes it hard to know whether we should follow the recommendation that "tabs activate automatically when they receive focus as long as their associated tab panels are displayed without noticeable latency". We should probably just require space or enter to be pressed across the board.

## Acceptance criteria

- [x] Form tab panels can be opened using the keyboard
- [x] Opening a tab panel _doesn't_ change where the focus is automatically
- [x] Pressing `tab` after opening a tab takes you to the first focusable element inside the tab panel
- [ ] If the tab panel has no focusable elements inside it, the tab panel itself should have `tabindex="0"` (if feasible). update - created new [issue](https://github.com/silverstripe/silverstripe-admin/issues/2014) for this
- [x] Check both entwine and react forms

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/18360720482

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2012
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1616
- https://github.com/silverstripe/developer-docs/pull/824

emteknetnz:
> Opening a tab panel doesn't change where the focus is automatically

There is an existing pattern where using left/right arrow keys on tab will automatically 'click' on the tab - this is done by jquery-ui tab https://jqueryui.com/tabs/ - at least for the secondary/panel navigation, and the primary nav for some things such as regular dataobjects

Because of this we should automatically "click on" the tabs as they are focused.

There is an existing issue where the some of the tabs don't do this:
- The primary nav for pages does not - due to this code https://github.com/silverstripe/silverstripe-admin/blob/3.0/client/src/legacy/LeftAndMain.js#L1434 - I tried removing this code but then AJAX just breaks, so I simply added new code in this PR to fix this
[...]

PR #2012, opened 2025-10-06
ENH Add keyboard navigation to tabs

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1966


client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -524,6 +524,44 @@ $.entwine('ss', function($){
     }
   });
 
+  /**
+   * Manually restore some functinality that's disabled in LeftAndMain.js
+   * in the `$('.cms-tabset').entwine({` block, because the default jquery-ui behaviour
+   * probably didn't work well with our custom tabs
```
ChloeSartorelli: Why is this being done here instead of just un-disabling the bit in `LeftAndMain.js`?
Also, I can't find where this is disabled there - there doesn't seem to be anything about `onkeydown` in the mentioned block?

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -524,6 +524,44 @@ $.entwine('ss', function($){
...
+        if (e.key === 'ArrowLeft') {
+          newIndex = index > 0 ? index - 1 : $tabs.length - 1;
+        } else {
+          newIndex = index < $tabs.length - 1 ? index + 1 : 0;
+        }
+        const $link = $tabs.eq(newIndex).find('a');
+        // Suffix #kb to the URL to indicate that the tab was changed through keyboard navigation
+        // This is used below to manually focus the active tab on page load
+        $link.attr('href', $link.attr('href').replace(/#kb$/, '') + '#kb');
+        $link.trigger('click');
```
ChloeSartorelli: I don't understand why we need this to be added to the URL? Feels like a bit of a smell.

> This is used below to manually focus the active tab on page load

Focusing on a tab has nothing to do with the URL??

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -524,6 +524,44 @@ $.entwine('ss', function($){
...
+
+  // If there is a #kb hash in the URL, focus on the currently active tab
+  $('.cms-tabset-nav-primary .ui-tabs-tab.ui-state-active').entwine({
+    onmatch: function() {
+      if (window.location.hash === '#kb') {
+        this.trigger('focus');
+      }
+      this._super();
+    },
+  });
```
ChloeSartorelli: Why is this based on the URL though? I don't understand.

What if I copy the URL and send it to someone? They'll get automatically focused on what to them is some random element. If they're relying on assistive technologies instead of sight that'll be really confusing.

tests/behat/features/accessibility.feature
```diff
@@ -65,3 +65,66 @@ Feature: Accessibility
     And the ".cms-container-skip-link-target" element should have focus
     When I press the "Tab" key globally
     And the ".ui-tabs-tab[aria-controls='Root_Main']" element should have focus
+@sboyd
```
ChloeSartorelli:
```suggestion

```

GuySartorelli (changes requested):
If I press "Home" or "End" it moves focus to the first/last tab respectively - but doesn't open that tab. This is only true for the top level tabs in admin sections (e.g. model admin tabs (i.e. swapping between `managed_models`), site tree content/settings/history tabs, archive admin Pages/other tabs).

Other tabs (e.g. inside the actual edit form) do seem to correctly open the tab when home/end is used.

For asset admin and the "Test React FormBuilder" section (presumably for all react tabs) the Home/End buttons don't do anything at all.

See also https://github.com/silverstripe/silverstripe-admin/issues/1966#issuecomment-3378863401

client/src/legacy/LeftAndMain.EditForm.js
emteknetnz: The pjax replaces lots of HTML, including the nav bar were using the keyboard on. We need to manually focus nav link otherwise you'll be focused on nothing.

You don't get this issue when using a mouse because there's no need to focus on things

client/src/legacy/LeftAndMain.EditForm.js
emteknetnz: If I undisable the bit in LeftAndMain.js, I get a bunch of server errors in toast messages

The fact it is disabled in LeftAndMain implies that back in the day people tried to get this to work, though gave up because there's some good reason why it does not work.

client/src/legacy/LeftAndMain.EditForm.js
emteknetnz: Updated to use sessionStorage instead

client/src/legacy/LeftAndMain.EditForm.js
emteknetnz: Updated to use sessionStorage

tests/behat/features/accessibility.feature
emteknetnz: Removed

client/src/legacy/LeftAndMain.EditForm.js
ChloeSartorelli: > Also, I can't find where this is disabled there - there doesn't seem to be anything about `onkeydown` in the mentioned block?

Can you please respond to that part? It's important for this review that I'm able to see what was disabled and where, so that I can understand for myself why we can't just enable that functionality again (whatever it was).

You've edited the comment a bit but it's still not 100% clear to me. Was it specifically the "open the tab panel when the tab itself has focus" part that got disabled?
If so that's fine. Just wanna make sure I do correctly understand that.

GuySartorelli (changes requested):
I haven't reviewed the code at all this round because the problems in https://github.com/silverstripe/silverstripe-admin/pull/2012#pullrequestreview-3312134321 still haven't been resolved.

Also page down and page up do work for some reason? Though I don't know if those are even meant to do anything at all? They're certainly not mentioned in [W3C's tabs pattern document](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/).

The "up" and "down" keys also don't work how I would expect at all - "up" goes to the _last_ tab and "down" goes to the _first_ tab? I would have expected just the previous and next tabs like left/right do. Or just do nothing per the W3C tabs pattern recommendations (looks like that would be easiest).

client/src/legacy/LeftAndMain.EditForm.js
emteknetnz: https://github.com/silverstripe/silverstripe-admin/blob/3/client/src/legacy/LeftAndMain.js#L1424-L1436 has the 'disabled auto load functionality'

beforeLoad return false stops a near duplicate ajax request (I think the only difference is it's missing an x-Pjax header) which will return a 400, which is what results in the error toast

beforeActivate is what disables the primary nav from autoloading. If I comment out `return false`, then I can use arrow keys and the tabs will autofocus. However the actual panels themselves won't load. Presumably back in the day people tried to get this to work, though they ran into issues with integrating it with the way the CMS is put together.

emteknetnz:
> Also page down and page up do work for some reason? 

PageUp/PageDown do work because I put in keybinds for them :-) We have those in the sitetree (go to the start/end of a group) so seems pretty easy to also have them here for consistency. They're easy enough to remove if we decide we don't like them.  Maybe it's a bad thing because in that context you expect the regular pageup/pagedown of scroll the whole window - https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ has them listed as (optional) in this context **just noticed they say home/end, not pageup/pagedown** - will change to home/end

(I'm OK also removing pageup/pagedown from the site tree, quick check in my linux file explorer shows that pageup/pagedown doesn't go to the top/bottom of subfolders, it does the whole entire window) 

> The "up" and "down" keys also don't work how I would expect at all - "up" goes to the last tab and "down" goes to the first tab?

Up does the same as left, down does the same as right.

Left/Up and Right/Down will both "wrap-around", which is possibly what you're experiencing? I'm just copying the behaviour from https://jqueryui.com/tabs/

https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ states - `(left arrow) If focus is on the first tab, moves focus to the last tab.` so the wrap around is standard

emteknetnz:
Just noticed on https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ it says `If the tab list is horizontal, it does not listen for Down Arrow or Up Arrow so those keys can provide their normal browser scrolling functions even when focus is inside the tab list.` - I'll remove the up/down binds

GuySartorelli:
I recommend following the example from W3C which we know is good practice rather than following jQuery UI's example.

Up/Down scroll the page, and as you noted page up/page down scroll the page further - so I think we shouldn't have key binds for those keys on our tabs (on _any_ tabs in the CMS, site tree or otherwise). As I mentioned in person I won't make a big fuss about up/down specifically, but page up and page down we should definitely remove.

Left and right wrapping around is what happens with W3C's examples too so that's fine.

All keys that are meant to do things to tabs should probably be mentioned in the changelog entry. I'm using that as my guide for testing, because that's what we're telling people the behaviour should be (and because the PR description has no information about what the behaviour should be).

Edit: Looks like whatever changes you made last round have removed the functionality that isn't recommended by W3C and fixed home/end functionality :+1:

GuySartorelli (approved):
LGTM, works well locally and follows relevant W3C suggestions

PR #2013, opened 2025-10-08
ENH Hide PHP font icons from screen readers

GuySartorelli:
> [!IMPORTANT]
> Requires https://github.com/silverstripe/silverstripe-framework/pull/11862 for CI to go green.
> May also require https://github.com/silverstripe/silverstripe-cms/pull/3119 for CI to go green.

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1957


client/src/components/Button/Button.js
```diff
@@ -9,7 +9,7 @@ const Button = ({ icon, className, noText, children, ...props }) =>
     aria-label={noText ? children : undefined}
     {...props}
   >
-    {icon && <span className={`font-icon-${icon}`} aria-hidden="true" />}
+    {icon && <span className={`btn__icon font-icon-${icon}`} aria-hidden="true" />}
```
ChloeSartorelli, line 12: `btn__icon` class corresponds with the `btn__title` (or sometimes `btn__text`) CSS class that the span holding text usually uses.

There are a few places that look specifically for this new `btn__icon` CSS class.

client/src/components/FormAction/FormAction.scss
```diff
@@ -90,6 +90,8 @@
 
 // Used buttons with text and icons, but you wan to hide the text only. eg. gridfield pagination
 // Not needed for inner spans holding the icons
+// This is only here for backwards compatibility - you should use `btn--no-text` instead and just not include text in the button.
+// Ideally a `title` and `aria-label` attribute are added for accessibility.
 .btn--hide-text[class*="font-icon-"] {
```
ChloeSartorelli, line 95: This isn't used anywhere anymore. It had been used specifically and only when there was a button with a font icon, and the text was being hidden with `display: none`.... which basically meant the text was useless.

Everywhere this was being used, we now add `title` and `aria-label` attributes instead. This surfaces the text for screen readers and when hovering over the button for sighted users, while keeping it out of the button itself (i.e. same behaviour as before but now it's accessible).

I opted to add this comment instead of deleting on the off chance someone is relying on this for something custom (it's not uncommon to look at what CSS classes are used in the CMS and just copy them into custom stuff). We have a standing action to look for the words "backwards compatibility" when doing a major, so it should be removed at that time.

client/src/components/PopoverOptionSet/PopoverOptionSet.scss
```diff
@@ -70,8 +70,15 @@ $divided-panel-padding-x: calc($panel-padding-x / 2);
...
+  // Removes the margin-right from the .btn__icon element since it's added to the
+  // :before pseudo element instead
+  & &__button .btn__icon {
+    margin-right: 0;
+  }
+
   // Adds necessary specificity to override default font-size and margins
-  & &__button:before {
+  & &__button:before,
+  & &__button .btn__icon:before {
```
ChloeSartorelli, line 81: I tried just having the margin on the `.btn__icon` element but it messed up the appearance for some reason - easier to just do this which effectively matches what it was before.

code/LeftAndMain.php
```diff
@@ -985,7 +985,7 @@ public function getEditForm($id = null, $fields = null)
                     $actions->push(
                         FormAction::create('save', _t(__CLASS__ . '.SAVE', 'Save'))
                             ->addExtraClass('btn btn-primary')
-                            ->addExtraClass('font-icon-add-circle')
+                            ->setIcon('add-circle')
```
ChloeSartorelli, line 988: `setIcon()` method was already there, it just didn't seem to do anything. It now is used to add the accessible `btn__icon` span.

themes/cms-forms/templates/SilverStripe/Forms/GridField/GridField_FormAction.ss
ChloeSartorelli, line 1: Done here to match the `FormAction.ss` template also added in this module's theme. Allows us to add the new icon span without affecting the (rare) case of someone using gridfield in the frontend (yes it does happen).

client/src/components/FormAction/README.md
```diff
@@ -15,7 +15,7 @@ Used for form actions. For example a submit button.
    * 'outline-secondary'
    * 'link'
    * 'danger'
- * `icon` (string): The icon to be used on the button, adds `font-icon-{icon}` class to the button. See available icons [here](../../../../fonts/incon-reference.html).
+ * `icon` (string): The icon to be used on the button, adds `font-icon-{icon}` class to a child element in the button. See available icons [here](../../../../fonts/incon-reference.html).
```
emteknetnz:
```suggestion
 * `icon` (string): The icon to be used on the button, adds `font-icon-{icon}` class to a child element in the button. See available icons [here](../../../../fonts/icon-reference.html).
```

emteknetnz (changes requested):

client/src/components/FormAction/README.md
ChloeSartorelli: lol there was a lot wrong with that link. Have fixed it.

emteknetnz (approved):

Issue #2008, opened 2025-10-01
6.2 REGRESSION: Split view is too small


emteknetnz:
installer 6.x-dev

The split view is too small

<img width="2237" height="1285" alt="Image" src="https://github.com/user-attachments/assets/90f3ff64-191b-4ed0-acd6-c3585c4e67ca" />

## PRs

- https://github.com/silverstripe/silverstripe-admin/pull/2017

GuySartorelli:
Note it looks like the left menu is also not as wide as it's meant to be when expanded.

PR #2017, opened 2025-10-14
FIX Restore correct widths to main CMS panel.

GuySartorelli:
The introduction of a new parent to .cms-main resulted in some of the CSS and JS that handles the width of the main panel being incorrect. Visually it appeared that the left and right panels were too narrow - but in reality that's just because the main was was too wide.

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/2008


client/src/legacy/LeftAndMain.js
```diff
@@ -353,7 +353,7 @@ $.entwine('ss', function($) {
       var mode = options.mode;
       this.clearViewMode();
 
-      var content = this.find('.cms-content');
+      var content = this.find('.cms-container-skip-link-target');
```
ChloeSartorelli, line 356: We need to target the parent here, or else even if the child is narrower, the parent will be too wide and push the preview panel out of the way.

client/src/styles/legacy/_style.scss
```diff
@@ -66,12 +66,6 @@ html, body {
   }
 }
 
-.cms-container-skip-link-target {
-  display: flex;
-  height: 100%;
-  width: 100vw;
-}
```
ChloeSartorelli, line 73: Replaced with standard CSS classes

templates/SilverStripe/Admin/LeftAndMain.ss
```diff
@@ -19,7 +19,7 @@
     <div class="cms-container" data-layout-type="custom">
         $Menu
         <main id="cms-container-skip-link-target"
-             class="cms-container-skip-link-target"
+             class="cms-container-skip-link-target flexbox-area-grow fill-height fill-width"
```
ChloeSartorelli, line 22: Standard CSS classes - this change makes the main panel stop squishing the left menu panel.

emteknetnz (approved):

Issue #1478, opened 2023-03-23
CMS5: Model Admin has doubled search button


maxime-rainville:
CMS 5 has doubled the Search button on Gridfield.

![image.png](https://images.zenhubusercontent.com/5ac42fb74b5806bc2bcd8e13/3da9ad62-4110-4d70-90b9-c2e5ad23ff05)

## Acceptance criteria
- Only the top button remains.
- Plain Gridfield search still works.

## Notes
- ~~Might be related to https://github.com/silverstripe/silverstripe-framework/pull/10733, but not sure.~~
- I've only seen it in CMS5, but didn't test it in the most up-to-date version of CMS4

## PRs
- https://github.com/silverstripe/silverstripe-framework/pull/10736

GuySartorelli:
Merged and confirmed it only affected CMS 5

PR #2020, opened 2025-10-16
FIX Don't add unnecessary column to grid field

GuySartorelli:
As noted by the comment in the deleted code, this functionality was adding a column for the legacy search button, which [was removed in CMS 5](https://github.com/silverstripe/silverstripe-admin/issues/1478). It serves no purpose anymore.

I'm only targeting `3` instead of as a patch on the off chance someone has some bizarre functionality that relies on this phantom column.

Note I don't think it's possible for a single row to have actions - seems like it's all or nothing - so I haven't paid any attention to the other acceptance criteria.

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11817


GuySartorelli:
PHPunit CI failures are happening on the `3` branch too: https://github.com/silverstripe/silverstripe-admin/actions/runs/18549159933

emteknetnz (changes requested):
Seems like the `<th class="main col-Actions">` is still there?  Also I'm not sure why we'd want to remove this any since we'd then be a `<th>` short?  On existing code, the `<th>` is hidden.

Perhaps I'm testing this wrong?

Make a MyDataObject, give it a child MySubDataObject - resulting GridField still has a  `<th class="main col-Actions">` despite no actions in the row

```php
// MyDataObject.php

use SilverStripe\Forms\GridField\GridFieldDeleteAction;
use SilverStripe\Forms\GridField\GridFieldEditButton;
use SilverStripe\Forms\GridField\GridFieldSortableHeader;

{

    private static $db = [
        'Title' => 'Varchar'
    ];

    private static $has_many = [
        'MySubDataObjects' => MySubDataObject::class
    ];

    {
        $gridField = new GridField('MySubDataObjects', 'MySubDataObjects', $this->MySubDataObjects());
        $config = GridFieldConfig_RelationEditor::create();
        // $config->removeComponentsByType(GridFieldSortableHeader::class);
        $config->removeComponentsByType(GridFieldEditButton::class);
        $config->removeComponentsByType(GridFieldDeleteAction::class);
        $gridField->setConfig($config);
        $fields->insertAfter('Title', $gridField);
        return $fields;
    }
}

// MySubDataObject.php

class MySubDataObject extends DataObject
{
    private static $table_name = 'MySubDataObject';

    private static $db = [
        'Title' => 'Varchar'
    ];

    private static $has_one = [
        'MyDataObject' => MyDataObject::class
    ];
}


// MyModelAdmin.php

{

        MyDataObject::class,
    ];
}
```

GuySartorelli:
Ooh interesting, I'll figure out what's causing it there. I was testing with the reports admin which has no actions already. Didn't try building a gridfield from scratch.

>  Also I'm not sure why we'd want to remove this any since we'd then be a `<th>` short? On existing code, the `<th>` is hidden.

If there's no actions, there should be no actions column so we would not be a `<th>` short.


Also rebased since CI is green in the base branch now.

GuySartorelli:
Added framework PR to tidy up that scenario.

emteknetnz (changes requested):
Still doesn't seem right with the framework PR added (I've got fresh version of all the PRs so pretty sure I'm testing the correct things)

Now I'm seeing the `<th class="main col-Actions">` with a `display:none` stylesheet rule on it, and that's whether or not there's actions in the column e.g. I commented out the `$config->removeComponentsByType(GridFieldEditButton::class);` line in my example code

GuySartorelli:
Ahh, I had modified your example to remove all actions by interface rather than the explicit ones you have, and forgot to change it back.
Versioned is adding a `GridFieldArchiveAction` which results in `$this->getItems($gridField)` being truthy inside the `GridField_ActionMenu`.

Of course for non-versioned records the archive action doesn't add any HTML to the gridfield, but it's too late by that point to tell the action menu not to include it in the action count.

The only solution to that which I can think of is to hook into the `onBeforeRenderHolder()` extension hook method and remove the archive action if `$gridField->getModelClass()` doesn't have the versioned extension. Does that sound sensible to you?

emteknetnz:
It's a bit messy, though we already do something sort of similar for the sudo mode button in GridField i.e. alter it all right before render based on some conditional logic - https://github.com/silverstripe/silverstripe-framework/blob/6/src/Forms/GridField/GridField.php#L554

GuySartorelli:
I'm gonna take that as a "in lieu of a better way to do it, that sounds like an acceptable approach" and implement it.

emteknetnz (changes requested):
Merge conflict

GuySartorelli:
Rebased

emteknetnz (approved):

Issue #1914, opened 2025-03-24
CMS `ui-dialog` close button not rendering correctly


jackkemmish:
### Module version(s) affected

2.3.5

### Description

Hi there, I've noticed that the close button for the `ui-dialog` in the CMS doesn't seem to be rendering correctly.

I only noticed this as I am using it as part of `GridFieldAddExistingSearchButton` in the `symbiote/silverstripe-gridfieldextensions` module, but am assuming this modal is standard Silverstripe UI.

### How to reproduce

For my uses case, you can install `symbiote/silverstripe-gridfieldextensions` module so you can use the `GridFieldAddExistingSearchButton` class on the GridField which triggers the `ui-dialog` component.

A few steps I found to reproduce:
1. Opening the modal no red close icon.
2. Clicking on the modal must focus it which then reveals the red close icon.
3. Hovering on the close button area then hides the red close icon.

When hovering on the close `button`, it looks like the `.ui-button .ui-icon` background image is being changed - I assume for a hover state, but this doesn't seem to be working either.

Here is a reduced code example of my set up to get the modal opening - I'm using `fromholdio/silverstripe-minigridfield` purely for a condensed view in the CMS:
```php
<?php

namespace Sdots\App\Elemental;

use DNADesign\Elemental\Models\BaseElement;
use Fromholdio\MiniGridField\Forms\MiniGridField;
use SilverStripe\Forms\GridField\GridFieldAddNewButton;
use SilverStripe\Forms\GridField\GridFieldConfig_RelationEditor;
use SilverStripe\Forms\GridField\GridFieldEditButton;
use SilverStripe\Forms\GridField\GridFieldFilterHeader;
use SilverStripe\Versioned\GridFieldArchiveAction;
use Symbiote\GridFieldExtensions\GridFieldAddExistingSearchButton;
use Symbiote\GridFieldExtensions\GridFieldOrderableRows;
use SilverStripe\Forms\FieldList;
use App\Projects\ProjectPage;

class ProjectCarouselBlock extends BaseElement
{
    private static string $table_name = 'ProjectCarouselBlock';
    private static string $singular_name = 'Project carousel block';
    private static string $plural_name = 'Project carousel blocks';
    private static string $description = '';
    private static string $icon = 'font-icon-block-carousel';

    private static array $many_many = [
        'Projects' => ProjectPage::class
    ];

    private static array $many_many_extraFields = [
        'Projects' => [
            'SortOrder' => 'Int'
        ]
[...]

PR #2025, opened 2025-10-22
FIX Make jQueryUI icon render nicely

GuySartorelli:
Before:

Screenshot from 2025-10-22 13-15-49

Note the icon only _sometimes_ displayed as noted in the issue - that has been fixed.

After:

Screenshot from 2025-10-22 13-18-00

It's not perfect, but it's not _broken_ which is the main thing.
Making it actually _nice_ will involve updating some of the styling to be closer to (if not properly match) the bootstrap modals we use almost everywhere else in the CMS, which should be done in a minor release.

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1914


emteknetnz (approved):
It's far from perfect, though I guess it's slightly better

Issue #2030, opened 2025-10-22
Fatal error passing string to ?int


NightJar:
https://github.com/silverstripe/silverstripe-admin/blob/b4a3aa0f7313a2b48e3aa2189468a2c64336d02c/code/LeftAndMain.php#L1034-L1035

Affects CMS 6 (admin v3) too.

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2033

GuySartorelli:
Please use the issue template in the future, which prompts for information that makes investigating and fixing bugs easier.

For now:

1. please provide some clear step-by-step instructions for how to reproduce this
2. You say "affects CMS 6 ... too" - do you mean to imply that this is a bug in both CMS 5 and CMS 6? Or only CMS 6?

NightJar:
I don't know how to reproduce it.
Just that my logs are reporting it.

NightJar:
I thought it'd be pretty on the nose.

This function returns a string.
https://github.com/silverstripe/silverstripe-framework/blob/5/src/Control/HTTPRequest.php#L704
https://github.com/silverstripe/silverstripe-framework/blob/6/src/Control/HTTPRequest.php#L706

And this function wants an integer.
https://github.com/silverstripe/silverstripe-admin/blob/2/code/LeftAndMain.php#L1788
https://github.com/silverstripe/silverstripe-admin/blob/3/code/LeftAndMain.php#L1232

And the highlighted code in the report sends the output of the former directly to the latter.
https://github.com/silverstripe/silverstripe-admin/blob/c8059d2007c94a9eb5b5748a60c032892fc666e7/code/LeftAndMain.php#L577-L581

I'm a little bit surprised static analysis hadn't picked it up 😮

GuySartorelli:
PR merged. This will be automatically tagged by GitHub actions

NightJar:
NIce one, thanks team 🙏

PR #2033, opened 2025-10-27
FIX Ensure arg is integer like

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2030

Original issue can be replicated by going to `/admin/pages/edit/show/def`

If a non integer value is passed, then it will simply use the last loaded value instead e.g. if I was previously on the "About us" page `/admin/pages/edit/show/2`, then going to `/admin/pages/edit/show/def` will show the "About us" page

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -1032,7 +1032,7 @@ class LeftAndMain extends Controller implements PermissionProvider
     public function show(HTTPRequest $request): HTTPResponse
     {
         if ($request->param('ID')) {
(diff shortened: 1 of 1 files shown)


GuySartorelli:
> going to `/admin/pages/edit/show/def` will show the "About us" page

It should give an error if the ID is not for a valid record.
For example if I go to `/admin/pages/edit/show/123456789`:

image

emteknetnz:
Fixed

GuySartorelli (approved):
LGTM, works as expected

Issue #2014, opened 2025-10-08
Accessibility: Set tabindex=0 on tab panels with no content


emteknetnz:
Follow on from https://github.com/silverstripe/silverstripe-admin/issues/1966

https://www.w3.org/WAI/ARIA/apg/patterns/tabs/ Keyboard Interaction Notes 4 states

`When the tabpanel does not contain any focusable elements or the first element with content is not focusable, the tabpanel should set tabindex="0" to include it in the tab sequence of the page`

An example of non-tabbable data might be a `LiteralField` that has `<p>lorem ipusm</p>` in it

We have 3x types of applicable tab panels
- The primary tabs (e.g. Content/Settings/History) that pjaxes in heaps of content
- Secondary tabs (e.g. getCMSFIelds()) that shows/hides content based using jquery
- React based tabs (`<TabNav>`) that do their own thing

## Acceptance criteria
- [x] Tab panels without tabbable content have tabindex="0" added to them

**[Kitchen sink CI](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/18894204748)**

## PRs
- https://github.com/silverstripe/silverstripe-frameworktest/pull/248
- https://github.com/silverstripe/silverstripe-admin/pull/2035
- https://github.com/silverstripe/developer-docs/pull/832

PR #2035, opened 2025-10-28
ENH Set tabindex=0 on tab panels with no focusable elements

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2014

Needs https://github.com/silverstripe/silverstripe-frameworktest/pull/248 for begat to pass

To test the react version - upload a file and view the 'Used on' tab, which has no focusable elements

To test the cms edit form tabs

```php
use SilverStripe\Forms\LiteralField;
use SilverStripe\Forms\TabSet;

class Page extends SiteTree
{
    {
        $fields = new FieldList([new TabSet('Root')]);
        // toggle this to try with and without secondary tab nav
        if (false) {
            $fields->addFieldToTab('Root.Main', new LiteralField('lf01', '<p id="lf01">Nothing to focus on</p>'));
        } else {
            $fields->addFieldToTab('Root.Main', new LiteralField('lf01', '<p id="lf01">Nothing to focus on</p>'));
            $fields->addFieldToTab('Root.Tab02', new LiteralField('lf02', '<p id="lf02" tabindex="0">Something to focus on</p>'));
        }
        return $fields;
    }
}

```


client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -656,6 +656,30 @@ $.entwine('ss', function($){
...
+        'button:not([disabled])',
+        'input:not([disabled]):not([type="hidden"])',
+        'select:not([disabled])',
+        'textarea:not([disabled])',
+        '[tabindex]:not([tabindex="-1"])',
+        'iframe',
+        'object',
+        'embed',
+      ].join(', ');
+      if (this.get(0).querySelectorAll(cssSelector).length === 0) {
```
ChloeSartorelli: Why use `querySelectorAll()` instead of jQuery's `find()`?

Using jQuery means we can get more accurate results, e.g. using [`:hidden`](https://api.jquery.com/hidden-selector/) for elements that have `display:none`, and using [`:disabled`](https://api.jquery.com/disabled-selector/) which takes a lot more into account such as being inside a disabled `fieldset`, and using [`:input`](https://api.jquery.com/input-selector/) instead of listing all the input-like elements.

While a little gross, we also have precedent for using jQuery inside react elements, and could get the same benefit there.

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -656,6 +656,30 @@ $.entwine('ss', function($){
...
+      const cssSelector = [
+        'a[href]',
+        'button:not([disabled])',
+        'input:not([disabled]):not([type="hidden"])',
+        'select:not([disabled])',
+        'textarea:not([disabled])',
+        '[tabindex]:not([tabindex="-1"])',
+        'iframe',
+        'object',
+        'embed',
```
ChloeSartorelli: Are these focusable? I wouldn't have thought so, but I could easily be wrong.

client/src/legacy/LeftAndMain.EditForm.js
```diff
@@ -609,7 +609,7 @@ $.entwine('ss', function($){
   });
 
   /**
-   * Add support for key navigation between tabs on the primary/main tab nav
+   * Add support for key navigation between tabs on the secondary nav
    *
    * Note that prior to this, arrow keys did work on secondary tabsets, presumably
    * set by jquery-ui tabs, and they did auto load the panel, though PageUp and PageDown did not.
```
ChloeSartorelli:
```suggestion
   * set by jquery-ui tabs, and they did auto load the panel, though Home and End did not.
```
Probably worth fixing while we're in here.

GuySartorelli (changes requested):

client/src/legacy/LeftAndMain.EditForm.js
emteknetnz: Updated

client/src/legacy/LeftAndMain.EditForm.js
emteknetnz: According to https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex they are

Based on that article I'm also added 'summary' and removed 'embed'. I did not add area or frame since they do not seem applicable in this context

client/src/legacy/LeftAndMain.EditForm.js
emteknetnz: Updated

emteknetnz:
CI failure is existing https://github.com/silverstripe/silverstripe-admin/actions/runs/19017583561/job/54307725239#step:12:971

client/src/legacy/LeftAndMain.EditForm.js
ChloeSartorelli: This now uses `find()` but doesn't use jQuery's selectors (`:hidden:`, `:disabled`, `:input`) and the react code wasn't updated not was the note about it responded to?

emteknetnz:
JS based solution needs to run on every tab panel render, while it would it works it's really suboptimal from a performance point of view

Issue #2037, opened 2025-10-29
Replace use of defaultProps with Javascript default parameters


emteknetnz:
When running any admin js with `yarn dev`, the console is fill with these sorts of warnings, which are really annoying

<img width="533" height="293" alt="Image" src="https://github.com/user-attachments/assets/95c3d4e2-dc11-4340-af52-7ce7fc7f1cb5" />

## Acceptance criteria
- [x] All functional components that use defaultProps are updated to used Javascript default parameters
- [x] All supported modules with javascript are updated
- [x] Target for patch release (not minor)

**[Kitchen sink CI](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/18927239854)**

## PRs
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1628
- https://github.com/silverstripe/silverstripe-admin/pull/2041
- https://github.com/silverstripe/silverstripe-elemental/pull/1383
- https://github.com/silverstripe/silverstripe-linkfield/pull/417
- https://github.com/silverstripe/silverstripe-mfa/pull/622

GuySartorelli:
PRs merged

PR #2041, opened 2025-10-29
ENH Use javascript default parameters instead of defaultProps

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2037


client/src/components/Button/Button.js
```diff
@@ -16,9 +21,4 @@ Button.propTypes = {
   noText: PropTypes.bool
 };
 
-Button.defaultProps = {
-  ...BaseButton.defaultProps,
```
emteknetnz, line 20: vendor/silverstripe/admin/node_modules/reactstrap/src/Button.js has no defaultProps, so this is safe

client/src/components/Button/IconHOC.js
```diff
@@ -18,7 +22,6 @@ const IconHOC = (Component) => {
     icon: PropTypes.string
   };
 
-  IconComponent.defaultProps = Component.defaultProps;
```
emteknetnz, line 21: IconHOC is only used on Button.js. 

Button.defaultProps did not include className or icon, so removing this has no effect

client/src/components/Button/BackButton.js
```diff
@@ -3,20 +3,21 @@ import Button from 'components/Button/Button';
...
+  children = i18n._t('Admin.BACK', 'Back'),
+  ...props
+}) => (
   <Button
     className={classnames(className, 'back-button')}
+    noText={noText}
+    icon={icon}
     {...props}
-  />
+  >{children}</Button>
```
emteknetnz, line 18: There's conditional logic on Button.js as to whether children will show or not

client/src/components/Button/Button.js
```diff
@@ -1,24 +1,18 @@
...
   (<BaseButton
     className={classnames(className, { 'btn--no-text': noText })}
     aria-label={noText ? children : undefined}
     {...props}
   >{noText ? undefined : children}</BaseButton>);
 
-Button.propTypes = {
-  ...BaseButton.propTypes,
-  noText: PropTypes.bool
-};
```
ChloeSartorelli, line 17: Why are the `propTypes` being removed? That seems unrelated?

client/src/components/Button/IconHOC.js
ChloeSartorelli: Since this isn't exposed in bundle.js it shouldn't be used by other modules either so this should be fine.

client/src/components/Search/SearchForm.js
```diff
@@ -68,11 +75,6 @@ SearchForm.propTypes = {
   forceLoadForm: PropTypes.bool,
 };
 
-SearchForm.defaultProps = {
-  formData: {},
```
ChloeSartorelli, line 72: `formData` not in the constructor? Looks like it wasn't used though so I assume it was omitted intentionally?

GuySartorelli (changes requested):

client/src/components/Button/Button.js
emteknetnz: Opps, reverted

client/src/components/Search/SearchForm.js
emteknetnz: Yup intentional

GuySartorelli (approved):
LGTM

PR #2042, opened 2025-10-30
ENH Remove interactivity from cells with child interactive items

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11820

Behat failures [are existing](https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/18929030532/job/54042156188)

You could argue that this PR is not actually needed as the `<tr>` is not tabbable, while the action-menu is. 

What this PR does is put the `cursor: pointer` inline style and click functionality on the `<td>` instead, and only if it that `<td>` does not have child interactive elements.

The css :hover style on the parent `<tr>` remains unchanged.

It's a little weird honestly though I'm not sure how else to retain the existing "click the `<tr>`" functionality, which needs to be retained, as it would be extremely annoying for mouse users if they had to open the action menu in order to edit a record

You can see the weirdness in action if one of data `<td>`s has an interactive child

```php
// app/src/MyModelAdmin.php

{

        MyDataObject::class,
    ];
}


// app/src/MyDataObject.php

use SilverStripe\Forms\GridField\GridFieldDataColumns;

{

    private static $db = [
        'Title' => 'Varchar'
    ];

    private static $has_many = [
        'MySubDataObjects' => MySubDataObject::class
    ];

    {
        $gridField = new GridField('MySubDataObjects', 'MySubDataObjects', $this->MySubDataObjects());
        $config = GridFieldConfig_RelationEditor::create();
        $dataColumns = $config->getComponentByType(GridFieldDataColumns::class);
        $dataColumns->setDisplayFields([
            'Title' => 'Title',
            'ThingDisplay' => 'ThingDisplay',
            'SomeOtherField' => 'Some Other Field'
        ]);
        $gridField->setConfig($config);
        $fields->insertAfter('Title', $gridField);
        return $fields;
    }
}


// app/src/MySubDataObject.php

use SilverStripe\Forms\LiteralField;

class MySubDataObject extends DataObject
{
    private static $table_name = 'MySubDataObject';

    private static $db = [
        'Title' => 'Varchar'
    ];

    private static $has_one = [
        'MyDataObject' => MyDataObject::class
    ];

    public function getThingDisplay()
    {
        return new LiteralField('link', "<a href='https://example.com'>example.com</a>");
    }
}
```


client/src/legacy/GridField.js
```diff
@@ -335,28 +335,42 @@ $.entwine('ss', function($) {
...
+      var cssSelector = [
+        'a[href]',
+        'button:not([disabled])',
+        'input:not([disabled]):not([type="hidden"])',
+        'select:not([disabled])',
+        'textarea:not([disabled])',
+        '[tabindex]:not([tabindex="-1"])',
+      ].join(', ');
+      var hasInteractiveElements = this.find(cssSelector).length > 0;
+      this.data('hasInteractiveElements', hasInteractiveElements);
```
emteknetnz, line 352: Using jquery .data() rather than data-* html attributes for better encapsulation as nothing than this code should need to use the state

client/src/legacy/GridField.js
```diff
@@ -335,28 +335,42 @@ $.entwine('ss', function($) {
...
+      this._super();
+      // Cache data on whether the parent row has interactive elements and any link to a form
+      var formLink = this.closest('tr').find('.edit-link, .view-link').prop('href');
+      var cssSelector = [
+        'a[href]',
+        'button:not([disabled])',
+        'input:not([disabled]):not([type="hidden"])',
+        'select:not([disabled])',
+        'textarea:not([disabled])',
+        '[tabindex]:not([tabindex="-1"])',
```
GuySartorelli:
```suggestion
        'a[href]',
        ':input:not(:disabled, :hidden)',
        '[tabindex]:not([tabindex="-1"])',
```
See jQuery docs:
- https://api.jquery.com/input-selector/
- https://api.jquery.com/disabled-selector/
- https://api.jquery.com/hidden-selector/

client/src/legacy/GridField.js
```diff
@@ -335,28 +335,42 @@ $.entwine('ss', function($) {
...
     onclick: function (event) {
-      if ($(event.target).closest('.action-menu__toggle').length) {
-        this._super(event);
-        return false;
-      }
-
-      if($(event.target).closest('.action').length) {
-        this._super(event);
-        return false;
+      // Disallow clicking if there are interactive elements inside this cell as this would be non-accessible
```
GuySartorelli:
```suggestion
      // Disallow clicking if there are interactive elements inside this cell for accessibility (WCAG 2.2 A 4.1.2)
```

GuySartorelli:
For future historians, "the weirdness" is that entire cells become unclickable if there's something interactive inside them. For example, in the below example because of the link in the `THINGDISPLAY` column, the entire grey region (except for the link itself) isn't clickable, but the rest of the row is clickable and opens the edit form:

image

I'm on the fence about this.

On the one hand, this is the technically correct thing to do. I can imagine fat-fingering the link or the actions `...` button without this PR and opening the edit form accidentally. Someone with mobility issues but who uses the mouse anyway might have this struggle.

On the other hand, for sighted mouse users without mobility issues on desktop (which is probably the vast majority of content authors using Silverstripe CMS if I had to guess) this is a reduction in usability. It's also confusing, since the cell the link is in is so wide, and there's no obvious visual barrier distinguishing "this is not clickable" and "this is clickable".

## Options

1. Merge this change
    - have really clear changelog entry about it
    - maybe add a config option to revert to the old behaviour?
    - maybe ask UX team to help us find a way to communicate within the CMS "this is the clickable region"?
1. Leave as is, accepting that this is one place where general purpose usability and accessibility clash
    - Would need product owner approval I think
1. Make the whole row unclickable
    - make the edit/view button _always visible outside_ of the `...` menu
    - have really clear changelog entry about it
    - maybe add a config option to revert to the old behaviour?
    
Do you have a preference out of those?

emteknetnz:
Probably 2 at this stage.

The other two options involve putting an entry in the changelog just to tell developers they can revert something we think is bad so that content editors don't get annoyed does not sound like a great way forward

A more technically correct implementation of this may involve removing the 'click the `<tr>` functionality entirely and adding a dedicated "view/edit record" button or link which isn't buried in a menu, possibly in the first column, designed in a way where the loss of clicking the `<tr>` doesn't feel annoying, though I'm not sure that's possible

Maybe for now just don't merge this PR and instead add the `role="row"` attribute to the `<tr>` and hope that screen readers ignore whatever javascript does to the row?

GuySartorelli:
> Maybe for now just don't merge this PR and instead add the role="row" attribute to the <tr> and hope that screen readers ignore whatever javascript does to the row?

Yeah that sounds sensible.

Probably still needs the product owner to have a quick glance though and say "yes that's fine", since the mobility issues problem still exists (clicking just slightly off the link by accident opens the edit form) which is an accessibility problem (i.e. accept that the current status quo isn't perfect but fixing it is probably more complex than it'd be worth right now)

SimulatedPanda:
Thanks for the information. I have reviewed this and am comfortable with not merging this PR, favouring general usability. It may be worth making a note that this could be considered further in the future if this proves to be an ongoing challenge. While this is less accessibility-optimal, the approach agreed here is pragmatic and suitable for the current situation.

PR #2045, opened 2025-10-30
ENH Inherit colour for disabled secondary btn

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11818


client/src/styles/legacy/_style.scss
```diff
@@ -1029,6 +1029,10 @@ html, body {
   color: $white;
 }
 
+.ui-widget-content a.btn-secondary.disabled {
+  color: inherit;
+}
```
ChloeSartorelli, line 1034: Move this into `client/src/styles/legacy/_uitheme.scss` which is where the colour we're overriding is set.

GuySartorelli (changes requested):

client/src/styles/legacy/_style.scss
emteknetnz: That file has the following comment up the top telling me to put this it in _style.scss

```
 * This file defines CMS-specific customizations to the jQuery UI theme.
 * Every rule should contain ONLY overwritten jQuery UI rules (with 'ui-' prefix).
 *
 * This file should be fairly short, as we're using our own custom jQuery UI theme already.
 *
 * Use _style.scss to add more generic style information,
 * and read the jQuery UI theming API: http://jqueryui.com/docs/Theming/API
```

ChloeSartorelli: > Every rule should contain ONLY overwritten jQuery UI rules (with 'ui-' prefix).
 
 You're literally overwriting one of the rules?
 Doesn't matter really, not worth going back and forth about I'll just merge.

Issue #2043, opened 2025-10-30
Convert UsedOnTable to functional component


emteknetnz:
We have a bunch of react class components that will need to be updated to functional components at some point

Start with the UsedOnTable since nothing extends it, unlike many other class components

[**Kitchen sink CI**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/19056023940)

Doing kitchen-sink-ci because the [behat test](https://github.com/silverstripe/silverstripe-asset-admin/blob/3/tests/behat/features/file-details.feature#L18) for this component is in asset-admin (this component should really also be in asset-admin, though not moving now as it would make this harder to review)

The behat failure is [existing](https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/19025147838/job/54327347130#step:12:972)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2047

GuySartorelli:
PR merged

PR #2047, opened 2025-11-03
ENH Convert UsedOnTable to functional component

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2043

I first added some new unit tests to fill out coverage and ensure they passed on the class component. I then refactored the class component to a functional component

Behat test failure is [existing](https://github.com/silverstripe/silverstripe-admin/actions/runs/19017583561/job/54307725239#step:12:971)


client/src/components/UsedOnTable/UsedOnTable.js
```diff
@@ -97,17 +57,55 @@ class UsedOnTable extends PureComponent {
         </td>
       </tr>
     );
-  }
+  };
+
+  const renderBody = () => {
```
emteknetnz, line 62: I had to move the function below renderRow() otherwise the linter would complain that renderRow() was not defined

Note that I'm using `const myMethod() { ... }` style not `function myMethod() {}` style which would hoist, because despite my [earlier preference](https://github.com/silverstripe/silverstripe-admin/blob/3/client/src/components/SudoModePasswordField/SudoModePasswordField.js) for the the latter, the former is the standard in the react community, and has the advantage of not create its own `this` context, which is important in react as callback are often defined as `onClick={this.myMethod}`. What that means is is that when a regular function is passed as a callback, it loses its connection to the component instance.

We don't define a style for this in https://docs.silverstripe.org/en/6/contributing/coding_conventions/javascript_coding_conventions/

client/src/components/UsedOnTable/tests/UsedOnTable-test.js
```diff
@@ -22,22 +21,6 @@ function makeProps(obj = {}) {
   };
 }
 
-test('UsedOnTable provideUsedOnData HOC should try to load data onMount', () => {
```
emteknetnz, line 25: I deleted this unit test because I also refactored the provideUsedOnData HOC, however this test used `.Component` which is not available in functional components

There's a behat test for used on table which does a much better job of validating this all work - https://github.com/silverstripe/silverstripe-asset-admin/blob/3/tests/behat/features/file-details.feature#L18

client/src/components/UsedOnTable/provideUsedOnData.js
```diff
@@ -1,51 +1,47 @@
...
       // - component is on the currently active tab
-      const doFetch =
-        oldProps.identifier !== this.props.identifier ||
-        !tabContext ||
-        tabContext.isOnActiveTab;
-
-      if (doFetch) {
-        this.fetchDataFromEndpoint();
+      if (!tabContext || (tabContext && tabContext.isOnActiveTab)) {
+        fetchDataFromEndpoint();
```
ChloeSartorelli: The `componentDidMount()` logic used to be based partially on the value of `haveFetchedData` being `undefined`. We explicitly only called `fetchDataFromEndpoint()` if `forceFetch` was true or if this was the "first" mount.

This will now fetch data if `tabContext` is falsy, or if `tabContext` and `tabContext.isOnActiveTab` are truthy - which seems like a different condition to me and doesn't even include `forceFetch` in the condition. 

Should probably be like this?

```suggestion
      if ((haveFetchedData !== undefined && tabContext?.isOnActiveTab) || forceFetch) {
        fetchDataFromEndpoint();
```

And then for the `useState` have

```diff
-    const [haveFetchedData, setHaveFetchedData] = useState(false);
+    const [haveFetchedData, setHaveFetchedData] = useState(undefined);
```

client/src/components/UsedOnTable/provideUsedOnData.js
```diff
@@ -1,51 +1,47 @@
...
       }
-    }
+    }, [identifier, tabContext, forceFetch, data, loadUsedOnAction]);
```
ChloeSartorelli: Why so many things being passed in here? Most of these don't seem relevant to what's happening in this hook.

GuySartorelli (changes requested):

client/src/components/UsedOnTable/provideUsedOnData.js
ChloeSartorelli: Looks like the usual replacement for `componentDidMount` is to have an empty array (i.e. the hook only ever fires _once_)

client/src/components/UsedOnTable/provideUsedOnData.js
emteknetnz: I don't understand why we would use `undefined` for something that's a bool?

The old code `const firstMount = this.haveFetchedData === undefined;` just seems like it was an anti-pattern. I ended up kind of manually refactoring a fair bit of this HOC cos it was kind of weirdly put together

I've manually tested that it works correctly:
- in its usual context where it's the 3rd tab in the file detail view
- it only makes the AJAX request when you click on the tab
- it will correctly only make a single ajax request when clicking the 2nd tab and then the 3rd tab again
- in a non-tab context by commenting out [these 2 lines](https://github.com/silverstripe/silverstripe-asset-admin/blob/3/code/Forms/FileFormFactory.php#L60-L61)

client/src/components/UsedOnTable/provideUsedOnData.js
emteknetnz: This is a replacement for componentDidUpdate()

The componentDidMount() method was redundant so has been removed (sorry should have left comment)

client/src/components/UsedOnTable/provideUsedOnData.js
ChloeSartorelli: > I don't understand why we would use undefined for something that's a bool?

Doesn't matter since you've pointed out before that this is replacing `componentDidUpdate` not `componentDidMount`

ChloeSartorelli: > I ended up kind of manually refactoring a fair bit of this HOC cos it was kind of weirdly put together

I was expecting a direct like-for-like - when you refactor things in these class-to-functional-component PRs _please_ point out every change that isn't a straight like-for-like move before marking the PR ready for review.

client/src/components/UsedOnTable/provideUsedOnData.js
```diff
@@ -1,51 +1,47 @@
...
       // - component is not in a tab
       // - component is on the currently active tab
-      const doFetch =
-        oldProps.identifier !== this.props.identifier ||
-        !tabContext ||
-        tabContext.isOnActiveTab;
-
-      if (doFetch) {
-        this.fetchDataFromEndpoint();
+      if (!tabContext || (tabContext && tabContext.isOnActiveTab)) {
```
ChloeSartorelli:
```suggestion
      if (!tabContext || tabContext.isOnActiveTab) {
```
`tabContext` must be truthy for `!tabContext` to evaluate false, so we don't need to check it a second time - won't hold up merging just for this though.

ChloeSartorelli: If `tabContext.isOnActiveTab` is false, but `identifier` has changed, previously this would refetch - but it looks like that won't happen now, because of this condition? i.e. the hook will fire, but then this condition will resolve to false and it won't refetch?

client/src/components/UsedOnTable/provideUsedOnData.js
ChloeSartorelli: Ahh yup I didn't notice the `componentDidUpdate()` function there.

GuySartorelli (changes requested):

client/src/components/UsedOnTable/provideUsedOnData.js
emteknetnz: Opps. Yeah this functional was a bit non-sensical, have re-written

GuySartorelli (approved):
LGTM

Issue #1969, opened 2025-08-11
Accessibility: Screen readers don't have context about pagination


GuySartorelli:
WCAG 2.1 1.3.1 4.1.2

Pagination "doesn't read as pagination" according to an accessibility review. There's no context telling a screen reader that the pagination is, in fact, pagination.
https://design-system.w3.org/components/pagination.html gives a good example of accessible pagination - there are boundless other examples online as well.

Main takeaways I can see:
- use the `nav` element as a parent of both a heading (if appropriate) and the pagination list (alternatively use `role="navigation"` which won't affect selectors)
- add either a heading inside the `nav` or a `aria-label` attribute to the `nav` element that has a visually-hidden text "pagination"
- "The current page is indicated by `aria-current="page"`"
- "Where there are multiple `<nav>` elements on a single page, they should all have a unique accessible name." I presume this also applies to elements with `role="navigation"`

## Notes

- The pagination in gridfields and ~campaign admin~ (Steve: I don't think campaign-admin paginates see my comment below) were explicitly flagged, but check _all_ pagination, e.g. in asset admin

## Acceptable criteria

- [ ] Pagination in the CMS is appropriately accessible to screen readers
- [ ] The changelog clearly calls out any changes to markup selectors so people can adjust their custom css/js as necessary

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/19217051568

## PRs
- https://github.com/silverstripe/silverstripe-framework/pull/11903
- https://github.com/silverstripe/silverstripe-admin/pull/2053
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/434
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1634
- https://github.com/silverstripe/developer-docs/pull/842

emteknetnz:
I don't think that campaign admin has pagination.

Here's a test script to create 200 campaigns, they all just show without paginating

```php
// create 200 test campaigns to show they do not paginate
use SilverStripe\Versioned\ChangeSet;
$existingCount = ChangeSet::get()->count();
if ($existingCount < 200) {
    for ($i = $existingCount + 1; $i <= 200; $i++) {
        $campaign = ChangeSet::create([
            'Name' => 'Test Campaign ' . $i,
            'Description' => 'Test campaign generated for pagination testing #' . $i,
[...]

PR #2053, opened 2025-11-05
ENH Improve screen reader support of pagination

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1969

This react paginator is used by asset-admin and versioned-admin


client/src/components/Paginator/Paginator.js
```diff
@@ -90,13 +100,13 @@ function Paginator(props) {
   }
 
   // Render the paginator
-  return <div className="paginator-footer">
+  return <nav aria-label={i18n._t('Admin.PAGINATION', 'Pagination')} className="paginator-footer">
```
ChloeSartorelli: We should probably allow for multiple of this component to be on the same page as it's likely to be used more if we want to react-ify all the things. In that case this label is too generic, as multiple pagination on the same page will end up having the same label.

See https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/navigation.html
> If a page includes more than one navigation landmark, each should have a unique label.

See https://design-system.w3.org/components/pagination.html

> Where there are multiple `<nav>` elements on a single page, they should all have a unique accessible name.

GuySartorelli (dismissed):
Screen reader experience seems good to me, just one gripe below.

client/src/components/Paginator/Paginator.js
emteknetnz: Same issue as 

[Paginator.js](https://github.com/silverstripe/silverstripe-admin/blob/3/client/src/components/Paginator/Paginator.js#L102) does not currently have props for a paginator title.

This would be much easier to retro fit than the GridField equivalent, because there we're auto-scaffolding gridfields based on relations where the name of the relation may be nonsensical to a screenreader.  

However as far as I know we only use this in file manger and versioned admin, so both of those scenarios this will be a unqiue paginator so there also isn't really a need for a unique name

ChloeSartorelli: It would be an easy prop to add.

> However as far as I know we only use this in file manger and versioned admin,

Again: 

> We should probably allow for multiple of this component to be on the same page as it's likely to be used more if we want to react-ify all the things.

emteknetnz: Added

client/lang/src/en.json
```diff
@@ -51,6 +52,7 @@
     "Admin.OWNED_WARNING_1": "You are unpublishing content that is being used in {count} other published section(s).",
     "Admin.OWNED_WARNING_2": "This could cause a published page to have missing components on the live site.",
     "Admin.OWNED_WARNING_3": "Do you want to unpublish anyway?",
+    "Admin.PAGINATION": "Pagination",
```
ChloeSartorelli, line 55:
```suggestion
    "Admin.PAGINATION": "Pagination",
    "Admin.PAGINATION_FOR_TITLE": "Pagination for {title}",
```

GuySartorelli (dismissed):
One thing to change

client/lang/src/en.json
emteknetnz: Updated

GuySartorelli:
Merge conflict

emteknetnz:
Rebased

GuySartorelli (approved):
LGTM

PR #2054, opened 2025-11-06
ENH Allow adding filter prefix to filter props

GuySartorelli:
In CMSMain and gridfield search filters, the `filters` prop is passed in from HTML data attributes, which already include the prefix.

In AssetAdmin, the filters NEVER include the prefix but don't need to, because they don't use SearchContext.

For CampaignAdmin, since we're not re-rendering the searchbox serverside every time we search (or ever) we can't rely on using some pre-set filters that include the prefixes - but since we're using SearchContext we DO need the prefix passed into the prop (even though we don't want it in the search POST request).

This is a little messy - but avoids reworking how _everything_ works. I've opened https://github.com/silverstripe/silverstripe-admin/issues/2055 to tidy it up in a future major release when BC is less of a concern.

## Issue

- https://github.com/silverstripe/silverstripe-campaign-admin/issues/365


emteknetnz (approved):

GuySartorelli:
Rebased due to merge conflict

emteknetnz (changes requested):
New merge conflict

GuySartorelli:
Rebased again

emteknetnz (approved):

PR #2056, opened 2025-11-11
NEW Add DependentCompositeField for complex dropdown updates.

GuySartorelli:
We have a need for a way to update one formfield based on the value of another - not just updating value or source but outright replacing one form field with a different type of form field.

Keeping this in admin instead of framework because:
1. It lives and dies on the JavaScript which we'll be keeping in admin due to conventions
2. We don't have a need for this directly in framework currently

Note that for now I'm _intentionally_ not mentioning this new field in the changelog, and have marked it `@internal`. This is because it doesn't work in non-react forms, which is the majority of forms in the CMS. A separate card has been raised to implement that functionality, but since it isn't required for moving elemental blocks and is definitely non-trivial, I've opted to deal with it separately.

I have split off https://github.com/silverstripe/silverstripe-admin/issues/2071 to deal with updating the new field to work in entwine forms.

## Issue

- https://github.com/silverstripe/silverstripe-elemental/issues/445


client/src/components/DependentCompositeField/DependentCompositeField.js
```diff
@@ -0,0 +1,138 @@
...
+            }
+            props.meta.dispatch(change(formid, fieldData.name, fieldData.value));
+          }
+        }
+      })
+      .catch((reason) => {
+        // @TODO check what this looks like and probably give better user feedback
+        throw reason;
+      });
+  }, 300), []);
```
ChloeSartorelli, line 119: 300ms feels about right when typing in a text field - but it feels sluggish when selecting items from a dropdown field.

If you have any suggestions about how to deal with that discrepancy better please let me know.

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -5,18 +5,23 @@ import schemaFieldValues, { schemaMerge } from 'lib/schemaFieldValues';
...
+      submittingAction: null,
+      // formFieldSchemaObj: null,
+    };
     this.submitApi = backend.createEndpointFetcher({
       url: schemaStructure.attributes.action,
       method: schemaStructure.attributes.method,
     });
     this.mapActionsToComponents = this.mapActionsToComponents.bind(this);
     this.mapFieldsToComponents = this.mapFieldsToComponents.bind(this);
+    this.normalizeFields = this.normalizeFields.bind(this);
```
ChloeSartorelli, line 20: Needed so it is bound for use in DependentCompositeField.js

client/src/components/DependentCompositeField/DependentCompositeField.js
```diff
@@ -0,0 +1,197 @@
...
+          newChild.message = initialChild.message;
+          hasChange = true;
+        }
+        newSchema.push(newChild);
+      }
+      // Return the original state if no changes were made.
+      // React won't trigger a rerender if it detects the state didn't change.
+      return hasChange ? newSchema : oldSchema;
+    });
+  }, [...initialChildrenSchema?.map(childField => [childField.name, childField.message?.rawValue]) ?? []]);
```
ChloeSartorelli: This is a kinda ugly dependency array but it's what we need to only update the field when message text updates.

The `...` operator seems to be needed or react fires the effect more often than it needs to.

code/Forms/DependentCompositeField.php
```diff
@@ -0,0 +1,560 @@
...
+     * Basically this is any composite-like field which handles its own children,
+     * since the onchange handler would be on the child field but we only know about
+     * the parent field name.
+     */
+    private static array $disallowed_field_types = [
+        CompositeField::class,
+        ConfirmedPasswordField::class,
+        DependentCompositeField::class,
+        MoneyField::class,
+    ];
```
ChloeSartorelli, line 43: tl;dr it would just be too hard to support these, and the benefit would be small.

code/Forms/DependentCompositeField.php
```diff
@@ -0,0 +1,560 @@
...
+    /**
+     * @var array<FormField>
+     */
+    private array $children = [];
+
+    /**
+     * Per-request caching of the results of getChildren() for this instance of the form field.
+     * Prevents dependency callbacks from being called unnecessarily.
+     */
+    private bool $childrenAreCached = false;
```
ChloeSartorelli, line 66: The actual cached values are in the `$children` array since we don't need to store a copy of the original fields separately.

client/src/components/DependentCompositeField/DependentCompositeField.js
emteknetnz: Make this thing a const above the useEffect hook, give it an inline comments explaining what's going on and make that the dependency

client/src/components/DependentCompositeField/DependentCompositeField.js
```diff
@@ -0,0 +1,197 @@
...
+   */
+  const handleChange = useCallback(debounce((value, state, schema, allChildrenNames) => {
+    setLoading(true);
+    // Call existing callback if there was one
+    if (typeof schema.onChange === 'function') {
+      schema.onChange(value);
+    }
+    // Get values from redux store but only for the relevant fields
+    const allValues = getFormValues(formid, getFormState)(state);
+    const onlyNecessaryValues = Object.fromEntries(Object.entries(allValues).filter(
```
emteknetnz, line 80: Need a comment explaining what this doing

client/src/components/DependentCompositeField/DependentCompositeField.js
```diff
@@ -0,0 +1,197 @@
...
+        setLoading(false);
+        throw reason;
+      });
+  }, 300), []);
+
+  /**
+   * Builds the legend for a fieldset if it is defined
+   * See CompositeField.getLegend()
+   *
+   * @returns {Component}
```
emteknetnz: or null?

emteknetnz:
I have the following notes from manual testing:

1. Move modal: Parent type - having a blank option makes no sense

2.  Move modal:
```
  When I select `Parent type` = (blank)
  And I select `Parent type` = "Elemental Behat Test Object"
  Then I see "Parent ID"
  When I select `Parent type` = "Multi Elemental Behat Test Object"
  Then I see "Parent"
  When I select `Parent type` = "Elemental Behat Test Object"
  Then I see "Parent" <<< But I expected to see "Parent ID"
```

3. Move modal:
```
   When I select Parent type = "Page"
   And I click on the Parent treedropdown
   Then I see a bunch of disabled pages << I'm not sure if  seeing all the disabled pages is a good thing or not.
```
   
4. Just a nice to have (could be a new card), have a link to the edit form of where the block was moved to as a hyperlink in the toast you see after moving the block

5. This is probably the correct behaviour, though if you have a block with a has_one File which it owns, then moving the block will unpublish the File. Could be an issue if the File is used in mutiple places, not just the block.

6. The following file causes AJAX issues when you create one an view it in the CMS, it doesn't cause issues on a regular sink install without these PRs

'Sorry, it seems you were trying to access a section or object that doesn't exist.`

Network panel shows its trying to access `/admin/elemental-area/api/readElements/0`

```php
<?php

namespace App\PageTypes;

use DNADesign\Elemental\Models\ElementalArea;
use DNADesign\Elemental\Extensions\ElementalAreasExtension;

class MutliElementalAreaPage extends SiteTree
{
    private static string $singular_name = 'Multi Elemental Area Page';

    private static array $extensions = [
        ElementalAreasExtension::class,
    ];

        'ElementalArea1' => ElementalArea::class,
        'ElementalArea2' => ElementalArea::class,
    ];

    private static array $owns = [
        'ElementalArea1',
        'ElementalArea2',
    ];

    private static array $cascade_duplicates = [
        'ElementalArea1',
        'ElementalArea2',
    ];
}
```

GuySartorelli:
> 1 Move modal: Parent type - having a blank option makes no sense

I can't find a way to remove it - if you know how I'll be happy to remove it.
There's no empty value in the sources array, and `hasEmptyDefault` is left at its default `false`.

> 2 Move modal:
> ...
> But I expected to see "Parent ID"

It should always be "Parent" - it saying "Parent ID" earlier is what's wrong. Fixed now.

> 3 Move modal:
> ...
> Then I see a bunch of disabled pages << I'm not sure if  seeing all the disabled pages is a good thing or not.

That's just how `TreeDropdownField` works. I can hide them instead, but then the children are also hidden whether they should be selectable or not.

> 4 Just a nice to have (could be a new card), have a link to the edit form of where the block was moved to as a hyperlink in the toast you see after moving the block

Have spun off as a [new card](https://github.com/silverstripe/silverstripe-elemental/issues/1392), this is already pretty complex :sweat_smile: 

> 5 This is probably the correct behaviour, though if you have a block with a has_one File which it owns, then moving the block will unpublish the File. Could be an issue if the File is used in mutiple places, not just the block.

This will depend on the way the relationship is set up. The following will publish the file when the block is published, but not unpublish it when the block is unpublished:

```php
private static $has_one = [
    'MyFile' => File::class,
];

private static $owns = [
    'MyFile',
];
```

> 6 The following file causes AJAX issues when you create one an view it in the CMS, it doesn't cause issues on a regular sink install without these PRs

I can't reproduce the error you saw. I created a page of that type, refreshed, tried clicking into it from `/admin/pages` in case it has to be loaded that way.... if you can reproduce again from a fresh installation with the PRs please let me know. Maybe it's related to data that you had from already playing around with the move feature.

GuySartorelli:
Rebased as there was a merge conflict in `bundle.css`

client/src/components/DependentCompositeField/DependentCompositeField.js
ChloeSartorelli: There is one:
> // Get values from redux store but only for the relevant fields

ChloeSartorelli: Have updated the comment to give a bit more detail but unless you tell me more specifically what's not clear about it I won't know what needs to be done here.

client/src/components/DependentCompositeField/DependentCompositeField.js
ChloeSartorelli: Yeah I guess, I just copied this from `CompositeField.js`. Have done what the internet says is correct for adding null, though vscode doesn't like it lol

emteknetnz (changes requested):
Merge conflict

GuySartorelli:
Rebased

Issue #1973, opened 2025-08-11
Implement "last saved `x` minutes ago" functionality


GuySartorelli:
In https://github.com/silverstripe/silverstripe-framework/issues/7877#issuecomment-535683397 it was proposed that edit forms get a clear indicator of when content was last saved.

This is already _sort-of_ implemented - the `...` menu shows you the _actual_ last time the record was saved, though this isn't very noticable, most people probably don't even know it's there - and isn't indicating how long since you saved _since you started editing the page_ which I think is the key thing.

## Notes

- There is a mockup in the above linked comment but get design to give it a once-over
- Don't forget elemental blocks. E.g. when you're editing an inline-editable block, probably we still put the indicator _on the page_. After saving the block, if there are no other unsaved changes, the indicator should go away. Depending on what we decide for "When should the indicated time be from?" below, the timer should start again if we start making new edits after saving.

## ~~To decide~~ Decided

We opted to go with "Guy's preferred option" for both of these.

- When should this be displayed?
    - before you have started making edits (i.e. as soon as the form is loaded)
    - only when the form is dirty
    - only after some configurable threshold, e.g. after 2 minutes since you started making edits (**this is Guy's preferred option**)
- When should the indicated time be from?
    - the last time it was _actually saved_ (could be "last saved 5 days ago" which may not be helpful)
    - the time since you opened the form
    - the time since the form went dirty (**this is Guy's preferred option**)

## Next steps
- We'd want this indicator on the DataObject edit form toolbar (entwine) and asset admin file edit form (react). That should cover all the relevant edit forms.
- Assuming with go with Guys preferred options, we need some UX guidance around the following:
  - What language should be used, so that it's not confused with the timestamp of when the DataObject was actually persisted (rather than time since the form went dirty)
  - Design so that it does not imply danger (i.e. not bright orange with a warning symbol)
- Once we have that we can put on some acceptance criteria

## Acceptance criteria

- [ ] There are two info states based on the number of minutes the form has been in a dirty state for
- [ ] The design mock up are used as is
- [ ] The "Last saved" mock up is NOT used i.e. keep it behind the "..." as it is now
- [ ] Regular edit form and react edit form i.e. asset-admin forms both have the same treatment
- [ ] Implement the aria role "live update?" that reads out when the state changes
- [ ] Can configure the the number of minutes for each state, or set to null to disble
- [ ] Developer docs make it clear how to do this

## Kitchen sink CI
[...]

PR #2058, opened 2025-11-12
NEW Create UnsavedChangesIndicator

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1973

The [design mock](https://github.com/silverstripe/silverstripe-admin/issues/1973#issuecomment-3478630625) had the indicator wrap underneath the save and publish buttons at small breakpoints, which would been hard to implement, and would have also vertically expanded the bottom toolbar which is something that doesn't happen right now, and would honestly look really bad.

Instead of doing that, I've implemented it so that the indicator gets horizontally smaller by having 3 different states depending on the current breakpoint, which means it won't wrap. There's a title attribute with the full message that can be seen on mouse over.

(Have slacked James seeking design feedback)

### Standard size
image

image

### Compact size
image

### Icon only showing mouse over state
image


client/src/boot/applyTransforms.js
```diff
@@ -55,6 +55,7 @@ const applyTransforms = () => {
 
           return togglePristineState(field, isPristine);
         });
+        form.mutateField('action_UnsavedChangesIndicator', (field) => ({ ...field, isDirty: form.isDirty() }));
```
emteknetnz: This sets isDirty prop on the component in a react context based of the form dirty state

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import i18n from 'i18n';
+import classnames from 'classnames';
+
+const UnsavedChangesIndicator = ({
```
emteknetnz, line 8: This component only deals with presentation, the timer logic is done separately in the UnsavedChangedIndicatorTimer component

emteknetnz:
Updated to use bootstrap variables for colors:

image

image

There's also a red version, though it's a bit much for a 10 minute warning IMO

image

jamessilverstripe:
Looks good to me!

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
```diff
@@ -0,0 +1,65 @@
...
+  level = 'none',
+  minutes = {
+    notice: 5,
+    warning: 10,
```
ChloeSartorelli: Why are we setting defaults here, when the defaults are already set in PHP?

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
```diff
@@ -0,0 +1,70 @@
...
+  margin-right: 8px;
+
+  // These use bootstrap colors
+  &--notice {
+    // --bs-info-text-emphasis
+    color: rgb(16.4, 68.4, 90.4);
+    // --bs-info-border-subtle
+    border-color: rgb(169.4, 221.4, 243.4);
+    // --bs-info-bg-subtle
+    background-color: rgb(212.2, 238.2, 249.2);
```
ChloeSartorelli: If these are already set in CSS variables why are we setting explicit values? Just use `var(--bs-info-text-emphasis)` for example. Or the bootstrap SCSS variables which I'm certain also exist.

For that matter why not use pre-existing CSS classes that already have these styles applied? e.g. `alert alert-info`

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicatorTimer.js
```diff
@@ -0,0 +1,75 @@
...
+/**
+ * A container component for UnsavedChangesIndicator that tracks the time since
+ * it was marked dirty, and updates the indicator level accordingly.
+ * By default it has a 5 minute notice level and 10 minute warning level.
+ */
+const UnsavedChangesIndicatorTimer = ({
+  isDirty = false,
+  minutes = {
+    notice: 5,
+    warning: 10,
```
ChloeSartorelli: We should only have the default values in one place - this is the _third_ place I've seen them. They should just be marked as required in proptypes instead.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
```diff
@@ -0,0 +1,65 @@
...
+  },
+}) => {
+  if (level === 'none') {
+    return null;
+  }
```
ChloeSartorelli: We should probably either:
1. throw an error if `level` is not one of the expected values, or
2. return `null` if `level` is not one of "warning" or "notice"

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
```diff
@@ -0,0 +1,65 @@
...
+    shortMessage = i18n.inject(
+      i18n._t('Admin.UNSAVED_CHANGES_PLURA_SHORT', '{mins} mins'),
+      { mins }
+    );
+  }
+  const icon = isWarning ? 'font-icon-attention' : 'font-icon-info-circled';
+  const className = classnames('unsaved-changes-indicator', {
+    'unsaved-changes-indicator--notice': level === 'notice',
+    'unsaved-changes-indicator--warning': isWarning,
+  });
```
ChloeSartorelli:
```suggestion
  const className = classnames('unsaved-changes-indicator', `unsaved-changes-indicator--${level}');
```
`level` should always be one of "notice" or "warning" so we can just use string interpolation here. Probably don't even need to use `classnames()` tbh

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
```diff
@@ -0,0 +1,65 @@
...
+  if (mins <= 1) {
+    message = i18n.inject(
+      i18n._t('Admin.UNSAVED_CHANGES_SINGULAR_FULL', 'Unsaved changes: {mins} min'),
+      { mins }
+    );
+    shortMessage = i18n.inject(
+      i18n._t('Admin.UNSAVED_CHANGES_SINGULAR_SHORT', '{mins} min'),
+      { mins }
+    );
+  } else {
```
ChloeSartorelli: We we expect people to be using such low values, we should be using seconds.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
```diff
@@ -0,0 +1,65 @@
...
+  minutes = {
+    notice: 5,
+    warning: 10,
+  },
+}) => {
+  if (level === 'none') {
+    return null;
+  }
+  const isWarning = level === 'warning';
+  const mins = isWarning ? minutes.warning : minutes.notice;
```
ChloeSartorelli: I was surprised to see the value in the alert stayed stagnant. I was expecting it to count up as time passed. I might ignore the warning when it pops up at 20 minutes, but if I look down and see it has counted up to 5 hours I might go "oh shit I need to save".

People might also think it's a bug if it stays at the value.

ChloeSartorelli: I think we can use the `aria-live="off"` outside of the changes between `level` to prevent bugging screen reader users every minute. Then set it back to `polite`  when the `level` changes (and arguably also every x period where x is the time difference between warning and notice).

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
```diff
@@ -0,0 +1,70 @@
+.unsaved-changes-indicator__container {
+  display: flex;
+}
```
ChloeSartorelli, line 9: Is this only used in the entwine? If so, why?

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
```diff
@@ -0,0 +1,70 @@
...
+  padding: 7px 8px 0;
+  border-radius: 3px;
+  border: 1px solid;
+  font-size: 13px;
+  white-space: nowrap;
+  margin-left: 4px;
+  margin-right: 8px;
+
+  // These use bootstrap colors
+  &--notice {
```
ChloeSartorelli:
```suggestion
}

.unsaved-changes-indicator--notice {
```
Using the full CSS class means it's easier to search for. If I search for `unsaved-changes-indicator--notice` I want to find the scss declaration for it.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
```diff
@@ -0,0 +1,70 @@
...
+    // --bs-warning-border-subtle
+    border-color: rgb(255, 230.2, 155.8);
+    // --bs-warning-bg-subtle
+    background-color: rgb(255, 242.6, 205.4);
+  }
+
+  .font-icon {
+    display: inline-block;
+    font-size: 14px;
+  }
```
ChloeSartorelli:
```suggestion
  .unsaved-changes-indicator__icon {
    display: inline-block;
    font-size: 14px;
  }
```
Should ideally use BEM instead of nesting.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
```diff
@@ -0,0 +1,70 @@
...
+  // full message
+  @include media-breakpoint-up(xl) {
+    .unsaved-changes-indicator__short {
+      display: none;
+    }
+
+    .unsaved-changes-indicator__full {
+      display: inline;
+    }
+  }
```
ChloeSartorelli: These should not be nested (that's one of the main purposes of BEM)

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
ChloeSartorelli: Why have them separated? Is there a way to use one without the other in a useful way?

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicatorTimer.js
```diff
@@ -0,0 +1,75 @@
...
+   * Check elapsed time and update level accordingly
+   */
+  useEffect(() => {
+    if (!isDirty || startTime === null) {
+      return () => {};
+    }
+    const checkElapsedTime = () => {
+      const elapsed = Date.now() - startTime;
+      const warningMs = minutes.warning ? minutes.warning * 60000 : null;
+      const noticeMs = minutes.notice ? minutes.notice * 60000 : null;
```
ChloeSartorelli: If the `minutes` values are both falsy we should just set level to "none" and not set an interval.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicatorTimer.js
```diff
@@ -0,0 +1,75 @@
...
+      if (warningMs && elapsed >= warningMs) {
+        setLevel('warning');
+      } else if (noticeMs && elapsed >= noticeMs) {
+        setLevel('notice');
+      } else {
+        setLevel('none');
+      }
+    };
+    const interval = setInterval(checkElapsedTime, 1000);
```
ChloeSartorelli: I'm a bit confused about this interval. Aren't we checking more often than we need to?
We know what level we're on, and we know when we starting timing, and we know how long we need to wait between each level. Why don't we just set a timeout for that specific amount of time?

client/src/legacy/UnsavedChangesIndicator/UnsavedChangesIndicatorEntwine.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+      const self = this;
+      const observer = new MutationObserver(() => {
+        const isChanged = self.hasClass('changed');
+        container.updateDirtyState(isChanged);
+      });
+      observer.observe(this[0], {
+        attributes: true,
+        attributeFilter: ['class'],
+      });
```
ChloeSartorelli, line 74: Why use a mutation observer? You can just use `form.changed` onmatch and onunmatch entwine handlers (entwine is already using a mutation observer under the hood for onmatch)

tests/behat/features/unsaved-changes-indicator.feature
```diff
@@ -0,0 +1,45 @@
+@javascript @retry @job2
+Feature: Unsaved changes indicator
+  As an author
+  I want to be alerted of unsaved changes
+
+  Background:
+    # Config will put ~2 second delay until the unsaved changes notice shows
+    # Note - intentionally not testing the warning time because there's likely to
+    # be flakiness with the behat test due to timing issue e.g. CI infra being slow
```
ChloeSartorelli: Why don't we just wait a couple of seconds longer than we expect to need to?

The likelyhood of flakiness with notice->warning is presumably the same likelyhood of flakiness for dirty->notice so I don't see why we would test one and not the other?

GuySartorelli (changes requested):

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
emteknetnz: I just found it cleaner to separate the timer and presentation out. Makes unit testing a lot cleaner and well as storybook which works better without a timer.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
emteknetnz: Removed, have made propType required

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
emteknetnz: Gone with 2

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
emteknetnz: Updated

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
emteknetnz: Without this the vertical alignment on the indicator will be wrong

emteknetnz:
@GuySartorelli I ended up almost rewriting this timer component since your last review because I wasn't happy with it. I also changed the presentation component to simply accept an elapsedMinutes prop instead of the minutes config

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
emteknetnz: Updated

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
emteknetnz: Updated

tests/behat/features/unsaved-changes-indicator.feature
emteknetnz: We already have comprehensive testing using react-testing-library for the time based state updates so shouldn't duplicate in behat which would only make things run slower. I've updated the comment to note this

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
emteknetnz: Updated

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicatorTimer.js
emteknetnz: Have removed

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
emteknetnz: We don't expect them to, the config does not encourage it (it's in minutes not seconds after all).

Also I've updated the logic to increment displayed time by a minute e.g. it'll tell you it's been 22, 23, 24 minutes since saved. If we should seconds then it would be logical to increment the time since saved every second, which is not intention at all.

The only reason this support for non int values is here is for is development and testing (behat), where it is extremely useful

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
emteknetnz: Have updated things so that they're not stagnant anymore

aria-live will update as suggested

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicatorTimer.js
emteknetnz: Updated

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicatorTimer.js
emteknetnz: This was honestly complex enough just using a interval. It could probably done with timeouts, but it would get confusing since you need to set two of them and like the intervals you'd need to keep cancelling them on onmount. It's just more to keep track of.

There's probably only a very minuscule performance gain from not using an interval, so keep as is

client/src/legacy/UnsavedChangesIndicator/UnsavedChangesIndicatorEntwine.js
emteknetnz: Tried that, doesn't work for some reason

emteknetnz:
@GuySartorelli To clarify the component rewrite - previously the `UnsavedChangesIndicator.schemaComponent` was `UnsavedChangesIndicatorTimer` - I've changed that to `UnsavedChangesIndicator`

In `UnsavedChangesIndicator.js` the default export is now wrapped in a HOC `withUnsavedChangesIndicator()` which will wrap the `UnsavedChangesIndicator` component in a `UnsavedChangesIndicatorTimer` component

code/Forms/GridFieldDetailFormPreviewExtension.php
```diff
@@ -31,6 +33,24 @@ protected function updateItemEditForm(Form $form): void
         }
     }
 
+    protected function updateFormActions(FieldList $actions): void
```
ChloeSartorelli: This is not appropriate to put on `GridFieldDetailFormPreviewExtension`. Someone might remove this extension because they don't want the preview stuff happening (or extend it or whatever) - but that functionality has nothing to do with the unsaved changes indicator. This should be in a separate extension.

ChloeSartorelli: If these should be combined in a major release when we can rename the extension to something that isn't explicitly about preview, we can open an issue to do that.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
```diff
@@ -0,0 +1,77 @@
...
+  if (mins <= 1) {
+    message = i18n.inject(
+      i18n._t('Admin.UNSAVED_CHANGES_SINGULAR_FULL', 'Unsaved changes: {mins} min'),
+      { mins }
+    );
+    shortMessage = i18n.inject(
+      i18n._t('Admin.UNSAVED_CHANGES_SINGULAR_SHORT', '{mins} min'),
+      { mins }
+    );
+  } else {
```
ChloeSartorelli: As per https://github.com/silverstripe/silverstripe-admin/pull/2058#discussion_r2540885374 we're not expecting such low values - I think given this is an unlikely configuration, it's fine to just ignore this scenario. "One minutes" isn't the worst thing in the world to see in the unlikely scenario it does happen.

Even if we did keep this, at least in english "zero minutes" is correct, and "zero minute" is not, and since our JS i18n has no pluralisation support trying to handle this for other languages is a non-starter.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
```diff
@@ -0,0 +1,77 @@
...
+      prevLevelRef.current = level;
+      return;
+    }
+    if (level !== prevLevelRef.current) {
+      setAriaLive('polite');
+      prevLevelRef.current = level;
+    } else {
+      setAriaLive('off');
+    }
+  }, [level, elapsedMinutes]);
```
ChloeSartorelli, line 26: Why do we need `elapsedMinutes` here as a dependency? It doesn't look like it's used in the effect. If we remove it, we can drop the weird `prevLevelRef` as well.

ChloeSartorelli: Ahhh I think I understand it better now. When the level changes, we need `setAriaLive('polite');` - but when the timer updates and we _aren't_ changing levels, we want `setAriaLive('off');`.

This should probably be a memo to avoid updating state every render.
```js
const ariaLive = useMemo(() => {
  // the same logic as the effect, but return the value instead of setting state
}, [level, elapsedMinutes]);
```

This means we can use the value immediately, rather than rendering with the wrong value this cycle and only using the correct value after the state gets actually updated next cycle.

See https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state and https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

Please also add some comment here explaining what this is doing, something like
> We want `aria-live` set to "polite" when swapping between levels, but any time the `elapsedTime` updates without the level changing, we want `aria-live` set to "none" to avoid annoying screen reader users.

ChloeSartorelli: I thought about the possibility of having two effects instead, like this:

```js
// Set aria-live to "off" when time updates
useEffect(() => {
  if (ariaLive !== 'off') {
    setAriaLive('off');
  }
}, [elapsedMinutes]);

// Set aria-live to "polite" whenever the level changes - happens after the above, so will take precedence
useEffect(() => {
  setAriaLive('polite');
}, [level]);
```

In theory the second effect will take precedence if both props change at the same time, but when only the `elapsedMinutes` changes it'll set the value to "off" - though this means using state again, and also is apparently not recommended due to the way react batches state - which apparently would do what we expect right now but is an implementation detail rather than an explicit intended feature so could change unexpectedly.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
ChloeSartorelli: Why is it only used in entwine, though? Why isn't this as a parent element just part of the react component?

code/Forms/UnsavedChangesIndicator.php
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * A visual indicator to show whether a form has unsaved changes.
+ *
+ * Note that this isn't strictly a FormAction as it doesn't perform any action when clicked.
+ * However, extending FormAction allows us to easily integrate with the existing form action
+ * architecture (e.g. placement in the form actions area).
+ */
+class UnsavedChangesIndicator extends FormAction
+{
+    private static array $minutes = [
```
ChloeSartorelli, line 19:
```suggestion
    /**
     * An associative array of time periods to determine when the notice displays, and when it updates to a warning.
     * notice: The number of minutes it takes to display the notice after starting to edit a form.
     * warning: The number of minutes it takes to update the display to a warning after displaying the notice.
     */
    private static array $minutes = [
```
Sorry, should have asked for this the first time around.

code/Forms/UnsavedChangesIndicator.php
```diff
@@ -0,0 +1,66 @@
...
+        'notice' => 5,
+        'warning' => 10,
+    ];
+
+    protected $schemaComponent = 'UnsavedChangesIndicator';
+
+    public function getSchemaDataDefaults()
+    {
+        // This is used for initialising the React component
+        // Refer to silverstrpie/admin UnsavedChangesIndicator.js
```
ChloeSartorelli:
```suggestion
        // Refer to client/src/components/UnsavedChangesIndicator
```
This is a bit of a nitpick - I won't delay merging if this change isn't made

code/Forms/UnsavedChangesIndicator.php
```diff
@@ -0,0 +1,66 @@
...
+        $data = parent::getSchemaDataDefaults();
+        $data['minutes'] = $this->getMinutes();
+        return $data;
+    }
+
+    public function getAttributes()
+    {
+        // This is used for entwine initialisation
+        // Refer to silverstrpie/admin UnsavedChangesIndicatorEntwine.js
```
ChloeSartorelli:
```suggestion
        // Refer to client/src/legacy/UnsavedChangesIndicatorEntwine
```
This is a bit of a nitpick - I won't delay merging if this change isn't made

code/Forms/UnsavedChangesIndicator.php
```diff
@@ -0,0 +1,66 @@
...
+        $attributes = parent::getAttributes();
+        $attributes['data-minutes'] = json_encode($this->getMinutes());
+        return $attributes;
+    }
+
+    private function getMinutes(): array
+    {
+        $minutes = UnsavedChangesIndicator::config()->get('minutes');
```
ChloeSartorelli:
```suggestion
        $minutes = static::config()->get('minutes');
```
Allows for the edge case of subclasses that behave differently for whatever reason

code/Forms/UnsavedChangesIndicator.php
```diff
@@ -0,0 +1,66 @@
...
+        $this->validateConfig($minutes);
+        return $minutes;
+    }
+
+    private function validateConfig(array $minutes): void
+    {
+        if (!isset($minutes['notice']) || !isset($minutes['warning'])) {
```
ChloeSartorelli:
```suggestion
        if (!array_key_exists('notice', $minutes) || !array_key_exists('warning', $minutes)) {
```
This is a bit of a nitpick - but technically with a value of `null` for either of these keys we should hit the "Config values must be numeric" exception since the keys _are_ present, just not with valid values.

It's not that important though - I won't block merging if this isn't changed.

code/Forms/UnsavedChangesIndicator.php
```diff
@@ -0,0 +1,66 @@
...
+        // Allow decimals here for partial minutes, as this is required for behat
+        // testing and develoment so that we don't have to wait long periods of time
+        if (!is_numeric($minutes['notice']) || !is_numeric($minutes['warning'])) {
+            throw new RuntimeException('Config values must be numeric');
+        }
+        if ($minutes['notice'] < 0 || $minutes['warning'] < 0) {
+            throw new RuntimeException('Config values must be positive or zero');
+        }
+        if ($minutes['warning'] !== 0 && $minutes['warning'] < $minutes['notice']) {
+            throw new RuntimeException('Config "warning" value must be greater than or equal to "notice" value');
```
ChloeSartorelli:
```suggestion
            throw new RuntimeException('minutes config must contain both "notice" and "warning" keys');
        }
        // Allow decimals here for partial minutes, as this is required for behat
        // testing and develoment so that we don't have to wait long periods of time
        if (!is_numeric($minutes['notice']) || !is_numeric($minutes['warning'])) {
            throw new RuntimeException('minutes config values must be numeric');
        }
        if ($minutes['notice'] < 0 || $minutes['warning'] < 0) {
            throw new RuntimeException('minutes config values must be positive or zero');
        }
        if ($minutes['warning'] !== 0 && $minutes['warning'] < $minutes['notice']) {
            throw new RuntimeException('minutes config "warning" value must be greater than or equal to "notice" value');
```
Just providing a little more context in the exception messages to make it clear _which_ config is incorrect. 

Yes this class only has one direct config currently, but someone who's in a rush and doesn't look at where the exception is being thrown might get frustrated because "which config? This site has a billion config parameters"

code/Forms/UnsavedChangesIndicator.php
```diff
@@ -0,0 +1,66 @@
...
+use SilverStripe\Forms\FormAction;
+
+/**
+ * A visual indicator to show whether a form has unsaved changes.
+ *
+ * Note that this isn't strictly a FormAction as it doesn't perform any action when clicked.
+ * However, extending FormAction allows us to easily integrate with the existing form action
+ * architecture (e.g. placement in the form actions area).
+ */
+class UnsavedChangesIndicator extends FormAction
```
ChloeSartorelli:
```suggestion
class UnsavedChangesIndicator extends DatalessField
```
And add `use SilverStripe\Forms\DatalessField;`

The reason it doesn't work with `LiteralField` (which you noted in https://github.com/silverstripe/silverstripe-framework/pull/11905#discussion_r2540454285 doesn't work) is because `LiteralField` overrides the `Field()` and `FieldHolder()` methods to return the `$this->content` value. 

It doesn't make sense to have a settable content for this indicator, so we can just subclass `DatalessField` directly. I have tested this and it worked for me in cmsmain, assetadmin, siteconfig, and a modeladmin which is all I tried before concluding it works.

Note that just in case other code in these PRs currently relies on a `action_$action` name I added the below constructor. If any code does expect that name that code should be updated to not expect that name as this isn't an action (i.e. don't use this constructor - I'm just adding it so you know what I changed when I tested this).

```php
public function __construct($action)
{
    parent::__construct("action_$action", '');
}
```

client/src/legacy/UnsavedChangesIndicator/UnsavedChangesIndicatorEntwine.js
ChloeSartorelli: From my testing, `onmatch` works, but `onunmatch` doesn't fire at all. It does fire if the element is removed, but not if it's just updated. That's clearly a bug. I've raised https://github.com/silverstripe/silverstripe-admin/issues/2063 to fix it and remove this workaround once it has been fixed - fine to keep it for now.

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicatorTimer.js
```diff
@@ -0,0 +1,93 @@
...
+      setStartTime(currentStartTime);
+    }
+
+    // Function to check elapsed time and update state
+    const checkTimeSinceDirty = () => {
+      const now = Date.now();
+      const elapsedMs = now - currentStartTime;
+      // Current elapsed minutes rounded down
+      const currentMinutes = Math.floor(elapsedMs / 60000);
+      // Access prev state here do need elapsedMinutes in the useEffect dependency array.
```
ChloeSartorelli: > Access prev state here do need elapsedMinutes in the useEffect dependency array.

This seems to say that `elapsedMinutes` should be in the useEffect dependency array, but it isn't? And probably shouldn't be - it's working as is.

GuySartorelli (changes requested):

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
emteknetnz: Updated to use ref + memo combo

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
emteknetnz: Have changed `<= 1` to `=== 1`

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.scss
emteknetnz: It's standard to have a `__container` class in entwine defined in PHP or the template to act as a target for the react component. The react component does not replace the target.

e.g. https://github.com/silverstripe/silverstripe-versioned-admin/blob/3/client/src/legacy/HistoryViewer/HistoryViewerEntwine.js#L11

> Why isn't this as a parent element just part of the react component?

In the entwine context, The `display: flex` needs to be on the element that's a direct child of `btn-toolbar` in order for the vertical alignment to work right.  In this case the direct child has to be the entwine target, not the react component.

emteknetnz: I've added a large inline comment explaining this

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicatorTimer.js
emteknetnz: Opps basically a typo, changed to `Access prev state here so don't need elapsedMinutes in the useEffect dependency array`

code/Forms/GridFieldDetailFormPreviewExtension.php
emteknetnz: Have split off to GridFieldDetailFormUnsavedChangesIndicatorExtension

code/Forms/UnsavedChangesIndicator.php
emteknetnz: Updated

code/Forms/UnsavedChangesIndicator.php
emteknetnz: Updated

code/Forms/UnsavedChangesIndicator.php
emteknetnz: Have updated though without the full path, you can easily just search for the filename

code/Forms/UnsavedChangesIndicator.php
emteknetnz: Same as above

code/Forms/UnsavedChangesIndicator.php
emteknetnz: Updated

code/Forms/UnsavedChangesIndicator.php
emteknetnz: Updated

code/Forms/UnsavedChangesIndicator.php
emteknetnz: Updated

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
ChloeSartorelli: I still don't think it's worth having entirely separate localisations for something we don't expect to hit. It feels like you ignored the whole first paragraph in my previous comment.

I'd prefer this to be removed, but it's also probably not worth ping-ponging about so I won't hold up merging just for this.

code/Forms/UnsavedChangesIndicator.php
```diff
@@ -0,0 +1,75 @@
...
+    ];
+
+    protected $schemaComponent = 'UnsavedChangesIndicator';
+
+    public function __construct($name, $title = null, $value = null)
+    {
+        // This field must be named 'UnsavedChangesIndicator' in order for
+        // applyTransforms.js to find and update it
+        return parent::__construct('UnsavedChangesIndicator', $title, $value);
+    }
```
ChloeSartorelli:
```suggestion
    public function __construct($title = null, $value = null)
    {
        // This field must be named 'UnsavedChangesIndicator' in order for
        // applyTransforms.js to find and update it
        return parent::__construct('UnsavedChangesIndicator', $title, $value);
    }
```
Why have a required `$name` parameter that we then don't use? We probably shouldn't have that parameter at all, but if having it makes tests easier or something then we can keep it but make it optional.

> This field must be named 'UnsavedChangesIndicator' in order for applyTransforms.js to find and update it

I don't think any other field has this requirement, I suspect it wouldn't be too hard to use something else to select the field instead and allow setting a variable name for the field like all the other fields - but it's debatable whether it'd be worth finding a way to do that.

I don't feel strongly about that part so I'll leave it up to you to decide if it's worth the extra time in order to keep this field's "name" treatment the same as all our other fields.

GuySartorelli (changes requested):
Sorry, forgot to hit start review. https://github.com/silverstripe/silverstripe-admin/pull/2058#discussion_r2557601515 is part of this review.

code/Forms/UnsavedChangesIndicator.php
emteknetnz: Have removed all the constructor arguments since they're not needed

client/src/components/UnsavedChangesIndicator/UnsavedChangesIndicator.js
emteknetnz: Opps sorry I did miss that, have removed the singular

GuySartorelli (approved):
LGTM, works well locally

PR #2059, opened 2025-11-14
FIX: Add default templates for SingleRecordAdmin (fixes #2057)

### LINKED ISSUE #2057: SingleRecordAdmin doesn't render the edit form properly
### Module version(s) affected

3.1.5

### Description

Following the docs results in a page with the form fields not visible:

Image

After copying the `SiteConfig` templates (https://github.com/silverstripe/silverstripe-siteconfig/tree/6/templates/SilverStripe/SiteConfig/Includes) and renaming them to match the custom admin class name, the fields show:

Image

### How to reproduce

```php
<?php

namespace App\Model\SelectLocation;

use SilverStripe\Assets\Image;

class SelectLocationConfig extends DataObject
{
    private static string $table_name = 'SelectLocationConfig';

    private static array $db = [
        'Title' => 'Varchar(255)',
    ];

        'HeroImage' => Image::class,
    ];

    private static array $owns = [
        'HeroImage',
    ];

    public function canDelete($member = null): bool
    {
        return false;
    }
}
```

```php
<?php

namespace App\Admin;

use App\Model\SelectLocation\SelectLocationConfig;
use SilverStripe\Admin\SingleRecordAdmin;

class SelectLocationAdmin extends SingleRecordAdmin
{
    private static string $url_segment = 'select-location';

    private static string $menu_title = 'Select Location';

    private static string $model_class = SelectLocationConfig::class;

    private static string $menu_icon_class = 'font-icon-globe';
}
```

### Possible Solution

Presumably if we just copy the two SiteConfig templates to the following paths, things should work as expected out of the box?

- `templates/SilverStripe/Admin/SingleRecordAdmin_Content.ss`
- `templates/SilverStripe/Admin/SingleRecordAdmin_EditForm.ss`

### Additional Context

_No response_

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## PRs

- https://github.com/silverstripe/developer-docs/pull/846
- https://github.com/silverstripe/silverstripe-admin/pull/2059

lozcalver:
## Description
Targeted this at the `3` branch because if we later opt to remove the matching templates from the siteconfig repo, we can then update the `composer.json` requirement to point to a minor release e.g. `"silverstripe/admin": "^3.2"`.  I suppose that constraint _could_ point to a patch release, but that feels weird to me for some reason... happy to be overruled on that.

## Manual testing steps
See reproduction steps in https://github.com/silverstripe/silverstripe-admin/issues/2057#issue-3617178304. Without these templates, the edit form won’t be visible.

## Issues

- #2057

## Pull request checklist

- [ ] The target branch is correct
- [x] CI is green

### DIFF
--- a/code/SingleRecordAdmin.php
+++ b/code/SingleRecordAdmin.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Admin;
 
 use SilverStripe\Forms\Form;
+use SilverStripe\Forms\HiddenField;
 use SilverStripe\ORM\DataObject;
 
 /**
@@ -25,6 +26,19 @@ abstract class SingleRecordAdmin extends LeftAndMain
 
     public function getEditForm($id = null, $fields = null): ?Form
     {
+        if (!$fields) {
+            if (!$id) {
+                $id = $this->currentRecordID();
+            }
+            $record = $this->getRecord($id);
+            if ($record) {
+                $fields = $record->getCMSFields();
+            }
+        }
+        if ($fields && !$fields->dataFieldByName('ID')) {
+            $fields->add(HiddenField::create('ID'));
+        }
+
         $form = parent::getEditForm($id, $fields);
         // Keep the tabs on the top row, rather than underneath
         if ($form?->Fields()->hasTabSet()) {
--- a/code/SingleRecordAdmin.php
+++ b/code/SingleRecordAdmin.php
@@ -28,6 +28,7 @@ abstract class SingleRecordAdmin extends LeftAndMain
         $form = parent::getEditForm($id, $fields);
         // Keep the tabs on the top row, rather than underneath
         if ($form?->Fields()->hasTabSet()) {
+            $form->addExtraClass('cms-tabset'); // Required for tabsets to work, see CMSTabSet.ss for info
             $form->Fields()->findOrMakeTab('Root')->setTemplate('SilverStripe/Forms/CMSTabSet');
         }
         return $form;


GuySartorelli:
We can have a constraint of `^3.2` in silverstripe/siteconfig even if this gets merged into `3.1` and released as a patch - I don't think composer constraints in a separate module is a convincing reason to leave this broken until April.

Please retarget to the `3.1` branch. You may need to [reset your commits](https://gist.github.com/GuySartorelli/74627843aeb75e12b777b1f44cb22d2f) after retargetting.

GuySartorelli:
I've ticked the relevant boxes in the PR checklist for you - but please note the checklist is there for _you_ not for me. The intention is that it helps you catch things that might need to change like commit messages etc before someone comes along to review it.

Also the commit prefix should technically not have a colon (i.e. should be `FIX Add default....` not `FIX: Add default...`) but it doesn't really matter, the changelog generator knows how to handle it so just a note for next time.

templates/SilverStripe/Admin/Includes/SingleRecordAdmin_Content.ss
```diff
@@ -0,0 +1,27 @@
+<div id="settings-controller-cms-content" class="flexbox-area-grow fill-height cms-content cms-tabset $BaseCSSClasses" data-layout-type="border" data-pjax-fragment="Content" data-ignore-tab-state="true">
```
ChloeSartorelli: That ID doesn't seem correct. I assume this was just a direct copy paste from siteconfig? That ID is correct for siteconfig but not for the general case.

GuySartorelli (changes requested):
After a bit of playing around, I have found that adding the `cms-tabset` class to the div in `LeftAndMain_Content.ss` fixes it, and doesn't seem to break anything anywhere else. No need for additional templates.

lozcalver:
Thanks Guy, I’ll take another look at this soon. I’ll also include the fix for the missing ID field in this PR but as a separate commit

lozcalver:
Okay after testing a bit further, adding `cms-tabset` to one of the _templates_ works on initial load, but the tabs fail to re-initialise after the PJAX fragment reload stuff. Adding it to the form instead does work on both initial load and PJAX reload. The upshot is that this is now a one-line fix, with no new templates required.

The PJAX fragment (CurrentForm) replaces the entire <form> HTML, and the code that re-initialises the tabset when this happens will only look at the form and any children, it won’t look up to any parent containers: https://github.com/silverstripe/silverstripe-admin/blob/ac4bb9d4cf0ffb76bf426561fc6ea2430b38cd48/client/src/legacy/LeftAndMain.EditForm.js#L220

I’ve also pushed a second commit with the suggested fix for missing `ID` fields

michalkleiner (approved):

michalkleiner:
Should we retarget at minor version branch for a patch fix?

michalkleiner (approved):
LGTM, but I'll let @GuySartorelli have one final look and then merge.

GuySartorelli (approved):
Looks good to me. Thanks for your patience on this and thanks @michalkleiner for giving this a review, makes it a lot easier to get it over the line.

Issue #2065, opened 2025-11-27
Convert some admin class components to functional


emteknetnz:
There are react class component(s) that needs to be converted to react functional component(s).

Convert the following:
- [x] InsertLinkModal.js
- [x] Breadcrumb.js
- [x] ViewModeToggle.js
- [x] MobileMenuToggle.js
- [x] Search.js
- [x] SearchBox.js

Note there are more component than this to convert, limiting number converted at a time to make peer review more manageable.

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/19840461145

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2066

GuySartorelli:
PRs merged.

PR #2066, opened 2025-11-27
ENH Convert some react class components to functional components

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2065


client/src/containers/InsertLinkModal/InsertLinkModal.js
```diff
@@ -1,68 +1,71 @@
...
 import { connect } from 'react-redux';
 import FormBuilderModal from 'components/FormBuilderModal/FormBuilderModal';
 import fileSchemaModalHandler from 'containers/InsertLinkModal/fileSchemaModalHandler';
 import * as schemaActions from 'state/schema/SchemaActions';
 
-class InsertLinkModal extends Component {
-  constructor(props) {
-    super(props);
+const InsertLinkModal = (props) => {
+  const {
```
emteknetnz, line 10: The following props are in propTypes, but not used in this component, though are passed through
- schemaUrl,
- actions,
- requireLinkText,
- currentPageID,

client/src/components/Breadcrumb/Breadcrumb.js
```diff
@@ -32,80 +30,73 @@ class Breadcrumb extends Component {
...
-  }
+  };
 
   /**
+   * @param {Array} icon
    * @returns {*}
    */
-  renderLastCrumb() {
-    const crumb = this.getLastCrumb();
+  const renderIcons = (icons) => icons.map((icon, i) => {
```
emteknetnz, line 39: Moved method up because it's called by something below

client/src/components/Search/SearchBox.js
```diff
@@ -12,131 +12,135 @@ import TagPropType from '../Tag/TagPropType';
...
+    left: true,
+  },
+  children
+}) => {
+  const [hasFocus, setHasFocus] = useState(false);
+  const [width, setWidthState] = useState(window.innerWidth - 180 - 55);
+  const [tagWidth, setTagWidth] = useState(0);
+  const nodeRef = useRef(null);
+
+  const getComponentWidth = () => {
```
emteknetnz, line 54: Moved method up because it's called by something below

client/src/components/Search/SearchBox.js
```diff
@@ -12,131 +12,135 @@ import TagPropType from '../Tag/TagPropType';
...
+      return 0;
+    }
+    return nodeRef.current.getBoundingClientRect().width;
+  };
 
-  componentDidUpdate() {
-    const width = this.getComponentWidth();
-    this.setWidth(width);
-  }
+  const setWidth = (newWidth) => {
```
emteknetnz, line 61: Moved method up because it's called by something below

client/src/components/Search/SearchBox.js
```diff
@@ -12,131 +12,135 @@ import TagPropType from '../Tag/TagPropType';
...
     // Don't return negative values
-    width = Math.max(width, 0);
+    calculatedWidth = Math.max(calculatedWidth, 0);
 
-    return width;
-  }
+    return calculatedWidth;
+  };
+
+  const focusOnLastTag = () => {
```
emteknetnz, line 122: Moved method up because it's called by something below

client/src/components/Search/Search.js
```diff
@@ -38,88 +38,88 @@ function hasFilters(filters) {
...
-
-  componentDidMount() {
-    this.setOverrides(this.props);
-  }
-
-  componentWillUnmount() {
-    this.setOverrides();
-  }
+const Search = (props) => {
+  const {
```
emteknetnz, line 42: The following props were removed - defined in propTypes but not used in code:
- tagHandlers,
- borders,

They are still passed into `<SearchBox>`

client/src/components/Search/Search.js
```diff
@@ -215,226 +214,203 @@ class Search extends Component {
...
+      actions.reduxForm.reset(identifierArg);
     }
-  }
+  };
+
+  /**
+   * Wrap up all the data into an object and call the onSearch method provided via the props.
+   * @param {Object} overrides Data to overrides over our existing form data.
+   */
+  const doSearch = (overrides = {}) => {
```
emteknetnz, line 247: Moved method up because it's called by something below

client/src/components/Search/Search.js
```diff
@@ -215,226 +214,203 @@ class Search extends Component {
...
     }
-    this.doSearch(clearables);
-  }
+    doSearch(clearables);
+  };
+
+  /**
+   * Expand fully form
+   */
+  const expand = () => {
```
emteknetnz, line 309: Moved method up because it's called by something below

client/src/components/Search/Search.js
```diff
@@ -215,226 +214,203 @@ class Search extends Component {
...
+  };
 
   /**
    * Show this field when the Search Toggle is click.
    */
-  show() {
-    this.setState({ forceLoadForm: true });
-    if (this.state.display !== DISPLAY.VISIBLE) {
-      this.setState({ display: DISPLAY.VISIBLE });
+  const show = () => {
```
emteknetnz, line 351: Moved method up because it's called by something below

client/src/components/Search/Search.js
```diff
@@ -215,226 +214,203 @@ class Search extends Component {
...
+    const selector = tagItem.focusSelector || `[name=${key}]`;
+    expand();
+    setTimeout(() => focusFirstFormField(selector), 50);
+  };
 
   /**
-   * Show the Search box and set the focus on it after a slight delay.
+   * Clear the Search component and focus on the main input field.
    */
-  open() {
```
emteknetnz, line 299: Removed because it's unused

client/src/components/Search/Search.js
```diff
@@ -215,226 +214,203 @@ class Search extends Component {
...
-    this.show();
-    this.focusInput();
-  }
+  const clearSearchBox = () => {
```
emteknetnz, line 329: Moved method up because it's called by something below

client/src/components/Search/SearchBox.js
```diff
@@ -12,131 +12,135 @@ import TagPropType from '../Tag/TagPropType';
...
-    this.focusOnLastTag = this.focusOnLastTag.bind(this);
-    this.focusOnInput = this.focusOnInput.bind(this);
-
-    this.state = {
-      hasFocus: false,
-      width: window.innerWidth - 180 - 55,
-      tagWidth: 0
-    };
-  }
+const SearchBox = ({
```
emteknetnz, line 22: The following props were removed - defined in propTypes but not used in code:
- filters
- formData
- term

client/src/components/Search/Search.js
```diff
@@ -38,88 +38,85 @@ function hasFilters(filters) {
...
+    actions,
+    // This prop is unused in this component, but has a default value and is passed into SearchBox
+    addFilterPrefix = false,
+  } = _props;
+
+  const props = {
+    ..._props,
+    // Use either the value passed into the component, or the default value
+    addFilterPrefix
+  };
```
ChloeSartorelli, line 79: I would prefer something like this which IMO is more obvious what's happening and doesn't require two separate `props` variables which means we can have simply `props` (no weird `_props`):

```js
if (!props.hasOwnProperty('addFilderPrefix')) {
  props.addFilterPrefix = false;
}
```
Unless there's a good reason to have two separate variables? I may just be missing something.

Also why does only `addFilderPrefix` get this treatment? What about the other defaults?

client/src/components/Search/Search.js
```diff
@@ -38,88 +38,85 @@ function hasFilters(filters) {
...
+    formSchemaUrl,
+    filters = {},
+    formData = {},
+    placeholder = i18n._t('Admin.SEARCH', 'Search'),
+    displayBehavior = BEHAVIOR.NONE,
+    term = '',
+    name = 'searchTerm',
+    filterPrefix = '',
+    forceFilters = false,
+    formIsDirty = false,
```
ChloeSartorelli: Where did the default value for `formIsDirty` come from?

client/src/components/Search/Search.js
```diff
@@ -215,226 +211,203 @@ class Search extends Component {
...
+    const tagItem = tagData[key];
     const clearables = { [key]: undefined };
 
-    const { schemaName, filters } = this.props;
-    this.props.actions.reduxForm.change(schemaName, key, '');
-    this.setOverrides({
-      ...this.props,
+    actions.reduxForm.change(schemaName, key, '');
+    setOverrides({
+      props,
```
ChloeSartorelli:
```suggestion
      ...props,
```
Used to be `...this.props`

client/src/components/Search/Search.js
```diff
@@ -445,74 +418,76 @@ class Search extends Component {
 
     // Convert the tag data to a plain array and remove un-needed attributes.
     return tagDataCopy ?
-      Object.values(tagDataCopy).map(({ key, label, value }) => ({ key, label, value })) :
+      Object.values(tagDataCopy).map(({ key: tagKey, label, value }) => ({ key: tagKey, label, value })) :
```
ChloeSartorelli: What is the `key: tagKey` bit here for? It doesn't look like it's necessary?

client/src/components/Search/Search.js
```diff
@@ -445,74 +418,76 @@ class Search extends Component {
...
+
+  const dirty = formIsDirty || searchTermIsDirty();
+  const data = getData();
+  const clearable = (Object.keys(data).length > 0);
+
+  return (
+    <Focusedzone onClickOut={onClickOut} className="search" ref={nodeRef}>
+      <SearchBox
+        {...props}
+        placeholder={placeholder}
```
ChloeSartorelli: Why was `placeholder` added?

client/src/components/Search/SearchBox.js
```diff
@@ -292,60 +284,64 @@ class SearchBox extends Component {
...
+    aria-label={i18n._t('Admin.CLOSE_SEARCH', 'Close search')}
+    className="btn--no-text btn--icon-lg search-box__cancel"
+    icon="cancel"
+    aria-controls={id}
+    aria-expanded="true"
+  />);
+
+  useEffect(() => {
+    // Do not run on first render to match previous behaviour from class component
+    // Note: This hook has no dependencies so will behave like componentDidMount() used to
```
ChloeSartorelli: There was no `componentDidMount` - there was only `componentDidUpdate`. Is that what you mean?

ChloeSartorelli: Also, there's no mention in the class component implementation of not running in the first render - can you please explain why we need to explicitly manage that now?

client/src/components/Search/SearchBox.js
```diff
@@ -12,131 +12,136 @@ import TagPropType from '../Tag/TagPropType';
...
+  },
+  children
+}) => {
+  const [hasFocus, setHasFocus] = useState(false);
+  const [width, setWidthState] = useState(window.innerWidth - 180 - 55);
+  const [tagWidth, setTagWidth] = useState(0);
+  const nodeRef = useRef(null);
+  const isFirstRender = useRef(true);
+
+  const getComponentWidth = () => {
```
ChloeSartorelli, line 54: Please add the doc comment back to this and any other functions you removed it from.

client/src/components/Search/SearchBox.js
```diff
@@ -12,131 +12,136 @@ import TagPropType from '../Tag/TagPropType';
...
-    const width = this.getComponentWidth();
-    this.setWidth(width);
-  }
+  const setWidth = (newWidth) => {
+    setWidthState((prevWidth) => {
+      if (prevWidth !== newWidth) {
+        return newWidth;
+      }
+      return prevWidth;
+    });
```
ChloeSartorelli:
```suggestion
    setWidthState(newWidth);
```
Is this not going to have an identical result?

client/src/components/Search/SearchBox.js
```diff
@@ -12,131 +12,136 @@ import TagPropType from '../Tag/TagPropType';
...
+  };
+
+  const onTagListResize = (dimensions) => {
+    const newTagWidth = dimensions.width;
+    setTagWidth((prevTagWidth) => {
+      if (prevTagWidth !== newTagWidth) {
+        return newTagWidth;
+      }
+      return prevTagWidth;
+    });
```
ChloeSartorelli:
```suggestion
    setTagWidth(newTagWidth);
```
Won't this end up being the exact same result?

client/src/components/Search/SearchBox.js
```diff
@@ -12,131 +12,136 @@ import TagPropType from '../Tag/TagPropType';
...
+  };
 
-  calculateInputRightPadding() {
-    // Right padding is used to set the alignedment of the "x" that shows in chrome
-    // when you type in the search box
-    return this.state.width < 576 ? 121 : 264;
-  }
+  // Right padding is used to set the alignedment of the "x" that shows in chrome
+  // when you type in the search box
+  const calculateInputRightPadding = () => (width < 576 ? 121 : 264);
```
ChloeSartorelli: If you're moving the comment to above the function please use `/**  */` style comments

client/src/components/Search/SearchBox.js
```diff
@@ -146,64 +151,57 @@ class SearchBox extends Component {
...
+      paddingRight: `${calculateInputRightPadding()}px`
     };
 
-    const mergedBorders = Object.assign({}, SearchBox.defaultProps.borders, borders);
+    const defaultBorders = {
+      top: false,
+      right: false,
+      bottom: true,
+      left: true,
+    };
```
ChloeSartorelli: Might be sensible to store this outside the component so it can be reused in the constructor and here - that way if we update the default values we don't have to remember that it's in two separate places.

client/src/containers/InsertLinkModal/InsertLinkModal.js
```diff
@@ -1,68 +1,71 @@
...
-    if (!props.isOpen) {
-      props.setOverrides(null);
+  const prevIsOpenRef = useRef(isOpen);
+
+  useEffect(() => {
+    if (!isOpen) {
+      setOverrides(null);
     }
-  }
+  }, []);
```
ChloeSartorelli, line 23: This was just happening in the constructor before - what's the reasoning for updating to being in the equivalent of `componentDidMount`?

client/src/containers/InsertLinkModal/InsertLinkModal.js
```diff
@@ -1,68 +1,71 @@
...
-    if ((props.isOpen && !oldProps.isOpen) || (!props.isOpen && oldProps.isOpen)) {
-      props.setOverrides(props.isOpen ? props : null);
+  useEffect(() => {
+    const prevIsOpen = prevIsOpenRef.current;
+    if ((isOpen && !prevIsOpen) || (!isOpen && prevIsOpen)) {
+      setOverrides(isOpen ? props : null);
     }
-  }
+    prevIsOpenRef.current = isOpen;
+  }, [isOpen]);
```
ChloeSartorelli, line 32: Since `isOpen` is a bool, isn't this just equivalent to the below?

```js
useEffect(() => {
  setOverrides(isOpen ? props : null);
}, [isOpen]);
```

GuySartorelli (changes requested):

client/src/components/Search/Search.js
emteknetnz: Should not have been added, have removed default

client/src/components/Search/Search.js
emteknetnz: Cannot modify props - props are immutable - https://react.dev/learn/passing-props-to-a-component#how-props-change-over-time `However, props are [immutable](https://en.wikipedia.org/wiki/Immutable_object) ...`

> Also why does only addFilderPrefix get this treatment? What about the other defaults?

This is mistake from me, we need to include all the props with default values to ensure they're passed through correctly, have updated. 

May have been some other PRs with passed through props where I did not do this, I'll check now

Created issue https://github.com/silverstripe/silverstripe-admin/issues/2067

client/src/components/Search/Search.js
emteknetnz: Fixed

client/src/components/Search/Search.js
emteknetnz: Fixed

client/src/components/Search/Search.js
emteknetnz: Removed

client/src/components/Search/SearchBox.js
emteknetnz: Done

client/src/components/Search/SearchBox.js
emteknetnz: Yup, updated

client/src/components/Search/SearchBox.js
emteknetnz: Yup, updated

client/src/components/Search/SearchBox.js
emteknetnz: Done

client/src/components/Search/SearchBox.js
emteknetnz: Moved to outside the component

client/src/components/Search/SearchBox.js
emteknetnz: Yes `componentDidUpdate` is what I meant, have updated comment

In the class component this code did not run in the first render. We need the `isFirstRender.current` check to ensure that it does not run in the first render here.

client/src/containers/InsertLinkModal/InsertLinkModal.js
emteknetnz: Because functional components do not have a `constructor()` method, this hook is the usual equivalent of it

Could technically use

```js
const InsertLinkModal = (...) => {
  const isFirstRenderRef = useRef(true);
  if (isFirstRenderRef.current) {
    setOverrides(null);
    isFirstRenderRef.current = false;
  }
  // ...
}
```

Though I don't see any reason to use that less standard syntax over the cleaner `useEffect(() => {...}, [])` hook

client/src/containers/InsertLinkModal/InsertLinkModal.js
emteknetnz: No, we do not want to call this on mount - this is done purely to match the previous behavior. I've added an inline comment for this 

There is a unit test "InsertLinkModal does not call setOverrides when isOpen is true on mount" that will fail if we make the recommended change

All unit tests that were added pass when the component was still a class component

client/src/components/Search/SearchBox.js
```diff
@@ -9,134 +9,141 @@ import ResizeAware from 'components/ResizeAware/ResizeAware';
...
+  name,
+  dirty,
+  clearable,
+  tagData = [],
+  borders = {
+    top: false,
+    right: false,
+    bottom: true,
+    left: true,
+  },
```
ChloeSartorelli:
```suggestion
  borders = defaultBorders,
```
This was the reason for moving that outside the component.

client/src/containers/InsertLinkModal/InsertLinkModal.js
ChloeSartorelli: I couldn't find any indication of a "standard" way to replace constructors.

I saw `useMemo()` used a few times in articles I found, the idea being that it runs immediately (before initial render, just like the constructor of the class component) - whereas the effect will run _after_ the first render, so you have to go through a second render pass to get to the intended state.

That said, our react is super sub-optimal anyway so I won't hold up merging just for this, which works fine.

client/src/containers/InsertLinkModal/InsertLinkModal.js
ChloeSartorelli: While I still don't think this additional complexity is necessary, it's not worth the ping pong so leave as is.

client/src/components/Search/SearchBox.js
ChloeSartorelli: I still don't think this additional complexity is strictly needed but it's not worth pingpong so leave as is.

GuySartorelli (changes requested):
Just the one change left

client/src/components/Search/SearchBox.js
emteknetnz: Updated

GuySartorelli (approved):

PR #2068, opened 2025-12-02
ENH Add roving tab index to tree

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/3135

Requires https://github.com/silverstripe/silverstripe-cms/pull/3136 for behat to pass (even though this is the admin module, the cms site tree is tested)


client/src/legacy/LeftAndMain.Tree.helpers.js
ChloeSartorelli, line 1: Why create a whole new file for these when they're only used in `LeftAndMain.Tree.js`? Feels like they should just live in there?

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -3,6 +3,9 @@
  */
 
 import $ from 'jquery';
+import { updateRovingTabindex, resetTabindexToCurrentPage } from './LeftAndMain.Tree.helpers';
+
+import '../../../thirdparty/jquery-entwine/jquery.entwine.js';
```
ChloeSartorelli: We shouldn't be importing entwine, it should be included in `dist/js/vendor.js` only (or however it's currently already added). Why do you need this here?

GuySartorelli (changes requested):
I've only done a very cursory glance at the code for now, most of this review is focused on local testing.

I have tested this locally and am using the following resources to base my expectations:
1. https://www.w3.org/WAI/ARIA/apg/practices/keyboard-interface/#kbd_roving_tabindex
1. https://www.w3.org/WAI/ARIA/apg/patterns/treeview/
1. https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/
1. https://github.com/silverstripe/developer-docs/issues/866

I am ignoring the screen reader experience for now, based on previous reviews with you on this. Please let me know if I should be testing screen reader experience as a part of this PR.

## Focus state

When something inside the tree has focus, there should (probably) be a dashed line around the whole roving tabindex container. This is both in our own notes for documenting (and is what the asset admin gallery will have) and in the recommendations in the W3C navigation tree example.

## Top element with focus should not be the arrow

It is visually strange that the root item (the name of the site) doesn't get focus, but the arrow next to it does. At all other levels with an arrow, it's the page name that gets focus.

This is going to be even more important when the screen reader testing is done, but even just visually it doesn't seem right.

It also feels a little strange that pressing "enter" on this node expands/collapses it - though the pattern does explicitly allow this as an action on parent nodes so we can leave that behaviour in.

## Need to expand if current page hidden

In the [W3C nav tree example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) they say:

> When focus leaves the tree, if the item representing the current page is in a branch that has been collapsed, that branch is expanded to ensure the current page item is both visible and exposed as active to assistive technologies.

We have decided to not always give that item tabindex=0, so _maybe_ this doesn't apply anymore, but I'd like Cassie's opinion on that before deciding to ignore the W3C advice.

## Possibly optional keys

The [W3C nav tree example](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/examples/treeview-navigation/) specifies the following keybinds that have not been applied:

- a-z, A-Z
- asterisk (*)

It doesn't specify that these are optional, though [the pattern](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/) does use this language:

> Type-ahead is recommended for all trees, especially for trees with more than 7 root nodes
> ...
> * (Optional): Expands all siblings that are at the same level as the current node.

Since the pattern explicitly marks asterisk as optional, we can leave that off for now.

Since the pattern only "recommends" type-ahead and doesn't require it, if you like I can spin up a separate card to add that as an enhancement instead of it being added in this PR.

## Batch actions

When I activate batch actions, pressing space on a node should select/deselect that node.

client/src/legacy/LeftAndMain.Tree.js
emteknetnz: Moved to LeftAndMain.Tree.js

emteknetnz:
> Focus state

Done

> Top element with focus should not be the arrow

Done

> Need to expand if current page hidden
> We have decided to not always give that item tabindex=0, so maybe this doesn't apply anymore, but I'd like Cassie's opinion on that before deciding to ignore the W3C advice.

Will send Cassie a message over slack, though I suspect it does not apply because in order to use the keyboard to collapse the branch you've then put focus on the branch. You could collapse it with the mouse, which will remove the tab focus entirely (just like it does with clicking anywhere else), though at that point it's not keyboard navigation

**Update: Cassie says this is fine (to ignore the W3C advice)**

> Possibly optional keys
> Since the pattern only "recommends" type-ahead and doesn't require it, if you like I can spin up a separate card to add that as an enhancement instead of it being added in this PR.

OK, I won't implement in this card though if you want to create a new card for that go for it

> Batch actions

Done

client/src/styles/legacy/_tree.scss
```diff
@@ -763,3 +769,12 @@ a .jstree-recordicon {
   }
 }
 /* stylelint-enable no-duplicate-selectors */
+
+/**
+ * Focus outline for tree container which has a roving index
+ * Displays a dashed outline around the entire tree when focus is within it.
+ */
+.cms-tree:focus-within {
```
ChloeSartorelli:
```suggestion
.cms-tree:has(:focus-visible) {
```

Lets the browser decide when the outline should be visible and when it shouldn't.

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -4,7 +4,45 @@
...
+  } else {
+    const $firstPage = $tree.find('li[data-id]').not('[data-id="0"]').first().find('> a').first();
+    if ($firstPage.length) {
+      $firstPage.attr('tabindex', 0);
+    }
+  }
+};
+
+try {
+  require('../../../thirdparty/jquery-entwine/jquery.entwine.js');
```
ChloeSartorelli: Why do we need to import entwine when we didn't previously?

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -4,7 +4,45 @@
...
+  $rootStrongs.attr('tabindex', -1);
+  if ($currentPage.length) {
+    $currentPage.attr('tabindex', 0);
+  } else {
+    const $firstPage = $tree.find('li[data-id]').not('[data-id="0"]').first().find('> a').first();
+    if ($firstPage.length) {
+      $firstPage.attr('tabindex', 0);
+    }
+  }
+};
```
ChloeSartorelli: Why are these defined out here instead of as functions inside the entwine block?

The entwine-y way to do this is to include them in the `$('.cms-tree').entwine({` block and call them as `this.resetTabindexToCurrentPage()` or (if inside a child or the tree for example) `this.closest('.cms-tree').resetTabindexToCurrentPage()`.

These functions then don't need to re-select `const $tree = $targetAnchor.closest('.cms-tree');` - they just use `this`.

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -4,7 +4,45 @@
...
 
-$.entwine('ss.tree', function($){
+const updateRovingTabindex = ($targetAnchor) => {
+  if (!$targetAnchor || !$targetAnchor.length) {
+    return;
+  }
+  const $tree = $targetAnchor.closest('.cms-tree');
+  const $links = $tree.length ? $tree.find('a') : $('.cms-tree a');
+  const $toggles = $tree.length ? $tree.find('ins') : $('.cms-tree ins');
+  const $rootStrongs = $tree.length ? $tree.find('#record-0 > strong') : $('.cms-tree #record-0 > strong');
```
ChloeSartorelli: If `$tree.length` is 0, doesn't this mean the target anchor isn't inside the tree? Under what circumstance would that happen and we still want to grab the tree and continue executing this function?

client/src/legacy/LeftAndMain.Tree.js
ChloeSartorelli: If this was for testability I'm okay with it, but mostly because it's likely we'll be removing this in 7 anyway.

client/src/legacy/LeftAndMain.Tree.js
```diff
@@ -154,13 +204,32 @@ $.entwine('ss.tree', function($){
...
               return null;
             };
 
-            // Find next visible node for ArrowDown
-            var getNextVisible = function($li) {
+            // Find next visible node for ArrowDown (can be toggle or anchor)
+            var getNextVisible = function($li, isOnToggle) {
+              // If on root strong element and root is open, go to first child
+              var $target = $(target);
+              var isOnRootStrong = $target.is('strong') && $li.attr('id') === 'record-0';
```
ChloeSartorelli: Not normally great to look for explicitly IDs like this, but:
1. we do it all over entwine code already
2. this specific ID is already required for the existing code to work anyway

tl;dr this is okay here.

GuySartorelli (changes requested):
>> Need to expand if current page hidden
>> ...
>> I'd like Cassie's opinion
>
> Will send Cassie a message over slack

Did you do this? I didn't see anything in the `cms-ux` channel.

> I suspect it does not apply because in order to use the keyboard to collapse the branch you've then put focus on the branch

I don't quite understand what you're trying to convey with this, since this is about what happens after you focus _away_ from the tree (and therefore nothing in the tree including that branch has focus).

I suspect the intention of the advice to expand the tree to show the current page is so that when you focus back onto the tree, you can use the up/down keys and find the current page without having to remember where the current page lies in the tree.

That can also be useful for users who _can_ see the tree, because it will be visually expanded for them as well. They won't need to remember what pages lead to that page if they want to do something with it in the tree. Though the benefit for them is less, since there are breadcrumbs in the edit form that are very easy to quickly glance at for sighted users.

## Tab index between navigations

Currently the site tree "remembers" what page I was focused on between navigations. This means when I'm editing a page and use the breadcrumb button to go back, the focus is on the previous page I focused on in the tree.

I'm unsure if that's the appropriate behaviour or not. It probably would be in a single page app, but given the whole tree is removed and recreated as part of the navigation I think it should _probably_ reset to the start. I'm not sure, though, that's just gut feel, I don't have an example to draw on for this scenario.

## "first" item to receive roving tabindex

I think the default item to have the roving tabindex (`tabindex="0"`) should probably be the root node when no edit form is open. This is especially important for screen reader users who cannot visually see the root of the tree.

## Mouse interaction should set roving tabindex

When I interact with something in the tree with the mouse (e.g. ticking a checkbox in batch actions mode) the roving tabindex should move to that item.

emteknetnz:
> Did you do this? I didn't see anything in the cms-ux channel.

My bad. Have had follow up conversation on that channel with refined question - answer there was to defer thinking about this until later and implement is separate card is we want

> Tab index between navigations ... Currently the site tree "remembers" what page I was focused on between navigations.

I don't think it does? I tried the page breadcrumb example you gave though that will change the currently edit page, which will set that to tabindex=0 in the site tree. The current experience feels correct to me.

> "first" item to receive roving tabindex .. I think the default item to have the roving tabindex (tabindex="0") should probably be the root node when no edit form is open. This is especially important for screen reader users who cannot visually see the root of the tree.

I really don't like this, simply because the root node shouldn't even be there IMO. It serves no purpose. When I tab into the container I would far prefer to be on the first actually useful thing i.e. the home page. If we are to change anything here it should be to remove it from the tabindex entirely and make it no longer toggle-able.

> Mouse interaction should set roving tabindex

I don't agree with this, this seems very jarring to me. The tabindex and the corresponding border around the focused element only shows up when using keyboard nav. To make the focus suddently jump on mouse interaction seems wrong IMO

client/src/legacy/LeftAndMain.Tree.js
emteknetnz: Can't think of any, have updated to just return if $tree wasn't found

client/src/legacy/LeftAndMain.Tree.js
emteknetnz: It's for testability

I've updated those functions to have a $tree as a firs param which is a variation of your suggestion to use `this`

client/src/styles/legacy/_tree.scss
emteknetnz: Updated

client/src/legacy/LeftAndMain.Tree.js
emteknetnz: Was for unit tests for the functions above. I've change this to a conditional check to only do the entwine stuff if entwine is present and added a comment

GuySartorelli:
## Tab index between navigations

>> Currently the site tree "remembers" what page I was focused on between navigations.
>> This means when I'm editing a page and use the breadcrumb button to go back, the focus is on the previous page I focused on in the tree.
>
> I don't think it does? I tried the page breadcrumb example you gave though that will change the currently edit page, which will set that to tabindex=0 in the site tree. The current experience feels correct to me.

Ahh weird, okay, so it doesn't quite remember it, but it does remember _something_ that it shouldn't be. Here's a video.

https://github.com/user-attachments/assets/d4af3aa2-10ab-462e-80cf-fb90a4139d38

You can see that it isn't remembering where _I_ put focus, instead it's for some reason remembering that gridfield test page _used to be_ open and is giving that the roving tabindex.

## "first" item to receive roving tabindex

>> I think the default item to have the roving tabindex (tabindex="0") should probably be the root node when no edit form is open.
>
> I really don't like this [...] If we are to change anything here it should be to remove it from the tabindex entirely and make it no longer toggle-able.

I'm okay with that.

## Mouse interaction should set roving tabindex

>> When I interact with something in the tree with the mouse (e.g. ticking a checkbox in batch actions mode) the roving tabindex should move to that item.
>
> I don't agree with this, this seems very jarring to me.

We already agreed about this - see the recorded decision in https://github.com/silverstripe/silverstripe-asset-admin/pull/1643#issuecomment-3781423666 and https://github.com/silverstripe/developer-docs/issues/866

Clicking on a checkbox usually gives focus (notably not focus-visible) to that checkbox. In this case the checkbox is part of the tree item, so the tree item should be given the roving tab index and pressing an arrow key should move focus to the appropriate next/previous tree item.

> The tabindex and the corresponding border around the focused element only shows up when using keyboard nav.

Tabindex should move with focus regardless of how focus is moved.

The border only appears with focus-visible - let the browser worry about what events should trigger that. That's not our job.

emteknetnz:
> Tab index between navigations

Fixed, have added behat test

> "first" item to receive roving tabindex

Have removed root node from tabindex

> Mouse interaction should set roving tabindex

I tried this but I simply couldn't get it to work properly. It largely comes down to the fact that we're using jstree. The `check_node` custom event that jstree fires on ticking a checkbox does not include information as to whether the event was triggered by a mouse event or a keyboard event. The checkbox also isn't really a checkbox it's an `<ins>` element that uses background-images to look like a checkbox.

While you want to blur() the parent `<a>` element of the checkbox that was mouse clicked, you don't want to do this with keyboard. [Here](https://github.com/silverstripe/silverstripe-admin/blob/3/client/src/legacy/LeftAndMain.Tree.js#L132) seemed like the correct place to try and implement this logic

If you really insist on this functionality, then please raise a follow up card to do this, I've thrown in the towel at this point.

GuySartorelli:
https://github.com/silverstripe/silverstripe-admin/issues/2108 created for mouse interaction

GuySartorelli (approved):
LGTM. Works as expected locally - just waiting for CI to pass

emteknetnz:
CI is green

Issue #2073, opened 2025-12-03
Convert more admin react components to functional


emteknetnz:
There are react class component(s) that needs to be converted to react functional component(s).

- [ ] Preview.js
- [ ] Badge.js
- [ ] FileStatusIcon.js
- [ ] FormAlert.js
- [ ] CircularLoading.js

Note there are more components than this to convert, limiting number converted at a time to make peer review more manageable.

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/19947492692

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2074

GuySartorelli:
PR merged

PR #2074, opened 2025-12-04
ENH Convert more react components to functional

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2073


client/src/components/Preview/Preview.js
```diff
@@ -8,54 +8,53 @@ import ActionMenu from '../ActionMenu/ActionMenu';
...
+  useEffect(() => {
     // Reset frame loaded state when the preview URL changes
-    if (!this.state.frameLoaded) {
+    if (!frameLoaded) {
       return;
     }
-
-    if (prevProps.previewUrl !== this.props.previewUrl) {
-      this.setFrameLoaded(false);
+    let previewUrl = null;
```
emteknetnz: I think the previous logic was just wrong.

There wasn't a `this.props.previewUrl`, instead previewUrl is defined in `this.props.itemLinks.preview.Stage.href` or `this.props.itemLinks.preview.Live.href`

If found 2x places where `<Preview>` is used - CampaignAdminList and HistoryViewerVersionDetail

I manually tested this by adding `...restProps` on the contstructor and `console.log(restProps);` - there was no `previewUrl` prop in campaign admin in split view nor when viewing a campaign with an image, nor in versioned-admin when viewing a page version with a has_one image in split view

client/src/components/Preview/Preview.js
```diff
@@ -125,45 +124,36 @@ class Preview extends Component {
...
-}
+    </div>
+  );
+};
 
 Preview.propTypes = {
   className: PropTypes.string,
   itemLinks: PropTypes.object,
   itemId: PropTypes.number,
-  onBack: PropTypes.func,
```
emteknetnz: This was unused, removed because eslint was complaining about it being in propTypes

client/src/components/Preview/Preview.js
```diff
@@ -8,54 +8,53 @@ import ActionMenu from '../ActionMenu/ActionMenu';
...
   /**
    * Update state to indicate that the preview page has completed loading
    *
    * @param {boolean} loaded
    */
-  setFrameLoaded(loaded = true) {
-    this.setState({ frameLoaded: loaded });
-  }
-
-  handleBackClick(event) {
```
emteknetnz: This was unused so deleted

client/src/components/Preview/Preview.js
```diff
@@ -8,54 +8,53 @@ import ActionMenu from '../ActionMenu/ActionMenu';
...
-    if (typeof this.props.onBack === 'function') {
-      event.preventDefault();
-      this.props.onBack(event);
-    }
-  }
+  const setFrameLoadedState = (loaded = true) => {
```
emteknetnz, line 50: Renamed function because there's a state variable with the same name

client/src/components/FileStatusIcon/FileStatusIcon.js
```diff
@@ -9,17 +9,25 @@ const cssClass = 'file-status-icon';
...
+  hasRestrictedAccess,
+  isTrackedFormUpload,
+  placement = 'auto',
+  extraClassName,
+  disableTooltip = false,
+  includeBackground
+}) => {
   /**
-   * @param {boolean} hasRestrictedAccess
+   * @param {boolean} hasRestrictedAccessParam
```
emteknetnz, line 22: Renamed param because it's the same name as a prop

client/src/components/FileStatusIcon/FileStatusIcon.js
```diff
@@ -9,17 +9,25 @@ const cssClass = 'file-status-icon';
...
    * @returns {Object}
    */
-  buildTrackedFormUpload(hasRestrictedAccess) {
-    const fontIconClass = hasRestrictedAccess
+  const buildTrackedFormUpload = (hasRestrictedAccessParam) => {
```
emteknetnz, line 25: Renamed param because it's the same name as a prop

client/src/components/FormAlert/FormAlert.js
```diff
@@ -39,54 +39,52 @@ class FormAlert extends Component {
       default:
         return 'danger';
     }
-  }
+  };
+
+  /**
+   * Handler for when the message box is dismissed and hidden
+   */
+  const handleClosed = () => {
```
emteknetnz, line 47: Moved method up because it's called by something below

client/src/components/Preview/Preview.js
```diff
@@ -8,54 +8,62 @@ import ActionMenu from '../ActionMenu/ActionMenu';
...
+      } else if (itemLinks.preview.Live) {
+        previewUrl = itemLinks.preview.Live.href;
+      }
+    }
+    if (lastPreviewUrlRef.current !== previewUrl) {
+      setFrameLoaded(false);
+      lastPreviewUrlRef.current = previewUrl;
     }
-  }
+  }, [frameLoaded, itemLinks]);
```
ChloeSartorelli: Why does `frameLoaded` need to be in the dependencies?

The old logic (if it used the correct variables to get the URL) only compared the URL. And the new logic also only compares the URLs.

I don't think we need to ever call `setFrameLoaded(false)` when `frameLoaded` changes, right?
Removing that from the dependencies should simplify this because we won't need the weird ref anymore.

ChloeSartorelli: We probably also don't need `itemLinks` as a whole - we can calculate the URL before the effect and then use the calculated URL as the dependency.

client/src/components/Preview/Preview.js
```diff
@@ -64,36 +72,36 @@ class Preview extends Component {
...
-        previewUrl = this.props.itemLinks.preview.Live.href;
-        previewType = this.props.itemLinks.preview.Live.type;
+    if (itemLinks && itemLinks.preview) {
+      if (itemLinks.preview.Stage) {
+        previewUrl = itemLinks.preview.Stage.href;
+        previewType = itemLinks.preview.Stage.type;
+      } else if (itemLinks.preview.Live) {
+        previewUrl = itemLinks.preview.Live.href;
+        previewType = itemLinks.preview.Live.type;
       }
```
ChloeSartorelli: If this is identical to what you're doing in the effect, this should be done once and then used in both places.
If it's not the same, please add code comments explaining why we need to do two different things to get two different URLs.

GuySartorelli (changes requested):

client/src/components/Preview/Preview.js
emteknetnz: `frameLoaded` is used in a conditional statement within the useEffect() block so needs to be in dependencies

Have removed `itemLinks` from dependencies and put in `previewUrl` in its place - am now deriving previewUrl in a new block above

client/src/components/Preview/Preview.js
emteknetnz: Moved to the top of the component

client/src/components/Preview/Preview.js
ChloeSartorelli: > frameLoaded is used in a conditional statement within the useEffect() block so needs to be in dependencies

My understanding is that it only needs to be in the dependencies if the effect should fire when the value changes - it doesn't have to be in dependencies just because we use the value in the effect... but it's not worth the ping pong to tidy this up so leave as is.

GuySartorelli (approved):
LGTM

Issue #1970, opened 2025-08-11
Accessibility: Review contrast of icons and text in left menu


GuySartorelli:
https://github.com/silverstripe/silverstripe-framework/issues/6408 raised some concerns about the contrast of icons in the left menu of the CMS.
Note that when the menu is collapsed, the icons are the only visual indicator of what each section is (without hovering)

At the time of that issue the icons were black - they're now even less contrasting with the background.

One option mentioned in comments was to increase the size of those icons to 20px (they're currently 17px)

## Acceptance criteria

- [x] Have the UX team review the contrast of the text and the icons in the left menu
- [x] Implement any recommended changes

## PRs

- https://github.com/silverstripe/silverstripe-admin/pull/2075

cassiezoest:
So WCAG requires a 3:1 minimum contrast ratio for graphics and UI components. Currently the icons do meet that requirement with the default menu item background, but not with the hover/active background colour. 

### Recommended/possible changes

I would update the icon fill colour to be #497792 which is slightly darker and has sufficient contrast with both states. 

We could bump up the size but 20px feels a touch too big, could do 19px? You might need to adjust some of the padding/margins to make sure everything looks okay after adjusting the size.

### Icons

Obviously we don't really have much control over what icons people use in the sidebar, but I thought it might be worth noting that some icons are better suited for use in the sidebar. Those that are filled/have thicker lines will be easier to distinguish at a smaller size. Some icons in the Silverstripe font are also a bit smaller so icon size isn't always consistent. For example (this is with darker colour + larger icon size): 

<img width="1355" height="893" alt="Image" src="https://github.com/user-attachments/assets/ef090565-d494-45ff-933d-795753713af3" />

I'm pretty sure that the icons which are styled as squares were designed specifically for elemental blocks where the icons are larger, hence thinner lines. Not as good in the sidebar where they're smaller

<img width="456" height="753" alt="Image" src="https://github.com/user-attachments/assets/779ff500-888a-4327-825e-effaca02c9ee" />

### Text

Text contrast is all good.

emteknetnz:
Linked PR merged

PR #2075, opened 2025-12-04
ENH Improve contrast of icons in left menu

GuySartorelli:
Before and after:

before
after

## Issue

- https://github.com/silverstripe/silverstripe-admin/issues/1970


client/src/components/Menu/Menu.scss
```diff
@@ -364,12 +364,12 @@
   position: absolute;
   margin-top: -1px;
   left: 13px;
-  font-size: $h3-font-size;
+  font-size: $h2-font-size;
```
ChloeSartorelli, line 367: 19px as per `_bootstrap-variables.scss`

client/src/components/Menu/Menu.scss
```diff
@@ -444,7 +444,7 @@
 
       .text {
         display: block;
-        margin-left: 30px;
+        margin-left: 32px;
```
ChloeSartorelli, line 447: Bump up the margin the same amount as we bumped up the icon size, which keeps the gap the same size.
Didn't seem like the other breakpoints matter for this PR.

client/src/components/Menu/Menu.scss
```diff
@@ -364,12 +364,12 @@
...
   margin-top: -1px;
   left: 13px;
-  font-size: $h3-font-size;
+  font-size: $h2-font-size;
   min-width: 16px;
   background-repeat: no-repeat;
   background-position: 50% 2px;
   transition: all $transition-speed-fast;
-  color: darken($base-menu-bg, 44%);
+  color: #497792;
```
ChloeSartorelli, line 372: Seems sensible to just hardcode rather than having some weird darken by an arbitrary amount.

emteknetnz (approved):

PR #2088, opened 2026-01-06
ENH Refactor react components from class to functional

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-elemental/issues/2086


client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -10,87 +10,80 @@ import PropTypes from 'prop-types';
...
-    // clicked inside the component's DOM and reset back to `false` in
-    // `handleDocumentClick`. This component knows when user clicked outside of
-    // its DOM by checking if `wasClicked` is false in `handleDocumentClick`.
-    // All of this works because of the order of event bubbling.
-    this.wasClicked = false;
+const Focusedzone = ({
+  children,
+  className = '',
+  onClickOut,
+  passBackContainer = () => {},
```
emteknetnz, line 31: passBackContainer is a newly added prop used to "pass back" the a ref to the "container" `<div>`

The only parent is Search.js, which has been updated to pass in this passback method

This used to work on the class component due to the lack of encapsulation on the class component where you could just access FocusZone.container from the Search component

client/src/components/Search/Search.js
```diff
@@ -78,7 +78,7 @@ const Search = (_props) => {
     addFilterPrefix
   };
 
-  const nodeRef = useRef(null);
+  const containerRef = useRef(null);
```
emteknetnz, line 81: This component needed to be modified to handle the inability to now go nodeRef.current.container (which used to access FocusedZone.container, though now longer works).

I renamed this variable to make it a little more obvious what it was referring to

client/src/stories/ValueTracker.js
```diff
@@ -1,72 +1,64 @@
...
-    if (value === null) {
-      React.Children.forEach(this.props.children, (child) => {
-        if (value === null || typeof value === 'undefined') {
-          value = child.props.value;
-        }
-      });
+const ValueTracker = ({
+  children,
+  value = null,
+  name,
```
emteknetnz, line 11: name is wasn't in the original propTypes though was used vis this.props.name

client/src/components/ResizeAware/ResizeAware.js
```diff
@@ -1,116 +1,134 @@
-import { createElement, Component, Children,
-  cloneElement, isValidElement } from 'react';
+import {
```
emteknetnz, line 1: This was a complex refactor due to lifecycle nuances and it ended up almost rewritten

client/src/stories/ValueTracker.js
```diff
@@ -1,72 +1,64 @@
...
+  // fallback to using a value the child was given
+  let val = value;
+  if (val === null) {
+    React.Children.forEach(children, (child) => {
+      if (val === null || typeof val === 'undefined') {
+        val = child.props.value;
+      }
+    });
+    if (val === null || typeof val === 'undefined') {
+      val = value;
```
ChloeSartorelli: Doesn't look like this condition was in the original?

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -10,87 +10,80 @@ import PropTypes from 'prop-types';
  * Example:
  *
  * ```jsx
- *    class TestComponent extends React.Component {
- *      ...
- *
- *      handleClickOut() {
+ *    function TestComponent() {
```
ChloeSartorelli:
```suggestion
 *    const TestComponent = (props) => {
```
The example should follow the same best practices as our actual code.

client/src/components/Focusedzone/Focusedzone.js
ChloeSartorelli: It looks like this breaks the example above, since the event handlers are no longer set unless this new prop is passed.

I think this goes a bit too far into the breaking changes territory for my liking, given we give an explicit example of how to use this component in a generic way.

I'll hold off reviewing this component and the search component for now since it sounds like they both rely on this change.

client/src/components/ResizeAware/ResizeAware.js
```diff
@@ -1,116 +1,134 @@
...
+  useEffect
+} from 'react';
 import ResizeObserver from 'resize-observer-polyfill';
 import PropTypes from 'prop-types';
 
 /**
  * Can be wrapped around a component to detect dimension changes.
- *
- * Adapted from https://github.com/FezVrasta/react-resize-aware created by Federico Zivolo.
+ * Adapted from https://github.com/FezVrasta/react-resize-aware
```
ChloeSartorelli: Is that still true, given your comment above that this component was basically rewritten?

client/src/components/ResizeAware/ResizeAware.js
```diff
@@ -1,116 +1,134 @@
...
-    this.handleResize = this.handleResize.bind(this);
-
-    this.state = { };
+const ResizeAware = ({
+  children,
+  onlyEvent,
+  component = 'div',
+  onResize,
+  widthPropName = 'width',
+  heightPropName = 'height',
```
ChloeSartorelli, line 23: Why are we accepting these as props instead of just hardcoding them? When would we want to change these?

client/src/components/ResizeAware/ResizeAware.js
```diff
@@ -1,116 +1,134 @@
...
   component: PropTypes.oneOfType([PropTypes.string, PropTypes.elementType]),
-  onResize: PropTypes.func
-};
-
-ResizeAware.defaultProps = {
-  component: 'div'
+  onResize: PropTypes.func,
+  widthPropName: PropTypes.string,
+  heightPropName: PropTypes.string,
+  onlyEvent: PropTypes.bool,
```
ChloeSartorelli: It's not clear from the code what `onlyEvent` is intended for. Can you please add a comment for this prop for its intended usage? It seems to have something to do with whether to include sizes when cloning something?

client/src/components/ResizeAware/ResizeAware.js
ChloeSartorelli: Based on this comment I'm ignoring the old code and treating this as a brand new file.

client/src/components/ResizeAware/ResizeAware.js
```diff
@@ -1,116 +1,134 @@
...
+    }
+  }, []);
+
+  // Use a Callback Ref to manage the node and the observer.
+  // This is safer than useEffect because it reacts immediately when the DOM node exists.
+  const handleNode = useCallback((node) => {
+    // Cleanup previous observer if the node changes or unmounts
+    if (observerRef.current) {
+      observerRef.current.disconnect();
+      observerRef.current = null;
```
ChloeSartorelli: The comment says "if the node changes or unmounts", but there doesn't seem to be a check for that. How do we know here if the node has changed or unmounted?

GuySartorelli (changes requested):

client/src/components/Focusedzone/Focusedzone.js
emteknetnz: That's not true - there's a `typeof passBackContainer === 'function'` check that's been added to pass the container el back to the parent. Passing in a method to do this is optional. It's just passed in by Search.js because previously just grabbed `this.container` directly and did some jquery stuff to it. 

containerRef.current being set is what the condtionally is for setting the handlers, and that el is in this component, not the parent

client/src/components/ResizeAware/ResizeAware.js
emteknetnz: Well it was originally adapted from it? Doesn't make sense having a comment "adapted from the old class component" :-)

client/src/components/ResizeAware/ResizeAware.js
emteknetnz: They were props you could change on the old class component with default values set in the render() function https://github.com/silverstripe/silverstripe-admin/blob/3/client/src/components/ResizeAware/ResizeAware.js#L79C24-L79C37

client/src/components/ResizeAware/ResizeAware.js
emteknetnz: Because observerRef.current will have been reassigned - it get assigned in the the equivalent of the render() functino

client/src/components/ResizeAware/ResizeAware.js
emteknetnz: I don't know, it was from the old component. It's usage hasn't changed. I've removed the propType

client/src/stories/ValueTracker.js
emteknetnz: It's to work around the classic "there's a prop with the same name as a state variable and a local variable". I'm using `val` because I can reassign to `value` which is a prop

client/src/components/Focusedzone/Focusedzone.js
emteknetnz: Updated

client/src/components/Focusedzone/Focusedzone.js
ChloeSartorelli: Ahh sorry I mixed up the new passback ref with the new `if (!containerRef.current) {` condition

client/src/components/ResizeAware/ResizeAware.js
ChloeSartorelli: Right, but you rewrote the component. So does your rewrite include anything that was adapted from that? Or did you work from scratch?

>  Doesn't make sense having a comment "adapted from the old class component" :-)

Naturally not. If there's no longer anything here that was adapted from the place this comment says it was, we should remove the comment.

client/src/stories/ValueTracker.js
ChloeSartorelli: I don't understand why that means you need to add a whole new condition? You already set `let val = value;` in the first line, so from there on it shouldn't matter that the name isn't the same as the prop name.

client/src/components/ResizeAware/ResizeAware.js
ChloeSartorelli: Sorry, I still don't understand. The condition we have here is `if (observerRef.current) {`. If that is ever true (i.e. if there is ever a ref there), we clean up the observer.

What it doesn't do is check if the node has changed or unmounted - which is what the comment says the condition should be.

client/src/components/Focusedzone/Focusedzone.js
```diff
@@ -10,87 +10,80 @@ import PropTypes from 'prop-types';
...
-    if (!this.wasClicked) {
-      this.props.onClickOut(evt);
+  // Pass back container reference to parent component
+  useEffect(() => {
+    if (typeof passBackContainer === 'function' && containerRef.current) {
+      passBackContainer(containerRef.current);
     }
-    this.wasClicked = false;
-  }
+  }, [containerRef, passBackContainer]);
```
ChloeSartorelli: The original just had `componentDidMount` and `componentWillUnmount` - please explain why we now have two effects which take a dependency args (i.e effectively `componentDidUpdate`)

And as a general reminder - for these refactor PRs whenever you make a change that isn't straight like for like, please add a comment pro-actively so I don't need to ask about it.

ChloeSartorelli: I think it would be better to make a function to set the refs like this:

```js
const setRefs = (element) => {
  containerRef.current = element;
  if (typeof passBackContainer === 'function') {
    passBackContainer(element);
  }
}
```

And then pass this function into the `ref` prop of the div below. This way we're setting refs in a more standard way and don't need to explain why there's a kinda weird effect.

GuySartorelli (changes requested):
I forgot to hit start review before replying to previous comments, so there's a few replies above that won't appear below.

client/src/components/Focusedzone/Focusedzone.js
emteknetnz: See https://github.com/silverstripe/silverstripe-admin/pull/2088#discussion_r2663632124

When Focuszone was a class coponent, Search.js used to just go [nodeRef.current.container](https://github.com/silverstripe/silverstripe-admin/blob/3/client/src/components/Search/Search.js#L183) - we can't do that any more so we need a mechanism for Search.js to access the container element in the Focusedzone component

client/src/components/ResizeAware/ResizeAware.js
emteknetnz: Removed

client/src/components/ResizeAware/ResizeAware.js
emteknetnz: Opps my bad, have added observerRef.current to dependencies array

client/src/stories/ValueTracker.js
emteknetnz: Sorry see what you mean now, have removed that bit

client/src/components/Focusedzone/Focusedzone.js
emteknetnz: Updated

client/src/components/ResizeAware/ResizeAware.js
emteknetnz: Actually reverted that, it caused an infinite loop

I hopefully reduce confusion I've changed the comment from "Cleanup previous observer if the node changes or unmounts" to just "Cleanup previous observer"

client/src/components/Focusedzone/Focusedzone.js
ChloeSartorelli: Sorry, I still don't understand why the `componentDidMount` was refactored into the equivalent of a `componentDidUpdate`. That doesn't seem to have anything to do with the `passBackContainer` ref.

emteknetnz: Sorry thought you were referring to the passBackContainer - you're referring to the useEffect() block that sets the events handlers

I've modified that block slightly changing the code inside it a little, and removed all the things from the dependency array except for onClickOut - that should be in the deps array as if it changes upstream it means that its usage in handleDocumentClick will be stale.

I think this wouldn't have been an issue in the class component because it used to use this.props.onClickOut and referencing props this way means they weren't ever stale (I could be wrong there though)

GuySartorelli (approved):
LGTM

PR #2089, opened 2026-01-06
ENH Update focus states to be consistent

GuySartorelli:
This PR uses CSS variables to set focus outline style, colour, width, and offset.
This has a few advantages over relying directly on SCSS variables:

- Developers can easily update the styles of their own components and even our components by simply updating CSS variable values
- We don't need to redeclare most of the `outline-*` CSS properties, which means the specificity remains very broad, making targetted updates very easy.

This PR updates all `focus-visible` styling to be consistent. Broadly speaking:

- Focused elements have a dark blue outline
- If the background is too dark for the outline to display, a white outline is used instead.
- If there is no room to display an outline outside the focused element, the outline is inset inside the element's boundaries

## Issue

- https://github.com/silverstripe/.github/issues/443


client/src/components/Paginator/Paginator.scss
```diff
@@ -20,12 +20,10 @@
     text-align: center;
     display: inline;
 
-    &:focus {
+    &:focus-visible {
```
ChloeSartorelli, line 23: focus-visible lets the browser decide when the focus outline is displayed. This means our styles are more inline with browser defaults :+1:

client/src/styles/_bootstrap-variables.scss
```diff
@@ -349,8 +349,8 @@ $input-btn-padding-y-sm: .3077rem; // 4px
 //$input-btn-line-height-lg:  1.5 !default;
 
 //$btn-font-weight:                $font-weight-normal !default;
-//$btn-box-shadow:                 inset 0 1px 0 rgba($white,.15), 0 1px 1px rgba($black,.075) !default;
-//$btn-focus-box-shadow:           0 0 0 3px rgba($primary, .25) !default;
+$btn-box-shadow: none;
+$btn-focus-box-shadow: none;
```
ChloeSartorelli, line 353: Removing a bunch of drop shadows in favour of the new focus outline

client/src/styles/legacy/_style.scss
```diff
@@ -263,7 +285,6 @@ html, body {
  * ----------------------------------------------------------------- */
 .breadcrumbs-wrapper {
   width: 250px; // to enable flexbox auto-grow
-  overflow: hidden;
```
ChloeSartorelli, line 266: This was originally added in https://github.com/silverstripe/silverstripe-admin/pull/304.
I tried adding a bunch of pages, some with long names, and didn't experience any difference with this vs without it - except that without it the focus outline displays correctly for links at the edges of the wrapper boundaries.

Here's a screenshot of my test, note it looked like this both with and without hiding the overflow.

Screenshot from 2026-01-06 16-32-06

ChloeSartorelli: Note that react breadcrumbs are handled separately. Because of course they are.

emteknetnz (changes requested):
Mostly good


## Double outline of inputs

Image

Double outline looks really bad IMO - happens whether I tab or just click on the field. This also happens on textareas e.g. metadata > meta description.
- The 'focus-visible' style 2px border shouldn't be happening on click
- On tab this is less bad, though probably should possibly the 2px border should be "brought in" so that overlay the existing border, so it only looks like there's a single thick border.

## Reports / gridfield missing border

Image

In reports section (probably like this on other gridfields) - left hand border missing

## Model admin border

Image

Border is 2px on top and left but 1px on bottom and right

## Asset-admin (there is no asset-admin PR)

Image

Border around title sort wrong size

GuySartorelli:
> The 'focus-visible' style 2px border shouldn't be happening on click

Wizard though I am, I can't control whether click vs tab gives you a focus-visible styling. That's the browser's job. I'll remove the unnecessary extra border though.

> On tab this is less bad, though probably should possibly the 2px border should be "brought in" so that overlay the existing border, so it only looks like there's a single thick border.

Seems more sensible to only have one outline in that case. I'll do that instead.

## Reports / gridfield missing border

I'll see what I can do about this, though it looks like you might have the wrong screenshot here.

## Model admin border

> Border is 2px on top and left but 1px on bottom and right

I'm not sure what you mean? You might have the wrong screenshot there.

Did you mean this one?

Image

If so, it looks like this for me in firefox:

image

Though it does look more like your screenshot in chrome.
If you can find a way to make them both look good I'll happily apply it. I fiddled with that for about an hour the other day and what I've got is as good as I was able to get.

## Asset admin

> Border around title sort wrong size

The size of the element is the width of the element that drops down. Arguably that might be wrong, but changing the width of that element is out of scope for this card.

emteknetnz:
Ah right I had my screenshots for report/model admin back to front

Those two looks like they can be fixed with changing `:root --focus-outline-offset-inset` (in _style.scss) from
`calc(max(1px, calc(var(--focus-outline-width) - 0.5px)) * -1)`
to
`-2px`

There's a big inline comment about the reason why it had that calculation, though I'd say it's probably now out of date

As for the inputs that trigger the `:focus-visible` state when you click on them ... this is really annoying. I spent a decent while trying to work out why it's trigger `:focus-visible` instead of just `:focus`, though I couldn't work it out. Maybe there's some javascript that calls `.focus()` and for whatever reason the browser thinks we should use `:focus-visible` at this point?

But honestly it's so bad I can't merge while it's still doing this. It looks really bad and is jarring seeing the 'I just tabbed to something' border when clicking on something. This is what happens when I click on a drop-down:

image

GuySartorelli:
> Those two looks like they can be fixed with changing :root --focus-outline-offset-inset (in _style.scss) from
> calc(max(1px, calc(var(--focus-outline-width) - 0.5px)) * -1)
> to
> -2px

Done, but using the var instead of hardcoded 2px. 

> But honestly it's so bad I can't merge while it's still doing this. It looks really bad and is jarring seeing the 'I just tabbed to something' border when clicking on something.

The chosen dropdowns don't let us control focus-within - we can only control "active" and "active :focus-within". There's a whole comment in there about how it's sub-optimal, but better than not having a focus state.

I don't know what you _want_ to happen here. I can't remove the focus styling, we need to have focus styling to meet accessibility requirements. I can't make it _only_ happen on `:focus-visible` for these elements because of constraints enforced on us by the third-party dep we're using to render the dropdown.

~~Though frankly what you're seeing looks consistent with what happens with a regular `select:focus-within` - try this example:~~

Now that I've corrected the example: You're right that with `focus-visible` there is no outline on a basic HTML `select` element.

```html
<html>
<head>
<style>
select:focus-visible {
  outline: 3px solid red;
}
</style>
</head>
<body>
<label for="pet-select">Choose a pet:</label>
<select name="pets" id="pet-select">
  <option value="">--Please choose an option--</option>
  <option value="dog">Dog</option>
  <option value="cat">Cat</option>
</select>
</body>
</html>
```

I have removed the extra border though, which I hadn't noticed was there separately from inputs.

emteknetnz:
That example above you gave uses `:focus-within` - which I think is essentially like `:focus` though if any child within the element gets focus it also gets trigger

If you change that example to `:focus-visible` then tabbing to the dropdown shows the border (correct) though clicking dropdown does not show the border (correct)

> I can't make it only happen on :focus-visible for these elements because of constraints enforced on us by the third-party dep we're using to render the dropdown.

What are those constraints?

GuySartorelli:
Oops yup I should have used `focus-visible` but my brain was confused. Fixed the example - you're right that a basic HTML `select:focus-visible` doesn't get an outline when you click on it.

> What are those constraints?

As I mentioned earlier:

> The chosen dropdowns don't let us control focus-within - we can only control "active" and "active :focus-within".

emteknetnz:
You can partially work around the third party restrictions with some css, e.g.:

```
.chosen-container:has(.chosen-search-input:focus-visible) .chosen-single {
  border: 2px solid red;
  outline: none;
}
```

I did `setInterval(() => console.log(document.activeElement), 1000);` and tabbed to the drop down to see work out what the focus element was (the non-visible input.chosen-search-input)

However ... it still triggers `:focus-visible` on click, so presumably something in javascript land is triggering the `:focus-visible` state within the input, similar to what I'm guessing is happening with the textfields and textareas (which clearly are behaving differently from the standard browser behaviour in your modified example above)

GuySartorelli:
Why are you _adding_ border and _removing_ outline? What is the actual change you want?

emteknetnz:
> Why are you adding border and removing outline?

It's just demonstrating that you're not restricted to using `active :focus-within` i.e. you can target a `:focus-visible` selector with css, though it doesn't solve the core issue that `:focus-visble` is being triggered on click when it shouldn't be

> What is the actual change you want?

Make it so that `:focus-visible` isn't triggered on textinput/textarea/chosen-select/etc on click, should only be triggered on tab

GuySartorelli:
>> What is the actual change you want?
>
> Make it so that :focus-visible isn't triggered on textinput/textarea/chosen-select/etc on click, should only be triggered on tab

To the best of my knowledge that's not possible. The chosen-select is a third party dep and I'm not touching the javascript it uses.

The input and textarea are controlled by the browser - we shouldn't try to fight the browser about whether the input and textarea should have `focus-visible` or not.

e.g. the same basic example as before, but now with input and textarea instead of select. Note they both get a red outline when clicking on them.

```html
<html>
<head>
  <style>
  input:focus-visible,
  textarea:focus-visible {
    outline: 3px solid red;
  }
  </style>
</head>
<body>
  <input type="text" />
  <textarea></textarea>
</body>
</html>
```

emteknetnz:
Oh that makes sense now. Text inputs where you're supposed to type with keyboard are treated as if you tabbed in to them when you click them, since they're assumed to be keyboard inputs (you could still paste in text via mouse though that's not the main way you use them)

The `<select>` not getting a triggered `:focus-visible` on click makes sense because you can easily use that with a keyboard

The chosen-drop down triggering the `:focus-visible` from the fake parent drop-downs `:focus-within` when you click on it may be to do with the non-visible input being a `<input type="text">`

I guess raise an new issue for the dropdown showing the border on click?

GuySartorelli:
I'm still not 100% sure what the problem is so probably better if you make the issue. If I made it it'd say something like "Steve said the dropdown shouldn't have an outline when it's focused" :sweat_smile:

emteknetnz:
https://github.com/silverstripe/silverstripe-admin/issues/2094

emteknetnz (approved):

PR #2093, opened 2026-01-12
ENH Update usage of deprecated event properties

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/472

While I was at it I also updated usage of [.which](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/which) - this happened because `.which` is mostly identical to .keyCode, though there were some ancient browser quirks differences.

However all the the instances of .which I found were to do with the also deprecated MouseEvent.which - so I've updated those to [.button](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button) or [.buttons](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons)


thirdparty/jquery-ui/jquery-ui.js
```diff
@@ -9660,7 +9688,7 @@ var widgetsMouse = $.widget( "ui.mouse", {
 				return this._mouseUp( event );
 
 			// Iframe mouseup check - mouseup occurred in another document
-			} else if ( !event.which ) {
+			} else if (event.buttons === 0) {
```
emteknetnz, line 9691: Note the use of .buttons rather than just .button

GuySartorelli (approved):
LGTM

Issue #2095, opened 2026-01-15
Convert yet more admin react components to functional


emteknetnz:
There are react class component(s) that needs to be converted to react functional component(s).

- [ ] FormBuilder.js
- [ ] ~FormBuilderLoader.js~ Proving hard, will do in its own pull-request
- [ ] fileSchemaModalHandler.js

Note there are more components than this to convert, limiting number converted at a time to make peer review more manageable.

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/21234488027

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2096

GuySartorelli:
PR merged

PR #2096, opened 2026-01-15
ENH Refactor yet more react components from class to functional

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2095

Multiple commits do not squash merge


client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
```diff
@@ -1,38 +1,32 @@
...
-
-  componentDidMount() {
-    this.setOverrides(this.props);
-  }
+  const { schemaUrl, actions } = props;
 
-  componentWillUnmount() {
-    this.setOverrides();
-  }
+  const schemaUrlRef = useRef(schemaUrl);
```
emteknetnz: This ref is being used so that when the useEffect() replacement for the componentWillUnmount() is called, it ensure that the `schemaUrlRef.current` is fresh rather than stale. There is `useEffect()` block with a `[schemaUrl] ` dependency tasked with keeping this up to date.

Previously with the class component this wasn't as issue as `this.props.schemaUrl` was always fresh

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
```diff
@@ -1,38 +1,32 @@
...
   /**
    * Compares the current properties with received properties and determines if overrides need to be
    * cleared or added.
    *
-   * @param {object} props
+   * @param {object} propsParam
    */
-  setOverrides(props = null) {
-    if (!props) {
+  const setOverrides = (propsParam = null) => {
```
emteknetnz: Renamed param was so as not to clash with props

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
```diff
@@ -1,38 +1,32 @@
...
 import PropTypes from 'prop-types';
 import { bindActionCreators } from 'redux';
 import * as schemaActions from 'state/schema/SchemaActions';
 import { connect } from 'react-redux';
 
-class FileSchemaHandler extends Component {
-  constructor(props) {
-    super(props);
+const FileSchemaHandler = (_props) => {
+  const {
```
emteknetnz, line 8: Doing all this destructing with `_props` and `props` so that we can render `<TargetComponent setOverrides={setOverrides} {...props} />;` later on

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -1,400 +1,374 @@
-import React, { Component } from 'react';
+import React, { useState, useRef, useMemo, useEffect } from 'react';
```
emteknetnz: This component was has had a pretty big refactor due it having redux-form (`<BaseFormComponent>`) as a child component. redux-form seems like it relies on strict reference equality for props it receives and passes to its own child fields. This seems to include functions passed in like `buildComponent`

In a functional component, functions are recreated when their dependencies change or every render if it has no dependency array. 

Finding this solution was hard, though after a lot of experimentation the following seems to work:

- Create a latestRef hook that holds the latest props and context. This allows the callbacks to read fresh data without needing to be recreated (which would change their reference identity)

- methods object: wraps several functions in a useMemo with an empty dependency array so that those methods are never recreated. This mimics Class Component methods which are only created once and never change

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -1,232 +1,274 @@
...
+/**
+ * If there is structural and state data available merge those data for each field.
+ * Otherwise just use the structural data. Ensure that keys don't conflict
+ * with redux-form expectations.
+ *
+ * @param {array} fields
+ * @param {Object} state Optional
+ * @return {array}
+ */
+const normalizeFields = (fields, state) => fields.map((field) => {
```
emteknetnz, line 18: This method doesn't need to be stable, though it's called by one of the stable methods and it seems a little cleaner to just have it fully outside the FormBuilder component rather than above the stable methods in the useEffect block i.e. in the middle of all the other hook definitions

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
```diff
@@ -1,39 +1,43 @@
...
   /**
    * Compares the current properties with received properties and determines if overrides need to be
    * cleared or added.
    *
-   * @param {object} props
+   * @param {object} propsParam
    */
-  setOverrides(props = null) {
-    if (!props) {
+  const setOverrides = useCallback((propsParam = null) => {
```
emteknetnz, line 23: Wrapping in a callback to ensure this method is not recreated on render. This was done as part of a fix after behat tests for htmleditor-tinymce as part of a kitchen-sink CI fail

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
```diff
@@ -1,39 +1,43 @@
...
+    ...props
+  } = _props;
+  const {
+    schemaUrl,
+    actions
+  } = props;
+
+  // Keep track of the latest values in refs to avoid stale closures
+  const schemaUrlRef = useRef(schemaUrl);
+  const actionsRef = useRef(actions);
```
emteknetnz: Also stablising the redux actions. This was done as part of a fix after behat tests for htmleditor-tinymce as part of a kitchen-sink CI fail

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -1,232 +1,274 @@
...
+  // without needing to be recreated.
+  const latestRef = useRef({ props, context });
+
+  // Update immediately for the current render cycle
+  latestRef.current = { props, context };
+
+  // Update the ref on every render so methods can access fresh data
+  useEffect(() => {
+    latestRef.current = { props, context };
+  });
```
ChloeSartorelli: I don't think we need both the effect (runs after every render) _and_ then "update immediately" (runs on/before every render)

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -1,232 +1,274 @@
...
+
+  const [submittingAction, setSubmittingAction] = useState(null);
+  const formDOMRef = useRef(null);
+  const schemaStructure = schema.schema;
+
+  // Stable API fetcher
+  const submitApi = useRef(backend.createEndpointFetcher({
+    url: schemaStructure.attributes.action,
+    method: schemaStructure.attributes.method,
+  })).current;
```
ChloeSartorelli:
```suggestion
  // Stable API fetcher
  const submitApi = useMemo(() => backend.createEndpointFetcher({
    url: schemaStructure.attributes.action,
    method: schemaStructure.attributes.method,
  }), [schemaStructure.attributes]);
```

This should be a `useMemo` with `schemaStructure.attributes.action` and `schemaStructure.attributes.method` (or just `schemaStructure.attributes`) as dependencies

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -1,232 +1,274 @@
...
+
+  // Create "stable methods" for redux-form that have a stable identity
+  // useMemo and empty dependency array [] is used so that these functions are only created once
+  // This is done to ensure these methods have a stable identify so that redux-form (<BaseFormComponent>)
+  // does cause re-renders when these methods change.
+  const methods = useMemo(() => {
```
ChloeSartorelli: Feels like we should wrap each of the individual functions in `useCallback()` instead - avoiding recreating a function is what `useCallback()` was made for.

Wrapping each one individually also avoids the possibility of someone coming in and dumping a new function in the middle of the memo that isn't meant to be memoised.

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -1,232 +1,274 @@
...
+    // Internal helper to access current scope
+    const getScope = () => {
```
ChloeSartorelli: What do you mean by "scope"? This isn't a concept that was in the component before.

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -236,165 +278,158 @@ class FormBuilder extends Component {
    * (originally retrieved through schemaFieldValues())
    * @return {Promise|null}
    */
-  handleSubmit(data) {
-    // Add form action data (or default to first action, same as browser behaviour)
```
ChloeSartorelli, line 240: Why was this comment removed? In particular "or default to first action, same as browser behaviour" seems like important context to retain.

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -236,165 +278,158 @@ class FormBuilder extends Component {
...
+    if (typeof validate === 'function') {
+      return validate(valuesParam);
+    }
 
-      return data;
-    });
-  }
+    // Use fresh props/context
+    const { schema: currentSchema } = latestRef.current.props;
+    const { injector } = latestRef.current.context;
```
ChloeSartorelli: This isn't one of the memoised functions, so won't it _already_ be referencing the current props/context without relying on this ref?

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -236,165 +278,158 @@ class FormBuilder extends Component {
...
-    const state = this.props.schema.state;
-    const BaseFormComponent = this.props.baseFormComponent;
-
-    // Map form schema to React component attribute names,
-    // which requires renaming some of them (by unsetting the original keys)
-    const attributes = {
-      ...schema.attributes,
-      className: schema.attributes.class,
-      encType: schema.attributes.enctype,
-      // Turn off HTML5 validation to rely on validateForm as the sole validator
```
ChloeSartorelli, line 341: Why is this comment removed?

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -236,165 +278,158 @@ class FormBuilder extends Component {
...
+
+    const validationMiddleware = injector.validate(identifier);
 
-  render() {
-    const schema = this.props.schema.schema;
-    const state = this.props.schema.state;
-    const BaseFormComponent = this.props.baseFormComponent;
-
-    // Map form schema to React component attribute names,
-    // which requires renaming some of them (by unsetting the original keys)
```
ChloeSartorelli, line 336: Why is this comment removed?

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -236,165 +278,158 @@ class FormBuilder extends Component {
...
-    );
-  }
-}
+    delete attrs.class;
+    delete attrs.enctype;
+    return attrs;
+  }, [schemaData.attributes]);
+
+  const formProps = {
+    form,
```
ChloeSartorelli:
```suggestion
    form, // required as redux-form identifier
```
This comment from the original component code is useful context

client/src/components/FormBuilder/FormBuilder.js
```diff
@@ -236,165 +278,158 @@ class FormBuilder extends Component {
...
+    actionHolder,
+    attributes,
+    data: schemaData.data,
+    initialValues,
+    onSubmit: handleSubmit,
+    valid: schemaState && schemaState.valid,
+    messages: (schemaState && Array.isArray(schemaState.messages)) ? schemaState.messages : [],
+    mapActionsToComponents,
+    // Uses the stable method
+    mapFieldsToComponents: methods.mapFieldsToComponents,
```
ChloeSartorelli: Why does `mapFieldsToComponents` use a "stable" method but `mapActionsToComponents` doesn't?

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
```diff
@@ -45,16 +49,19 @@ class FileSchemaHandler extends Component {
       };
       // set overrides into redux store, so that it can be accessed by FormBuilder with the same
       // schemaUrl.
-      this.props.actions.schema.setSchemaStateOverrides(props.schemaUrl, overrides);
+      actionsRef.current.schema.setSchemaStateOverrides(propsParam.schemaUrl, overrides);
     }
-  }
+  }, []);
```
ChloeSartorelli: Did you try adding `schemaUrl` and `actions` in as dependencies before resorting to refs?

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
```diff
@@ -1,39 +1,43 @@
...
+  const {
+    schemaUrl,
+    actions
+  } = props;
+
+  // Keep track of the latest values in refs to avoid stale closures
+  const schemaUrlRef = useRef(schemaUrl);
+  const actionsRef = useRef(actions);
+
+  // Update refs synchronously (during render) or inside an effect.
```
ChloeSartorelli:
```suggestion
  // Update refs synchronously (during render).
```
It's not inside an effect.

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
```diff
@@ -1,39 +1,43 @@
...
+  const {
+    Component: TargetComponent,
+    ...props
+  } = _props;
+  const {
+    schemaUrl,
+    actions
+  } = props;
+
+  // Keep track of the latest values in refs to avoid stale closures
```
ChloeSartorelli: Please add something here about why the closure is relying on refs instead of a dependency array, since it's likely that seeing this comment in the future it will be tempting to go "well why don't we just use the dependency array instead?"

GuySartorelli (changes requested):

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
ChloeSartorelli: > There is useEffect() block with a [schemaUrl] dependency tasked with keeping this up to date.

No there isn't - but presumably assigning the `.current` value each render does the same thing without relying on effects so no need for a change.

client/src/components/FormBuilder/FormBuilder.js
emteknetnz: "Inner scope" for the "stable methods". Separate from the "outer scope"

The class component didn't have this concept because props and methods are constant / stable (created once) whereas in functional components they can get stale.

client/src/components/FormBuilder/FormBuilder.js
emteknetnz: Added back in

client/src/components/FormBuilder/FormBuilder.js
emteknetnz: Added back in

client/src/components/FormBuilder/FormBuilder.js
emteknetnz: Added back in

client/src/components/FormBuilder/FormBuilder.js
emteknetnz: Added back

client/src/components/FormBuilder/FormBuilder.js
emteknetnz: mapActionsToComponents calls the stable method methods.buildComponent

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
emteknetnz: Have rewritten comment for this block to be more descriptive

emteknetnz:
@GuySartorelli I've updated some comments, though as mentioned offline, this thing is fragile due to the difference in how class and functional components deal with references. I got to where I go to largely based on lots of trial and error and I really don't want to have change any code just to make things a tiny bit tidier at this stage.

For any comments where I haven't replied, assume my response is "it currently works, though it's fragile and I don't want change it". Feel free to experiment with tidying up the code yourself, though please:
- ensure that any tests pass (excluding existing failures)
- kitchen-sink passes
- asset-admin, elemental inline editing and versioned-admin all pass a smoke test

client/src/components/FormBuilder/FormBuilder.js
GuySartorelli:
```suggestion
    mapActionsToComponents,
    // Uses the stable method methods.buildComponent
    mapFieldsToComponents: methods.mapFieldsToComponents,
```

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
GuySartorelli: Note to self: Part of test 3

GuySartorelli:
I have tested the suggested changes in https://github.com/creative-commoners/silverstripe-admin/tree/pulls/3/refactor-test3

These passed a [kitchen sink CI run](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/21229385443).

You can see the diff between that branch and this PR at https://github.com/creative-commoners/silverstripe-admin/compare/pulls/3/refactor-yet-more...pulls/3/refactor-test3

Please apply these changes unless you have specific objections to any of them.

client/src/components/FormBuilder/FormBuilder.js
ChloeSartorelli: See https://github.com/silverstripe/silverstripe-admin/pull/2096#issuecomment-3781833272

client/src/components/FormBuilder/FormBuilder.js
ChloeSartorelli: See https://github.com/silverstripe/silverstripe-admin/pull/2096#issuecomment-3781833272

client/src/components/FormBuilder/FormBuilder.js
ChloeSartorelli: See https://github.com/silverstripe/silverstripe-admin/pull/2096#issuecomment-3781833272

client/src/components/FormBuilder/FormBuilder.js
ChloeSartorelli: No longer needed, see https://github.com/silverstripe/silverstripe-admin/pull/2096#issuecomment-3781833272

client/src/components/FormBuilder/FormBuilder.js
ChloeSartorelli: See https://github.com/silverstripe/silverstripe-admin/pull/2096#issuecomment-3781833272

client/src/components/FormBuilder/FormBuilder.js
ChloeSartorelli: Comment no longer needed, see https://github.com/silverstripe/silverstripe-admin/pull/2096#issuecomment-3781833272

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
ChloeSartorelli: See https://github.com/silverstripe/silverstripe-admin/pull/2096#issuecomment-3781833272

client/src/containers/InsertLinkModal/fileSchemaModalHandler.js
ChloeSartorelli: No longer needed, see https://github.com/silverstripe/silverstripe-admin/pull/2096#issuecomment-3781833272

emteknetnz:
Cool your addition looks good, I've added your PR to this PR as a separate commit (so don't squash merge)

I've updated the kitchen sink and it's all green. Have smoked tests in elemental, asset-admin, versioned-admin all seems fine

GuySartorelli (approved):
LGTM

Issue #2100, opened 2026-01-22
Convert FormBuilderLoader to functional component


emteknetnz:
## Acceptance criteria
- [ ] Component is converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/21271207803
Failing elemental behat test is existing

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2101

GuySartorelli:
PR merged

PR #2101, opened 2026-01-23
ENH Refactor react components from class to functional

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2100


client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -74,11 +73,57 @@ class FormBuilderLoader extends Component {
...
+  const getIdentifier = (propsParam = propsRef.current) => createFormIdentifierFromProps(propsParam);
+
+  /**
+   * Given a submitted schema, ensure that any errors property is merged safely into
+   * the state.
+   *
+   * @param {Object} schema - New schema result
+   * @return {Object}
+   */
+  const reduceSchemaErrors = (schema) => {
```
emteknetnz, line 87: Moved above handleSubmit which calls this

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -218,49 +215,85 @@ class FormBuilderLoader extends Component {
...
+          })
+      );
+    });
+
+  /**
+   * Convert error to a json object to pass to onLoadingError
+   *
+   * @param {Object} error
+   */
+  const normaliseError = (error) => {
```
emteknetnz, line 265: Moved to be above fetchSchema which calls this

GuySartorelli:
A general reminder to please point out all changes that you make in these refactoring PRs that aren't just converting lifecycle events or functions to arrow format. For example please point out renamed variables or functions for these PRs in future, as you have in the past.

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -404,32 +399,47 @@ class FormBuilderLoader extends Component {
...
-      return <Loading containerClass="loading--form flexbox-area-grow" />;
+  }, []);
+
+  useEffect(() => {
+    if (
+      prevPropsRef.current.schemaUrl !== schemaUrl
+      || prevPropsRef.current.refetchSchemaCriteria !== refetchSchemaCriteria
+    ) {
+      fetchSchema();
     }
```
ChloeSartorelli: This condition isn't needed - it will always be true in this effect because those are the only two variables in the dependency array.

ChloeSartorelli: Looks like that was the only reason for the `prevPropsRef` so that can be removed as well.

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -27,42 +27,41 @@ import getFormState from 'lib/getFormState';
...
+const FormBuilderLoader = (props) => {
+  const {
+    autoFocus,
+    loading,
+    loadingComponent: LoadingComponent,
+    onSubmit,
+    onSubmitSuccess,
+    refetchSchemaCriteria,
+    refetchSchemaOnMount = true,
+    schema: schemaProp,
```
ChloeSartorelli, line 44: Why did this prop get renamed? In other refactors I think you only renamed a prop if it conflicted with state. IMO it's best if props retain their name, and other variables (e.g. function parameters) get renamed instead - but if there's a good reason for it please just state the reason and leave as is.

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -27,42 +27,41 @@ import getFormState from 'lib/getFormState';
...
+
+  const [didError, setDidError] = useState(false);
+  const prevPropsRef = useRef({ schemaUrl, refetchSchemaCriteria });
+
+  // Persist the latest props to simulate class component `this.props` behavior and prevent stale closures.
+  // This ensures asynchronous callbacks (e.g. fetchSchema, handleSubmit) always access
+  // the latest props, even if the component re-renders while the request is pending.
+  const propsRef = useRef(props);
+  // Update current props on each render before running any effects or callbacks
+  propsRef.current = props;
```
ChloeSartorelli, line 56: This wasn't there in the original - I presume there was some failure when trying to convert without this? What was the failure?

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
```diff
@@ -404,32 +399,47 @@ class FormBuilderLoader extends Component {
...
-}
+
+  const builderProps = Object.assign({}, props, {
+    refetchSchemaOnMount,
+    form: getIdentifier(),
+    onSubmitSuccess,
+    onSubmit: handleSubmit,
+    onAutofill: handleAutofill,
+    autoFocus
+  });
```
ChloeSartorelli, line 442: `refetchSchemaOnMount` wasn't here before, why is it being added?

GuySartorelli (changes requested):

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
emteknetnz: Putting all the props into a propsRef was done to solves the "stale closure" problem that I experience a lot with the FormBuilder refactorig. This will emulate this.props in class components which is mutable, unlike functional components where the props param is immutable. This means that things can end up stale and we have to have a somewhat convoluted approach of wrapping everything in useCallaback() and useMemo() to ensuring the dependency array is correct. This approach greatly simplifies things since we don't need to do that as propsRef.current is mutable like this.props was

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
emteknetnz: It's to make it act like componentDidUpdate() used to - useEffect() will always run on mount, no matter what's in the dependency array

I've added inline comments to explain this

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
emteknetnz: `schema` is used in several other places and function params and variables - we could ever rename all of those or just rename this, seems easier to just leave this one as is.

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
emteknetnz: builderProps is a combination of `props` (`this.props` on the class component) and those other things in the object e.g. autoFocus

The class component had 

```
FormBuilderLoader.defaultProps = {
  refetchSchemaOnMount: true,
};
```

We're now defining default value in the object destructuring:

```
const FormBuilderLoader = (props) => {
  const {
    autoFocus,
    loading,
    loadingComponent: LoadingComponent,
    onSubmit,
    onSubmitSuccess,
    refetchSchemaCriteria,
    refetchSchemaOnMount = true,
    schema: schemaProp,
    schemaUrl,
  } = props;
```

However it's possible that `refetchSchemaOnMount` isn't actually passed to the component, so we need to add it to this object to make it behave the same that the defaultProp used to

client/src/containers/FormBuilderLoader/FormBuilderLoader.js
ChloeSartorelli: I don't see the comments - looks like you didn't push?

emteknetnz: Opps forgot to push

GuySartorelli (approved):
LGTM

PR #2111, opened 2026-02-10
ENH Change tree ins tags to span

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-cms/issues/3122


client/src/legacy/tests/JSTreeToggleElement-test.js
```diff
@@ -0,0 +1,33 @@
+/* global jest, test, expect */
+
+import $ from 'jquery';
+
+global.jQuery = $;
+global.$ = $;
+require('../../../../thirdparty/jstree/jquery.jstree');
+
+test('jstree uses span toggle icons', (done) => {
```
ChloeSartorelli: I really don't think we need tests that check the markup. That seems excessive (a markup change would have to be intentional, so regressions are very unlikely) and annoying (any time we want to change the markup, we now have to change this test).

We don't usually have tests asserting the markup like this.

GuySartorelli (changes requested):
The actual markup looks correct, and I like that you've gone with class names instead of node types for selectors where appropriate.
I mostly skim-read the changes to jstree but they all look about right. 
Works as expected locally, except:

If I have nested pages, and I have enough pages that it needs to lazyload some of the pages, I end up with two arrows on the element whose children got lazyloaded:

Image

client/src/legacy/tests/JSTreeToggleElement-test.js
emteknetnz: Removed

emteknetnz:
> If I have nested pages, and I have enough pages that it needs to lazyload some of the pages, I end up with two arrows on the element whose children got lazyloaded:

Fixed - issue was whitespace in the templates which jstree is sensitive to :laughing:

GuySartorelli (approved):

PR #2112, opened 2026-02-10
Make admin JS resilient to missing base tag

sminnee:
If the base_tag isn't present in LeftAndMain.ss, there are a number of things that break. This PR fixes those, while also continuing to work if a base tag is provided.

## Summary
- Adds `getBaseUrl()` helper in `lib/urls.js` that falls back to `window.location.href` when no `<base>` tag is present
- Replaces all `$('base').attr('href')` calls with `getBaseUrl()` across GridField, LeftAndMain, LeftAndMain.Content, and LeftAndMain.Menu
- Fixes `AdminRootController::admin_url()` to include `Director::baseURL()` prefix
- Amend tests to expect URLs that start with "/admin" instead of "admin"
- Recreate site-relative URLs for the react router props & DependentCompositeForm

Relates to silverstripe/silverstripe-framework#11961 and silverstripe/silverstripe-framework#11964

## Notes

There were a few places in the code that broke because admin_url() started returning "/admin" instead of "admin". It does beg the question of whether other modules' code will also break.

That said, the places we had to fix this were, for the most part, in our "base classes" that other modules/projects should be extending from.

We may need to note it in the 6.2 (admin 3.2) release notes though?

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11961


michalkleiner:
@sminnee just a thing that came to mind re: this work - do we still support installation in a subfolder, and how/if it would impact this work?

sminnee:
> @sminnee just a thing that came to mind re: this work - do we still support installation in a subfolder, and how/if it would impact this work?

Good question! But, yes.

Director::baseUrl() is prefixed to all URLs, so the URL might be written as `/mysite/admin/assets` instead. This is why this patch needs some Director::makeRelative() calls, which turns the URL back into "admin/assets"

This approach of using "root-relative URLs" by default is already present elsewhere, e.g. form actions on front-end forms and theme resource insertion.

client/src/legacy/LeftAndMain.Menu.js
```diff
@@ -208,7 +208,7 @@ $.entwine('ss', function($){
       var item = this.getMenuItem();
 
       var url = this.attr('href');
-      if(!isExternal) url = joinUrlPaths($('base').attr('href'), url);
+      if(!isExternal) url = $.path.makeUrlAbsolute(url, getBaseUrl());
```
ChloeSartorelli: What is the difference between `$.path.makeUrlAbsolute` and `joinUrlPaths`?

client/src/legacy/LeftAndMain.js
```diff
@@ -63,7 +64,7 @@ window.ss.formatTabStateUrl = function(url) {
   return url
     .replace(/\?.*/, '')
     .replace(/#.*/, '')
-    .replace(new RegExp(`^${escapeRegExp($('base').attr('href'))}/?`), '');
+    .replace(new RegExp(`^${escapeRegExp(window.ss.config.absoluteBaseUrl)}/?`), '');
```
ChloeSartorelli, line 65: Why use this and not `getBaseUrl()` here?

client/src/lib/urls.js
```diff
@@ -1,3 +1,9 @@
+/**
+ * Get the base URL for resolving relative URLs.
+ * Uses the <base> tag if present, otherwise falls back to window.location.href.
+ */
+export const getBaseUrl = () => document.querySelector('base')?.getAttribute('href') || window.location.href;
```
ChloeSartorelli: When I'm on `https://example.com/admin/pages`, these give two different values:

|`$('base')?.attr('href')`|`window.location.href`|
|---|---|
|"https://pure-cms6-sink.ddev.site/"|"https://pure-cms6-sink.ddev.site/admin/pages"|

Should we be using `window.location.origin`? That gives the same value as `document.querySelector('base')?.getAttribute('href')`

code/AdminRootController.php
```diff
@@ -56,7 +56,7 @@ public static function get_admin_route(): string
      */
     public static function admin_url(string $action = ''): string
     {
-        return Controller::join_links(AdminRootController::get_admin_route(), $action);
+        return Controller::join_links(Director::baseURL(), AdminRootController::get_admin_route(), $action);
```
ChloeSartorelli: Why is this change needed?

Will this (or other changes) affect sites that have their app installed in a subdirectory (e.g. `www.example.com/silverstripe/` as the root of their Silverstripe app)?
Based on https://github.com/silverstripe/silverstripe-admin/pull/2112#issuecomment-3881606958 it looks like it'll add the subdirectory name where it wasn't before. Won't that cause problems for anyone who was using this method for their own purposes?

code/Forms/DependentCompositeField.php
```diff
@@ -429,9 +430,10 @@ private function getChildren(): array
         // For example the value for SearchableDropdownField will be a weird associative array
         // instead of a raw ID value.
         $form = $this->getForm();
+        // Make relative because HTTPRequest::setUrl() strips the base URL and leading slashes
```
ChloeSartorelli, line 433: I'm probably just missing something here, but I don't understand why stripping the base URL means we need to make the base URL relative.

code/AdminController.php
```diff
@@ -193,11 +194,11 @@ public function getClientConfig(): array
         $name = static::config()->get('section_name') ?: static::class;
         // Trim leading/trailing slash to make it easier to concatenate URL
         // and use in routing definitions.
-        $url = trim($this->Link(), '/');
+        $url = trim(Director::makeRelative($this->Link()), '/');
         $clientConfig = [
             'name' => $name,
             'url' => $url,
-            'reactRoutePath' => preg_replace('/^' . preg_quote(AdminRootController::admin_url(), '/') . '/', '', $url),
+            'reactRoutePath' => preg_replace('/^' . preg_quote(trim(Director::makeRelative(AdminRootController::admin_url()), '/'), '/') . '/', '', $url),
```
ChloeSartorelli, line 201: Are these `Director::makeRelative()` calls required because of the change to `AdminRootController::admin_url()`? That kinda feels like we're going one step forward, one step back.

GuySartorelli (commented):
Seems to work locally with a quick click around and a few CRUD actions - I'll want to throw the kitchen sink at it before merging, but a few things stuck out when looking at the code so I'll ask those first, and throw the sink at it once I understand the below a little better.

client/src/legacy/LeftAndMain.Menu.js
sminnee: joinUrlPaths just joins segments. makeUrlAbsolute is aware of host-relative links (i.e. start with /)

so both `joinUrlPaths("http://example.com/one/two", "/three/four")` and `joinUrlPaths("http://example.com/one/two", "three/four")` return `"http://example.com/one/two/three/four"`.

makeUrlAbsolute does what it says on the tin: converts any relative URL to an absolute on, using the 2nd argument as the current ULR.

so 
- `makeUrlAbsolute("/three/four", "http://example.com/one/two")` returns `"http://example.com/three/four"`
- `makeUrlAbsolute("three/four", "http://example.com/one/two")` returns `"http://example.com/one/two/three/four"`

if you only ever use relative URLs, they return the same result, which is why joinUrlPaths used to work.

client/src/legacy/LeftAndMain.js
sminnee: absoluteBaseUrl is returning the site-root. this is returned by getBaseUrl() only if there is a base tag present.

client/src/lib/urls.js
sminnee: getBaseUrl() is the URL for resulting relative URLs. This is *not* the origin if there isn't a base tag present. The point of the code is to make scripts operate the same  as `<a href="something">` to avoid weird errors - you don't want onClick handler and right-click, "open in new tab" to go to different places.

code/AdminController.php
sminnee: As mentioned above, we've taken a step towards consistency with the rest of the system. These changes mean that react continues to operate on root-relative URLs. I'm not sure the react-router code could be changed to avoid this — the hard part would be getting it to work on silverstripe sites hosted in a subdirectory — but I opted leave any such changes out.

code/Forms/DependentCompositeField.php
sminnee: The value of getUrl() will be "admin/assets" not "/admin/assets". So we need to convert "/admin/assets" back to "admin/assets" for comparison, which is what makeRelative does.

code/AdminRootController.php
sminnee: In most cases, baseURL returns "/". In the example you describe it returns "/silverstripe/". but it's needed in both cases.

In the absence of a base tag, if i'm on http://example.com/admin/pages, I need to link to /admin/assets.

Other link generation, notably SiteTree:Link() does this; it's the pattern that link generation in the system should follow, we just missed it here. I tried to find where we added this to SiteTree, but it was more than 18 years ago and the commit was lost in svn.

sminnee:
OK @GuySartorelli I responded point by point, hopefully it makes more sense!

client/src/lib/urls.js
ChloeSartorelli: > getBaseUrl() is the URL for resulting relative URLs. This is not the origin if there isn't a base tag present.

Ahhh okay. I had to sit with that for a bit to internalise it - I think the word "base" is what's tripping me up here, since we're removing the `<base>` tag which has a slightly different meaning of "base" than what this function has.

Looks like we should be able to just use `document.baseURI` instead of adding this new function.

code/AdminRootController.php
ChloeSartorelli: Regardless of how _we're_ using this method, I don't think we should be changing what this method is outputting, because as a public method anyone can be calling it.

Changing this now could break code that expects it to output how it's currently outputting (which we already notice, since we need to adjust for it in https://github.com/silverstripe/silverstripe-admin/pull/2112#discussion_r2796117347) - i.e. a value that's relative to the website root URL.

I think we should leave this method alone, and instead if we need to adjust for it downstream, lets do that.

Alternatively we could change the output here based on the value of the `SSViewer.enable_base_tag` config, though that feels like weird coupling and would then require anyone using this method to adjust how they consume it based on that configuration, which doesn't sound fun.

ChloeSartorelli: One use case I can imagine is someone linking to the CMS from the frontend - they might consume `AdminRootController::admin_url()` to do that. Changing the output could break those links.

GuySartorelli (dismissed):
Thanks for the explanations.
Just https://github.com/silverstripe/silverstripe-admin/pull/2112#discussion_r2810011386 and https://github.com/silverstripe/silverstripe-admin/pull/2112#discussion_r2810014286 to address.

Looks like this also needs to be rebased to resolve a merge conflict.

client/src/lib/urls.js
sminnee: Cool - I'll replace with references to document.baseURI.

code/AdminRootController.php
sminnee: So, if someone has make a link to the CMS, disabled the base tag, and we *don't* change the output of admin_url, it's more likely to break things. This change is being made on the assumption that this is the method that everyone uses to generate a URL to admin areas.

In particular, CMS plugins generating URLs to the different modules will need to have "admin/mymodule" changed to "/admin/mymodule"

One approach would be to change the output admin_url only if base tags have been disabled? It's a bit messy but it would limit the scope of this change to only those sites where the base tag was disabled.

sminnee:
Cool, so:

 * [x] Replace getBaseUrl() with references to document.baseURI.
 * [x] Change admin_url to only prefix with Director::baseUrl() if base tags have been disabled. (**pending Guy's agreement**)
 * [x] Rebase & rebuild client/dist/

code/AdminRootController.php
ChloeSartorelli: If I already don't use a `<base>` tag in my front-end template and I'm using this method to generate a URL to the CMS (perhaps massaging it a bit to make it a valid link, if that's needed), and then we merge this PR, won't that break my link?
If yes, then we shouldn't make this change as-is.

> One approach would be to change the output admin_url only if base tags have been disabled? It's a bit messy but it would limit the scope of this change to only those sites where the base tag was disabled.

That's what I meant by changing it based on the value of the `SSViewer.enable_base_tag` config. Though I think in the scenario I mentioned above where someone just outright isn't using `<base>` tags already, they'll still end up having to fix their links after the change because that config won't affect them - but the change in output will.

That said, making it conditional on `SSViewer.enable_base_tag` being set `false` does at least mean they have to explicitly set config before their links break, which is better than the links breaking on `composer update`. So that does seem better than the current change IMO.

If you do that please add a comment next to the conditional logic that says something like `// this condition is for backwards compatibility` - that way we'll find it when doing CMS 7 changes (we have a process to look for a bunch of things including "backwards compatibility" for things that need tidying up) and can remove the condition.

sminnee: To put it another way: the docblock says "Returns the root admin URL for the site with trailing slash". If I don't make this change, that is no longer true. The absence of base tags requires the baseURL prefix in order to be a valid URL.

The presence of base tags also allows for this, which is why default Link() implementations include baseURL() - they are the flexible result to provide.

sminnee:
OK @GuySartorelli the code changes are in a separate commit here c68a870ab0d948a84ffba8bb74ee068a628d967d but we probably want to squash that in before merge, right?

GuySartorelli:
I can squash at merge time (when I remember to lol) so while it's nice to have commits squashed before I merge it's not strictly necessary.

GuySartorelli (commented):
Thanks for your work and patience on this.
As I mentioned earlier I'll throw the kitchen sink at it before merging, but for now there's no further changes to make to this or the framework PR.

sminnee:
All good, thanks for your work on the opposite side of the bench!

GuySartorelli:
Looks like there's some [unit tests failing in kitchen sink](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/22047543505/job/63700886713) with this PR which don't fail [without this PR](https://github.com/silverstripe/recipe-kitchen-sink/actions/runs/22047927641/job/63700094402) - can you please look into that? (ignore the behat failures, since they're happening without this PR too)

The only thing off the top of my head that might be causing most of them would be an extension, e.g. if fluent has a `updateLink()` extension hook applied to `LeftAndMain` or something along those lines.

sminnee:
Right so i think those changes were a consequence of us changing the AdminRootController rules. I had update the test to use "/admin" instead of "admin", but when I changed that behaviour, I forgot to revert changes to my test. That's done now.

sminnee:
To get this passing outside of kitchen sink, we'll need to merge the framework PR and re-run this PR's tests.

GuySartorelli:
The [js CI run](https://github.com/silverstripe/silverstripe-admin/actions/runs/22051358070/job/63710071257?pr=2112) is showing `vendor.js` isn't correct - can you please try running `yarn build`?

I'll rerun kitchen sink in the meantime.

GuySartorelli:
Kitchen sink shows the unit test failures will go green: https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/22082727246

So just the js CI run to resolve.

sminnee:
Sorry about that! It's passing now.

GuySartorelli (approved):

PR #2114, opened 2026-02-12
FIX Ensure site tree expands to show current child

ChloeSartorelli:
There are various bugs related to expanding the site tree when the current item isn't loaded into the tree, which whis PR should fix.

## Fixed scenarios

- If the currently loaded page is one level deeper in the tree than what is currently loaded, the node will be loaded from the server and added to the tree but the tree won't expand.
This PR fixes it by expanding the relevant parent after the child node is added to the tree.
- If the currently loaded page is at least two levels deeper in the tree than what is currently loaded, `self.jstree('load_node', -1);` is called but results in an error (the main problem the linked issue is about)
This PR fixes it by reinstating the mandatory callback passed to `load_node` which was removed in https://github.com/silverstripe/silverstripe-admin/pull/932

## Remaining scenarios

- If the currently loaded page is at least **_three_** levels deeper in the tree than what is currently loaded, `self.jstree('load_node', -1, callback);` is called but nothing happens. Looks like that's not recursive.

## Issue

- https://github.com/silverstripe/silverstripe-cms/issues/3144


GuySartorelli:
I've spent way too long trying and failing to get jstree to recursively load and render items in the tree when you're viewing a page that's too deeply nested to load directly. I'm giving up on the remaining scenario for now, at least we can see an improvement with the first two scenarios.

I'll add some behat test for the two scenarios that are being improved - we'll just need to keep the issue open after merging this PR.

GuySartorelli:
Actually nevermind, I'm ditching this PR because it still has the problem where the page is often duplicated when opening the site tree.

Issue #2117, opened 2026-02-18
Convert admin components that will retain legacy support


emteknetnz:
There are react class component(s) that needs to be converted to react functional component(s).

- LiteralField

Only one in this PR because it's a demonstration of the keeping legacy support for a subset of admin form field components where the old class components in retained separately. I want to ensure this way of retaining support is agreed upon before doing the rest.

Note there are more components than this to convert, limiting number converted at a time to make peer review more manageable.

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved
- [ ] Legacy class components are retained (if applicable i.e. is admin form field component)

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2118
- https://github.com/silverstripe/webpack-config/pull/90
- https://github.com/silverstripe/developer-docs/pull/882

GuySartorelli:
@emteknetnz Might be better to do one like `TextField` that we know gets extended

emteknetnz:
I choose LiteralField simply because it's such a simple component and nothing in core extends it (unlike TextField), that it meant the real focus of this PR could be how to do the webpack-config, changelog, etc, and basically not worry about the functional component refactor

GuySartorelli:
Okay. We'll see where we get with it. It may be difficult to judge it without actually seeing how it affects components that are already extending the existing component though.

emteknetnz:
I've got test code in https://github.com/silverstripe/silverstripe-admin/pull/2118 which I used to validate that it seems to work.

ChloeSartorelli:
PRs merged

PR #2118, opened 2026-02-18
ENH Convert LiteralField to functional component with legacy support

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2117

I've tested the legacy component works in conjuction with the the webpack-config PR by reaplce the code BackButton.js in asset-admin with this:

```js
import LegacyLiteralField from 'legacy/react-components/LegacyLiteralField';

class BackButton extends LegacyLiteralField {
  getContent() {
    return { __html: 'i am a cool back button' };
  }
}

export { BackButton as Component };

export default BackButton;
```

And manually update asset-admins node_modules webpack-config externals.js with what's in the webpack-config PR linked on this issue - works as expected


client/src/components/LiteralField/LiteralField.js
```diff
@@ -1,41 +1,42 @@
-import React, { Component } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
 
-class LiteralField extends Component {
+const LiteralField = ({
+  // React considers "undefined" as an uncontrolled component.
```
ChloeSartorelli: ~~I'm not quite following what this comment is trying to tell me. I assume you don't mean it thinks the value `undefined` is literally a component... is this trying to explain why `className` and `extraClass` have default values? Maybe "If there are no passed in props, React assumes your component is an uncontrolled component"?~~

~~Do you have any docs you can link to (for me, not necessarily in the comment itself) about this and what that actually affects? My understanding was that controlled vs uncontrolled components was more a concept rather than something that React actually cares about in its runtime code.~~

I wrote all that before I realised you were just keeping the existing comment. Do you know what the comment is trying to convey? If not I propose we remove it, since it doesn't really make any sense to me.

GuySartorelli (commented):
In terms of approach, this doesn't tell us a lot because we don't have an existing example of a class that extends `LiteralField`.

For example, using your `BackButton` code - how would we then convert that class in a way that avoids duplicating all the code that is currently inherited? Or would we just accept that the code is duplicated? How many classes have to extend something before we say "we don't want this duplicated anymore"?

emteknetnz:
> For example, using your BackButton code - how would we then convert that class in a way that avoids duplicating all the code that is currently inherited? Or would we just accept that the code is duplicated? How many classes have to extend something before we say "we don't want this duplicated anymore"?

Just accept that it's duplicated, otherwise it's spaghetti code. Keep in mind we virtually never update these fields (I cannot remember the last time we updated one of these), so risk of them getting "out of sync" is very low. Also we're deleting them all in CMS 7 so ongoing maintenance isn't really a concern IMO.

> How many classes have to extend something before we say "we don't want this duplicated anymore"?

Not sure that's relevant? We're converting all fields to functional components which don't really so things like DatetimeField (extends DateField, extends TextField, extends InputField) are likely all just going to have all their logic in the single field, maybe some shared helpers or some HOC thing or something, but functional components cannot subclass.

If you were referring to legacy components - they'll just keep extended the other legacy components i.e. LegacyDateField extends LegacyTextField

client/src/components/LiteralField/LiteralField.js
emteknetnz: Deleted

GuySartorelli:
Judging by the "we're deleting them all in CMS 7" comment I think I wasn't clear with what I meant - I'm not talking about the legacy class components since their inheritance chain is unbroken. I'm talking about the functional components, since they can't use inheritance after conversion, but do use inheritance before conversion.

Lets use the real example you pointed out - `DateField` is a class component that extends the `TextField` class component. When these are both converted into functional components, this inheritance gets broken. This means anything in `TextField` that is to be reused in `DateField` can't just be taken for granted anymore.

In our new functional component, how will `DateField` be able to reuse logic from `TextField`? Will we duplicate code between these two functional components? Or have a HoC that shares some logic between them? Or.... something else?

Ignore the legacy class components, I think what you're doing in that regard is fine. And ignore examples like `LiteralField` which don't currently exist in an inheritance chain. I'm only concerned about:
1. What do _we_ do about any logic we're currently inheriting, after we convert to functional components?
2. What do we suggest _other developers_ do if they want to avoid duplicating code when converting their own `extends TextField` class components into functional components?

> functional components [...] are likely all just going to have all their logic in the single field, maybe some shared helpers or some HOC thing or something

Right. It's figuring out what that will be that seems to be important here, in terms of validating the approach. This PR doesn't show that.

> Just accept that it's duplicated, otherwise it's spaghetti code.

That's certainly one option, but I'd like to see it in action so we can properly evaluate the approach - and I'd like to evaluate the approach before merging any PRs that convert form field components. Here's a few related inheritance chains that we'll need to consider, and an example of the amount of duplication we might expect:
- NumberField < InputField
- TinyMceHtmlEditorField < TextField < InputField
- DateTimeField < DateField < TextField < InputField
- TimeField < DateField < TextField < InputField

emteknetnz:
What we do with functional components in terms of preventing code duplication is kind of irrelevant to this PR. We'll work that out on the PRs that update things like DateField.

The purpose of this set of PRs is to demonstrate how we handle the legacy class components and agree on that, not how we handle functional components

GuySartorelli:
Okay, fair enough. This looks like a good approach for that narrow scope. I'll hold off merging this for now, pending PRs that show how we'll deal with the functional components of pre-existing class component extension chains as I'd like to know we have a good approach for that. Otherwise we risk merging this only to discover there are problems with converting the rest of them and having to revert.

Will merge when the rest of the approach has been demonstrated.

ChloeSartorelli:
I can't remember what specific PRs I was waiting for with my above comment, this is probably okay to be merged, just needs the conflict resolved.

emteknetnz:
Done

ChloeSartorelli (approved):
LGTM

Issue #2119, opened 2026-02-23
Convert more admin components that will retain legacy support


emteknetnz:
Convert various admin components that are related to TextField to functional components

There are react class component(s) that needs to be converted to react functional component(s).
- [ ] InputField.js
- [ ] NumberField.js (extends InputField)
- [ ] TextField.js (extends InputField)
- [ ] TinyMceHtmlEditorField.js (extends TextField)
- [ ] DateField.js (extends TextField)
- [ ] DatetimeField.js  (extends DateField)
- [ ] TimeField.js (extends DateField)

Note there are more components than this to convert, limiting number converted at a time to make peer review more manageable.

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved
- [ ] Legacy class components are retained (if applicable i.e. is admin form field component)

## Kitchen sink CI
https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/23471997515

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2120
- https://github.com/silverstripe/silverstripe-htmleditor-tinymce/pull/41
- https://github.com/silverstripe/webpack-config/pull/91
- https://github.com/silverstripe/developer-docs/pull/882
- https://github.com/silverstripe/webpack-config/pull/92

GuySartorelli:
I've linked the changelog PR here as it's a lot more applicable to these changes than to the change in https://github.com/silverstripe/silverstripe-admin/issues/2117

GuySartorelli:
PR merged

PR #2120, opened 2026-02-23
ENH Convert more admin components to functional

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2119

Solved the "functional components don't do inheritance" by exporting then things that were previously inherited and importing then on the functional component "subclasses" 

I think behat is failing because https://github.com/silverstripe/silverstripe-htmleditor-tinymce/pull/41 needs to be merged first

The recipe-kitchen-sink run linked with both PRs on the issue is green


client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,147 @@
...
+    type,
+    // `parse()` of redux-form `Field` should be used for parsing the
+    // localised input value to iso format to pass to redux store but `Field`
+    // is not accessible in this context.
+    value,
+    placeholder,
+  };
+};
+
+const DateField = (_props) => {
```
emteknetnz, line 123: A `_props` intermediate variable was introduced — defaults are applied via destructuring, but the fully-merged props still need to be passed through to shared InputField helpers and the wrapper render.

client/src/components/DateField/DateField.js
```diff
@@ -132,11 +152,29 @@ DateField.propTypes = {
...
-  modernizr,
-  data: {},
-};
-
 export { DateField as Component };
 
 export default fieldHolder(DateField);
+
+// Exported for use by DatetimeField, TimeField which override this method
+export { getInputProps };
```
emteknetnz: `getInputProps` (and related helpers) exported as named exports because `DatetimeField`/`TimeField` previously overrode these methods via class inheritance.

client/src/components/DatetimeField/DatetimeField.js
```diff
@@ -1,75 +1,127 @@
...
+    ...inputProps,
+    type,
+    value,
+    placeholder,
+  };
+};
+
+const render = (props, inputProps) => renderInputField(props, inputProps);
+
+const DatetimeField = (_props) => {
```
emteknetnz, line 83: A `_props` intermediate variable was introduced — defaults are applied via destructuring, but the fully-merged props still need to be passed through to shared InputField helpers and the wrapper render.

client/src/legacy/ReactComponents/tests/LegacyInputField-test.js
ChloeSartorelli, line 1: This should just be `InputField-test.js`.
Applies to a few other test files too.

client/src/legacy/ReactComponents/LegacyNumberField.js
ChloeSartorelli, line 1: Move the old test files in here too. For any that didn't used to have tests but do now, include those new tests here since they should be like-for-like anyway.

These won't be included in behat tests anymore so their jest tests are even more important than they used to be for validating they still work. Since we're keeping them around, they're supported for bug fixes etc.

client/src/bundles/bundle.js
```diff
@@ -48,7 +48,14 @@ import 'expose-loader?exposes=ListGroup!components/ListGroup/ListGroup';
 import 'expose-loader?exposes=ListGroupItem!components/ListGroup/ListGroupItem';
 import 'expose-loader?exposes=Loading!components/Loading/Loading';
 import 'expose-loader?exposes=CircularLoading!components/Loading/CircularLoading';
+import 'expose-loader?exposes=InputField!components/InputField/InputField';
 import 'expose-loader?exposes=TextField!components/TextField/TextField';
+import 'expose-loader?exposes=LegacyInputField!legacy/ReactComponents/LegacyInputField';
```
ChloeSartorelli, line 53: We probably don't need to expose the legacy class component here since we weren't already exposing it before, but I won't block merging this PR if you choose to keep it.

This applies to all class components we weren't previously exposing:

- `InputField`
- `NumberField`
- `TimeField`

While we do need to keep `LegacyInputField` for inheritance in our own class components, the others won't be extended by us _or_ by anyone else, so we can just ditch those legacy class components.

If you do ditch them, please update the changelog entry to make it clear that only the previously exposed form fields are being kept as legacy components.

client/src/components/InputField/InputField.js
```diff
@@ -7,86 +7,136 @@ import PropTypes from 'prop-types';
...
+    name,
+    disabled,
+    readOnly,
+    value: value || '',
+    placeholder,
+    autoFocus,
+    maxLength: data && data.maxlength,
+    type: type || null,
+    onBlur,
+    onFocus
```
ChloeSartorelli:
```suggestion
  const inputProps = {
    className: `${props.className} ${props.extraClass}`,
    id: props.id,
    name: props.name,
    disabled: props.disabled,
    readOnly: props.readOnly,
    value: props.value || '',
    placeholder: props.placeholder,
    autoFocus: props.autoFocus,
    maxLength: props.data && props.data.maxlength,
    type: props.type || null,
    onBlur: props.onBlur,
    onFocus: props.onFocus,
```

Deconstructing here is distracting - it's only really useful when the deconstructed variables are used in multiple places or if it's just a couple of variables that get references a few times, neither of which is the case here.

client/src/components/InputField/InputField.js
```diff
@@ -7,86 +7,136 @@ import PropTypes from 'prop-types';
...
+  };
+
+  if (attributes && !Array.isArray(attributes)) {
+    Object.assign(inputProps, attributes);
```
ChloeSartorelli:
```suggestion
  if (props.attributes && !Array.isArray(props.attributes)) {
    Object.assign(inputProps, props.attributes);
```

client/src/components/InputField/InputField.js
```diff
@@ -7,86 +7,136 @@ import PropTypes from 'prop-types';
...
-    }
-
-    if (!this.props.readOnly) {
-      Object.assign(props, {
-        onChange: this.handleChange,
-      });
-    }
-
-    return props;
+  if (!readOnly) {
```
ChloeSartorelli:
```suggestion
  if (!props.readOnly) {
```

client/src/components/InputField/InputField.js
```diff
@@ -7,86 +7,136 @@ import PropTypes from 'prop-types';
...
+    onBlur,
+    onChange,
+    onFocus,
+    placeholder,
+    readOnly,
+    tip,
+    title,
+    type,
+    value,
+  };
```
ChloeSartorelli:
```suggestion
const InputField = (_props) => {
  const defaultProps = {
    attributes = {},
    className = '',
    extraClass = '',
    type = 'text',
    value = '',
  };
  const props = {
    ...defaultProps,
    ..._props,
  };
```

Again deconstructing here isn't that useful - just means we have two long lists to look at.
The class component didn't cherry-pick props to pass into `getInputProps` or `render`, so we shouldn't do that here either.

client/src/components/InputField/InputField.js
```diff
@@ -101,17 +151,21 @@ InputField.propTypes = {
...
 export default InputField;
+
+// Exported for use by NumberField, TextField, DateField, DatetimeField, TimeField,
+// TinyMceHtmlEditorField which override this method
+export { getInputProps };
+// Exported for use by DateField which overrides this method
+export { handleChange };
+// Exported for use by DateField, TinyMceHtmlEditorField which override this method
+export { render };
+// Exported for use by TextField to retain propTypes inheritance
```
ChloeSartorelli:
```suggestion
// Exported for use other form fields which can override or reuse this logic
export { getInputProps };
export { handleChange };
export { render };
```
We don't need to keep a running list of which components use this logic, as that could easily end up getting stale.

client/src/components/InputField/InputField.js
ChloeSartorelli: TextField is a great example where we've avoided deconstructing props unnecessarily and it keeps things much shorter.

client/src/components/TextField/TextField.js
```diff
@@ -1,36 +1,66 @@
...
+  } = _props;
+
+  const props = {
+    ..._props,
+    attributes,
+    className,
+    extraClass,
+    type,
+    value,
+  };
```
ChloeSartorelli:
```suggestion
  const defaultProps = {
    attributes = {},
    className = '',
    extraClass = '',
    type = 'text',
    value = '',
  };

  const props = {
    ...defaultProps,
    ..._props,
  };
```
IMO this makes it a bit more obvious what's going on, and it a little more succinct.

Since these are identical to `InputField`'s default props we could also export the default props object from there and import it here so it's not re-declared, but I won't hold up merging if you choose not to do that.

client/src/components/TextField/TextField.js
```diff
@@ -1,36 +1,66 @@
...
+
+TextField.propTypes = inputFieldPropTypes;
 
 export { TextField as Component };
 
 export default fieldHolder(TextField);
+
+// Exported for use by DateField, DatetimeField, TimeField, TinyMceHtmlEditorField which override this method
+export { getInputProps };
+// Exported for use by DateField, DatetimeField, TimeField which override this method
```
ChloeSartorelli:
```suggestion
// Exported for use other form fields which can override or reuse this logic
export { getInputProps };
```

client/src/components/NumberField/NumberField.js
```diff
@@ -1,22 +1,45 @@
...
+    value = '',
+  } = _props;
+  const props = {
+    ..._props,
+    attributes,
+    className,
+    extraClass,
+    type,
+    value,
+  };
```
ChloeSartorelli:
```suggestion
  const defaultProps = {
    attributes = {},
    className = '',
    extraClass = '',
    type = 'text',
    value = '',
  };
  const props = {
    ...defaultProps,
    ..._props,
  };
```
Same comment as with TextField default props applies here.

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,147 @@
 import fieldHolder from 'components/FieldHolder/FieldHolder';
-import moment from 'moment';
+import momentLib from 'moment';
```
ChloeSartorelli, line 2: ~~Why did this change names?~~
Looks like this changed names to avoid conflicts with the declaration of a function with the same name

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,147 @@
...
 import modernizr from 'modernizr';
 import i18n from 'i18n';
 import PropTypes from 'prop-types';
-import { Component as TextField } from '../TextField/TextField';
+import { getInputProps as getInputFieldProps, render as renderInputField } from '../InputField/InputField';
```
ChloeSartorelli:
```suggestion
import { getInputProps as getInputFieldProps, render } from '../InputField/InputField';
```
And remove the declaration of the `render` function below, since it just passes the arguments directly through to this - it's just the same function.

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,147 @@
...
 
-  /**
-   * Check if this field has native html5 date support
-   *
-   * @return {Boolean}
-   */
-  hasNativeSupport() {
-    return this.props.modernizr.inputtypes.date;
-  }
+const momentWithProps = (props, ...args) => {
```
ChloeSartorelli:
```suggestion
const moment = (props, ...args) => {
```
You changed the name we're using for the import to avoid a conflict with this function - so we may as well keep the original function name.

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,147 @@
...
-      isoValue = enteredValue;
-    } else {
-      isoValue = this.convertToIso(enteredValue);
-    }
+  return isoDate;
+};
 
-    if (typeof this.props.onChange === 'function') {
-      this.triggerChange(event, isoValue);
+const convertToLocalised = (props, isoDate) => {
```
ChloeSartorelli, line 54: With some functions `props` has become the first argument, and with others its the last - can you please update these to either consistently always be first or consistently always be last? Unless there's a reason some need to be first but others need to be last.

It would be nice for example to be able to say in the changelog "for all of these new exported functions, props is the first argument"

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,147 @@
...
+/**
+ * Handles changes to the text field's value.
+ *
+ * @param {Event} event
+ */
+const handleChange = (event, props, {
+  asHTML5Fn = asHTML5,
+  convertToIsoFn = convertToIso,
+  triggerChangeFn = triggerChange,
+} = {}) => {
```
ChloeSartorelli: What is this `{/*...*/} = {}` syntax? Is that deconstructing from an empty object to give default values?

Why do we even need an object there, instead of just having regular default args?

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,147 @@
...
-    return localDate;
-  }
-}
+const getInputProps = (props) => {
+  const placeholder = i18n.inject(
+    i18n._t('Admin.FormatExample', 'Example: {format}'),
+    { format: momentWithProps(props).endOf('month').format(localFormat) }
+  );
+
+  const value = asHTML5(props)
```
ChloeSartorelli: In `handleChange()` we allow passing in a custom implementation of `asHTML5()` - why don't we do that here?

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,147 @@
...
+  const props = {
+    ..._props,
+    attributes,
+    className,
+    data,
+    extraClass,
+    modernizr: modernizrValue,
+    type,
+    value,
+  };
```
ChloeSartorelli:
```suggestion
  const defaultProps = {
    attributes = {},
    className = '',
    data = {},
    extraClass = '',
    modernizr,
    type = 'text',
    value = '',
  };

  const props = {
    ...defaultProps,
    ..._props,
  };
```
The same comment about default props from `TextField` applies here

client/src/components/DateField/DateField.js
```diff
@@ -132,11 +152,29 @@ DateField.propTypes = {
...
-DateField.defaultProps = {
-  modernizr,
-  data: {},
-};
-
 export { DateField as Component };
 
 export default fieldHolder(DateField);
+
+// Exported for use by DatetimeField, TimeField which override this method
```
ChloeSartorelli:
```suggestion
// Exported for use other form fields which can override or reuse this logic
```
And please remove the other comments below, following the same changes I requested for `InputField` and `TextField`

client/src/components/DatetimeField/DatetimeField.js
```diff
@@ -1,75 +1,127 @@
...
+import {
+  Component as DateField,
+  asHTML5 as asHTML5DateField,
+  getLang as getLangDateField,
+  getLocalisedValue as getLocalisedValueDateField,
+  handleChange as handleDateFieldChange,
+} from '../DateField/DateField';
+import { getInputProps as getInputFieldProps, render as renderInputField } from '../InputField/InputField';
+
+/* eslint-disable no-use-before-define */
```
ChloeSartorelli: Which lines is this for? We should avoid disabling linting rules for an entire file if we can

client/src/components/DatetimeField/DatetimeField.js
```diff
@@ -1,75 +1,127 @@
...
-    } else {
-      this.props.onChange(event, { id: this.props.id, value });
-    }
-  }
+const getLang = (props) => getLangDateField(props, hasNativeSupport);
+
+const momentWithProps = (props, ...args) => {
+  moment.locale(getLang(props));
+  return moment(...args);
+};
```
ChloeSartorelli: This is identical to the implementation in `DateField` - presumably it's here because `getLang` is a different implementation, but the way that has been handled elsewhere is to allow passing in the replacement function as an argument.

client/src/components/DatetimeField/DatetimeField.js
```diff
@@ -1,75 +1,127 @@
...
+  const props = {
+    ..._props,
+    attributes,
+    className,
+    data,
+    extraClass,
+    modernizr: modernizrValue,
+    type,
+    value,
+  };
```
ChloeSartorelli: Please apply the same change here as with all the other default props.

client/src/components/DatetimeField/DatetimeField.js
```diff
@@ -1,75 +1,127 @@
...
+};
+
+const getInputProps = (props) => {
+  const placeholder = i18n.inject(
+    i18n._t('Admin.FormatExample', 'Example: {format}'),
+    { format: momentWithProps(props).endOf('month').format(localFormat) }
+  );
+  const value = asHTML5(props)
+    ? props.value
+    : getLocalisedValue(props);
```
ChloeSartorelli, line 72: This condition seems to be new - what's the reason for adding it? Also doesn't look like it was part of the returned value originally.

client/src/components/TimeField/TimeField.js
```diff
@@ -3,54 +3,97 @@ import moment from 'moment';
...
-      { format: moment().endOf('month').format(localFormat) }
-    );
-    const type = this.asHTML5() ? 'time' : 'text';
-    return {
-      ...super.getInputProps(),
-      type,
-      placeholder,
-    };
-  }
+/* eslint-disable no-use-before-define */
```
ChloeSartorelli: As with datetimefield - what lines is this needed for? We should avoid disabling linting rules for entire files if we can.

client/src/components/TimeField/TimeField.js
```diff
@@ -3,54 +3,97 @@ import moment from 'moment';
...
+  const props = {
+    ..._props,
+    attributes,
+    className,
+    data,
+    extraClass,
+    modernizr: modernizrValue,
+    type,
+    value,
+  };
```
ChloeSartorelli: Please give this the same treatment as other default props

client/src/components/TimeField/TimeField.js
```diff
@@ -3,54 +3,97 @@ import moment from 'moment';
...
+};
+
+const getInputProps = (props) => {
+  const placeholder = i18n.inject(
+    i18n._t('Admin.FormatExample', 'Example: {format}'),
+    { format: moment().endOf('month').format(localFormat) }
+  );
+  const value = asHTML5(props)
+    ? props.value
+    : getLocalisedValue(props);
```
ChloeSartorelli, line 55: Presumably the reason for this is the same as the reason it's added in datetimefield, so this comment is just here to remind me the condition was added in both places so I keep that in mind when the response comes.

GuySartorelli (changes requested):

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,157 @@
...
 
-  getLocalisedValue() {
-    return this.convertToLocalised(this.props.value);
-  }
+  return isoDate;
+};
 
-  isMultiline() {
-    return false;
+const convertToLocalised = (isoDate, props) => {
```
emteknetnz: Moved `convertToLocalised` above `getLocalisedValue` so the helper is defined before the caller.

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,157 @@
...
 
-  /**
-   * Check if this field has native html5 date support
-   *
-   * @return {Boolean}
-   */
-  hasNativeSupport() {
-    return this.props.modernizr.inputtypes.date;
-  }
+const convertToIso = (localDate, props) => {
```
emteknetnz: Moved `convertToIso` above `handleChange` so the helper is defined before the caller.

client/src/components/DateField/DateField.js
```diff
@@ -1,127 +1,157 @@
...
-      this.triggerChange(event, isoValue);
-    }
+const isMultiline = () => false;
+
+/**
+ * Handles changes to the text field's value.
+ *
+ * @param {Event} event
+ */
+const handleChange = (event, ...args) => {
```
emteknetnz: Moved `handleChange` above `getInputProps` so the handler is defined before the caller.

client/src/components/TimeField/TimeField.js
```diff
@@ -3,54 +3,84 @@ import moment from 'moment';
...
-  }
-
-  convertToLocalised(isoTime) {
-    let localTime = '';
-    if (isoTime) {
-      const timeObject = moment(isoTime, 'HH:mm:ss');
-      if (timeObject.isValid()) {
-        localTime = timeObject.format(localFormat);
-      }
+const convertToLocalised = (isoTime) => {
```
emteknetnz: Moved `convertToLocalised` above `getLocalisedValue` so the helper is defined before the caller.

client/src/components/TimeField/TimeField.js
```diff
@@ -3,54 +3,84 @@ import moment from 'moment';
...
+};
 
-  convertToIso(localTime) {
-    let isoTime = '';
-    if (localTime) {
-      const timeObject = moment(localTime, localFormat);
-      if (timeObject.isValid()) {
-        isoTime = timeObject.format('HH:mm:ss');
-      }
+const convertToIso = (localTime) => {
```
emteknetnz: Moved `convertToIso` above `handleChange` so the helper is defined before the caller

client/src/components/TimeField/TimeField.js
```diff
@@ -3,54 +3,84 @@ import moment from 'moment';
...
     }
-    return isoTime;
   }
-}
+  return isoTime;
+};
+
+const getLocalisedValue = (props) => getLocalisedValueDateField(convertToLocalised, props);
+
+const handleChange = (event, props) => {
```
emteknetnz: Moved `handleChange` above `getInputProps` so the handler is defined before the caller.

client/src/legacy/ReactComponents/LegacyNumberField.js
emteknetnz: Have created legacy -test.js files where are just the old test files

client/src/bundles/bundle.js
emteknetnz: Done

emteknetnz: Removed

client/src/components/InputField/InputField.js
emteknetnz: Done

client/src/components/InputField/InputField.js
emteknetnz: Done

client/src/components/InputField/InputField.js
emteknetnz: Done

client/src/components/InputField/InputField.js
emteknetnz: Done

client/src/components/InputField/InputField.js
emteknetnz: Done

client/src/components/TextField/TextField.js
emteknetnz: Done

client/src/components/TextField/TextField.js
emteknetnz: Done

client/src/components/NumberField/NumberField.js
emteknetnz: Done

client/src/components/DateField/DateField.js
emteknetnz: Done

client/src/components/DateField/DateField.js
emteknetnz: Done

client/src/components/DateField/DateField.js
emteknetnz: Moved to first param

client/src/components/DateField/DateField.js
emteknetnz: Changed

client/src/components/DateField/DateField.js
emteknetnz: Can now pass is

client/src/components/DateField/DateField.js
emteknetnz: Done

client/src/components/DateField/DateField.js
emteknetnz: Done

client/src/components/DatetimeField/DatetimeField.js
emteknetnz: Removed

client/src/components/DatetimeField/DatetimeField.js
emteknetnz: now importing and using the implementation from datafield

client/src/components/DatetimeField/DatetimeField.js
emteknetnz: Done

client/src/components/DatetimeField/DatetimeField.js
emteknetnz: Added an inline comment

client/src/components/TimeField/TimeField.js
emteknetnz: Removed

client/src/components/TimeField/TimeField.js
emteknetnz: Done

client/src/legacy/ReactComponents/tests/LegacyInputField-test.js
emteknetnz: I want to keep the 'Legacy' prefix otherwise we'll have 2x InputField-test.js - which is confusing. Much better for the LegacyInputField.js to have a LegacyInputField-test.js IMO

ChloeSartorelli: Yup that's fine. The comment was made when the file was `client/src/components/InputField/tests/InputField-functional-test.js`

client/src/bundles/bundle.js
ChloeSartorelli: The legacy components for anything we're not exposing should be removed as well.

client/src/components/TextField/TextField.js
ChloeSartorelli: Please respond to the following part of the comment, either by actioning it or saying you've chosen not to. Otherwise it's impossible to know whether you actually saw the comment and intentionally chose not to action it, or just missed it.

> Since these are identical to InputField's default props we could also export the default props object from there and import it here so it's not re-declared, but I won't hold up merging if you choose not to do that.

client/src/components/DatetimeField/DatetimeField.js
ChloeSartorelli: FWIW that probably should have just been a comment here rather than in the code. "since this component no longer delegates to DateField.getInputProps()" is nonsensical outside of the context of this conversion PR. I won't block merging just for this though.

GuySartorelli (changes requested):

client/src/bundles/bundle.js
emteknetnz: Removed LegacyNumberField and LegacyTimeField

client/src/components/DatetimeField/DatetimeField.js
emteknetnz: Will leave as is

client/src/components/TextField/TextField.js
emteknetnz: Leave as is

client/src/bundles/bundle.js
```diff
@@ -48,8 +48,12 @@ import 'expose-loader?exposes=ListGroup!components/ListGroup/ListGroup';
 import 'expose-loader?exposes=ListGroupItem!components/ListGroup/ListGroupItem';
 import 'expose-loader?exposes=Loading!components/Loading/Loading';
 import 'expose-loader?exposes=CircularLoading!components/Loading/CircularLoading';
+import 'expose-loader?exposes=InputField!components/InputField/InputField';
```
emteknetnz, line 51: Seems like this is required at least otherwise admin won't build

I'm not 100% sure if LegacyInputField is also needed here, though the downside of including it is zero so I've left it in.

emteknetnz:
Sorry though I'd replied, I may have forgotten to click submit.

> Since these are identical to InputField's default props we could also export the default props object from there and import it here so it's not re-declared, but I won't hold up merging if you choose not to do that.

Will keep as is.

> FWIW that probably should have just been a comment here rather than in the code. "since this component no longer delegates to DateField.getInputProps()" is nonsensical outside of the context of this conversion PR. I won't block merging just for this though.

Keep as is

> The legacy components for anything we're not exposing should be removed as well.

I've keep InputField and LegacyInputField exported as InputField at least seems like it needed to be exported otherwise things didn't work. I'm not sure if LegacyInputField though the 'cost' of adding to the export list is zero so no harm in it.

client/src/bundles/bundle.js
ChloeSartorelli: The downside is slightly more than 0 - it becomes an implicit part of our API and is also a globally-spaced variable name that can't be used by anything else. But it's probably fine.

ChloeSartorelli: I'm also curious about what "won't build" specifically means - it's possible whatever error you were seeing has a solution that can be resolved relatively simply... but I won't bother with the pingpong for that.

GuySartorelli (approved):
Seems okay, will merge when all other code PRs are ready.

GuySartorelli:
Or I'll just merge it now because of muscle memory lol. Oops

PR #2126, opened 2026-02-23
FIX Correctly set frame loaded state for react preview

GuySartorelli:
I'll admit I don't 100% know how this fixes the bug - but I noticed this effect was a bit messy and figured it was probably related. After tidying it up, lo-and-behold the bug is gone.

## Issues

- https://github.com/silverstripe/silverstripe-versioned-admin/issues/463
- https://github.com/silverstripe/silverstripe-versioned-admin/issues/464


client/src/components/Preview/Preview.js
```diff
@@ -33,14 +32,8 @@ const Preview = ({
...
-    if (!frameLoaded) {
-      return;
-    }
-    if (lastPreviewUrlRef.current !== previewUrl) {
-      setFrameLoaded(false);
-      lastPreviewUrlRef.current = previewUrl;
-    }
-  }, [frameLoaded, previewUrl]);
+    setFrameLoaded(false);
+  }, [previewUrl]);
```
ChloeSartorelli, line 36: This effect ultimately boils down to "if the URL changes, the frame hasn't loaded so update the state".

We can tell if the URL changes by having that as the only dependency - and we don't need to return early if `frameLoaded` is false, because react is smart enough to realise that `false === false` so setting the state to the value it already was doesn't affect the lifecycle of the component.

emteknetnz (approved):

PR #2131, opened 2026-02-24
FIX Normalise reactRoutePath to always provide leading slash

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-asset-admin/issues/1655


GuySartorelli (approved):

Issue #2107, opened 2026-02-04
Cannot set (author) permissions for SingleRecordAdmin


RoyalPulp:
### Module version(s) affected

3.1.8

### Description

In security section I can set permission for each ModelAdmin section. But for SingleRecordAdmin Section, unfortunately I cannot, the SingleRecordAdmins don't appear in securty / group permissions

### How to reproduce

Log into CMS 
- with admin privilege: You see SingleRecordAdmin sections
- with author privilege: You cannot see any SingleRecordAdmin
- try to allow authors to see /edit SingleRecordAdmin section: not possible

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

## Kitchen-sink CI
https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/23465357685

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2132
- https://github.com/silverstripe/silverstripe-siteconfig/pull/199
- https://github.com/silverstripe/developer-docs/pull/891

GuySartorelli:
@emteknetnz Please also add a changelog entry and see if there's a need to add/update documentation elsewhere.

emteknetnz:
Had added changelog entry

GuySartorelli:
PRs merged. This will be included in the October minor release

PR #2132, opened 2026-02-25
FIX Add SingleRecordAdmin to permission management

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2107

- Add skip_permission_generation config on LeftAndMain with CMSProfileController set to true
- Set CMS access permission generation to rely on the skip config for ModelAdmin and SingleRecordAdmin subclasses
- Update SingleRecordAdminTest to cover permission generation skip behavior
- ~Conflict siteconfig <=3.3 which is set to skip permission generation like CMSProfileController~

### DIFF
--- a/code/LeftAndMain.php
+++ b/code/LeftAndMain.php
@@ -4,14 +4,15 @@ namespace SilverStripe\Admin;
 
 use InvalidArgumentException;
 use LogicException;
-use ReflectionClass;
-use SilverStripe\CMS\Controllers\CMSMain;
+use SilverStripe\Admin\Forms\UnsavedChangesIndicator;
 use SilverStripe\Admin\Navigator\SilverStripeNavigator;
+use SilverStripe\Admin\SingleRecordAdmin;
+use SilverStripe\CMS\Controllers\CMSMain;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
-use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\HTTPResponse_Exception;
+use SilverStripe\Control\HTTPResponse;
 use SilverStripe\Control\PjaxResponseNegotiator;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
@@ -19,6 +20,8 @@ use SilverStripe\Core\Convert;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Manifest\ModuleResourceLoader;
 use SilverStripe\Core\Manifest\VersionProvider;
+use SilverStripe\Core\Validation\ValidationException;
+use SilverStripe\Core\Validation\ValidationResult;
 use SilverStripe\Dev\TestOnly;
 use SilverStripe\Forms\DropdownField;
 use SilverStripe\Forms\FieldList;
@@ -28,22 +31,19 @@ use SilverStripe\Forms\HiddenField;
 use SilverStripe\Forms\LiteralField;
 use SilverStripe\Forms\PrintableTransformation;
 use SilverStripe\i18n\i18n;
+use SilverStripe\Model\ArrayData;
 use SilverStripe\Model\List\ArrayList;
+use SilverStripe\Model\List\SS_List;
 use SilverStripe\ORM\CMSPreviewable;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\FieldType\DBField;
 use SilverStripe\ORM\FieldType\DBHTMLText;
 use SilverStripe\ORM\Hierarchy\Hierarchy;
-use SilverStripe\Model\List\SS_List;
-use SilverStripe\Core\Validation\ValidationException;
-use SilverStripe\Core\Validation\ValidationResult;
-use SilverStripe\Admin\Forms\UnsavedChangesIndicator;
 use SilverStripe\Security\Permission;
 use SilverStripe\Security\PermissionProvider;
 use SilverStripe\Security\Security;
 use SilverStripe\Security\SecurityToken;
 use SilverStripe\SiteConfig\SiteConfig;
-use SilverStripe\Model\ArrayData;
 use SilverStripe\View\Requirements;
 use SilverStripe\View\SSViewer;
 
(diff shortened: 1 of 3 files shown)


code/LeftAndMain.php
```diff
@@ -1442,22 +1442,28 @@ public function providePermissions()
...
-            }
-            if (ClassInfo::classImplements($class, TestOnly::class)) {
+        // Remove ModelAdmin::class and SingleRecordAdmin::class from $classes
+        $classes = array_diff(
+            $classes,
+            [ModelAdmin::class, SingleRecordAdmin::class]
+        );
+        foreach ($classes as $class) {
+            if (ClassInfo::classImplements($class, TestOnly::class)
+                || Config::inst()->get($class, 'ignore_menuitem')
```
ChloeSartorelli: Why was `ignore_menuitem` added here? This doesn't seem related to the issue and wasn't there before. Just because we don't want a link to the admin section in the left menu doesn't necessarily mean we don't want a permission entry for it.

As hinted at in the comment about about siteconfig, we probably need a separate explicit configuration to skip adding permissions instead.

GuySartorelli (changes requested):
Should this be targeting `3.1`? Seems like a bug fix that doesn't add any new API.

This PR is adding a new "Access to 'Settings' section" permission - though siteconfig adds an explicit permission in the "Roles and access permissions" area at the bottom of the permissions tab called "Manage site configuration"

We shouldn't have both of these. Since `providePermissions` is public and can be overridden, we should probably keep the existing one and omit the new one.

code/LeftAndMain.php
emteknetnz: Removed

GuySartorelli:
> Should this be targeting 3.1? Seems like a bug fix that doesn't add any new API.

Please respond to this. It seems to be this bugfix should target `3.1`.

tests/php/SingleRecordAdminTest.php
```diff
@@ -114,4 +118,35 @@ public function testGetRecord(
             $this->assertNull($result);
         }
     }
+
+    public function testProvidePermissionsSkipsExplicitPermission(): void
+    {
+        $permissions = LeftAndMain::singleton()->providePermissions();
+        $code = SiteConfigLeftAndMain::getRequiredPermissions();
```
ChloeSartorelli: `silverstripe/siteconfig` isn't a dependency of admin. We should avoid referencing it in unit tests.

I know we kinda break this rule a bit with behat (we shouldn't but it's too much work with our current team size to resolve that) but at least in unit tests lets respect the dependencies of the repo.

code/LeftAndMain.php
```diff
@@ -1442,29 +1442,62 @@ public function providePermissions()
...
             ]
         ];
-
-        // Add any custom ModelAdmin subclasses. Can't put this on ModelAdmin itself
-        // since its marked abstract, and needs to be singleton instanciated.
-        foreach (ClassInfo::subclassesFor(ModelAdmin::class) as $i => $class) {
-            if ($class === ModelAdmin::class) {
+        $explicitPermissionCodes = [];
+        $permissionProviders = ClassInfo::implementorsOf(PermissionProvider::class);
+        foreach ($permissionProviders as $provider) {
```
ChloeSartorelli: I don't think we should be looping through permission providers if we can help it. I still think having an explicit way to say "skip this class" is a better approach both from a code cleanliness perspective as well as a flexibility perspective.

code/LeftAndMain.php
```diff
@@ -1442,29 +1442,62 @@ public function providePermissions()
...
+        foreach ($classes as $class) {
             if (ClassInfo::classImplements($class, TestOnly::class)) {
                 continue;
             }
-
+            if (ClassInfo::hasMethod($class, 'providePermissions')
+                && !ClassInfo::has_method_from($class, 'providePermissions', LeftAndMain::class)
+            ) {
+                continue;
+            }
```
ChloeSartorelli: What's this new condition in here for?

GuySartorelli (changes requested):
We shouldn't be adding a permission for the `CMSProfileController`  - it handles its own permissions for the most part, and ultimately anyone who can access the CMS _at all_ should be allowed to access their own profile.

So again just being able to set a "skip this class" config on `CMSProfileController` and `SiteConfigLeftAndMain` would let us skip these two specific classes (which handle their permissions in different ways) without adding a bunch of new logic here.

emteknetnz:
@GuySartorelli If you want the 'skip check' solution that'll require new API (target 6). Otherwise to target 6.1 we'll need the messy dynamic stuff this PR has now, let me know which you'd like

GuySartorelli:
The best scenario would be a clean solution that doesn't require new API that can be in a patch.
If we can't do that and it's between new API vs messy dynamic stuff, I'd opt for the new API.

emteknetnz:
Gone for new API

composer.json
```diff
@@ -30,6 +30,9 @@
         "silverstripe/standards": "^1",
         "phpstan/extension-installer": "^1.3"
     },
+    "conflict": {
+        "silverstripe/siteconfig": "<=3.3"
+    },
```
ChloeSartorelli: I don't think we should add this. Sure people with older versions of siteconfig will still exhibit this bug, but everything will still _work_. We don't tend to add conflicts when we add API - usually instead we bump the dependency in the modules that use the new API instead.

```suggestion
```

code/LeftAndMain.php
```diff
@@ -1442,22 +1446,29 @@ public function providePermissions()
...
+        );
+        // Can't put this on ModelAdmin itself
         // since its marked abstract, and needs to be singleton instanciated.
-        foreach (ClassInfo::subclassesFor(ModelAdmin::class) as $i => $class) {
-            if ($class === ModelAdmin::class) {
+        // Remove ModelAdmin::class and SingleRecordAdmin::class from $classes
+        $classes = array_diff(
+            $classes,
+            [ModelAdmin::class, SingleRecordAdmin::class]
+        );
```
```suggestion
```
Pass `false` as the second argument to `ClassInfo::subclassesFor` instead.

GuySartorelli (changes requested):

GuySartorelli:
Also has a merge conflict

composer.json
emteknetnz: Removed

emteknetnz:
Rebased

code/LeftAndMain.php
emteknetnz: Done

GuySartorelli (approved):
LGTM

PR #2140, opened 2026-03-31
FIX Strip Content-Type parameters before matching in Backend.js decode()

lerni:
## Description

Servers that append `; charset=utf-8` to the `Content-Type` header (e.g. Apache with `AddDefaultCharset`, nginx, reverse proxies) cause `decode()` in `Backend.js` to throw `"Can't decode format"` because the `switch` uses exact string matching.

This broke LinkField saves. Server-side save succeeded but the JS error swallowed the response, showing a false "Failed to save link" toast. The modal stayed open even though the link was persisted.

The fix strips Content-Type parameters (everything after `;`) before the switch, so `application/json; charset=utf-8` correctly matches the `application/json` case.

## Manual testing steps
1. Configure your web server to append `charset=utf-8` to JSON responses (e.g. Apache `AddDefaultCharset UTF-8`, or nginx `charset utf-8;`)
2. Add a `LinkField` to a page type with a `has_one` to `Link`
3. In the CMS, edit a page → click the LinkField → fill in the link form → save
4. **Before fix:** a "Failed to save link" toast appears and the modal stays open (link is actually saved server-side)
5. **After fix:** the link saves and the modal closes without error

## Issues
- silverstripe/silverstripe-linkfield#349

## Pull request checklist
- [ ] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [x] CI is green


lerni:
This fix also resolves a broken re-authentication flow in the admin. When editing sensitive user data (e.g. on /admin/myprofile), the re-auth POST succeeds server-side (session is established) but the decode() error swallows the response — so nothing happens visually. On page reload the user is authenticated, confirming the server-side action completed. Same root cause, same fix.

lerni:
Just discovered one more cases affected: creating a folder in Assets also silently fails - nothing happens until page reload.

https://github.com/user-attachments/assets/b1dc3131-ce65-478a-8757-7360dba1e044

lerni:
Major distributions like RHEL, CentOS, and Fedora explicitly override Apache's defaults by shipping with AddDefaultCharset UTF-8 turned on inside `httpd.conf`. This setting is mirrored across many shared-hosting stacks, meaning **the fix resolves broken behavior for an estimated 20–30% of all deployed sites**.
@emteknetnz or Guy - is there something I can do to drive this ahead?

emteknetnz (changes requested):
This PR is targetting 2.4 (CMS 5.4) - as per our [support timeline](https://www.silverstripe.org/software/roadmap) that's now in partial support so only receiving fixes so security issues that have a CVE of high or above, so you'll need to aimed it at 3.2 (CMS 6.2) instead

I'd also update it to be a bit more constrained with the fix to reduce chance of weird side-effect - e.g. specifically remove the `charset=utf-8` from the end of the string

lerni:
Thank you for the feedback - I need to sort-out some git-foo retargetting it, but hopefully 'll get there.

[RFC 9110 §8.3](https://datatracker.ietf.org/doc/html/rfc9110#section-8.3) defines that the media type is the token before the first `;` - all parameters follow it. Stripping everything after `;` is what a HTTP client library should do. Targeting just `charset=utf-8` would still break with different encodings/encoding-casing, or with more parameters like `boundary` cos of the same root cause.

emteknetnz (changes requested):
Could you add a test in to `client/src/lib/tests/Backend-test.js` - this will fail before the code fix and passes with your code change

```js
it('should decode a response with a parameterised Content-Type header', () => {
  const charsetMock = getBackendMock({
    text: () => Promise.resolve('{"status":"ok"}'),
    headers: {
      get: () => 'application/json; charset=utf-8',
    },
  });
  const endpoint = charsetMock.createEndpointFetcher({
    url: 'http://example.com',
    method: 'get',
    responseFormat: 'json',
  });
  return endpoint({}).then((data) => {
    expect(data).toEqual({ status: 'ok' });
  });
});
```

lerni:
Test added - thank you for the suggestion ☝️

emteknetnz (approved):

Issue #2148, opened 2026-05-21
Convert more admin fields to functional components


emteknetnz:
There are react class component(s) that needs to be converted to react functional component(s).
- CheckboxSetField.js
- CompositeField.js
- FieldGroup.js
- FieldHolder.js
- FormAction.js
- HiddenField.js

Note there are more components than this to convert, limiting number converted at a time to make peer review more manageable.

## Acceptance criteria
- [ ] Component(s) are converted to a functional component using hooks
- [ ] Jest test coverage is increased
- [ ] All existing functionality is preserved
- [ ] Legacy class components are retained (if applicable i.e. is admin form field component)

## Kitchen sink CI
- https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/35061828240

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/2149
- https://github.com/silverstripe/silverstripe-elemental/pull/1434
- https://github.com/silverstripe/webpack-config/pull/93

ChloeSartorelli:
@emteknetnz PRs are in draft - are these ready to be reviewed?

emteknetnz:
Ready now

ChloeSartorelli:
PRs merged

PR #2149, opened 2026-05-21
ENH Convert more admin fields to functional components

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/2148


client/src/bundles/bundle.js
```diff
@@ -73,6 +76,11 @@ import 'expose-loader?exposes=UnsavedFormsActions!state/unsavedForms/UnsavedForm
 import 'expose-loader?exposes=Badge!components/Badge/Badge';
 import 'expose-loader?exposes=VersionedBadge!components/VersionedBadge/VersionedBadge';
 import 'expose-loader?exposes=CheckboxSetField!components/CheckboxSetField/CheckboxSetField';
+import 'expose-loader?exposes=LegacyCheckboxSetField!legacy/ReactComponents/LegacyCheckboxSetField';
+import 'expose-loader?exposes=CompositeField!components/CompositeField/CompositeField';
+import 'expose-loader?exposes=LegacyCompositeField!legacy/ReactComponents/LegacyCompositeField';
+import 'expose-loader?exposes=FieldGroup!components/FieldGroup/FieldGroup';
+import 'expose-loader?exposes=LegacyFieldGroup!legacy/ReactComponents/LegacyFieldGroup';
```
ChloeSartorelli: I don't think we need the legacy fields for things that weren't previously exposed, since they could only directly be subclassed by us or by someone doing things in a weird unexpected way.
IMO remove these expose lines and delete the legacy components for these.

If the legacy classes are needed for inheritance into other components we use, it's okay to keep the legacy class but we still don't need to expose them here.

ChloeSartorelli (changes requested):

client/src/bundles/bundle.js
emteknetnz: Agreed for LegacyCompositeField and LegacyFieldGroup. Neither was in bundle.js before this PR and neither is a parent of an exported component, so [per the rule in the 6.3.0 changelog](https://docs.silverstripe.org/en/6/changelogs/6.3.0/#react-functional-components) (previously defined in the `client/src/bundles/bundle.js`) they don't qualify. I've delete these two legacy components and their tests, and dropped the two expose lines. I've also removed 

I'll keep LegacyCheckboxSetField though. CheckboxSetField was already exposed in bundle.js, so it's the same situation as LegacyTextField and friends and the changelog rule says it should have a legacy copy.

The non-legacy CompositeField and FieldGroup exposes need to stay, [elemental pr](https://github.com/silverstripe/silverstripe-elemental/pull/1434) imports getLegend and getClassName from them.

Have updated [webpack-config](https://github.com/silverstripe/webpack-config/pull/93) to remove the two matching externals entries and resolve the conflicts at the same time.

ChloeSartorelli:
This PR is failing in different places than the regular CI run for the target branch: https://github.com/silverstripe/silverstripe-admin/actions/runs/35449965376

Can you please validate if these are related to this PR, and if not why the failures are different here?

emteknetnz:
Rerunning failed sporadic behat failures - they should pass on rerun - behat has been extremely flaky as of late

ChloeSartorelli:
Has conflicts

emteknetnz:
rebased

ChloeSartorelli (approved):
LGTM

PR #2158, opened 2026-06-16
FIX SingleRecordAdmin::currentRecordID() returns URL ?ID= param, breaking TreeDropdownField AJAX.

RVXD:
## Description
When a TreeDropdownField is placed inside a SingleRecordAdmin (e.g. SiteConfigLeftAndMain) and the tree has enough nodes to trigger AJAX lazy-loading, expanding a node returns a 404:

I can't handle sub-URLs on class SilverStripe\Forms\FormRequestHandler.
The field works when the tree is small enough to be pre-populated without AJAX. The bug does not affect page editing in CMSMain. This is a regression from 6.1.x.

## Manual testing steps
See: https://github.com/silverstripe/silverstripe-framework/issues/11984

## Issues
See: https://github.com/silverstripe/silverstripe-framework/issues/11984

### DIFF
--- a/code/SingleRecordAdmin.php
+++ b/code/SingleRecordAdmin.php
@@ -24,6 +24,18 @@ abstract class SingleRecordAdmin extends LeftAndMain
      */
     private static bool $allow_new_record = true;
 
+    /**
+     * Single-record admins never use a URL-based record ID — the record is always
+     * fetched via getSingleRecord(). Returning null here prevents query parameters
+     * such as ?ID= (e.g. from a TreeDropdownField AJAX subtree request) from being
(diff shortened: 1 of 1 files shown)


emteknetnz:
The issue with this approach is that returning null means that any call to `setCurrentRecordID()` will be ignored. In CMSProfileController which extends SingleRecordAdmin (used for `/admin/myprofile`) this would break as that calls `$this->setCurrentRecordID(Security::getCurrentUser()->ID);` in `init()`

Closing in favor of https://github.com/silverstripe/silverstripe-admin/pull/2169

PR #2161, opened 2026-07-14
FIX Open the site tree to the current page when deep-linked

pjayme:
## Description
Opening the CMS deep-linked to a page whose ancestors haven't been lazy-loaded into the site tree yet — e.g. following the front-end `SilverStripeNavigator` "edit" link, or a bookmarked `admin/pages/edit/show/<id>` URL — leaves the pages tree **collapsed on the root** instead of expanding onto the edited page.

**Cause:** `LeftAndMain.Tree.updateFromEditForm()` falls back to `updateNodesFromServer([id])` when the record isn't in the loaded tree. For a deep record that hits `jstree('load_node', -1)`, which (a) reloads the **root** — never the record's ancestor path — and (b) passes **no success callback**, so the html_data plugin's `load_node` (`s_call.call(this)`) throws `TypeError: Cannot read properties of undefined (reading 'call')` and aborts the open.

**Fix:** add `openTreeToRecord()`, called from `updateFromEditForm()`. It walks the ancestor chain via the `updatetreenodes` endpoint (each response carries `ParentID`) up to a loaded node / the root, opens each level top-down with `open_node` (which lazy-loads children and fires its callback when done), then selects the record — or the nearest visible ancestor for records hidden from the tree (Lumberjack children, `show_in_sitetree = false`). Includes the rebuilt `client/dist`.

A record whose parent exceeds `node_threshold_total` needs one more step: the parent renders a **"too many records"** placeholder instead of its children, so `open_node` never loads the target and the walk would otherwise stop at that parent. `spliceNodeIntoParent()` handles it — it fetches the single node from `updatetreenodes` and inserts it under the loaded parent via `createNode()`, so the target is revealed and selected while the placeholder and its "show as list" link stay in place for the remaining children.

## Manual testing steps
1. On a site with a deep page hierarchy (grandchild or deeper), log into the CMS.
2. Go straight to `admin/pages/edit/show/<id>` for a deep page (or use the front-end "edit in CMS" navigator link). Before this change the tree stays collapsed on the root; after it, the tree expands to and selects the page. ✅
3. Regression: open a top-level page — it still selects, no change.
4. Open a tree-hidden page (e.g. a Lumberjack child / `show_in_sitetree = false`) — the tree selects the nearest visible ancestor, with no console error.
5. Open `admin/pages/edit/show/<id>` for a page whose parent has more than `node_threshold_total` children (a parent that shows "too many records"). Before this change the tree stops at that parent; after it, the target page is spliced in and selected, its own children stay expandable, and the parent's "show as list" link remains. ✅
6. Confirm the browser console shows **no** `TypeError: Cannot read properties of undefined (reading 'call')` on load.

## Issues
- #2162

## Pull request checklist
- [x] The target branch is correct (`3` — bugfix on the current major)
- [x] The PR follows our contribution guidelines
- [x] Code changes follow our coding conventions
- [x] This change is covered with tests — the ancestor-chain resolution is extracted into an exported, dependency-injected `buildAncestorChain()` helper and unit tested (`client/src/legacy/tests/LeftAndMain.Tree-test.js`): loaded-ancestor stop, walk-to-top, top-level, and already-loaded (no server calls) cases. The `open_node` orchestration and the `spliceNodeIntoParent()` reveal remain entwine/jstree-side and are verified manually (steps above), consistent with the other untested entwine methods in this file.
- [ ] Any relevant User Help/Developer documentation is updated — _n/a (behavioural bugfix, no documented behaviour changes)_
- [ ] CI is green — _pending CI run_


emteknetnz:
I tried this locally, it didn't work right away after running the build task defined on this PR to replicate - https://github.com/silverstripe/silverstripe-admin/pull/2164 - closing this in favor of that PR instead

PR #2166, opened 2026-08-18
MNT Use yarn 4

emteknetnz:
Issue https://github.com/silverstripe/.github/issues/496


package.json
```diff
@@ -12,7 +12,7 @@
     "node": "^18.x"
   },
   "scripts": {
-    "build": "yarn && yarn lint && yarn test && rm -rf client/dist/* && NODE_ENV=production webpack --mode production --bail --progress",
+    "build": "yarn && yarn lint && yarn test && rm -rf client/dist && NODE_ENV=production webpack --mode production --bail --progress",
```
emteknetnz, line 15: Yarn 4 runs scripts through its own shell, which errors if a glob matches nothing, so `rm -rf client/dist/*` fails in CI where the dist directory has already been deleted.

.gitignore
```diff
@@ -2,6 +2,7 @@
 /vendor/
 /resources/
 /node_modules/
+/.yarn/
```
blueo, line 5: we prevent yarn.lock + other files from being part of the 'dist' install via gitattributes so we should do the same with the new `.yarnrc.yml`: https://github.com/silverstripe/silverstripe-admin/blob/3.2/.gitattributes#L29

blueo (changes requested):
mostly looking good just a consistency thing with including yarn files in the dist installation of the module

.gitignore
emteknetnz: I'm confused by this? What would we use .gitattributes when we're using .gitignore for everything else? We've commited yarn.lock for a long time and we need to so that we're using the same set of deps on every install. need to commit .yarnrc.yml file it includes the cooldown so if anyone updates deps outside of the `gha-update-js` workflow the cooldown will still apply

blueo: this still allows them to be committed - we do both atm. This is just for when it gets turned into an install tar ('exported') - eg when you do composer install it will not include yarn.lock + other files in .gitattributes with `export-ignore` comments as they're dev only files. To get it you need `composer install --prefer-source`. Its so you can have something committed but then excluded from the 'package'. I'm suggesting we add to .gitattributes but that's not part of the diff so I couldn't add a comment on it.

emteknetnz: oh right sorry I misunderstood, have added

PR #2176, opened 2026-08-30
FIX Guard entwine.warningLevel assignment against undefined entwine

tomstgeorge:
While clicking into a record in a ModelAdmin (an AJAX panel load), `LeftAndMain.Panel.js` runs:

```js
\$.entwine.warningLevel = \$.entwine.WARN_LEVEL_BESTPRACTISE;
```

In some load orderings the jQuery instance this module resolves does not (yet) have `entwine` attached, so `\$.entwine` is `undefined`. The assignment then throws:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'WARN_LEVEL_BESTPRACTISE')
```

The throw happens inside the panel bootstrap that is `globalEval`'d from the panel AJAX response, aborting panel initialisation. The practical symptom is that **CMS left-hand menu navigation stops responding** (menu carets / menu links dead) after opening a detail view.

### Fix

Null-check `\$.entwine` before assigning `warningLevel`. If entwine is not present on the resolved jQuery, skip the assignment rather than throwing and aborting init. Applied in both `LeftAndMain.Panel.js` and `LeftAndMain.js` (same pattern in both).

### Notes
- Purely defensive; no behaviour change when entwine is present.
- Reproduced on admin 3.2.x consuming projects with several third-party CMS modules.

### Issues
- (none filed yet)

### Pull request checklist
- [ ] Documentation (if relevant)
- [ ] CI is green

### DIFF
--- a/client/src/legacy/LeftAndMain.Panel.js
+++ b/client/src/legacy/LeftAndMain.Panel.js
@@ -3,7 +3,12 @@ import $ from 'jquery';
 $.entwine('ss', function($) {
 
   // setup jquery.entwine
-  $.entwine.warningLevel = $.entwine.WARN_LEVEL_BESTPRACTISE;
+  // Guard against entwine being unavailable on the jQuery instance this module
+  // resolves during an AJAX panel load — assigning to an undefined entwine here
+  // throws and aborts panel init, freezing CMS menu navigation.
+  if ($.entwine) {
+    $.entwine.warningLevel = $.entwine.WARN_LEVEL_BESTPRACTISE;
+  }
 
   /**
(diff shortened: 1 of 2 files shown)


tomstgeorge:
Closing — this was based on a misdiagnosis on our side. The issue we were chasing (CMS menu-group carets not toggling after navigating into a detail view) turned out to be caused by our own module double-binding a click handler on every pjax:end, not by an undefined $.entwine. The warningLevel assignment does not actually fail on a stock install, so this guard isn't needed upstream. Apologies for the noise.

PR #2177, opened 2026-08-30
FIX Skip non-element nodes in entwine matching (comment nodes)

tomstgeorge:
Opening a ModelAdmin list view (and other CMS interactions) can throw:

```
Uncaught TypeError: el.getAttribute is not a function
```

from inside the compiled entwine selector matcher, reached via a MutationObserver / the entwine event proxies.

### Root cause

`handleMutation()` filters added mutation nodes with:

```js
if (node.nodeName === '#text') { continue; }
```

That skips text nodes but **not comment nodes** (or any other non-element node). A comment node then reaches the compiled matcher (`SelectorBase.matches`), whose body calls `el.getAttribute(...)`. A `Comment` has no `getAttribute`, so it throws. The throw aborts the `EntwineElementsAdded` batch, so entwine stops (re)binding behaviour.

The same class of node can also arrive via the entwine event proxies (`selector_proxy` / `property_proxy`), which run `matcher.matches(e.target)` where `e.target` may be a text/comment node for bubbled DOM events.

Repro on any entwine page:

```js
document.body.appendChild(document.createComment('x'));
```

### Fix

Two defensive guards, both restricting selector matching to element nodes (`nodeType === 1`):

1. `handleMutation()` — filter on `nodeType !== 1` instead of only `nodeName === '#text'` (mirrors `addNodeChildrenToList`, which already does this).
2. The compiled `SelectorBase.matches` body — guard `if (!el || el.nodeType !== 1) return false;` so **every** caller (including the event proxies) is covered, not just the mutation path.

### Notes
- Only element nodes can match a CSS selector, so returning `false`/skipping for non-elements is correct.
- `thirdparty/jquery-entwine/jquery.entwine.js` is a vendored copy of `jquery.entwine`; happy to redirect this upstream to that library if preferred, but admin builds from this copy so the fix is applied here.

### Pull request checklist
- [ ] Documentation (if relevant)
- [ ] CI is green


michalkleiner:
Hi @tomstgeorge, thanks for taking the time to open the PR.
Since this is a fix, it should target a minor version branch, in this case `3.2`.
Can you also squash the commits into a single commit with the FIX prefix?

tomstgeorge:
Thanks @michalkleiner! I've retargeted the PR to 3.2 and squashed everything into a single commit with the FIX prefix. Let me know if you'd like any further changes.

ChloeSartorelli:
I've run `yarn build` to build out the dist file. Assuming CI goes green I'll be happy to merge this

ChloeSartorelli:
There's a failing behat test. Can you please investigate that? This failure is not happening in a run directly on the `3.2` branch.
If this doesn't appear to be related to the changes in this PR, it's possible this is something that has been fixed after you created your branch, so rebasing may resolve it.

tomstgeorge:
Thanks @ChloeSartorelli! The failing Behat test was down to my branch being based on 3.2 from before #2169 (the SingleRecordAdmin fix) merged — so the diff was effectively reverting those changes. I've rebased onto the latest 3.2, which pulls #2169 back in, and force-pushed. Your `yarn build` dist changes are preserved. CI should be green now.

ChloeSartorelli (approved):
LGTM. Thanks for your work on this

Issue #1965, opened 2025-08-10
Accessibility: jQuery UI accordions use `tablist` and `tabpanel` roles incorrectly.


GuySartorelli:
[ARIA 1.2](https://www.w3.org/TR/wai-aria-1.2/)[WCAG 2.2 A 1.3.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/content-structure-separation-programmatic.html) tabpanel elements must not appear inside the tablist and are usually placed immediately after the tablist

The `ToggleCompositeField` is controlled with a [jQuery UI accordion](https://jqueryui.com/accordion/). That results in a parent element with `role="tablist"`, inside of which both the toggle button (`role="tab"`) and collapsible content (`role="tabpanel"`) are children.

This may be a good opportunity to start moving away from jQuery UI. We shouldn't remove jQuery UI in case people are relying on it for their customisations, but we can stop using it for this.

## Notes

- `ToggleCompositeField` is called out explicitly but check for other cases where we're using jQuery UI for accordion behaviour
- Tabs are probably not the appropriate roles to use for accordion-like behaviour, so check out some existing accessible accordion examples before deciding how to best implement this.
- The copy of jQuery UI we're using is in the `thirdparty` directory of `silverstripe/admin` because we've already made alterations to it. We _could_ just change the roles or markup through edits to jQuery UI directly. However, this is really just kicking the can down the road, as we should be aiming to stop using this library.
- This is an opportunity to provide a JS implementation of this field that would work on the frontend. Consider that option.

## Related

- https://github.com/silverstripe/.github/issues/470

## Acceptance criteria

- [ ] We stop using jQuery UI for any accordion-like behaviour (if sensible - see notes)
- [ ] Accordion-like markup follow accessibility guidelines
- [ ] The changelog clearly calls out any changes to markup selectors so people can adjust their custom css/js as necessary
- [ ] Check `ToggleCompositeField` in both entwine and react forms

emteknetnz:
I had a relatively quick go at just hacking around jQuery UI, it didn't prove as easy as I'd like.

Similar to [replace chosen-js](https://github.com/silverstripe/silverstripe-admin/issues/2027#issuecomment-3483771015), by replacing jQuery UI for accordion behaviour, we're risking double handling here as we're likely to just replace this jquery compnent with another jquery component, only to get rid of it later with a react refactor

[...]

Issue #1752, opened 2024-05-22
Validation errors don't highlight the tab in `GridField` like they do in `SiteTree` edit forms


GuySartorelli:
Found while working on https://github.com/silverstripe/silverstripe-cms/issues/2944

The edit form for `SiteTree` records (i.e. `CMSMain`) has the following functionality which is not present for `GridField` edit forms:
1. When there is a validation error, the tabs which have errors get an icon next to them indicating there's an error _in that tab_ specifically
2. Add a message to the top of the form indicating there are errors to be resolved

![image](https://github.com/silverstripe/silverstripe-admin/assets/36352093/c5860d1b-84b0-4d2e-95ee-374692f99d0e)

## Acceptance criteria
- Implement the above-mentioned functionality for `GridField` edit forms

## PRs
- https://github.com/silverstripe/silverstripe-admin/pull/1753

PR #2179, opened 2026-09-03
FIX Mark every enclosing tab when a nested tabset holds a validation error

### LINKED ISSUE #2178: Validation error indicator is not applied to the outer tab when a `TabSet` is nested inside another `TabSet`
## Summary

When a CMS edit form nests a `TabSet` inside another `TabSet`, the "this tab has validation errors" indicator (`.tab-attention`) never lands on the outer tab. The editor gets the generic red banner, a clean-looking outer tab, and the field-level error hidden inside a collapsed sub-tab pane. Finding the invalid field means opening every sub-tab in turn.

| Form | Tab structure | Outer tab | Inner sub-tab |
|---|---|---|---|
| Page edit form | flat (control) | marked correctly | n/a |
| Page edit form | nested `TabSet` | **not marked** | marked |
| GridField detail form | flat (control) | marked correctly | n/a |
| GridField detail form | nested `TabSet` | **not marked** | **not marked** |

## Affected versions

`silverstripe/admin` 3.2.1, still present on `3.2` (3.2.3). Reproduced on CMS 6.2, `silverstripe/framework` 6.2.2, PHP 8.3, Chrome - the fault is in selector logic so it isn't browser specific.

Branch `2.4` (CMS 5) carries the same defect. The selection block at `client/src/legacy/LeftAndMain.EditForm.js` lines 153-171 is byte-identical between `2.4` and `3.2`, only the icon markup differs (`2.4` appends `<i class="font-icon-attention-1 tab-attention">` and `sr-only`, `3.2` a `<span>` and `visually-hidden`).

## Steps to reproduce

<details>
<summary>Show steps with code snippets</summary>

Add these classes to a CMS 6 installer project and run `sake db:build --flush`.

**1. Flat tabs, the control.** Marks the right tab. Build this one first, it's the behaviour the other cases are measured against.

```php
<?php

namespace App\Test;

use Page;
use SilverStripe\Forms\TextField;
use SilverStripe\Forms\Validation\CompositeValidator;
use SilverStripe\Forms\Validation\RequiredFieldsValidator;

class FlatTabPage extends Page
{
    private static $table_name = 'FlatTabPage';

    private static $db = [
        'AlphaValue' => 'Varchar(255)',
        'BetaValue' => 'Varchar(255)',
    ];

    public function getCMSFields(): FieldList
    {
        $fields->removeByName(['AlphaValue', 'BetaValue']);
        $fields->addFieldToTab('Root.Alpha', TextField::create('AlphaValue', 'Alpha value'));
        $fields->addFieldToTab('Root.Beta', TextField::create('BetaValue', 'Beta value'));
        return $fields;
    }

    public function getCMSCompositeValidator(): CompositeValidator
    {
        $validator = parent::getCMSCompositeValidator();
        $validator->addValidator(RequiredFieldsValidator::create(['BetaValue']));
        return $validator;
    }
}
```

Create a `FlatTabPage`, give it a title, leave `Beta value` empty, leave `Main content` as the active tab, and publish. Validation fails and the `Beta` tab is marked.

**2. Page edit form, nested `TabSet`.** Marks the wrong tab. Same two fields and same required rule, with `Alpha` and `Beta` replaced by a nested `TabSet`.

```php
<?php

namespace App\Test;

use Page;
use SilverStripe\Forms\Tab;
use SilverStripe\Forms\TabSet;
use SilverStripe\Forms\TextField;
use SilverStripe\Forms\Validation\CompositeValidator;
use SilverStripe\Forms\Validation\RequiredFieldsValidator;

class NestedTabPage extends Page
{
    private static $table_name = 'NestedTabPage';

    private static $db = [
        'AlphaValue' => 'Varchar(255)',
        'BetaValue' => 'Varchar(255)',
    ];

    public function getCMSFields(): FieldList
    {
        $fields->removeByName(['AlphaValue', 'BetaValue']);

        /** @var TabSet $root */
        $root = $fields->fieldByName('Root');
        $root->push(
            TabSet::create(
                'Questions',
                Tab::create('SectionA', TextField::create('AlphaValue', 'Alpha value')),
                Tab::create('SectionB', TextField::create('BetaValue', 'Beta value'))
            )
        );

        return $fields;
    }

    public function getCMSCompositeValidator(): CompositeValidator
    {
        $validator = parent::getCMSCompositeValidator();
        $validator->addValidator(RequiredFieldsValidator::create(['BetaValue']));
        return $validator;
    }
}
```

Create a `NestedTabPage`, give it a title, leave `Beta value` empty, leave `Section A` as the active sub-tab, and publish. Validation fails - look at the top-level tab strip.

**3. GridField detail form, nested `TabSet`.** Marks nothing at all. Same structure on a `DataObject` behind a `ModelAdmin`:

```php
public function getCMSFields(): FieldList
{
    return FieldList::create(
        TabSet::create(
            'Root',
            Tab::create('Main', TextField::create('Title', 'Record name')),
            TabSet::create(
                'Questions',
                Tab::create('SectionA', TextField::create('AlphaValue', 'Alpha value')),
                Tab::create('SectionB', TextField::create('BetaValue', 'Beta value'))
            )
        )
    );
}

public function getCMSCompositeValidator(): CompositeValidator
{
    $validator = parent::getCMSCompositeValidator();
    $validator->addValidator(RequiredFieldsValidator::create(['BetaValue']));
    return $validator;
}
```

Create a record, clear `Beta value`, leave `Section A` active, and save.

**4. The other nesting shape,** a `TabSet` added inside an ordinary `Tab` rather than pushed onto `Root`:

```php
$fields->addFieldToTab(
    'Root.Other',
    TabSet::create(
        'Questions',
        Tab::create('SectionA', TextField::create('AlphaValue', 'Alpha value')),
        Tab::create('SectionB', TextField::create('BetaValue', 'Beta value'))
    )
);
```

Same outcome, so this isn't specific to how the nested `TabSet` is attached. The DOM does differ, which matters for any fix - pushing onto `Root` gives an outer panel that is an `.ss-tabset` div with no `.tab-pane`, while adding inside a `Tab` gives a `.tab-pane` containing an `.ss-tabset`. The ids differ too, `Root_Questions_set_SectionB` against a bare `Questions_SectionB`, because `TabSet::ID()` only prefixes when the `TabSet` is a direct child of another `TabSet`.

</details>

## Expected

The tab strip points the editor at the invalid field, which is what flat tabs already do. In case 1 the required `Beta value` is empty and the `Beta` tab holding it carries the icon, visible from the top of the form without opening anything:

Flat tabs: the top-level strip reads Main content, Alpha, Beta, and Beta carries the attention icon

Nesting those two fields under a `Questions` tab shouldn't change that. `Questions` carries the icon, and so does `Section B` inside it, so the path to the error stays visible from the top level.

## Actual

Same two fields, same required rule, same failed publish, nested this time. The top-level strip now reads `Main content` and `Questions`, and `Questions` carries no icon:

Nested tabset: the top-level strip reads Main content and Questions, and neither carries an icon

The icon went one level down, onto the inner `Section B` sub-tab. That nav sits inside the `Questions` panel, which isn't the active tab, so nothing above the fold shows it. The editor has to open `Questions` on a hunch to find it.

The GridField detail form is worse - no icon anywhere, in neither nav.

In both form types the field-level `"Beta value" is required` message renders correctly, but inside a hidden sub-tab pane. Both screenshots use synthetic data.

## Cause

<details>
<summary>Show detailed cause information</summary>

`client/src/legacy/LeftAndMain.EditForm.js`, in the `'from .cms-tabset': { onafterredrawtabs: ... }` handler, lines 153-171 on `3.2`:

```js
// Find tab-pane's with decedent validation .alert's
const $invalidTabPanes = this.find('.tab-pane .alert-danger, .tab-pane .alert.error').closest('.tab-pane');
if (!$invalidTabPanes.length) {
  // If we are at this point it's probably a failed DataObject::validate()
  // where there was a general (non-field) error added via ValidationResult::addError()
  return;
}

// Get the tabs for this form
const $gridfieldTabs = this.find('.cms-content-header-tabs.cms-tabset-nav-primary li[role="tab"]');
const $ssTabSet = $invalidTabPanes.closest('.tab-content').closest('.ss-tabset');
let getTabLi = null;
if ($gridfieldTabs.length > 1) {
  // GridField logic
  getTabLi = (invalidTabPaneId) => $gridfieldTabs.filter(`[aria-controls="${invalidTabPaneId}"]`);
} else if ($ssTabSet.length) {
  // SiteTree logic
  getTabLi = (invalidTabPaneId) => $ssTabSet.find(`#tab-${invalidTabPaneId}`).closest('li');
}
```

`.closest('.tab-pane')` resolves to the innermost pane, so `invalidTabPaneId` is always the sub-tab's. Each lookup then searches exactly one nav and stops:

- `$ssTabSet` resolves to the innermost enclosing tabset. `#tab-Root_Questions_set` is sitting in the outer nav, but the search never reaches it.
- `$gridfieldTabs` is restricted to `.cms-tabset-nav-primary`, where no `li` has `aria-controls="Root_Questions_set_SectionB"`, so nothing matches and no icon is appended.

Neither walks the ancestor chain, so neither crosses more than one level of tabset.

The two lookups can't be swapped for one another either, because the navs expose different hooks. The page edit form renders its nav inside `#Root` from `TabSet.ss`, giving every anchor an `id="tab-<panelId>"`. The GridField detail form promotes the top level into `.cms-content-header-tabs.cms-tabset-nav-primary` from `LeftAndMain_EditForm.ss`, where anchors have no `id="tab-..."` and jQuery UI assigns generated `ui-id-N` ids instead. Nested tabsets in both form types keep the `tab-<panelId>` anchors.

One hook does hold uniformly. jQuery UI's `tabs()` decorates every tab `li` with `role="tab"` and `aria-controls="<panelId>"`, and every panel with `role="tabpanel"`, including a nested `.ss-tabset` div acting as a panel. A nested outer panel is therefore reachable by the same `li[role="tab"][aria-controls="<panelId>"]` lookup the GridField branch already uses.

</details>

## Prior art

This is the nested-tab gap left open by [#1752](https://github.com/silverstripe/silverstripe-admin/issues/1752), "Validation errors don't highlight the tab in `GridField` like they do in `SiteTree` edit forms" (closed 2024-05-22), which added the `$gridfieldTabs` branch quoted above. That fix brought flat tabs in `GridField` into line with `SiteTree`, nested tabsets weren't covered and fail in both form types.

edwilde:
Fixes #2178

## The problem

When a `TabSet` is nested inside another `TabSet`, the "this tab has validation errors" icon never lands on the outer tab. On a page edit form only the inner sub-tab is marked, and that nav sits inside a panel the editor hasn't opened. On a GridField detail form nothing is marked at all. Either way you get a red banner and no pointer to the invalid field.

## What changed

The old code resolved one pane against one nav, so the pane it found was always the innermost one and the nav lookup never searched past a single tabset.

Each alert now walks all of its enclosing panels, innermost first, so one invalid field marks the sub-tab holding it and every outer tab above it. The walk tests for `role="tabpanel"` rather than `.tab-pane` - a tab which is itself a `TabSet` has no `.tab-pane`, because `TabSet.ss` renders `$FieldHolder` straight into `.tab-content`, so the nested `.ss-tabset` _is_ the outer tab's panel. A `.tab-pane` walk would skip the outer level, which is the level most worth flagging.

Nav lookup is now a single `li[role="tab"][aria-controls="<panelId>"]` search from the form root, which reaches a GridField primary nav and a nested tabset alike, so the branch on nav shape goes away. Both the panel walk and the nav lookup fall back to the template markup (`id="tab-<panelId>"`, the panel hash in `href`, direct child of `.tab-content`) for redraws that run before jQuery UI has decorated anything. That matters for #1965 - this only ever reads those roles, it doesn't depend on them staying.

## Behaviour note for reviewers

The banner is now set whenever a field inside a tab is invalid, rather than only when a tab nav was found. The old guard was false on a single-tab GridField detail form, so the banner never appeared there - which is what #1752 asked for. The tradeoff is that a specific `ValidationResult::addError()` message on that one shape is now replaced by the generic one, as it already was on every other shape.

Marking is idempotent, since a field can produce more than one alert and outer tabs are shared.

## Tests

`markTabValidationErrors` is extracted to module level and exported so it can be unit tested without booting entwine, following `LeftAndMain.Tree.js`. The entwine block is deliberately not re-indented to keep the diff on the logic.

12 new jest tests against six fixtures captured from a real CMS 6.2 after a failed save, covering both nesting shapes on a page form and a GridField form plus flat controls. All six reproduce the reported behaviour under the pre-fix code. The flat paths assert total icon counts so an over-eager walk fails.

Full suite 1043 passing across 105 suites, `yarn lint` clean, `client/dist/js/bundle.js` rebuilt.

## Before and after

Same nested form in the same state - the editor has opened `Questions`, then `Section B`, which before this change is the only way to reach the field that failed.

### Before

Only the inner `Section B` sub-tab is marked, and that nav lives inside the `Questions` panel.

Before the fix: the outer strip reads Main content and Questions with no icon on Questions, and only the inner Section B sub-tab carries the attention icon

### After

`Questions` and `Section B` are both marked, so the trail down to the invalid field is visible without opening anything on spec.

After the fix: the outer Questions tab and the inner Section B sub-tab both carry the attention icon

## Scope

Base branch is `3.2`. `2.4` carries the same defect - the selection block is byte-identical - but CMS 5.4 is in partial support, so whether it's worth backporting is the maintainers' call. A cherry-pick would need the icon markup adjusted, `2.4` emits `<i>` plus `sr-only` where `3.2` emits `<span>` plus `visually-hidden`.

`ToggleCompositeField` is out of scope and unchanged. An accordion header is `h4[role="tab"]`, not an `li`, so the nav lookup can't reach it. Confirmed by moving an alert into an accordion on the flat page fixture, which still marks `Root_Main` and only `Root_Main`.


edwilde:
CI has failed twice on this branch, in two different jobs and two different feature files, neither of them anything this change touches. Noting it here so it isn't read as a regression.

**Run 1** ([33714171964](https://github.com/silverstripe/silverstripe-admin/actions/runs/33714171964), `d21155ec`) - `8.5 mysql84 endtoend root job3` failed with `UnexpectedAlertOpenException: unexpected alert open: {Alert text : }` in `validation.feature:28`. That exact scenario passed on `8.3 mysql80 job3` in the same run, 35/35 scenarios and 416/416 steps.

**Run 2** ([33716322773](https://github.com/silverstripe/silverstripe-admin/actions/runs/33716322773), `08eeb994`) - `8.3 mysql80 endtoend root job2` failed twice in `accessibility.feature`: `.cms-tabset-nav-primary .nav-item:nth-of-type(1)` not found at :223, and `.breadcrumbs-wrapper` not found after the back button at :263. Both are element-not-found races after a redraw.

`08eeb994` is an empty commit, so run 2 tested the same tree as run 1. The `8.3 mysql80 job2` that failed in run 2 passed in run 1 - byte-identical code, an hour apart.

The behat coverage for this change passed in both runs: `multitab-validation.feature` and the GridField tab icon scenario in `validation.feature`.

`3.2` fails the same way without this branch. Run [32481120746](https://github.com/silverstripe/silverstripe-admin/actions/runs/32481120746) hit the same empty-alert exception in `multitab-validation.feature`, and the two runs before it failed in `gridfield-navigation.feature` and `site-settings.feature`. Different feature each time, always an endtoend job, always an element or alert timing race.

This is from a fork so I can't re-run the failed job - would someone with triage rights mind giving it a kick?

ChloeSartorelli (approved):
LGTM, thanks for tackling this!

Issue #1841, opened 2024-10-24
Standardise how actions are pulled through to admin sections


GuySartorelli:
Right now:
1. `LeftAndMain` gets actions from `$record->getCMSActions()` and only adds 'save' and 'delete' if there are _no_ actions in the result.
1. `CMSMain` gets actions from `$record->getCMSActions()` and uses it outright, after some normalisation of empty action menus
1. `GridFieldDetailForm_ItemRequest` which powers `ModelAdmin` completely ignores `$record->getCMSActions()`, building its own list from scratch

These should be normalised, so developers can rely on a consistent behaviour for how form actions are produced.

## Acceptance Criteria
- [ ] All admin sections include actions returned from the `getCMSActions()` method of the record
- [ ] All admin sections include a "create" action if the user has create permissions and this is a new record
- [ ] All admin sections include a "save" action if the user has edit permissions and this is a pre-existing record
- [ ] All admin sections include a "delete" action if the user has delete permissions and this is a pre-existing record
- [ ] silverstripe/versioned-admin replaces the "delete" action with a "archive" action for versioned records
- [ ] silverstripe/versioned-admin adds a "publish" action for versioned records if the user has publish permissions and the record has not been published
- [ ] silverstripe/versioned-admin adds a "unpublish" action for versioned records if the user has unpublish permissions and the record has been published
- [ ] The actions added by the admin sections will be displayed _before_ actions returned from the record
- [ ] The actions won't be added if they're already in the list returned from the record
- [ ] The actions should appear essentially the same as they currently do in the /admin/pages section

PR #2183, opened 2026-09-10
FIX Ensure SingleRecordAdmin always offers a Save action for editable records

### LINKED ISSUE #2182: SingleRecordAdmin loses the Save button when an extension adds a CMS action
### Module version(s) affected

3.x for CMS 6

### Description

# SingleRecordAdmin loses the Save button when an extension adds a CMS action

## Affected version(s)
- silverstripe/admin 3.x (CMS 6). Reproduced on a clean `silverstripe/recipe-cms ~6.2` install, PHP 8.4.

## Description
On any `SingleRecordAdmin` section (e.g. the built-in **Settings**/`SiteConfig` admin), the **Save** button disappears as soon as *any* extension adds an action through `updateCMSActions()`.

Root cause: `LeftAndMain::getEditForm()` only adds the default `save`/`delete` actions when the record's `getCMSActions()` returns an **empty** list:

```php
// vendor/silverstripe/admin/code/LeftAndMain.php (getEditForm)
if ($record->hasMethod('getAllCMSActions')) {
    $actions = $record->getAllCMSActions();
} else {
    $actions = $record->getCMSActions();
    // add default actions if none are defined
    if (!$actions || !$actions->count()) {
        if ($record->hasMethod('canEdit') && $record->canEdit()) {
            $actions->push(FormAction::create('save', ...));
        }
        ...
    }
}
```

`SiteTree` is immune because it defines its own `getCMSActions()` that already contains a Save action, so its list is never empty. But `SiteConfig` (and single-record models in general) define **no** `getCMSActions()`, so they rely entirely on this "empty list" fallback. The moment an extension pushes a single action, the list is non-empty and the Save button is silently dropped — the record can no longer be saved from the UI.

This is not specific to any one module; it happens with **any** action added via `updateCMSActions()`. In the wild it is triggered by, for example, `s2hub/silverstripe-autotranslate` (which adds an "Auto Translate" button to every translatable record, including `SiteConfig`).

## Steps to reproduce (clean install, no third-party modules)
1. On a fresh `silverstripe/recipe-cms ^6` install, add an extension that pushes one action onto `SiteConfig`:

```php
<?php

namespace App;

use SilverStripe\Core\Extension;
use SilverStripe\Forms\FormAction;
use SilverStripe\SiteConfig\SiteConfig;

/**
 * @property SiteConfig $owner
 */
class DemoSiteConfigActionExtension extends Extension
{
    public function updateCMSActions(FieldList $actions): void
    {
        $actions->push(
            FormAction::create('demoAction', 'Demo action')->setUseButtonTag(true)
        );
    }
}
```

```yaml
# app/_config/demo.yml
SilverStripe\SiteConfig\SiteConfig:
  extensions:
    - App\DemoSiteConfigActionExtension
```

2. `dev/build?flush=all`
3. Open **Settings** (`/admin/settings`).

## Expected
The **Save** button is present (the record is editable), alongside the custom action.

## Actual
Only the custom action ("Demo action") is shown. **No Save button** — the SiteConfig can no longer be saved.

Removing the extension (or its `push()`) brings the Save button back, confirming the cause.

### Possible Solution

The root cause is in `LeftAndMain::getEditForm()`: the default `save`/`delete` actions are only added when the record's `getCMSActions()` returns an **empty** list. Single-record models such as `SiteConfig` define no actions of their own, so any action added via `updateCMSActions()` makes the list non-empty and the Save button is dropped.

A minimal, backward-compatible fix scoped to `SingleRecordAdmin` (admin 3.x / CMS 6): after `parent::getEditForm()`, guarantee a Save action for editable records when one isn't already present.

```php
// SingleRecordAdmin::getEditForm(), after $form = parent::getEditForm($id, $fields);
$record = $this->getRecord($id ?: $this->currentRecordID());
if ($form && $record && $record->hasMethod('canEdit') && $record->canEdit()
    && !$form->Actions()->fieldByName('action_save')
) {
    $form->Actions()->unshift(
        FormAction::create('save', _t(LeftAndMain::class . '.SAVE', 'Save'))
            ->addExtraClass('btn btn-primary')
            ->setIcon('add-circle')
    );
}
```

This mirrors the Save action LeftAndMain already creates (same translation key, classes, icon), only adds it when the member can edit and no action_save exists (no duplicates; records that define their own Save are untouched). The broader standardisation across LeftAndMain/CMSMain/GridFieldDetailForm_ItemRequest remains 
[silverstripe/silverstripe-admin#1841](https://github.com/silverstripe/silverstripe-admin/issues/1841)
 (type/api-break, CMS 7.0). Happy to open a PR for this scoped fix.

## Relationship to #1841
#1841 ("Standardise how actions are pulled through to admin sections") covers this at the framework level for **all** admin sections, but is marked `type/api-break` / milestone **CMS 7.0**. This issue asks for a **backward-compatible fix scoped to `SingleRecordAdmin`** so the CMS 6 line (admin 3.x) is not left with an unsaveable Settings screen whenever a module adds a CMS action. A single-record admin whose sole purpose is editing one record should always offer a Save action for editable records.

A minimal, non-breaking fix (PR to follow): in `SingleRecordAdmin::getEditForm()`, after `parent::getEditForm()`, add a `save` action when `canEdit()` and none is present.

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

RVXD:
# Ensure SingleRecordAdmin always offers a Save action for editable records

Fixes #2182. Scoped, backward-compatible companion to #1841.

## Problem
`LeftAndMain::getEditForm()` only adds the default `save`/`delete` action when the record's `getCMSActions()` returns an empty list. Single-record models such as `SiteConfig` define no actions of their own, so any extension that adds an action via `updateCMSActions()` (e.g. a "translate" button) makes the list non-empty and the Save button silently disappears — leaving the Settings screen with no way to save.

`SiteTree` is unaffected because it defines its own `getCMSActions()` containing a Save action.

## Change
In `SingleRecordAdmin::getEditForm()`, after `parent::getEditForm()`, guarantee a `save` action for editable records when one is not already present:

```php
$record = $this->getRecord($id ?: $this->currentRecordID());
if ($form && $record && $record->hasMethod('canEdit') && $record->canEdit()
    && !$form->Actions()->fieldByName('action_save')
) {
    $form->Actions()->unshift(
        FormAction::create('save', _t(LeftAndMain::class . '.SAVE', 'Save'))
            ->addExtraClass('btn btn-primary')
            ->setIcon('add-circle')
    );
}
```

## Why this is safe
- Mirrors the Save action LeftAndMain already creates, using the same translation key, classes and icon.
- Only adds it when the member can edit the record **and** no `action_save` already exists (so records that provide their own Save are untouched, and no duplicate is created).
- Scoped to `SingleRecordAdmin`, whose entire purpose is editing a single record, so it should always be saveable.

## Scope / relationship to #1841
Deliberately minimal: this only restores the **Save** action for single-record admins in the CMS 6 (admin 3.x) line. The broader standardisation of how actions (create/save/delete + ordering) are pulled through across `LeftAndMain`, `CMSMain` and `GridFieldDetailForm_ItemRequest` remains #1841 (`type/api-break`, CMS 7.0).

## Testing
- Add an extension that pushes one action onto `SiteConfig` via `updateCMSActions()`, `dev/build?flush=all`, open `/admin/settings`.
  - Before: only the custom action shows, no Save.
  - After: Save is present alongside the custom action, and saving works.
- No change for records that already define a Save action (e.g. `SiteTree`), and none for members without edit permission.


RVXD:
Re-pushed to re-trigger CI: the previous red was an unrelated flaky end-to-end shard (the base HEAD b194dd32 failed a different shard while job3 passed there), not this change — which is a no-op for the "Form sudo mode" feature anyway, since LeftAndMain already adds the Save action and sudo mode makes it read-only before the SingleRecordAdmin guard runs.

code/SingleRecordAdmin.php
```diff
@@ -40,6 +41,25 @@ public function getEditForm($id = null, $fields = null): ?Form
...
+        // - for example a "translate" button - the list is no longer empty and
+        // the Save button silently disappears, leaving no way to save the
+        // record. Guarantee a Save action here for editable records.
+        // See https://github.com/silverstripe/silverstripe-admin/issues/1841.
+        $record = $this->getRecord($id ?: $this->currentRecordID());
+        if ($form && $record && $record->hasMethod('canEdit') && $record->canEdit()
+            && !$form->Actions()->fieldByName('action_save')
+        ) {
+            $form->Actions()->unshift(
+                FormAction::create('save', _t(LeftAndMain::class . '.SAVE', 'Save'))
```
andrewandante: I think in order to keep the buttons in-sync, it'd be worth making a function on `LeftAndMain` that returns the action, i.e.

```php
public static function getDefaultSaveAction(): FormAction
{
  return FormAction::create('save', _t(__CLASS__ . '.SAVE', 'Save'))
                            ->addExtraClass('btn btn-primary')
                            ->setIcon('add-circle');
```

and an equivalent for the Delete action (though it won't be used by `SingleRecordAdmin`), so that you can call that action directly from `SingleRecordAdmin` and ensure it's the same, without risk of them drifting apart.

code/SingleRecordAdmin.php
```diff
@@ -40,6 +41,25 @@ public function getEditForm($id = null, $fields = null): ?Form
         }
 
         $form = parent::getEditForm($id, $fields);
+
+        // LeftAndMain::getEditForm() only adds the default Save (and Delete)
```
andrewandante: This comment is overly verbose and specific - recommend reducing this down to something more concise, ideally a line or two.

code/SingleRecordAdmin.php
```diff
@@ -40,6 +41,25 @@ public function getEditForm($id = null, $fields = null): ?Form
...
+        // action when the record's getCMSActions() returns an EMPTY list. A
+        // single-record model (e.g. SiteConfig) usually defines no actions of
+        // its own, so as soon as any extension adds one via updateCMSActions()
+        // - for example a "translate" button - the list is no longer empty and
+        // the Save button silently disappears, leaving no way to save the
+        // record. Guarantee a Save action here for editable records.
+        // See https://github.com/silverstripe/silverstripe-admin/issues/1841.
+        $record = $this->getRecord($id ?: $this->currentRecordID());
```
andrewandante: If we are _always_ needing a `$record` now then we can move this higher up, where we are already kind of running this check. No need to duplicate

andrewandante (changes requested):
Thanks for the PR @RVXD - a couple of pieces of feedback I'd like to see.

code/SingleRecordAdmin.php
RVXD: Thanks @andrewandante! Addressed all three in c649eb4:

Extracted LeftAndMain::getDefaultSaveAction() and getDefaultDeleteAction() — LeftAndMain::getEditForm() and SingleRecordAdmin now both use them, so the Save action can't drift from the default (Delete added for parity, though SingleRecordAdmin doesn't use it).
Trimmed the comment down to two lines. Let me know if you want less comments.
SingleRecordAdmin::getEditForm() now resolves $id/$record once at the top instead of duplicating the lookup.

andrewandante (approved):
Thanks, looks good to me! Will look to get this merged shortly

PR #2189, opened 2026-09-17
fix: upgrade path-to-regexp to patched version (CVE-2024-45296)

anupamme:
## Summary
Fix CVE-2024-45296 (path-to-regexp ReDoS) by patching the vulnerable `path-to-regexp@0.1.7` pin (used by `page.js`) up to `0.1.13`, using a **descriptor-scoped** Yarn resolution rather than a blanket one, so unrelated dependencies keep tracking their own version ranges normally.

## History of this PR
1. Initially resolved to `8.4.0` — too broad (7 majors ahead, incompatible API).
2. Revised to a blanket `"path-to-regexp": "0.1.10"` — patches CVE-2024-45296, but forced a downgrade of `express`'s own `~0.1.12`-declared dependency below what it needs.
3. Revised to a blanket `"path-to-regexp": "0.1.13"` — fixes the downgrade (and a second advisory, CVE-2026-4867, that `0.1.12` doesn't fully address), but still a *blanket* override: it would permanently clamp `express`'s independently-declared `~0.1.12` range, so this repo would stop tracking future patch releases (e.g. a hypothetical `0.1.14`) unless someone manually bumped the resolutions value again.
4. **Now using a scoped/descriptor override**: `"path-to-regexp@0.1.7": "0.1.13"`. This only overrides requests matching the exact vulnerable range `page.js` declares (`0.1.7`), forcing those to the patched `0.1.13`. `express`'s own `~0.1.12` range is left untouched and resolves normally against the registry — it happens to land on `0.1.13` today, and will pick up future compatible patches on its own going forward.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | CVE-2024-45296 |
| **Severity** | HIGH |
| **Scanner** | trivy |
| **Rule** | `CVE-2024-45296` |
| **File** | `yarn.lock` (dependency: `path-to-regexp`) |

**Description**: path-to-regexp: Backtracking regular expressions cause ReDoS

## Changes
- `package.json`: `resolutions` entry changed from blanket `"path-to-regexp": "0.1.13"` to scoped `"path-to-regexp@0.1.7": "0.1.13"`
- `yarn.lock`: regenerated via `yarn install`

## Verification
- `yarn why path-to-regexp` shows `page.js` resolving `path-to-regexp@0.1.13` **via the override** (`via npm:0.1.13`), and `express` resolving `path-to-regexp@0.1.13` **via its own natural range** (`via npm:~0.1.12`) — confirming the override no longer touches express's independent resolution.
- No `0.1.7` (vulnerable) version remains anywhere in the tree.
- `yarn build` completes successfully: lint, full test suite (109 suites / 1102 tests) pass, and webpack rebuilds cleanly.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*


ChloeSartorelli:
This is not the procedure for reporting security issues. Please refer to https://docs.silverstripe.org/en/6/contributing/issues_and_bugs/#reporting-security-issues for any future security related issues you want to raise (and don't raise them for something you haven't actually confirmed is a problem).

Also the vulnerable version of `path-to-regexp` that is used in production code is 0.1.7. According to the vulnerability report, the correct version to use is 0.1.10 (not 8.4.0 which is 7 major versions higher and likely to cause things to not work).

Please update this PR to set the version to 0.1.10, and make sure you run `yarn build` to build out the change instead of only updating the yarn lock file.

anupamme:
Thanks for catching this, @ChloeSartorelli — you're right that forcing `path-to-regexp` to `8.4.0` was too broad for this remediation.

I've pushed an update that:
- Sets the `path-to-regexp` resolution to `0.1.10` (the patched release on the vulnerable `0.1.x` line) instead of `8.4.0`
- Regenerated `yarn.lock` by running `yarn install` rather than hand-editing it
- Ran `yarn build` so `client/dist` reflects the change

`yarn why path-to-regexp` now shows both `express` and `page.js` resolving to `0.1.10`, with no `0.1.7` left anywhere, and the full test suite (109 suites / 1102 tests) plus the production build pass cleanly.

Noted on the security disclosure process as well — will follow the documented process for any future reports.

yarn.lock
```diff
@@ -13422,17 +13422,10 @@ __metadata:
...
   linkType: hard
 
-"path-to-regexp@npm:0.1.7":
-  version: 0.1.7
-  resolution: "path-to-regexp@npm:0.1.7"
-  checksum: 10c0/50a1ddb1af41a9e68bd67ca8e331a705899d16fb720a1ea3a41e310480948387daf603abb14d7b0826c58f10146d49050a1291ba6a82b78a382d1c02c0b8f905
-  languageName: node
-  linkType: hard
-
-"path-to-regexp@npm:~0.1.12":
```
emteknetnz, line 13432: Looks there's two version being installed - path-to-regexp 0.17 and 0.1.13 (from something requiring ~0.1.12 it looks like - using `resolutions` in package.json means that this 0.1.13 version is being downgraded to 0.1.10 which means that some other dependency isn't getting a new enough version of path-to-regex

emteknetnz (changes requested):

yarn.lock
anupamme: Good catch, @emteknetnz — you're right, and it's worth being precise about what was actually happening:

- The "two versions" (`0.1.7` and `0.1.13`) you're seeing in the diff hunk is the **pre-fix** state (left side of the diff): before this PR, `page.js` resolved to the vulnerable `0.1.7` and `express` resolved to `0.1.13` on its own via its `~0.1.12` range. After the fix there's genuinely only one version installed (confirmed via `yarn why path-to-regexp` and checking `node_modules` directly) — but you were right that it was the *wrong* one.
- Forcing everything to `0.1.10` did downgrade `express` below what it declares (`~0.1.12`) and, more importantly, below what it actually needs: express's own `History.md` shows it moved `0.1.10` → `0.1.12` specifically to **"Fix backtracking protection"**, i.e. express's maintainers found `0.1.10`'s ReDoS fix incomplete. There's also a newer advisory, **CVE-2026-4867 (GHSA-37ch-88jc-xwx2)**, showing `0.1.12`'s protection is still exploitable with 3+ path parameters — fixed only in `0.1.13`.

I've pushed an update bumping the resolution to **`0.1.13`**, which satisfies express's own `~0.1.12` range naturally (no more forced downgrade), keeps `page.js` off the vulnerable `0.1.7`, and patches both CVE-2024-45296 and CVE-2026-4867. `yarn why path-to-regexp` now shows both `express` and `page.js` resolving to `0.1.13`, and the full test suite (109 suites / 1102 tests) plus production build pass cleanly.

ChloeSartorelli: This will mean when `0.1.14` is released we won't be benefiting from that with the dependency that currently declares `~0.1.12`.

I recommend using the syntax `"path-to-regexp@1.2.1": "1.9.0"` as recommended in https://github.com/visionmedia/page.js/issues/616

ChloeSartorelli:
Also, for the sake of anyone reading this and being worried about a vulnerability in their projects: the exploit relies on a routing pattern that we aren't actually using. While it will be good to get this patched, it doesn't actually affect projects that use Silverstripe CMS unless they're doing some funky custom client-side routing which is highly unlikely.

yarn.lock
anupamme: Thanks, that's a much better approach — pushed.

`resolutions` is now:
```json
"path-to-regexp@0.1.7": "0.1.13"
```
instead of the blanket `"path-to-regexp": "0.1.13"`. This only overrides requests where the requested range is exactly `0.1.7` (page.js's vulnerable pin), and leaves `express`'s own `~0.1.12` range completely unforced.

Confirmed via `yarn why path-to-regexp`:
```
├─ express@npm:4.22.2
│  └─ path-to-regexp@npm:0.1.13 (via npm:~0.1.12)
│
└─ page.js@npm:4.13.3
   └─ path-to-regexp@npm:0.1.13 (via npm:0.1.13)
```
`express` is resolving `0.1.13` via its own `~0.1.12` range (not the override), so it'll pick up `0.1.14` or later on its own once available — `page.js` is the only one being forced, and only off its specific vulnerable pin. Test suite (109 suites / 1102 tests) and build still pass.

ChloeSartorelli (approved):
Thanks for sorting this out. As mentioned above this vulnerability doesn't directly affect Silverstripe CMS but it is nice to get rid of it so it doesn't pop up in warnings etc.

